PR #49, opened 2022-12-07
Snowpark packaging improvements

jamesweakley:
## Reason for the change

Currently snowcli can examine a requirements.txt file, identify which packages are available on Anaconda, and download the remainder from PyPi to include in the app.zip artefact.

All of the dependencies of those PyPi packages are included, because pip has no way of knowing that Anaconda could provide some packages to the Snowpark object.

So this change is aimed at providing a second chance at leveraging Anaconda for packages not directly listed in requirements.txt.

## What's included

This change adds a boolean parameter `--check-anaconda-for-pypi-deps` which performs an optional second Anaconda check over the full list of packages pip downloaded. This naturally only comes into effect if a pip install occurs. The default value is True which is a change in behaviour, I figured this is probably desirable and the project is still in concept phase.

It does this by:
1. Scanning the downloaded package files, determining their package names and list of files/folders included
2. Comparing the list of packages with the Anaconda channel
3. For all matches, add the Anaconda package to the snowpark packages list and delete the local files/folders before they are zipped up.

As part of the change, I added a couple of other parameters to prevent prompts from occurring during packaging. I did this mainly because I'm using parts of snowcli in another app and needed programmatic control.
- `--package-native-libraries` allows you to pre-answer 'yes' or 'no' to the question of native libraries. Defaults to 'ask' so that current behaviour continues.
- `--pypi-download` allows you to pre-answer 'yes' or 'no' to the question of downloading missing Anaconda libraries from PyPi. . Defaults to 'ask' so that current behaviour continues.

## Alternate approaches

I could have leaned more on pip to find out about the dependant packages, maybe via a wrapper library like [johnnydep](https://pypi.org/project/johnnydep/). For example, you could pip freeze, filter the resulting requirements file, then purge/reinstall without dependencies.

However, I didn't want to risk interfering with snowcli itself by removing actual packages. By parsing the package metadata directly, it's more easily confined to the .packages folder.

### DIFF
--- a/src/snowcli/__about__.py
+++ b/src/snowcli/__about__.py
@@ -1,3 +1,3 @@
 from __future__ import annotations
 
-VERSION = "0.1.8"
+VERSION = "0.1.9"
--- a/src/snowcli/cli/function.py
+++ b/src/snowcli/cli/function.py
@@ -7,6 +7,9 @@ from shutil import copytree
 
 import pkg_resources
 import typer
+from snowcli.cli.snowpark_shared import CheckAnacondaForPyPiDependancies
+from snowcli.cli.snowpark_shared import PackageNativeLibrariesOption
+from snowcli.cli.snowpark_shared import PyPiDownloadOption
 from snowcli.cli.snowpark_shared import snowpark_create
 from snowcli.cli.snowpark_shared import snowpark_describe
 from snowcli.cli.snowpark_shared import snowpark_drop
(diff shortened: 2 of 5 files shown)


sfc-gh-jhollan:
This is awesome! looks great. I'll test this out manually for now as well to make sure things still play nice but on first glance seems great. I wonder if in the next release we should make this the default as I imagine most folks would want this behavior. I'd even be open to making this the default for now but either works

sfc-gh-jhollan (approved):
LGTM

jamesweakley:
@sfc-gh-jhollan I applied the pre-commit suggested fixes, are you able to release the new build?

sfc-gh-jhollan:
Awesome thank you! Yep let me create the release

PR #56, opened 2023-01-09
Use filename when creating streamlit app, not full path

### LINKED ISSUE #54: snow streamlit create with relative path to file fails
I ran the following commands:
```sh
snow streamlit create --file folder/file.py app_name
snow streamlit deploy --file folder/file.py app_name -o
```
Which created a new stage and uploaded file.py to the new stage, but put it in the base folder of the stage, while the streamlit app expected it to be located at `/folder/file.py`

SQL query
```sql
create streamlit app_name
ROOT_LOCATION = @DB.SCHEMA.app_name_stage
MAIN_FILE = '/folder/file.py'
QUERY_WAREHOUSE = WAREHOUSE;
```

The Streamlit app never ran, and showed this error 
image

I then went back and uploaded (through the UI) `file.py` into a new folder `/folder/` in the stage, and it worked

It seems that, if creating a streamlit only looks at the file name when uploading it to a stage, then it should also only use at the raw file name when

sfc-gh-zblackwood:
Fixes #54 

This makes `streamlit create --file relative/path.py` work the same as `streamlit deploy --file relative/path.py`, so that they both only look at the filename `path.py`.

Previously, `streamlit deploy` would create the file `path.py` in the new stage, but `streamlit create` would create a streamlit app that was expecting a path to be located at `relative/path.py`

### DIFF
--- a/src/snowcli/cli/streamlit.py
+++ b/src/snowcli/cli/streamlit.py
@@ -79,7 +79,7 @@ def streamlit_create(
             role=env_conf.get('role'),
             warehouse=env_conf.get('warehouse'),
             name=name,
-            file=str(file),
+            file=file.name,
         )
         print_db_cursor(results)
 


sfc-gh-jhollan:
image

Beautiful

sfc-gh-jhollan (approved):

PR #58, opened 2023-01-10
Fix streamlit deploy without host

sfc-gh-zblackwood:
If host is not set, fall back to old behavior of return the embedded url of the app


sfc-gh-jhollan (approved):
LGTM - @sfc-gh-jroes lmk if you have any thoughts before I merge this in and publish

PR #66, opened 2023-01-20
Cache SSH connections locally

### LINKED ISSUE #19: Allow connection re-use
You can technically authenticate via oauth today. If you create a config setting like the below in `~/.snowsql/config` it will prompt you to login with a browser and run the command:

```
[connections.myaccount]
username = jondoe
accountname = snow123
authenticator = externalbrowser
rolename = ACCOUNTADMIN
```

HOWEVER, it will make you login every single time you run any command. Would be great if the CLI could cache the tokens / connection and re-use as log as that session is valid.

sfc-gh-zblackwood:
It does seem to fix #19 once the secure-local-storage version has been installed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ dynamic = ["version"]
 requires-python = ">=3.8"
 description = "Snowflake CLI"
 dependencies = [
-  "snowflake-connector-python",
+  "snowflake-connector-python[secure-local-storage]",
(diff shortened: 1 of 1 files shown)


sfc-gh-jhollan:
Oh wow! I had no idea this could be so easy. Thank you

sfc-gh-jhollan (approved):

PR #69, opened 2023-01-25
Add role usage to create and describe streamlits

sfc-gh-tsimons:
Create Streamlit and Describe Streamlit are currently failing because no role is used in the SQL command even though it is passed in the python functions.


sfc-gh-zblackwood (approved):
LGTM :)

PR #73, opened 2023-01-31
Add drop/create stage, drop streamlit commands

sfc-gh-zblackwood:
Adds:
* `snow streamlit drop`
* `snow stage drop`
* `snow stage create`

### DIFF
--- a/src/snowcli/sql/drop_streamlit.sql
+++ b/src/snowcli/sql/drop_streamlit.sql
@@ -1,0 +1,6 @@
+use database {database};
+use schema {schema};
+use warehouse {warehouse};
+
+drop stage {name}_stage;
+drop streamlit {name};
--- a/src/snowcli/sql/drop_streamlit.sql
+++ b/src/snowcli/sql/drop_streamlit.sql
@@ -2,5 +2,6 @@ use database {database};
 use schema {schema};
 use warehouse {warehouse};
 
-drop stage {name}_stage;
+{drop_command}
+
 drop streamlit {name};
--- a/src/snowcli/cli/stage.py
+++ b/src/snowcli/cli/stage.py
@@ -120,3 +120,47 @@ def stage_put(
             parallel=parallel,
         )
         print_db_cursor(results)
+
+
+@app.command("create")
+def stage_create(
+    environment: str = EnvironmentOption,
+    name: str = typer.Argument(..., help="Stage name"),
+):
+    """
+    Create stage if not exists
+    """
+    env_conf = AppConfig().config.get(environment)
+
+    if config.isAuth():
+        config.connectToSnowflake()
+        results = config.snowflake_connection.createStage(
+            database=env_conf.get("database"),
+            schema=env_conf.get("schema"),
+            role=env_conf.get("role"),
+            warehouse=env_conf.get("warehouse"),
+            name=name,
+        )
+        print_db_cursor(results)
+
+
+@app.command("drop")
+def stage_drop(
+    environment: str = EnvironmentOption,
+    name: str = typer.Argument(..., help="Stage name"),
+):
+    """
+    Drop stage
+    """
+    env_conf = AppConfig().config.get(environment)
+
+    if config.isAuth():
+        config.connectToSnowflake()
+        results = config.snowflake_connection.dropStage(
+            database=env_conf.get("database"),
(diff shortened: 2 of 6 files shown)


src/snowcli/sql/drop_streamlit.sql
```diff
@@ -0,0 +1,6 @@
+use database {database};
+use schema {schema};
+use warehouse {warehouse};
+
+drop stage {name}_stage;
```
sfc-gh-jroes, line 5: This may be problematic if they are using the stage for other things. I don't have a good recommendation for how to resolve, but just pointing it out. I don't think we support specifying the stage when creating a streamlit today in the CLI.

sfc-gh-zblackwood: I could add an extra argument that would allow you to not drop the stage. Or I could leave this step out, since you could drop the stage separately. WDYT?

sfc-gh-jhollan: my 2c is either option is ok and has pros and cons. I keep going back and forth. I personally may want it "opt out" because I would probably always want to drop stage at same time, but can see a very valid other side. @sfc-gh-jroes not sure if you have a strong opinion other way but a bit of a coin flip for me on options

sfc-gh-jroes:
I'm open as well. @sfc-gh-lwilby said she'd take a look, maybe we could get
her $0.02.

src/snowcli/sql/drop_streamlit.sql
sfc-gh-tsimons: Once [embedded stage](https://snowflakecomputing.atlassian.net/browse/SNOW-698167) lands, does it become less likely that the stage will be used for other things?

sfc-gh-lwilby: I'm not completely sure how you should handle this, but here is some background on how stages are being handled elsewhere and what will change with embedded stages! 

Currently in the UI we are handling this by automatically creating a stage when the user creates a streamlit. 

That stage is associated with the streamlit using a system function (`SYSTEM$ASSOCIATE_STAGE_WITH_STREAMLIT(<streamlit_name>, <stage_name>)`). 

When the streamlit is dropped, the stage is dropped in the backend. 

This is an interim solution to simplify the UX while embedded stages is developed. 

There are some downsides, like the user could see this stage if they did `SHOW STAGES` and be confused about where these stages came from. They could change the ownership of the stage but not the streamlit causing issues with running/editing the streamlit. They could delete the stage. They may not have permission to create a stage in the same db/schema as the streamlit and creating the streamlit will fail. For the UI, we decided simplifying the UX was worth these risks. 

Embedded stages removes the requirement to specify a stage when you are creating a streamlit. You will actually not be able to specify an external stage that will be used by the streamlit. 

You can do ... 

`CREATE STREAMLIT <name> FROM <stage>` 

But the files will be copied from the provided stage and added into the embedded stage. 

The embedded stage is nested in the streamlit, not a db/schema, so it will not be returned by `SHOW STAGES`. The user will be able to do stage operations, but they will reference a new url format that points to the streamlit. 

All these restrictions will make reusing the stage for other things less likely @sfc-gh-tsimons . They could still put some files in the stage, but the access would be through the streamlit and the sub-directories they can use will be restricted, so it should look like a bad idea to reuse this for some other purpose. 

With that in mind, it may be better to do something similar to the UI and create the stage in the background? That might be more forward compatible with embedded stage?

sfc-gh-lwilby: On the other hand, in the CLI the user experience may be MORE confusing since these users are probably more familiar with stages and would be more likely to encounter the potential problems?

sfc-gh-jhollan: Thanks much @sfc-gh-lwilby - agree on at least mirroring UI here, which for sake of this would me doing what is currently described here of dropping stage by default, and @sfc-gh-zblackwood we could add a flag to opt-out of that behavior if folks did re-use. And once embedded stages shows up we no longer need to opt-out.

That said- given above I'm ok to leave code as-is (which matches UX).

sfc-gh-jhollan (approved):
Feel free to add a patch to create an opt-out flag for dropping stages, or if you want to just start with PR as-is. I'm personally cool either way

PR #82, opened 2023-02-23
Fix typing error

sfc-gh-zblackwood:
For some reason, despite the `from __future__ import annotations`, this is causing issues with python 3.8 when running `snow streamlit` commands with 0.1.21.

I tested a standalone script with `list[str]` and it worked fine in python 3.8, so there must be some way that Typer is using that annotation in a way that doesn't actually use the `from future...`

This could possible also be fixed by updating versions of Typer, but this is an easier fix for now.

I also went ahead and added some tox tests to make sure that we catch issues like this with python 3.8, and am setting them up to run as GHA.

I did have to drop pyupgrade, which I think is pretty nice to have, but it doesn't seem to have any way to ignore a specific line, and it isn't playing nicely with the limitations of Typer


sfc-gh-jhollan (approved):

sfc-gh-jhollan (approved):
Even added an action. Impressive

sfc-gh-zblackwood:
Still some build failures, but the tests are all passing now (and I confirmed that they did fail with the old version), so I'm happy with this PR now.

PR #87, opened 2023-02-23
Fix stage name case since quotes are now added

sfc-gh-zblackwood:
Without the quotes it seems it automatically upper-cased the stage name, but with the quotes it doesn't, causing "does not exist" errors.

### DIFF
--- a/src/snowcli/snow_connector.py
+++ b/src/snowcli/snow_connector.py
@@ -500,7 +500,7 @@ class SnowflakeConnector:
         drop_stage: bool = True,
     ) -> SnowflakeCursor:
 
-        drop_command = f'drop stage "{name}_stage";' if drop_stage else ""
+        drop_command = f'drop stage "{name}_STAGE";' if drop_stage else ""
 
         return self.runSql(
             "drop_streamlit",


jeffhollan (approved):

sfc-gh-jhollan (approved):
Approving with right account ;)

thomaseibner:
If anything should you not be leaving out the quotes completely? 

`snowcli.py`:
```
    def uploadFileToStage(
        self,
        file_path,
        destination_stage,
        path,
        role,
        database,
        schema,
        overwrite,
    ):
        self.cs.execute(f"use role {role}")
        self.cs.execute(f"use database {database}")
        self.cs.execute(f"use schema {schema}")
        self.cs.execute(
            f"create stage if not exists {destination_stage} "
            'comment="deployments managed by snowcli"',
        )
# ..
        self.uploadFileToStage(
            file_path,
            f"{name}_stage",
            stage_path,
            role,
            database,
            schema,
            overwrite,
        )
# ..
        drop_command = f'drop stage "{name}_STAGE";' if drop_stage else ""
```
Now you have a mix of a quoted identifier for the drop and unquoted identifier for the create? If the code ever has to support quoted identifiers in the name for the stage it might be better to keep them consistent.

sfc-gh-zblackwood:
@thomaseibner I agree that it should be consistent. The specific reason for including the quotes in this context is that if have a streamlit with a space in the name, quotes seem to be the only reasonable way to handle that.

PR #90, opened 2023-03-03
Code coverage wrapper for stored procs

jamesweakley:
This PR introduces a new flag to `snow procedure create` and `snow procedure update` named `--install-coverage-wrapper`.

When you enable it, wraps the proc handler with some coverage.py code that, after each proc invocation, writes a code coverage report back to the stage.

You can then run `snow procedure coverage report` and it'll download them, merge them, and generate a HTML code coverage report.

You can run `snow procedure coverage clear` to delete the coverage files from the stage to start a new recording session.

Feedback welcome.


src/snowcli/python_templates/snowpark_coverage.py.jinja
```diff
@@ -0,0 +1,22 @@
...
+    cov = coverage.Coverage(data_file=coverage_file)
+    cov.start()
+    # call the original handler
+    import {{handler_module}}
+    result = {{handler_module}}.{{handler_function}}(*args,**kwargs)
+    cov.stop()
+    cov.save()
+    session = args[0]
+    session.file.put(coverage_file,f'@{{coverage_reports_stage}}/{{coverage_reports_stage_path}}',auto_compress=False)
+    return result
```
sfc-gh-kbregula, line 22:
```suggestion
    # call the original handler
    import {{handler_module}}
    try:
        return {{handler_module}}.{{handler_function}}(*args,**kwargs)
    finally:
        try:
            cov.stop()
            cov.save()
            session = args[0]
            session.file.put(coverage_file,f'@{{coverage_reports_stage}}/{{coverage_reports_stage_path}}',auto_compress=False)
        except Exception:
            pass
```
This will allow us to track negative cases as well.

jamesweakley:
@sfc-gh-kbregula I noticed it's trying to build against Python 3.9 and 3.10, but since I introduced Snowpark for a simpler interface to staging files that's causing the build to fail.
What's the best way forward do you think? Do I need to remove Snowpark since it's limited to 3.8?

sfc-gh-jhollan:
Oh good point. Right now we’ve built to run on 3.8/3.9/3.10. Snowpark would mean we’d have to make only 3.8 compatible for now (at least until Snowpark supports newer runtimes which is on roadmap). 

Looked briefly and looks like you’re using snowpark to get files, anything else? I wonder if it’s not too much work if we just use python connector for now? Id be open to exploring moving to 3.8 exclusive.

jamesweakley:
it wouldn't be too much work to just use the python connector, I'll switch it to that

sfc-gh-jhollan (approved):
good for this to merge?

PR #102, opened 2023-03-29
Add render template command

sfc-gh-turbaszek:
Given template like:
```
CREATE ROLE APP_ROLE;

CREATE SCHEMA APP;
GRANT USAGE ON SCHEMA APP TO ROLE APP_ROLE;

-- procedures
CREATE OR REPLACE PROCEDURE PUBLIC.FOO()
    RETURNS VARCHAR
    LANGUAGE JAVASCRIPT
    EXECUTE AS OWNER
AS
$$
{{ "proc.js" | procedure_from_js_file }}
$$;
```
you can do:
```
➜ snow render template setup.sql
CREATE ROLE APP_ROLE;

CREATE SCHEMA APP;
GRANT USAGE ON SCHEMA APP TO ROLE APP_ROLE;

-- procedures
CREATE OR REPLACE PROCEDURE PUBLIC.FOO()
    RETURNS VARCHAR
    LANGUAGE JAVASCRIPT
    EXECUTE AS OWNER
AS
$$
var module = {};
var exports = {};
module.exports = exports;
(function() {
function foo() {

}
module.exports = foo

})()
return module.exports.apply(this, arguments);
$$;
```

### DIFF
--- a/src/snowcli/cli/__init__.py
+++ b/src/snowcli/cli/__init__.py
@@ -9,7 +9,16 @@ from rich import print
 from .. import __about__
 from ..config import AppConfig
 from ..snowsql_config import SnowsqlConfig
-from . import connection, function, package, procedure, stage, streamlit, warehouse
+from . import (
+    connection,
+    function,
+    package,
+    procedure,
+    render,
+    stage,
+    streamlit,
+    warehouse,
+)
 
 app = typer.Typer(context_settings={"help_option_names": ["-h", "--help"]})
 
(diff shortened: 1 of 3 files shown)


src/snowcli/cli/__init__.py
```diff
@@ -137,6 +146,8 @@ def default(
 app.add_typer(warehouse.app, name="warehouse")
 app.add_typer(stage.app, name="stage")
 app.add_typer(package.app, name="package")
+app.add_typer(render.app, name="render", hidden=True)
```
sfc-gh-jhollan, line 149: @sfc-gh-turbaszek any reason you are hiding this? Just to do a slower rollout?

sfc-gh-jhollan (approved):

PR #105, opened 2023-03-30
Add simple sql command allowing to execute any query from cli

sfc-gh-turbaszek:
This PR adds `snow sql` command that allows executing query from string option, file or stdin:
```
➜ snow sql --help                                                       
                                                                                                                                                                                                                        
 Usage: snow sql [OPTIONS]                                                                                                                                                                                              
                                                                                                                                                                                                                        
 Executes Snowflake query.                                                                                                                                                                                              
 Query to execute can be specified using - query option - filename option (all queries from file will be executed) - stdin by piping output, for example `snow render template my.sql | snow sql`                       
                                                                                                                                                                                                                        
╭─ Options ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --query       -q      TEXT  Query to execute. [default: None]                                                                                                                                                        │
│ --filename    -f      FILE  File to execute. [default: None]                                                                                                                                                         │
│ --connection  -c      TEXT  Connection to be used [default: None]                                                                                                                                                    │
│ --verbose     -v            Prints information about executed queries [default: True]                                                                                                                                │
│ --help        -h            Show this message and exit.                                                                                                                                                              │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

```

Allowing stdin input allows using `snow render template` for example:
```
snow render template create_ml_schema.sql -D "db_name=foo_bar" | snow sql
```

### DIFF
--- a/src/snowcli/cli/__init__.py
+++ b/src/snowcli/cli/__init__.py
@@ -15,6 +15,7 @@ from . import (
     package,
     procedure,
     render,
+    sql,
     stage,
     streamlit,
     warehouse,
@@ -146,7 +147,8 @@ app.add_typer(connection.app, name="connection")
 app.add_typer(warehouse.app, name="warehouse")
 app.add_typer(stage.app, name="stage")
 app.add_typer(package.app, name="package")
-app.add_typer(render.app, name="render", hidden=True)
+app.add_typer(render.app, name="render")
+app.command("sql")(sql.execute_sql)
 
 
 if __name__ == "__main__":
--- a/src/snowcli/cli/render.py
+++ b/src/snowcli/cli/render.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 import json
 from pathlib import Path
 from textwrap import dedent
-from typing import Optional
+from typing import List, Optional
 
 import jinja2
 import typer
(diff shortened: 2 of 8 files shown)


jeffhollan:
Overall this is really good. Over time we talked about having 'snow sql' mirror what snowsql does today (backwards compatible). @sfc-gh-turbaszek do you know if these flags match snowsql already so we could still move to that model?

sfc-gh-turbaszek:
> Overall this is really good. Over time we talked about having 'snow sql' mirror what snowsql does today (backwards compatible). @sfc-gh-turbaszek do you know if these flags match snowsql already so we could still move to that model?

I agree. Query, filename and connection options are aligned with `snowsql`. The `--verbose` is the only new flag. `snowsql` verbosity is controlled by `output_format` option which is much more complex than what's proposed in PR and that can be added later one. 

Accepting `stdin` is a new addition that can be worked around in `snowsql` using `xargs`.

sfc-gh-jhollan (approved):

PR #106, opened 2023-03-30
Use jinja for rendering sql templates

sfc-gh-turbaszek:


requirements-dev.txt
```diff
@@ -1,3 +1,4 @@
 pytest==7.2.0
+syrupy==2.3.0
```
sfc-gh-turbaszek, line 2: Used for snapshot testing, we used it in other project for SQL comparison.

sfc-gh-turbaszek:
Note to reviewers: use this comparison https://github.com/Snowflake-Labs/snowcli/pull/106/commits/f48b262056bd0c37d5ceb02c00e4634cc13066ce as information about what changes in the queries. I've added the tests before switching to jinja so we can see that everything seems ok after the change.

sfc-gh-jhollan (approved):
Wow thank you!

PR #109, opened 2023-04-04
implement dynamic cli app importing

sfc-gh-mkeller:
In this PR I added dynamic importing to all sub-modules inside `snowcli.cli`. So now every file will be added that defines a global `app`.
This will make it so that we don't need to manually import new modules and call `add_typer` on their apps


sfc-gh-jhollan (approved):
Thank you this is a nice improvement

PR #112, opened 2023-04-06
Added SNOWCLI_EXTRA_LIBS option for shared libs

sfc-gh-cconner:


sfc-gh-jhollan:
Thank you for creating this! In general I like it - a few thoughts (and happy to try to add to take to finish line):

1. I wonder if we can have this work with directories as well as individual files. `PYTHONPATH` today I think does so by directories so could be consistent.
2. Naming. I wonder if we call this like `SNOWCLI_INCLUDE_LIBS` or `SNOWCLI_INCLUDE_PATHS` or something?

sfc-gh-cconner:
Sounds good, I'll play around with it some.

sfc-gh-cconner:
Made the recommended changes!

sfc-gh-jhollan (approved):
This is awesome! Thank you

PR #118, opened 2023-04-29
Minor Linting, Reworking Response Status Return, Config File Tweaks

IndexSeek:
- Performed flake8 checks for quality checks; more info on F401 and F451 at [Error / Violation Codes](https://flake8.pycqa.org/en/latest/user/error-codes.html)
- Added cache folders for mypy and ruff to gitignore
- Exit [src/snowcli/utils.py](https://github.com/Snowflake-Labs/snowcli/commit/88f3a0f7615d331269401c4da92785d9f7d0928f#diff-b3e7f12457b86d4c37c642c8026f99672c7857e94da58c9204d843ba17298767) earlier to reduce conditional nesting. 
- `os.path.expanduser(path)` being used multiple times in `src/snowcli/snowsql_config.py`, initializing that with the class.
- Single command `touch` in README for Unix systems

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ If you do not want to install SnowSQL, you can add Snowflake account credentials
 
 1. In your home directory, make `.snowsql` directory with a `config` file:
 
-   `touch ~/.snowsql/config`
+   `mkdir -p ~/.snowsql && touch ~/.snowsql/config`
 
 2. Open the config file for editing:
 
--- a/README.md
+++ b/README.md
@@ -94,9 +94,7 @@ If you do not want to install SnowSQL, you can add Snowflake account credentials
 
 1. In your home directory, make `.snowsql` directory with a `config` file:
 
-   `mkdir .snowsql`
-   `cd .snowsql`
-   `touch config`
+   `touch ~/.snowsql/config`
 
 2. Open the config file for editing:
 
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ dist/
 .idea
 .tox
 .packages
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -94,9 +94,7 @@ If you do not want to install SnowSQL, you can add Snowflake account credentials
 
 1. In your home directory, make `.snowsql` directory with a `config` file:
 
-   `mkdir .snowsql`
-   `cd .snowsql`
-   `touch config`
+   `touch ~/.snowsql/config`
```
sfc-gh-jroes, line 97: Think we still need the mkdir; touch won't create a directory if it doesn't exist. At least, not the version on OSX from my testing:

```shell
λ touch /tmp/some/long/path/test.txt
touch: /tmp/some/long/path/test.txt: No such file or directory

λ touch /tmp/some/path/test.txt 
touch: /tmp/some/path/test.txt: No such file or directory

λ touch /tmp/path/test.txt 
touch: /tmp/path/test.txt: No such file or directory
```

sfc-gh-twhite: Can we just swap it to use `mkdir ~/.snowsql && touch ~/.snowsql/config` in that case?

I probably had already created it and tried to break it down just to use the `touch` with the parent directory being present as you are correct, I don't think it would create the directory. 

In my experience, I find the fewer steps to get people started, even if it's just one or two fewer commands, makes the library seem more "digestable".

sfc-gh-jroes: Yeah for sure I think that would be good. Might need `mkdir -p` to make sure it doesn't return non-zero if the dir exists.

PR #123, opened 2023-05-16
Change ROOT_LOCATION to FROM

sfc-gh-tsimons:
Per @sfc-gh-asaini, [the DDL for CREATE STREAMLIT has changed. ](https://snowflake.slack.com/archives/C03E8TLQ7LP/p1684167544391709)

We need to update the command to match this new format. This change is a breaking change from the previous version so ROOT_LOCATION will no longer work. 

However, since we have moved to embedded stages, it may be easier to keep creating them so we can upload other files there. For that reason, I am not removing the CREATE STAGE portion.

### DIFF
--- a/src/snowcli/cli/streamlit.py
+++ b/src/snowcli/cli/streamlit.py
@@ -165,10 +165,6 @@ def streamlit_share(
 def streamlit_drop(
     environment: str = EnvironmentOption,
     name: str = typer.Argument(..., help="Name of streamlit to be deleted."),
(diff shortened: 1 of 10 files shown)


sfc-gh-tsimons:
@sfc-gh-zblackwood Same for this PR -- please merge it in if you do end up approving it.

sfc-gh-zblackwood (approved):
👍

sfc-gh-mkeller (approved):
:ship: :ship:

PR #130, opened 2023-05-18
Fix create streamlit syntax

sfc-gh-zblackwood:
One more small fix to get the FROM STAGE working


sfc-gh-mkeller (approved):
:ship: :ship:

PR #131, opened 2023-05-19
Add dependabot

sfc-gh-astus:


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
```
sfc-gh-pjob, line 2: I wonder if we shouldn't also configure dependabot for github actions (https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot).

@sfc-gh-astus , @sfc-gh-turbaszek WDYT?

I'm not very familiar with github actions, I've been using GitLab for last few years, but I see that we have already a few different versions of the same action (checkout) used in the pipeline configurations:
* checkout v1 & v2 -> https://github.com/Snowflake-Labs/snowcli/blob/main/.github/workflows/build.yaml
* checkout v3 -> https://github.com/Snowflake-Labs/snowcli/blob/main/.github/workflows/publish.yaml

sfc-gh-astus: I think it is a good idea. I will add it.

About different versions in actions I think we should update them. I will create new PR with it.

sfc-gh-pjob: > About different versions in actions I think we should update them. I will create new PR with it.

Hmm, I thought dependabot will help us with it :)

sfc-gh-pjob (approved):

.github/dependabot.yml
sfc-gh-pjob: After private discussion with Adam I understood that a separate PR for GH actions upgrade is a good idea to avoid flooding with PRs containing initial upgrades created by dependabot separately for each action. 

I've also dismissed my previous PR approval because in our opinion we have to think about a similar PR for python dependencies upgrade before merging the dependabot configuration.

sfc-gh-jsikorski (approved):

sfc-gh-pjob (approved):
@sfc-gh-astus I've seen your comment about 5 PRs limit by default, so I'm reapproving this PR.

PR #141, opened 2023-05-23
Changed function names in utils.py

sfc-gh-jsikorski:
Changed names of some functions in utils.py to make them consistent with python naming conventions.
I predict no behaviour change, as none of the functions is called directly by the user.

### DIFF
--- a/src/snowcli/cli/package.py
+++ b/src/snowcli/cli/package.py
@@ -36,7 +36,7 @@ def package_lookup(
     """
     Check to see if a package is available on the Snowflake anaconda channel.
     """
-    packageResponse = utils.parseAnacondaPackages([name])
+    packageResponse = utils.parse_anaconda_packages([name])
     ## if list has any items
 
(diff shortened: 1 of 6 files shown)


sfc-gh-pjob:
Changes look good but I have two questions / comments:
1. IMO it's a good idea to merge main/master -> feature branch before creating a PR. This branch is 8 commits behind the main.
2. Maybe we should think about some auto-formatter or style checker? @sfc-gh-turbaszek WDYT about it?

sfc-gh-turbaszek:
> 2. Maybe we should think about some auto-formatter or style checker? @sfc-gh-turbaszek WDYT about it?

I agree with that. We should enable more checks in ruff configuration. Probably using [default](https://beta.ruff.rs/docs/settings/#select) `["E", "F"]` will be a good start. At the beginning we can disable some rules to avoid to many changes at once.

sfc-gh-pjob (approved):

PR #142, opened 2023-05-23
Refactor existing commands

sfc-gh-astus:
* move function, package, procedure under snowpark command
* move stage, warehouse under core command
* hide render command


src/snowcli/cli/core/__init__.py
```diff
@@ -0,0 +1,11 @@
+import typer
+
+from . import (
```
sfc-gh-turbaszek, line 3: I would suggest using absolute import `from snowcli.cli.core.stage import app as stage_app`. Absolute imports behave better when code is refactored and structure is changing. This can be our case

src/snowcli/cli/core/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+    stage,
+    warehouse,
+)
+
+app = typer.Typer(context_settings={"help_option_names": ["-h", "--help"]})
+
+app.add_typer(stage.app, name="function")
+app.add_typer(warehouse.app, name="package")
```
sfc-gh-turbaszek, line 11: Are those names correct? Shouldn't id be `stage` and `warehouse`?

src/snowcli/cli/core/stage.py
```diff
@@ -49,7 +49,7 @@ def stage_get(
     environment: str = EnvironmentOption,
     name: str = typer.Argument(..., help="Stage name"),
     path: Path = typer.Argument(
-        Path("."),
+        Path(".."),
```
sfc-gh-turbaszek, line 52: Why? This should download the file to current working directory of user

sfc-gh-turbaszek: Maybe using `Path.cwd()` will be more explicite?

src/snowcli/cli/core/__init__.py
sfc-gh-astus: copy paste 😢

src/snowcli/cli/core/stage.py
sfc-gh-astus: Idea changed it 😮 , I didn't even notice it. Changed to `Path.cwd()`

src/snowcli/cli/core/stage.py
```diff
@@ -49,7 +49,7 @@ def stage_get(
     environment: str = EnvironmentOption,
     name: str = typer.Argument(..., help="Stage name"),
     path: Path = typer.Argument(
-        Path("."),
+        Path.cwd(),
```
sfc-gh-turbaszek, line 52: Should we add basic test for this method? Probably assuring this path is same before and after refactor will be enough. WDYT?

src/snowcli/cli/snowpark/__init__.py
```diff
@@ -0,0 +1,11 @@
+import typer
+
+from snowcli.cli.snowpark.function import app as function_app
+from snowcli.cli.snowpark.package import app as package_app
+from snowcli.cli.snowpark.procedure import app as procedure_app
+
+app = typer.Typer(context_settings={"help_option_names": ["-h", "--help"]})
+
+app.add_typer(function_app, name="function")
```
sfc-gh-turbaszek, line 9:
```suggestion
app.add_typer(function_app)
```
Will this work? The name will be taken from type instance

sfc-gh-turbaszek (approved):

PR #144, opened 2023-05-25
Jsikorski/snow 824873 add more checks in ruff

sfc-gh-jsikorski:
Added another check in pre-commit, using ruff.
As i don`t want to autofix functon, variable etc. names - it is performed as a separate step, after all linting and formatting.

Also, as many names in the code were not compliant with PEP-8 guidelines, i took a liberty to reformat them to stick with the rules.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,10 +11,10 @@ repos:
       - id: debug-statements
       - id: check-ast
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: "v0.0.269"
+    rev: "v0.0.270"
     hooks:
       - id: ruff
-        args: [--fix, --exclude, "**/tests/"]
+        args: [--fix, --exclude, "**/tests/", --unfixable, "N"]
   - repo: https://github.com/psf/black
     rev: 22.12.0
     hooks:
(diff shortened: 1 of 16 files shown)


src/snowcli/cli/snowpark/package.py
```diff
@@ -37,10 +37,10 @@ def package_lookup(
     """
     Check to see if a package is available on the Snowflake anaconda channel.
     """
-    packageResponse = utils.parse_anaconda_packages([name])
-    ## if list has any items
+    package_response = utils.parse_anaconda_packages([name])
```
sfc-gh-astus, line 40: Please separate task part and refactor. Create new PR for it

.pre-commit-config.yaml
```diff
@@ -24,3 +24,8 @@ repos:
     hooks:
       - id: mypy
         additional_dependencies: [types-all]
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    rev: "v0.0.270"
+    hooks:
+      - id: ruff
+        args: [--select, "N"]
```
sfc-gh-astus, line 31: Do we need to have this repo twice?

src/snowcli/cli/snowpark/package.py
sfc-gh-turbaszek: I think the refactor is part of this task. Introducing precommit check without refactoring code will result in precommit raisning error for others (or main branch)

PR #159, opened 2023-05-31
Fixed regex in get_package_name_from_metadata

sfc-gh-jsikorski:
### Pre-review checklist

### Changes description
Tests for this change are part of SNOW-821329_add_missing_unit_tests

So far, function get_package_name_from_metadata in utils module, was using a regex:
`re.search("^Name: (.*)$", contents, flags=re.MULTILINE)`

When checked with example files:
https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html
https://gist.github.com/bbengfort/a77dd0ff610fd10f40926f7426a89486

it appeared, the code has two flaws:
1. It searches only for the occurence at the beginning of a line, while in example files, "name" is indented with two spaces
2. It looks for uppercase "Name", while example files use lowercase "name"

To fix it, i added IGNORECASE flag, and allowed 0+ spaces at the beggining of the line.

### DIFF
--- a/src/snowcli/utils.py
+++ b/src/snowcli/utils.py
@@ -237,10 +237,10 @@ def get_package_name_from_metadata(metadata_file_path: str) -> str | None:
     """
     with open(metadata_file_path, encoding="utf-8") as metadata_file:
         contents = metadata_file.read()
-        results = re.search("^Name: (.*)$", contents, flags=re.MULTILINE)
+        results = re.search("^( *)name: (.*)$", contents, flags=re.MULTILINE | re.IGNORECASE)
         if results is None:
             return None
-        return results.group(1)
+        return results.group(2)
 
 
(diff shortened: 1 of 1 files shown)


sfc-gh-turbaszek (changes requested):
I think the current code is correct. The METADATA we parse is file that is added to wheel package and it looks like this:
```
Metadata-Version: 2.1
Name: something_name
Version: 0.0.1
Requires-Dist: requests===2.28.1
Requires-Dist: snowflake-connector-python==3.0.2
Requires-Dist: snowflake-snowpark-python==1.1.0
Provides-Extra: dev
Requires-Dist: pytest; extra == 'dev'
```

Reference: 
https://peps.python.org/pep-0491/#file-contents
https://packaging.python.org/en/latest/specifications/core-metadata/

PR #163, opened 2023-06-02
Added support for environment variables in config

sfc-gh-astus:
### Pre-review checklist

### Changes description
* Add support for environment variables in connection parameters
* Hide ConnectionConfigs config field and allow access only by methods


src/snowcli/snowsql_config.py
```diff
@@ -28,3 +29,12 @@ def add_connection(self, connection_name, entry):
...
             self.config.write(f)
+
+    def _replace_with_env_variables(self):
+        connection_sections = [
+            s for s in self.config.sections() if s.startswith("connections.")
+        ]
+        for section in connection_sections:
+            for (name, value) in self.config[section].items():
+                if value.startswith("$SNOWCLI_"):
+                    self.config[section][name] = os.environ[value[1:]]
```
sfc-gh-turbaszek, line 40: Hm, I'm wondering if wouldn't it be better to resolve the variables when we read them?

sfc-gh-astus: You mean just before pass them to snowflake connector?

sfc-gh-pjob: 1. Why do we have to limit env substitution only to `connections` sections? IMO the code will be simpler if we apply env variables resolution to whole config (even if we don't need the rest of the config).
2. Do I understand the code correctly that it won't be possible to use env variable in the middle of string value (that value has to start with "$SNOWCLI_"?
3. Maybe we can somehow add env vars resolution to configparser (I found a few ideas here: https://stackoverflow.com/questions/26586801/configparser-and-string-interpolation-with-env-variable)?  It might be more consistent with the syntax of other interpolations already possible to use in the config?

tests/config/connection_config
```diff
@@ -0,0 +1,7 @@
+[connections.test]
```
sfc-gh-pjob, line 1: Does this file is meant to be used by other tests also or is it special only for the test of env vars substitution?
If it's the second case, I think we should think about more specific name to not mislead us and others in the future (I'd think that this is some common connection configuration used in tests).

sfc-gh-pjob (changes requested):

tests/config/connection_config
sfc-gh-pjob: I forgot to mention that I'm talking about the filename, not the section name.

src/snowcli/snowsql_config.py
sfc-gh-astus: ad1. Because we only need and use connections part
ad2. Yes, it has start with `$SNOWCLI_`. But why someone would like to use it in the middle of string? 
ad3. Configparser with os.environ loads all env variables into default section in config object. I don't like it :/

tests/config/connection_config
sfc-gh-astus: For now only one test is using it, so changed name to more specific

src/snowcli/snowsql_config.py
sfc-gh-pjob: 1. Ok, so if we use only connections part, then maybe the config should not be named `SnowsqlConfig`? I understand naming of this class can be outside of the scope of this PR, but then maybe we can at least filter out all unused sections just after creation of config object instead of doing it in the function responsible for env vars resulution (which imo shouldn't have to care which sections are used and which not).
2. If we care only about currently used settings, then I can't give any example but I can imagine using it in paths, urls, ... So, if the goal is to handle env vars only for current form of snowsql connection configuration, then the solution is ok, but if the goal is to prepare mechanism for the future usage, then IMO we're limiting this usage.
3. Two notes here:
    * `ConfigParser(os.environ)` is only one of solutions.
    * We can filter env vars dictionary searching for some prefix.

sfc-gh-astus: 1. Changed `SnowsqlConfig` to `ConnectionConfigs`
2. We can extend this in future, for now it is not needed
3. Checked this solution a bit more and configparser coping default variables into all sections. It creates only mess in dict

src/snowcli/snow_connector.py
```diff
@@ -28,15 +26,10 @@ class SnowflakeConnector:
 
     def __init__(
         self,
-        snowsql_config: SnowsqlConfig,
-        connection_name: str,
+        connection_config: dict,
```
sfc-gh-astus, line 29: Changed it but don't like it. I think it should be dataclass, because now we don't know what should be int this dict. WDYT?

src/snowcli/snowsql_config.py
sfc-gh-turbaszek: I think we should start simple, applying this only to connection sections is fine. Later on we can consider extending this to more than env vriables - for example reading values from external vaults.

Here are something that can inspire us https://github.com/apache/airflow/blob/bc1d68a6eb01919415c399d678f491e013eb9238/airflow/configuration.py#L622-L667

> You mean just before pass them to snowflake connector
Yes. Basically every time we access a config variable we should evaluate if it's not present in environment.

src/snowcli/cli/__init__.py
```diff
@@ -57,7 +57,7 @@ def login(
         print(f"Path to snowsql config does not exist: {snowsql_config_path}")
         raise typer.Abort()
 
-    cfg_snowsql = SnowsqlConfig(snowsql_config_path)
+    cfg_snowsql = ConnectionConfigs(snowsql_config_path)
```
sfc-gh-pjob, line 60: IMO we should change variable name (cfg_snowsql) because we changed class name.

src/snowcli/config.py
```diff
@@ -46,15 +46,13 @@ def save(self):
 def connect_to_snowflake(connection: Optional[str] = None, **overrides):  # type: ignore
     global snowflake_connection
     cfg = AppConfig()
-    snowsql_config = SnowsqlConfig(path=cfg.config.get("snowsql_config_path"))
+    snowsql_config = ConnectionConfigs(path=cfg.config.get("snowsql_config_path"))
```
sfc-gh-pjob, line 49: IMO we should change variable name (snowsql_config) because we changed class name.

src/snowcli/snow_connector.py
sfc-gh-pjob: I agree with you but I don't think your changes are making it worse. Do you think so?

src/snowcli/connection_config.py
```diff
@@ -1,16 +1,16 @@
...
 
-import configparser
 import os
 
+from configparser import ConfigParser
 
-class SnowsqlConfig:
+
+class ConnectionConfigs:
     def __init__(self, path="~/.snowsql/config"):
```
sfc-gh-pjob, line 9: I wonder if we shouldn't change also arg name from `path` to `snowsql_config_path`. I see it this way: 
* We have class ConnectionConfigs which holds configs of connections
* We create instances of this class using values from SnowSql config
* So we require snowsql config path in the constructor (not any path, but exactly snowsql config path)

IMO it would make the code more self-explanatory.

src/snowcli/connection_config.py
```diff
@@ -28,3 +28,12 @@ def add_connection(self, connection_name, entry):
 
         with open(os.path.expanduser(self.path), "w+") as f:
             self.config.write(f)
+
+    def _replace_with_env_variables(self):
+        connection_sections = [
```
sfc-gh-pjob, line 33: WDYT about moving this filtering outside of `_replace_with_env_variables` function? IMO the function shouldn't have to worry about which configs it should operate on. IMO it should just do what the name states -> replace env variables for "configs" known in the context of the class. IMO we should do filtering before calling `_replace_with_env_variables`. The class in named `ConnectionConfigs` so we want to keep only connection configs and `_replace_with_env_variables` doesn't seem to me to be a good selection to be responsible for this. 

WDYT?

src/snowcli/snowsql_config.py
sfc-gh-pjob: @sfc-gh-astus 
1. Ok. I left a few other separate comments related to this rename.
2. Ok.
3. Ok.

src/snowcli/snow_connector.py
sfc-gh-astus: Yes, but @sfc-gh-turbaszek PR will fix it :)

src/snowcli/cli/__init__.py
sfc-gh-astus: Right, changed

src/snowcli/config.py
sfc-gh-astus: changed

src/snowcli/connection_config.py
sfc-gh-astus: You are right, but this class is responsible for adding new connections too. We have to keep whole config, because we write everything into file. For now I would extract filter part into separate function and leave the rest as is.

sfc-gh-pjob: A few notes from our meeting (I decided to write them down to not forget about anything):
* We have to change the implementation because currently after invocation of "add connection" we will save resolved env values to the file instead of keeping env variables keys.
* We can fix it by moving env vars resolution to "get connection".
* But probably even better solution would be to move it to new method "get connections". The method can be responsible for:
    * Config sections filtering
    * Doing all necessary replacements
* "get connections" can be invoked outside "ConnectionConfigs" class to get info about all possible connections (currently 1 usage) - this will allow to remove one of underlying sqlconfig leaks
* Old "get connection" method can invoke "get connections" and search only for the one connection with the given name. IMO it's worth to think about making "get connection" returning optional value (to allow to remove the second sqlconfig leak)
    * Alternative solution to remove this leak is to have additional method "containsConnection" or sth similar which will check if there is a connection with the name
* When we will have all of it, IMO it should be possible to try to make underlying sqlconfig a private field (I don't know how it will work with tests)
* Sometime in the future, when we switch from dictionary to dataclass for single connection config representation, I hope we won't have any leaks of underlying sqlconfig and we'll be able to simply switch from one underlying storage to another if there will be such need
* Not everything above has to be done in this PR. As far as we fix the issue from the first point, other points can be moved to separate task.

sfc-gh-pjob: + It's worth to think about a test confirming that we fixed the issue mentioned above.

README.md
```diff
@@ -117,6 +117,27 @@ password = hunter2
...
+```
+
+will be replaced with environment variables:
+
+```ini
+SNOWCLI_ACCOUNT = "account"
+SNOWCLI_USER = "user"
+SNOWCLI_PASSWORD = "password"
+```
+
```
sfc-gh-astus, line 140: Does it make sense?

src/snowcli/cli/connection.py
```diff
@@ -29,15 +29,11 @@ def list():
...
-    for (connection_name, v) in cfg.config.items():
-        if connection_name.startswith("connections."):
-            connection_name = connection_name.replace("connections.", "")
-            if "account" in v:
-                table.add_row(connection_name, v["account"], v["username"])
-            if "accountname" in v:
-                table.add_row(connection_name, v["accountname"], v["username"])
+    for (connection_name, v) in connection_configs.get_connections().items():
+        connection_name = connection_name.replace("connections.", "")
+        table.add_row(connection_name, v["account"], v["user"])
```
sfc-gh-astus, line 36: Removed `ifs` because `get_connection` removes `name` from keys

src/snowcli/connection_config.py
```diff
@@ -0,0 +1,52 @@
...
+from snowcli.exception import EnvironmentVariableNotFoundError
+
+
+class ConnectionConfigs:
+    def __init__(self, snowsql_config_path="~/.snowsql/config"):
+        self.snowsql_config_path = snowsql_config_path
+        self._config = ConfigParser(inline_comment_prefixes="#", interpolation=None)
+        self._config.read(os.path.expanduser(snowsql_config_path))
+
+    def is_connection_exists(self, connection_name):
```
sfc-gh-turbaszek, line 16:
```suggestion
    def connection_exists(self, connection_name: str):
```
WDYT?

sfc-gh-astus: I'm for it, but we should decide to add types everywhere or not

tests/test_connection_config.py
```diff
@@ -0,0 +1,69 @@
...
+        "SNOWCLI_PASS": "password",
+        "SNOWCLI_HOST": "localhost",
+        "SNOWCLI_PORT": "8080",
+        "SNOWCLI_PROTOCOL": "http",
+    },
+)
+def test_missing_account_in_environment_variables():
+    result = ConnectionConfigs(CONFIG_PATH)
+
+    with pytest.raises(EnvironmentVariableNotFoundError):
```
sfc-gh-pjob, line 48: IMO we should use exceptions not related to ClickException only if sth really unexpected happens but I don't even have an idea what it can be. IMO such exceptions will be thrown by external libraries.

In all other cases I would try to choose one of exceptions from `ClickException` or create own subclass of it (but only if it will give any benefit). Using `ClickException` allows Click to handle an exception and convert it to a appropriate result code and nicely format the error message. 

See `test_sql_fails_for_both_stdin_and_other_query_source` for an example of test presenting the difference (exception isn't propagated outside the framework). You can also try to execute both cases on your own and see actual difference in output visible in your terminal.

WDYT?

src/snowcli/connection_config.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def _get_connection(self, connection_full_name: str):
+        connection = self._config[connection_full_name]
+        # Remap names to appropriate args in Python Connector API
+        connection = {
+            k.replace("name", ""): v.strip('"') for k, v in connection.items()
+        }
+        return self._replace_with_env_variables(connection)
+
+    def _replace_with_env_variables(self, connection_sections: dict):
```
sfc-gh-pjob, line 45: I think you forgot to change arg name (`connection_sections`) - now you don't have multiple sections here.

sfc-gh-pjob:
Shouldn't we write a summary of changes in PR description?

src/snowcli/connection_config.py
sfc-gh-astus: right

sfc-gh-astus: Changed to `connection_parameters`

tests/test_connection_config.py
sfc-gh-astus: I agree. `EnvironmentVariableNotFoundError` extends `ClickException` now

PR #164, opened 2023-06-02
Snow 821329 add missing unit tests for utils.py

sfc-gh-jsikorski:
### Pre-review checklist

### Changes description
Added unit tests for functions in utils.py


tests/unit/test_utils.py
```diff
@@ -0,0 +1,312 @@
+from pathlib import Path, PosixPath
```
sfc-gh-turbaszek, line 1: My personal preference would be to place this into `tests/` not `tests/unit` I think usually `tests` are considered units. WDYT? @sfc-gh-astus ?

tests/unit/test_utils.py
```diff
@@ -0,0 +1,312 @@
...
+    TEMP_DIR_FOR_APP_ZIP = "temp_dir"
+    TEMP_TEST_DIRECTORY = ".tests"
+
+    @pytest.mark.parametrize("argument", utils.YesNoAskOptions)
+    def test_yes_no_ask_callback_with_correct_argument(self, argument: str):
+        result = utils.yes_no_ask_callback(argument)
+
+        assert result == argument
+
+    @pytest.mark.parametrize("argument", bad_arguments_for_yesnoask)
```
sfc-gh-turbaszek, line 35: Do we use `bad_arguments_for_yesnoask` more than once? In tests my personal preference is to keep parametrize input inline so developers can easily see what's being tests. WDYT?

tests/unit/test_utils.py
```diff
@@ -0,0 +1,312 @@
...
+                req_file.writelines(req + "\n")
+        yield path
+
+    @pytest.fixture
+    def streamlit_requirements_txt(self, temp_test_directory: str) -> Generator:
+        path = os.path.join(temp_test_directory, self.REQUIREMENTS_SNOWFLAKE)
+        with open(path, "w") as dummy_file:
+            for req in requirements:
+                dummy_file.writelines(req + "\n")
+        yield path
```
sfc-gh-turbaszek, line 272: Can we abstract this part? It seems that it's repeated few times

tests/unit/test_utils.py
```diff
@@ -0,0 +1,312 @@
...
+    ):
+        utils.add_file_to_existing_zip(correct_app_zip, correct_requirements_txt)
+        zip_file = ZipFile(correct_app_zip)
+
+        assert os.path.basename(correct_requirements_txt) in zip_file.namelist()
+
+    def test_install_packages(self):
+        pass  # todo: add this
+
+    def test_recursive_zip_packages(
```
sfc-gh-turbaszek, line 184: I'm wondering if this should be done here or in a followup PR but I think we need to group those tests. `utils` is very ambiguous  and we may end up with single test file with 100s of test cases for unrelated functionalities

tests/unit/test_utils.py
sfc-gh-astus: In java projects we always separated tests into `test/unit` and `test/integration`. 

In @sfc-gh-turbaszek approach we will have something like this:
- config (folder for unit)
- integration (folder with integrations)
- output (folder with units)
- some unit .py

It will be a mess for me

sfc-gh-astus: I will move `tests/integration` to `tests-integration` and it will be fine :)

tests/unit/test_utils.py
sfc-gh-jsikorski: Probably you`re right. But my proposition is to leave it as it is - as currently we have to setup any structure for unit tests and refactor it, as the test-base grows

tests/unit/test_utils.py
sfc-gh-jsikorski: Idea behind this, was to separate all the test data from test, to make everything clearer. Probably, in this case it would be easier to understand if the list was passed directly, but on the other hand it would mean mixing both solutions - if we`re ok with that, i`ll pass the argument directly

tests/unit/test_utils.py
sfc-gh-jsikorski: I moved it to separate function

tests/unit/test_utils.py
sfc-gh-turbaszek: Mixing is fine with me

sfc-gh-turbaszek (approved):

src/snowcli/utils.py
```diff
@@ -405,6 +407,10 @@ def install_packages(
 
 
 def recursive_zip_packages_dir(pack_dir: str, dest_zip: str) -> bool:
+    # TODO:
+    # 1. why this is -> bool
+    # 2. Reformat it to with open...
+    # 3. Reformat it to get list od dirs and then iterate through it, adding files to zip
```
sfc-gh-astus, line 413: `get list of dirs`

tests/unit/test_utils.py
```diff
@@ -0,0 +1,336 @@
...
+        assert len(anaconda_packages.get("snowflake")) == 2
+
+    def test_anaconda_packages_streamlit(self):
+        packages.append("streamlit==1.2.3")
+        anaconda_packages = utils.parse_anaconda_packages(packages)
+        assert "snowflake" in anaconda_packages.keys()
+        assert "other" in anaconda_packages.keys()
+        assert "streamlit" not in anaconda_packages.get("other")
+        # TODO: above tests rely on correct response from anaconda.com
+        #  - shouldn`t it be mocked?
```
sfc-gh-astus, line 105: it should be mocked

tests/unit/test_utils.py
```diff
@@ -0,0 +1,336 @@
...
+        result = utils.generate_streamlit_package_wrapper(
+            "example_stage", "example_module", False
+        )
+
+        assert os.path.exists(result)
+        with open(result, "r") as f:
+            assert 'importlib.reload(sys.modules["example_module"])' in f.read()
+
+    def test_get_downloaded_package_names(self):
+        pass  # todo: prepare a fixture to test it
```
sfc-gh-astus, line 151: This is for next PR? :)

tests/unit/test_utils.py
```diff
@@ -0,0 +1,336 @@
...
+    def test_add_file_to_existing_zip(
+        self, correct_app_zip: str, correct_requirements_txt: str
+    ):
+        utils.add_file_to_existing_zip(correct_app_zip, correct_requirements_txt)
+        zip_file = ZipFile(correct_app_zip)
+
+        assert os.path.basename(correct_requirements_txt) in zip_file.namelist()
+
+    def test_install_packages(self):
+        pass  # todo: add this
```
sfc-gh-astus, line 187: and this? :)

tests/unit/test_utils.py
```diff
@@ -0,0 +1,336 @@
...
+            == expected_resource_dict
+        )
+
+    # Setup functions
+    # These functions are used to set up files and directories used in tests
+    # and delete them, after the tests are performed
+
+    @pytest.fixture
+    def temp_test_directory(self) -> Generator:
+        print(os.getcwd())
```
sfc-gh-astus, line 264: This print is needed?

tests/unit/test_utils.py
```diff
@@ -0,0 +1,336 @@
...
+        temp_test_directory: str,
+        correct_app_zip: str,
+        temp_directory_for_app_zip: str,
+    ):
+        result = utils.prepare_app_zip(correct_app_zip, temp_directory_for_app_zip)
+
+        assert result == temp_directory_for_app_zip + "/app.zip"
+
+    def test_prepare_app_zip_if_exception_is_raised_if_no_source(
+        self, temp_directory_for_app_zip
```
sfc-gh-astus, line 64: Why not `tmp_path`? https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html

pyproject.toml
```diff
@@ -39,6 +39,7 @@ dev = [
   "pytest==7.2.0",
   "syrupy==4.0.2",
   "tox==4.5.2",
+  "requests-mock==1.10.0"
```
sfc-gh-astus, line 42: Can we use it? 🤔

tests/unit/test_utils.py
sfc-gh-jsikorski: Done

tests/unit/test_utils.py
sfc-gh-jsikorski: No :)

pyproject.toml
sfc-gh-jsikorski: I don`t now, now i switched to unittest patch

tests/unit/test_utils.py
sfc-gh-jsikorski: tmp_path is designed for use inside single test. Every fixture we currently use is called 31(max) to 2(min) times - so there is no point. 
We could use tmp_path_factory, which is more suitable for fixtures, but introducing it would change exactly one line in every fixture code, and i can`t see any gain from it.

src/snowcli/cli/connection.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
```
sfc-gh-turbaszek, line 1: This is not neccessary 
```suggestion
```

tests/unit/test_data/test_data.py
```diff
@@ -0,0 +1,80 @@
+from requirements.requirement import Requirement
```
sfc-gh-turbaszek, line 1: Haven't we agreed to have
```
tests
tests_integrations
```

In such approach we don't need `unit` directory. WDYT?

src/snowcli/cli/connection.py
sfc-gh-jsikorski: Deleted

tests/unit/test_data/test_data.py
sfc-gh-jsikorski: Refactored

PR #165, opened 2023-06-02
Fix about path in pyproject.toml

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,7 @@ dev = [
 snow = "snowcli.cli:app"
 
 [tool.hatch.version]
-path = "src/snowcli/__about__.py"
+path = "src/__about__.py"
 
 [tool.hatch.build.targets.sdist]
 exclude = [


pyproject.toml
```diff
@@ -49,7 +49,7 @@ dev = [
 snow = "snowcli.cli:app"
 
 [tool.hatch.version]
-path = "src/snowcli/__about__.py"
+path = "src/__about__.py"
```
sfc-gh-pjob, line 52: There is no such file.
And the checks are failing.
---
This is the case why I think it's a good idea to make PR drafts firstly and then convert them to "ready to review".

sfc-gh-jsikorski: Am i missing something? As the path looked ok, as it was?

sfc-gh-turbaszek: Good catch, I was inspired by different fork

Issue #116, opened 2023-04-24
Return versions from native packages in installPackages


tvdboom:
Hi,

I am using the InstallPackages function in `/src/snowcli/utils.py` directly in Python (not through cli calls) to create a directory with all dependencies of a certain library (to later upload to a stage and use it in a snowpark udf). Ideally, I would use `perform_anaconda_check=True` and install the anaconda native libraries separately, but I noticed that the function only returns (second argument) the native library's names and not the exact version of those libraries. Getting only the name is not sufficient since I require an exact version to be sure the conda libraries will work properly with the rest of the (non-native) libraries. 

Would it be possible to refactor the function to return the versions together with the libraries, e.g. return `numpy==1.24.0` instead of just `numpy`.

sfc-gh-jhollan:
makes sense - or I wonder if ideally it could return a `dict` that has like "name" and "version" keys.

tvdboom:
That would also do. It does mean that the `parseAnacondaPackages` function in `utils.py` should also check if the exact required version is in the anaconda channel.

jamesweakley:
@tvdboom I opened a PR which partially covers this, but at the moment it doesn't check for the exact version in the anaconda channel. It should be possible to implement, but would require extra HTTP requests since the list of versions isn't in the main channeldata.json response (I think it's just got the latest version).

sfc-gh-turbaszek:
Done in #781

PR #167, opened 2023-06-05
Preserve version specifiers in requirements.txt

jamesweakley:
### Pre-review checklist

### Changes description
This change improves handling of requirements.txt when packaging for Snowpark.
Specifically, when requirements.txt is split into 'snowflake' and 'other' requirements, the version specifiers are preserved.

This feature was first requested in #116.

The key changes are:
- Instead of passing around List[str] of package names, we use the more complete Requirement class provided by the requirements-parser package
- Instead of a dict with 'snowflake' and 'other' packages, we use a new dataclass named SplitRequirements
- There is another new dataclass named RequirementWithFiles with associates a Requirement with the paths of locally downloaded files. This is used when pruning pip dependencies which were found on the Snowflake channel.

This means:
- `utils.parse_requirements` no longer returns a list of package names, instead it returns a list of Requirements
- `utils.parse_anaconda_packages` takes a list of Requirements and returns SplitRequirements
- `utils.install_packages` returns SplitRequirements for the second chance results (dependencies of downloaded pip packages found on Snowflake channel)
- `utils.get_downloaded_packages` returns a dict of package names to RequirementWithFiles so that within install_packages, downloaded dependencies and be pruned without losing version specifers.

*Note: Currently the version specifier is not used when checking Anaconda. While possible, this would involve more HTTP requests. Optimistically it's assumed Anaconda would have the full version history available*


src/snowcli/utils.py
```diff
@@ -89,62 +103,72 @@ def parse_requirements(requirements_file: str = "requirements.txt") -> list[str]
...
         if excluded_anaconda_deps is not None:
             print(f"""Excluded dependencies: {','.join(excluded_anaconda_deps)}""")
-            requirements = [
-                line for line in requirements if line not in excluded_anaconda_deps
+            snowflake_requirements = [
+                r
+                for r in snowflake_requirements
+                if r.name not in excluded_anaconda_deps
             ]
         # remove duplicates, remove comments, remove snowflake-connector-python
```
sfc-gh-turbaszek, line 162: Is this comment still true? Do we remove duplicates? Can you add tests for this?

jamesweakley: we do via the `list(set())`, but I should probably move that into parse_requirements so that callers don't need to.
I'll do that now and write a test to cover it.

jamesweakley: created `utils.deduplicate_and_sort_reqs` with test case

tests/test_data/anaconda_channel_data.json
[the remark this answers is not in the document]
jamesweakley: Haha, fair call. I didn't notice it was that large

sfc-gh-turbaszek (approved):

PR #183, opened 2023-06-13
Global exception handling with debug mode

sfc-gh-pjob:
### Pre-review checklist

### Changes description
* Added `OutsideTyperGlobalContext` responsible for holding the context visible outside Typer/Click command execution.
* Added `SnowCliMainTyper` extending `Typer`. It is responsible for adding global exception handling to `app()` call.
* This global exception handling is not tested because CliRunner used in tests runs commands on its own (it doesn't call `app()`).

### DIFF
--- a/src/snowcli/cli/main/outside_typer_global_context.py
+++ b/src/snowcli/cli/main/outside_typer_global_context.py
@@ -1,0 +1,6 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class OutsideTyperGlobalContext:
+    debug_logs_and_tracebacks: bool
--- a/src/snowcli/cli/main/snow_cli_main_typer.py
+++ b/src/snowcli/cli/main/snow_cli_main_typer.py
@@ -1,0 +1,31 @@
+import click
+from typer import Typer
+
+from snowcli.cli import DEFAULT_CONTEXT_SETTINGS, OutsideTyperGlobalContext
+
+
+class SnowCliMainTyper(Typer):
+    def __init__(self, outside_typer_global_context: OutsideTyperGlobalContext):
+        super().__init__(
+            context_settings=DEFAULT_CONTEXT_SETTINGS,
+            pretty_exceptions_show_locals=False,
+        )
+        self.outside_typer_global_context = outside_typer_global_context
+
+    def __handle_exception(self, exception: Exception):
+        debug = self.outside_typer_global_context.debug_logs_and_tracebacks
+        if debug:
+            raise exception from None
+        else:
+            click.echo(
+                "An unexpected exception occurred. Use --debug option to see a traceback.\n"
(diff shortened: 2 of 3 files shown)


src/snowcli/cli/main/outside_typer_global_context.py
```diff
@@ -0,0 +1,6 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class OutsideTyperGlobalContext:
+    debug_logs_and_tracebacks: bool
```
sfc-gh-astus, line 6: Do we need this class? Maybe just pass value in SnowCliMainTyper constructor?

sfc-gh-pjob: It's a wrapper for values which need to be accessible outside of Typer but which can be modified after Typer instance start (for example after reading `--debug` flag).

I'm not sure how I would be able to modify the value (after reading `--debug` flag) without using a wrapper.

src/snowcli/cli/main/snow_cli_main_typer.py
```diff
@@ -0,0 +1,31 @@
+import click
+from typer import Typer
+
+from snowcli.cli import DEFAULT_CONTEXT_SETTINGS, OutsideTyperGlobalContext
+
+
+class SnowCliMainTyper(Typer):
```
sfc-gh-turbaszek, line 7: Do we plan to have other than `Main` typers?

src/snowcli/cli/main/outside_typer_global_context.py
sfc-gh-turbaszek: I think you can use logic like this:
```
context = click.get_current_context()
debug = context.find_root().params.get("debug")
```
You can check if the flag was present or not, without need for additional state. WDYT? That's what we do for format flag

sfc-gh-pjob: I tried it before sending changes to review and it was throwing an exception - probably because being outside typer command execution.

src/snowcli/cli/main/snow_cli_main_typer.py
sfc-gh-pjob: I imagine we can have one main typer (entrypoint) and other typers (added by typer.add_typer) for purpose of command groups.

Main typer may be responsible for global exception handling (like here) but also for top-level settings (like these in the constructor).

sfc-gh-pjob:
I'm closing this PR because I created a new one with better exception handling: https://github.com/Snowflake-Labs/snowcli/pull/191

sfc-gh-pjob:
We decided to return to the idea presented in this PR but we made it in a new PR: https://github.com/Snowflake-Labs/snowcli/pull/192

PR #184, opened 2023-06-13
Decouple snowcli and snowsql configuration

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
Refactor snowcli configuration to use snowflake-connector-python configuration module. In this way snowcli can be configured using global config file.


src/snowcli/connection_config.py
```diff
@@ -1,37 +1,30 @@
...
 
     def connection_exists(self, connection_name: str):
         return self._config.sections().__contains__(f"connections.{connection_name}")
 
     def get_connections(self):
-        connections_names = [
-            n for n in self._config.sections() if n.startswith("connections.")
-        ]
-        return {n: self._get_connection(n) for n in connections_names}
+        return self._config["connections"]
```
sfc-gh-astus, line 19: This will return KeyError. In config we have 

```
{
   "connection.a": {},
   "connection.b": {},
   "other_sections": {}
}
```

sfc-gh-turbaszek: It will not because the config structure is changed

src/snowcli/connection_config.py
```diff
@@ -1,37 +1,30 @@
...
-        self.snowsql_config_path = snowsql_config_path
-        self._config = ConfigParser(inline_comment_prefixes="#", interpolation=None)
-        p = Path(snowsql_config_path).absolute().expanduser()
-        self._config.read(p)
+    def __init__(self):
+        self._config = CONFIG_PARSER
+        self._config.read_config()
 
     def connection_exists(self, connection_name: str):
         return self._config.sections().__contains__(f"connections.{connection_name}")
```
sfc-gh-astus, line 16: `self.get_connections().__contains__(connection_name)` ?

src/snowcli/cli/connection.py
```diff
@@ -60,17 +52,13 @@ def add(
...
+    connection_configs = ConnectionConfigs()
     connection_entry = {
         "account": account,
         "username": username,
         "password": password,
     }
     connection_configs.add_connection(connection, connection_entry)
     print(
-        f"Wrote new connection {connection} to {connection_configs.snowsql_config_path}"
+        f"Wrote new connection {connection} to {connection_configs._config.file_path}"
```
sfc-gh-astus, line 63: Maybe add method get_file_path?

src/snowcli/connection_config.py
```diff
@@ -1,37 +1,30 @@
...
 import os
-from configparser import ConfigParser
-from pathlib import Path
+
+from snowflake.connector.config_manager import CONFIG_PARSER
+
 from snowcli.exception import EnvironmentVariableNotFoundError
 
 
 class ConnectionConfigs:
```
sfc-gh-astus, line 10: Maybe `SnowCliConfig`?

src/snowcli/utils.py
```diff
@@ -567,23 +566,7 @@ def convert_resource_details_to_dict(function_details: list[tuple]) -> dict:
...
-            if config_section in app_config:
-                ctx.default_map.update(
-                    app_config.get(config_section),
-                )  # type: ignore
-            if value in app_config:
-                # TODO: Merge the config dict into default_map
-                # type: ignore
-                ctx.default_map.update(app_config.get(value))
-        except Exception as ex:
-            raise typer.BadParameter(str(ex))
```
sfc-gh-astus, line 586: Why you removed this code?

src/snowcli/config.py
```diff
@@ -1,65 +1,30 @@
...
-        snowsql_config_path=cfg.config.get("snowsql_config_path")
-    )
 
+def connect_to_snowflake(connection_name: str, **overrides):  # type: ignore
     # If there's no user-provided connection then read
     # the one specified by configuration file
-    connection_name = connection or cfg.config.get("snowsql_connection_name")
-    connection_config = connection_configs.get_connection(connection_name)
-    snowflake_connection = SnowflakeConnector(connection_config, overrides=overrides)
+    connection_name = connection_name  # or cfg.config.get("snowsql_connection_name")
```
sfc-gh-astus, line 19: Maybe just remove this line?

src/snowcli/utils.py
sfc-gh-turbaszek: Honestly, I don't know what this code is doing. I don't think we use the click context anywhere. WDYT?

src/snowcli/cli/sql.py
```diff
@@ -64,7 +64,7 @@ def execute_sql(
         readable=True,
         help="File to execute.",
     ),
-    connection: Optional[str] = typer.Option(
+    connection: str = typer.Option(
```
sfc-gh-pjob, line 67: I don't think I understand this change, especially seeing changes in `test_sql_overrides_connection_configuration`. Could you explain me that change?

src/snowcli/cli/__init__.py
```diff
@@ -33,101 +31,6 @@ def version_callback(value: bool):
         raise typer.Exit()
 
 
-@app.command()
-def login(
```
sfc-gh-pjob, line 37: 1. Do I correctly understand that now we won't have to login and configure connection and instead of this we will have to specify connection for each command using it if we don't want to use "default" connection?
2. If that's true, then IMO we should update README.

tests/test_connection.py
```diff
@@ -0,0 +1,57 @@
+from tempfile import NamedTemporaryFile
+from textwrap import dedent
+
```
sfc-gh-pjob, line 3: What about environment variables in config? Do we still support them? Don't we need tests for it (maybe they are tested in python connector project?)?

If we still support them, then is their syntax still the same? Won't changes in README necessary?

sfc-gh-pjob:
Please go through PR checklist.

src/snowcli/cli/__init__.py
sfc-gh-turbaszek: That's correct. I will update the README

tests/test_connection.py
sfc-gh-turbaszek: Those are supported by snowflek-connector-python implementation. So we basically get this for free.

src/snowcli/utils.py
sfc-gh-astus: Right, let's remove it

pyproject.toml
```diff
@@ -17,8 +17,9 @@ dependencies = [
   "rich==13.4.2",
   "requests==2.28.1",
   "requirements-parser==0.5.0",
-  "snowflake-connector-python==3.0.4",
-  "toml==0.10.2",
+#  "snowflake-connector-python==3.0.4",
```
sfc-gh-astus, line 20: commented

src/snowcli/cli/sql.py
sfc-gh-turbaszek: Good catch, there's not explicite need for connection if we have a default one and overrides.

pyproject.toml
sfc-gh-turbaszek: That's deliberate, we need to uncomment this asap

src/snowcli/cli/sql.py
sfc-gh-astus: There is any reason why you removed optional only for one parameter and left for the others?

src/snowcli/config.py
```diff
@@ -1,67 +1,123 @@
...
+    def __init__(self, file_path: Path = CONFIG_FILE):
+        super().__init__(name="SNOWCLI_PARSER", file_path=file_path)
+        self._add_options()
+
+    def from_context(self, config_path_override: Path):
+        self.file_path = config_path_override
+        if not self.file_path.exists():
+            self.initialize_connection_section()
+            self._dump_config()
+            print(f"Created Snowflake configuration file at {cli_config.file_path}")
```
sfc-gh-astus, line 32: Should be log

src/snowcli/config.py
```diff
@@ -1,67 +1,123 @@
...
+        while idx < len(path):
+            section = section[path[idx]]
+            idx += 1
+        return section
 
-        if p == p.home():
-            return None
-        else:
-            return next(p.glob("app.toml"))
+    def has_section(self, *path) -> bool:
```
sfc-gh-astus, line 53: `section_exists`? To be consistent

sfc-gh-astus (approved):

sfc-gh-pjob (approved):

PR #191, opened 2023-06-20
SnowCliTyper with common logic of exception handling + SnowCliGlobalContext

sfc-gh-pjob:
…gic common for all commands and callbacks. Added SnowCliGlobalContext.

### Pre-review checklist

### Changes description
...


src/snowcli/cli/__init__.py
```diff
@@ -16,17 +16,30 @@
...
 from snowcli.connection_config import ConnectionConfigs
 
-app = typer.Typer(
+app = SnowCliTyper(
     context_settings=DEFAULT_CONTEXT_SETTINGS,
     pretty_exceptions_show_locals=False,
 )
 
 
+def setup_global_context(debug: bool):
```
sfc-gh-turbaszek, line 35: Can you please add short docstring about what this function does?

src/snowcli/cli/common/snow_cli_global_context.py
```diff
@@ -0,0 +1,41 @@
...
+class LoggingAndExceptionHandlingContext:
+    enable_tracebacks: bool
+
+
+@dataclass
+class SnowCliGlobalContext:
+    logging_and_exception_handling: LoggingAndExceptionHandlingContext
+
+
+class SnowCliGlobalContextHolder:
```
sfc-gh-turbaszek, line 16:
```suggestion
class SnowCliGlobalContextManager:
```
WDYT? I've never meet `holder` concept, is this a common practice?

src/snowcli/cli/common/snow_cli_typer.py
```diff
@@ -0,0 +1,227 @@
...
+    Default,
+)
+
+from snowcli.cli.common.snow_cli_global_context import snow_cli_global_context_holder
+
+
+def _are_tracebacks_enabled() -> bool:
+    return (
+        snow_cli_global_context_holder.get_global_context_copy().logging_and_exception_handling.enable_tracebacks
+    )
```
sfc-gh-turbaszek, line 20: Should this be a method of the holder class?

src/snowcli/cli/common/snow_cli_typer.py
```diff
@@ -0,0 +1,227 @@
...
+            raise exception from None
+        else:
+            raise ClickException(
+                "An unexpected exception occurred:\n"
+                + f"{exception.__class__.__name__}: {exception}\n"
+                + "Use --debug option to see the full traceback."
+            ) from None
+
+
+class SnowCliTyper(Typer):
```
sfc-gh-turbaszek, line 39: Can you document the class? Why we introduce it? How it differes from parent? Should we add also docs for the methods?

src/snowcli/cli/common/snow_cli_typer.py
```diff
@@ -0,0 +1,227 @@
...
+            deprecated=deprecated,
+            add_completion=add_completion,
+            rich_markup_mode=rich_markup_mode,
+            rich_help_panel=rich_help_panel,
+            pretty_exceptions_enable=pretty_exceptions_enable,
+            pretty_exceptions_show_locals=pretty_exceptions_show_locals,
+            pretty_exceptions_short=pretty_exceptions_short,
+        )
+
+    def callback(
```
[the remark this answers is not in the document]
sfc-gh-pjob: `I would consider using *args, **kwargs instead of all parameters to indicate that we simply pass them to super method.`

I thought about it but doesn't it break type hints?

src/snowcli/cli/common/snow_cli_typer.py
sfc-gh-pjob: I'm not sure about it.

I'd like to avoid adding more responsibilities to holder/manager. Let's have it only to provide and keep the context.

src/snowcli/cli/common/snow_cli_global_context.py
sfc-gh-pjob: Manager is a good name. I'll change it.

src/snowcli/cli/common/snow_cli_typer.py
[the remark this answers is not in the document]
sfc-gh-pjob: `Do we need all methods?`

I overridden 3 methods which in my opinion we need to override.
* command - to handle exceptions in commands
* callback - to handle exceptions before commands (in callbacks)
* add_typer - to provide common logic (including exception handling) to all typer instances (command subgroups) and forbid using simple `Typer`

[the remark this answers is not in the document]
sfc-gh-pjob: `How we plan to maintain compatibility with typer/click?`

Do you think adjusting our interface on Typer's interface change will be a problem? I hope it won't happen often.

src/snowcli/cli/common/snow_cli_typer.py
sfc-gh-pjob: Ok, no problem.

Just one question. I believe the questions are only the topic of documentation, they are not your questions to me, right?

src/snowcli/cli/__init__.py
sfc-gh-pjob: Ok, I'll do it.

src/snowcli/cli/common/snow_cli_typer.py
sfc-gh-turbaszek: With renaming to manager we get a good argument to move the method there. In future we may have more `is_something` methods and having them in one class may be better than multiple loose functions. WDYT? + @sfc-gh-astus @sfc-gh-jsikorski

sfc-gh-pjob: If we go this way, then we should also avoid calling `get_...` and `update_...` directly and wrap all reads and updates in helper methods.

sfc-gh-astus: I agree with @sfc-gh-turbaszek, this function fits better in manager

src/snowcli/cli/common/snow_cli_global_context.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+@dataclass
+class LoggingAndExceptionHandlingContext:
+    enable_tracebacks: bool
+
+
+@dataclass
+class SnowCliGlobalContext:
+    logging_and_exception_handling: LoggingAndExceptionHandlingContext
```
sfc-gh-astus, line 13: We will have some more fields soon? Because double wrapping boolean is overkill now

src/snowcli/cli/common/snow_cli_typer.py
sfc-gh-pjob: I don't understand why is that.

I created this function only to make the code more readable (to avoid long method execution chain).
I don't see any advantage of duplicating all fields stored in global context to manager's methods.

src/snowcli/cli/common/snow_cli_global_context.py
sfc-gh-pjob: I can agree to flatten the structure, however I don't think it's overkill. It introduces some convention from the very beginning.

src/snowcli/cli/common/snow_cli_typer.py
sfc-gh-astus: 1. Because this module has to know global context structure
2. `bool` is immutable so it can be return by manager and it won't be necessary to deepcopy whole context to get one value

src/snowcli/cli/common/snow_cli_global_context.py
sfc-gh-turbaszek: I think this is possible over-optimisation. I would +1 simpler, flat context. It will be easier to reason about what's in there.

src/snowcli/cli/__init__.py
sfc-gh-pjob: As we agreed to not use the solution from this PR, I made changes from this comment in new PR: https://github.com/Snowflake-Labs/snowcli/pull/192

src/snowcli/cli/common/snow_cli_global_context.py
sfc-gh-pjob: As we agreed to not use the solution from this PR, I made changes from this comment in new PR: https://github.com/Snowflake-Labs/snowcli/pull/192

src/snowcli/cli/common/snow_cli_typer.py
sfc-gh-pjob: As we agreed to not use the solution from this PR, I made changes from this comment in new PR: https://github.com/Snowflake-Labs/snowcli/pull/192

src/snowcli/cli/common/snow_cli_typer.py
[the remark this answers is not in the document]
sfc-gh-pjob: We agreed to not use the idea of overriding Typer's interface as it can become problematic in the future and as the result we returned to previous solution: https://github.com/Snowflake-Labs/snowcli/pull/192

src/snowcli/cli/common/snow_cli_typer.py
sfc-gh-pjob: As we agreed to not use the solution from this PR, I made changes from this comment in new PR: https://github.com/Snowflake-Labs/snowcli/pull/192

src/snowcli/cli/common/snow_cli_global_context.py
sfc-gh-pjob: As we agreed to not use the solution from this PR, I made changes from this comment in new PR: https://github.com/Snowflake-Labs/snowcli/pull/192

sfc-gh-pjob:
We agreed to not use the solution from this PR and we returned to previous solution, but in new PR: https://github.com/Snowflake-Labs/snowcli/pull/192

I'm closing this PR.

PR #193, opened 2023-06-21
Add integration with internal Jira

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
...

### DIFF
--- a/.github/workflows/jira_issue.yml
+++ b/.github/workflows/jira_issue.yml
@@ -1,0 +1,50 @@
+name: Jira creation
+
+on:
+  issues:
+    types: [opened]
+  issue_comment:
+    types: [created]
+
+jobs:
+  create-issue:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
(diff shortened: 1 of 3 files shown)


.github/workflows/jira_issue.yml
```diff
@@ -0,0 +1,50 @@
...
+    types: [opened]
+  issue_comment:
+    types: [created]
+
+jobs:
+  create-issue:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    if: ((github.event_name == 'issue_comment' && github.event.comment.body == 'recreate jira' && github.event.comment.user.login == 'sfc-gh-turbaszek') || (github.event_name == 'issues' && github.event.pull_request.user.login != 'whitesource-for-github-com[bot]'))
```
sfc-gh-pjob, line 14: Does it mean that only you will be able to recreate jira?
It's not the problem IMO, but I'm just asking if it is behaviour which you wanted?

sfc-gh-turbaszek: Updated to list of current codewowners

sfc-gh-pjob (approved):

PR #194, opened 2023-06-22
Add missing unit tests for package.py

sfc-gh-jsikorski:
### Pre-review checklist
   * [x ] I've described my changes in the section below.

### Changes description
Added unit tests for package.py


tests/test_package.py
```diff
@@ -0,0 +1,68 @@
...
+
+import pytest
+
+from snowcli.cli.snowpark import package
+
+from tests.test_data import test_data
+
+
+class TestPackage:
+    LOGGER = logging.getLogger(__name__)
```
sfc-gh-astus, line 16: it is not needed

sfc-gh-jsikorski: You`re right. It`s a sole survivor of a different approach :D

tests/test_package.py
```diff
@@ -0,0 +1,68 @@
...
+        mock_response.json.return_value = test_data.anaconda_response
+        mock_requests.get.return_value = mock_response
+
+        monkeypatch.setattr("sys.stdin", io.StringIO("N"))
+
+        with caplog.at_level(logging.DEBUG, logger=argument[2]):
+            result = package.package_lookup(argument[0], install_packages=True)
+
+        assert caplog.text
+        assert argument[1] in caplog.text
```
sfc-gh-astus, line 46: assert logs is the only way to test it?

sfc-gh-jsikorski: It felt, that functions in package.py rely on so many outside connections (anaconda, pip, etc.), that it makes more sense to test them in integration tests. But i'm open to new test case propositions.

tests/test_package.py
```diff
@@ -0,0 +1,66 @@
...
+            ),
+            (
+                "some-weird-package-we-dont-know",
+                "not found in Snowflake anaconda channel...",
+                "snowcli.utils",
+            ),
+        ],
+    )
+    @patch("tests.test_package.package.utils.requests")
+    def test_package_lookup(self, mock_requests, caplog, argument, monkeypatch):
```
sfc-gh-astus, line 32: Lookup function has a lot of ifs, these 2 cases cover all of them?

tests/test_package.py
sfc-gh-astus: Don't see other way ☹️

tests/test_package.py
```diff
@@ -0,0 +1,66 @@
...
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from snowcli.cli.snowpark import package
+
+from tests.test_data import test_data
+
+
+class TestPackage:
```
sfc-gh-astus, line 15: Tests for `package_upload` are missing

sfc-gh-jsikorski: Yes, `package_upload` return type is None, and it`s purpose is in side-effect of uploading the zipped package to Snowflake.
Of course, i could mock every stage, but the test would be really checking if mocks return correct values - therefore i think it`s rather a candidate for an. integration test

sfc-gh-astus (approved):

PR #202, opened 2023-06-23
Fix config path and update tests

sfc-gh-zblackwood:
### Pre-review checklist

### Changes description
I was getting this error when trying to do any snow command
```
Exception message:
'connections.snowhouse'
```

I finally discovered that the config lookup was not correctly expanding the relative path, and was instead creating a path like this 
```
/Users/zblackwood/projects/snowcli/~/.snowsql/config
```

Instead of 
```
/Users/zblackwood/.snowsql/config
```

I fixed this by switching the order of `absolute` and `expanduser`, and added a test to check

### DIFF
--- a/src/snowcli/connection_config.py
+++ b/src/snowcli/connection_config.py
@@ -3,14 +3,19 @@ from __future__ import annotations
 import os
 from configparser import ConfigParser
 from pathlib import Path
+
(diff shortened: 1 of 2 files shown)


sfc-gh-turbaszek:
This change is no longer valid after change of configuration approach

sfc-gh-zblackwood:
No longer applicable

Issue #177, opened 2023-06-12
SNOW-847301: `standard_zip_dir()` tries to zip its own output directory


lewinfox:
Thanks for this awesome tool!

I ran into an issue today trying to create a function; `standard_zip_dir()` creates its output zip file before building the list of files to zip. This appears (for me at least) to cause it to recurse until the zip file size limit is hit.

I propose excluding `dest_zip` when zipping:

![image](https://github.com/Snowflake-Labs/snowcli/assets/25552362/8403405e-33c9-4248-ac62-9aee62e9dee3)

Happy to submit a PR if that's helpful.

sfc-gh-jsikorski:
Hi, sorry for the delay - later in the day we will file a PR solving this, along with some refactors

sfc-gh-jsikorski:
recreate jira

PR #204, opened 2023-06-26
Refactor of zipping functions

sfc-gh-jsikorski:
### Pre-review checklist

### Changes description
Fix for #177 with refactoring of zip functions.
I separated tasks of creating a list of files to pack, and creating an archive, also moving duplicated code to separate function


src/snowcli/utils.py
```diff
@@ -478,66 +477,59 @@ def install_packages(
...
-                    ):
-                        zipf.write(file, arcname=os.path.relpath(file, directory))
-
-    # close the zip file object
-    zipf.close()
+    files_to_pack = get_list_of_files_to_pack(pack_dir, True)
+
+    with ZipFile(dest_zip, "w", ZIP_DEFLATED, allowZip64=True) as package_zip:
+        for file in files_to_pack:
+            package_zip.write(file[0], arcname=os.path.relpath(file[0], file[1]))
```
sfc-gh-turbaszek, line 484: Do we need to use `file[0]` ? Can't we use `file.name` or something to be more explicite?

src/snowcli/utils.py
```diff
@@ -478,66 +477,59 @@ def install_packages(
...
+
+    files: List[Tuple[Path, Optional[str]]] = []
+
+    def filenames_filter(file: Path) -> bool:
+        return (
+            not file.name.startswith(".")
+            and not str(file).startswith(".")
+            and not file.match("*.pyc")
+            and not file.match("*__pycache__*")
+            and file not in map(lambda x: x[0], files)
```
sfc-gh-turbaszek, line 509:
```suggestion
            and file not in map(lambda x: x[0], excluded_files)
```
WDYT? Also, `[f[0] for f in files]` would be probably more pythonic

src/snowcli/utils.py
sfc-gh-jsikorski: I added File dataclass - now it's file.name

src/snowcli/utils.py
sfc-gh-turbaszek: MAybe we can use `set` instead of list? It automatically resolves the duplicates. WDYT?

sfc-gh-jsikorski: I thought about it, but, when added in different steps, Files can have the same name and different relpath - therefore a simple change from list to set would potentially leave some duplicates.
And, as we`re already checking the filename, i thought adding the check there would the simplest solution.

src/snowcli/utils.py
```diff
@@ -478,66 +477,67 @@ def install_packages(
...
-
-    # close the zip file object
-    zipf.close()
+    files_to_pack = get_list_of_files_to_pack(pack_dir, True)
+
+    with ZipFile(dest_zip, "w", ZIP_DEFLATED, allowZip64=True) as package_zip:
+        for file in files_to_pack:
+            package_zip.write(
+                file.name, arcname=os.path.relpath(file.name, file.relpath)
+            )
```
sfc-gh-astus, line 486: Great candidate to extract it to separate function. In `standard_zip_dir` you have the same code

sfc-gh-jsikorski: Good catch - created separate function

PR #222, opened 2023-07-03
Environment variables in integration tests adapted to new configuration

sfc-gh-astus:
### Pre-review checklist

### Changes description
Changed getting env variables
Adapted integration test configuration
Adapted snowpark connection
Using default ("dev") connection in integration tests

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,6 +31,30 @@ To execute only unit tests run `pytest`.
 
 To execute only integration tests run `pytest -m integration`.
 
+### Connection parameters for integration tests
+
+Parameters have to be set in environment parameters and must use the following format:
+
+``SNOWFLAKE_CONNECTIONS_INTEGRATION_<key>=<value>``
+
+where ``<key>`` is the name of the key
+
+For example: SNOWFLAKE_CONNECTIONS_INTEGRATION_ACCOUNT="my-account"
+
+List of required parameter keys:
+- account
+- user
+- password
+- host
+- port
+- protocol
+
+Optional parameter keys:
+- db
+- schema
+- warehouse
+- role
+
 ## Remote debugging with PyCharm or IntelliJ
 
 Snowflake CLI can connect to a remote debug server started in PyCharm or Intellij.
--- a/src/snowcli/config.py
+++ b/src/snowcli/config.py
@@ -2,14 +2,16 @@ from __future__ import annotations
 
 import os
 from pathlib import Path
-from typing import Optional, Any
+from typing import Optional, Any, Dict
 
 import tomlkit
 from tomlkit import dump, table, TOMLDocument
 from tomlkit.exceptions import NonExistentKey
+from tomlkit.items import Table
+from tomlkit.container import Container
 import logging
 
-from snowcli.exception import MissingConfiguration
+from snowcli.exception import MissingConfiguration, UnsupportedConfigSectionTypeError
 from snowflake.connector.constants import CONFIG_FILE
 from snowflake.connector.config_manager import ConfigManager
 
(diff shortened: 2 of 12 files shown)


tests_integration/snowflake_connector.py
```diff
@@ -1,21 +1,42 @@
...
+    config = {k: v for k, v in config.items() if v is not None}
     connection = connector.connect(**config)
     yield connection
     connection.close()
+
+
+def _get_from_env(parameter_name: str, required: bool) -> str | None:
+    try:
+        return os.environ[f"{_ENV_PARAMETER_PREFIX}{parameter_name}"]
+    except KeyError:
```
sfc-gh-jsikorski, line 36: Maybe, instead of catching errors, we could do os.environ.get() - and when the result is None, check if variable is required?

sfc-gh-astus: Good idea, changed 👍

src/snowcli/config.py
```diff
@@ -54,17 +54,20 @@ def section_exists(self, *path) -> bool:
         except NonExistentKey:
             return False
 
-    def _get_from_env(self, *path, key: str):
-        env_variable_name = (
-            "SNOWFLAKE_" + "_".join(p.upper() for p in path) + f"_{key.upper()}"
+    @staticmethod
+    def _get_from_env(*path, connection_name: str) -> dict:
```
sfc-gh-pjob, line 58:
```suggestion
    def _get_env_variables_for_section(*path, section_name: str) -> dict:
```

src/snowcli/config.py
```diff
@@ -54,17 +54,20 @@ def section_exists(self, *path) -> bool:
...
+            + f"_{connection_name.upper()}"
         )
-        return os.environ.get(env_variable_name)
+        return {
+            k.replace(f"{env_variable_prefix}_", "").lower(): os.environ[k]
+            for k in os.environ.keys()
+            if k.startswith(env_variable_prefix)
+        }
 
     def get(self, *path, key: str, default: Optional[Any] = None) -> Any:
```
sfc-gh-pjob, line 70:
```suggestion
    def get_from_config_file(self, *path, key: str, default: Optional[Any] = None) -> Any:
```

sfc-gh-pjob: I think we should also consider similar change for `get_section` and `section_exists`.

sfc-gh-turbaszek (approved):

src/snowcli/config.py
```diff
@@ -39,63 +84,40 @@ def _add_options(self):
...
 
-    def section_exists(self, *path) -> bool:
-        try:
-            self.get_section(*path)
-            return True
-        except NonExistentKey:
-            return False
+    def _merge_section_with_env(self, section: Table, *path) -> Dict[str, str]:
+        merged_dict = {}
+        for k in section:
```
sfc-gh-pjob, line 97: Does it mean all env variables have to be also specified in toml? For example if I want to set password in env variable, then I have to provide some dummy password in toml?

src/snowcli/exception.py
```diff
@@ -18,3 +18,8 @@ def format_message(self):
 class SnowflakeConnectionError(ClickException):
     def __init__(self, snowflake_err: Exception):
         super().__init__(f"Could not connect to Snowflake. Reason: {snowflake_err}")
+
+
+class UnsupportedConfigSectionType(ClickException):
```
sfc-gh-pjob, line 23: IMO we should extend ClickException only for expected exceptions - like connection error, invalid configuration etc.

This new exception can be thrown only in case of a bug (in our or 3rd party plugin's code) and I'm not sure if we should hide traceback for it in `--debug` mode (ClickExceptions are handled by click and they are not propagated further, so we can't get traceback).

tests/test_config.py
```diff
@@ -19,25 +17,94 @@ def test_empty_config_file_is_created_if_not_present():
...
+    },
+    clear=True,
+)
 def test_environment_variables_works_if_config_value_not_present(test_snowcli_config):
     cm = CliConfigManager(file_path=test_snowcli_config)
     cm.read_config()
 
-    with pytest.raises(Exception):
-        assert cm.get("connections", "2o3nei23ne2oixn", key="warehouse") == "xs"
+    assert cm.get_connection("empty") == {
```
sfc-gh-pjob, line 71: this test doesn't check what its name tells

src/snowcli/config.py
sfc-gh-astus: Yes

sfc-gh-pjob: Don't we remove some part of functionality which we had before (or thought we had)?: https://github.com/Snowflake-Labs/snowcli#use-environment-variables-for-snowflake-credentials

src/snowcli/config.py
```diff
@@ -39,63 +84,40 @@ def _add_options(self):
...
+            value = self._get_from_env(*path, key=k)
+            if value:
+                merged_dict[k] = value
+            else:
+                merged_dict[k] = section[k]
+        return merged_dict
```
sfc-gh-jsikorski, line 103: Maybe i'm missing something, but what happens if user won't put the required key in toml? In integration tests config, there's a dummy config, that holds required values.
So, if there's no "account" field in the config file, do we catch it during config parsing, or do we rely on python connector's messages?

CONTRIBUTING.md
```diff
@@ -31,6 +31,30 @@ To execute only unit tests run `pytest`.
...
+where ``<key>`` is the name of the key
+
+For example: SNOWFLAKE_CONNECTIONS_INTEGRATION_ACCOUNT="my-account"
+
+List of required parameter keys:
+- account
+- user
+- password
+- host
+- port
```
sfc-gh-jsikorski, line 49: Is port really required? I think we can get a connection without it currently

sfc-gh-astus: It is needed for local dev

tests/test_config.py
sfc-gh-astus: fixed

src/snowcli/config.py
sfc-gh-astus: I returned to previous solution

src/snowcli/config.py
sfc-gh-astus: Paweł asked same question https://github.com/Snowflake-Labs/snowcli/pull/222#discussion_r1260937675

sfc-gh-jsikorski (approved):

sfc-gh-turbaszek (approved):

PR #236, opened 2023-07-13
Nix flake: init

sfc-gh-vtimofeenko:
### Pre-review checklist
   * [X] I've added new automated tests to verify correctness of my new code.

### Changes description

This commit adds a `contrib/nix` directory which contains a Nix flake that allows a user to run snowcli using a [Nix flake](https://nixos.wiki/wiki/Flakes).


sfc-gh-vtimofeenko:
Fixed the whitespace and force-pushed

sfc-gh-vtimofeenko:
Will add a GH action to check `nix build`.

sfc-gh-vtimofeenko:
I added a Github action that builds snowcli using Nix <=> there are changes in `contrib/nix` directory. Since Nix pins all dependencies in a single lock file, there is no need to test that build unless the definitions themselves change.

One note about the created action: the package is defined for 4 architectures: 

```nix
      systems = [ "x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin" ];
```

The defined action only tests one of Linux variants. On Linux the checks run for ~a minute. MacOS could be longer and I don't want the action to waste compute if testing other systems is not deemed necessary.

sfc-gh-pjob:
Hello, thank you very much for your involvement in the form of this PR. Thanks for fulfilling all the checks from the checklist and including the workflow testing SnowCLI build with Nix.

At first sight the PR looks good but we'd like to look at it in more details and unfortunatelly it will take us at least a few days because none of us uses Nix, so we have to find some time to read a little bit about it, think about it and then finally review the PR. I think you shouldn't run into any merge conflicts because almost all of your changes are in a new directory.

sfc-gh-pjob:
> Understandable. Let me know if it would be helpful if I were to |provide some materials for the background.
>
> Just in case -- I will add some comments to the files in this PR.

I think additional comments will be enough at the moment. Thanks!\
If we meet troubles with understanding the PR, we will simply ask for more explanation :)

> I would prefer rebase my branch and bump the nix dependencies before merging though

👍

sfc-gh-vtimofeenko:
> I think additional comments will be enough at the moment

Done. Probably the only somewhat non-standard component is flake-parts, but it's there just to reduce the boilerplate

sfc-gh-bwarsaw:
Can we have a conversation about this?  If nix is an optional addition, it might be fine, but I am looking at a holistic approach to client-side Python, and want to use more native tooling for our Python code.  E.g. [PDM](https://pdm.fming.dev/latest/) is the native package manager I want to use.

sfc-gh-vtimofeenko:
In that sense Nix would be an optional addition.

If the package were to be migrated over to PDM, I could then rewrite the Nix build scripts to use PDM without changing any interfaces that Nix users use -- i.e. `nix run github:Snowflake-Labs/snowcli` would behave exactly the same way from the user's perspective

sfc-gh-vtimofeenko:
Hi! I have rebased my code onto `main` and pinned the package that Nix installs at v1.1.0, all checks pass:

https://github.com/sfc-gh-vtimofeenko/snowcli/actions

sfc-gh-vtimofeenko:
This PR is probably too out of date at this point. 2.x version is maintained [here](https://github.com/sfc-gh-vtimofeenko/snowcli-nix-flake).

PR #245, opened 2023-07-17
[SNOW-851220] Integration tests of snowpark function commands

sfc-gh-pjob:
### Pre-review checklist

### Changes description
* Fixed `snow snowpark function update` command execution.
* Added integration test for all commands from `snowpark function` group.
* Added some test utils in `tests_integration/testing_utils` directory.
* Changed temporary integration test database prefix from `db_` to `snowcli_integration_tests_db_` to make our trash be easier to identify.
* Fixed imports in `test_sql.py` and `test_warehouse.py`.
* Added `tests_integration` directory to the list of exclusions for `trailing-whitespace` hook.


tests_integration/conftest.py
```diff
@@ -37,6 +37,12 @@ def invoke(self, *a, **kw):
         return super().invoke(self.app, *a, **kw)
 
     def invoke_with_config(self, *args, **kwargs):
+        return self.invoke(
+            ["--config-file", self.test_snowcli_config, *args[0]],
+            **kwargs,
+        )
+
+    def invoke_with_config_and_integration_connection(self, *args, **kwargs):
```
sfc-gh-astus, line 45: I think we will only use this method, so maybe rename this one to `invoke` and remove `invoke_with_config`? WDTY?

sfc-gh-pjob: We can't use only this method because there are commands which do not have `-c`/`--connection` option.

tests_integration/test_snowpark_function.py
```diff
@@ -0,0 +1,417 @@
...
+        ]
+        adjusted_expected_full_function_signatures = [
+            function.upper() for function in expected_full_function_signatures
+        ]
+        assert set(actual_function_signatures) == set(
+            adjusted_expected_full_function_signatures
+        )
+
+    def assert_that_no_files_are_staged_in_snowflake(self) -> None:
+        self.assert_that_only_these_files_are_staged_in_snowflake()
```
sfc-gh-astus, line 408: Alias for method? 🤔

sfc-gh-pjob: yes, to reflect in method name what exactly I want to assert

tests_integration/testing_utils/assertions/test_file_assertions.py
```diff
@@ -0,0 +1,18 @@
...
+from syrupy import SnapshotAssertion
+
+
+def assert_that_current_working_directory_contains_only_following_files(
+    *filenames: str,
+) -> None:
+    assert set(os.listdir()) == set(filenames)
+
+
+def assert_that_file_content_is_equal_to_snapshot(
```
sfc-gh-astus, line 12: Should we care about files content?

sfc-gh-pjob: IMO we should to avoid changing it by mistake (for example by replacing strings globally).

tests_integration/testing_utils/file_utils.py
```diff
@@ -0,0 +1,6 @@
+def replace_text_in_file(file_path: str, to_replace: str, replacement: str) -> None:
+    with open(file_path, "r") as file:
```
sfc-gh-astus, line 2: Maybe just single open with `w+`?

tests_integration/testing_utils/assertions/test_result_assertions.py
```diff
@@ -0,0 +1,91 @@
...
+def assert_that_result_is_successful_and_json_output_contains_value_at_path(
+    result: Result, path: List[Any], expected_value: Any
+) -> None:
+    assert_that_result_is_successful(result)
+    assert_that_json_output_contains_value_at_path(
+        result=result, path=path, expected_value=expected_value
+    )
+
+
+def assert_that_result_is_usage_error(
```
sfc-gh-astus, line 85: Not used function?

tests_integration/testing_utils/assertions/test_result_assertions.py
```diff
@@ -0,0 +1,91 @@
...
+
+def assert_that_result_is_successful_and_output_contains(
+    result: Result,
+    expected_output: str,
+    ignore_case: bool = False,
+) -> None:
+    assert_that_result_is_successful(result)
+    actual = result.output.lower() if ignore_case else result.output
+    expected = expected_output.lower() if ignore_case else expected_output
+    assert expected in actual
```
sfc-gh-astus, line 19: Maybe mock print and assert like in other IT tests?

tests_integration/test_snowpark_function.py
```diff
@@ -0,0 +1,417 @@
...
+    def query_stages_in_database(self) -> List[Dict[str, Any]]:
+        return self.sql_test_helper.execute_single_sql(
+            f"SHOW STAGES IN DATABASE {self.snowflake_session.database}"
+        )
+
+    def query_files_uploaded_in_this_test_case(self) -> List[Dict[str, Any]]:
+        stages_in_database = [
+            stage["name"].lower() for stage in self.query_stages_in_database()
+        ]
+        if "deployments" in stages_in_database:
```
sfc-gh-astus, line 158: I hope other commands don't use `deployments` stage 🙏

tests_integration/test_snowpark_function.py
```diff
@@ -0,0 +1,417 @@
...
+            f"SHOW USER FUNCTIONS LIKE '{self.object_name_prefix.upper()}%'"
+        )
+        return results
+
+    def query_stages_in_database(self) -> List[Dict[str, Any]]:
+        return self.sql_test_helper.execute_single_sql(
+            f"SHOW STAGES IN DATABASE {self.snowflake_session.database}"
+        )
+
+    def query_files_uploaded_in_this_test_case(self) -> List[Dict[str, Any]]:
```
sfc-gh-astus, line 154: There is easier way to do this:

```
row_from_snowflake_session(execute_single_sql("ls @deployments"))
```

This will return list of files in stage

sfc-gh-astus: in dictionary

tests_integration/test_snowpark_function.py
```diff
@@ -0,0 +1,417 @@
...
+        arguments="()",
+        expected_value=1,
+    )
+
+    _test_steps.snowpark_function_drop_should_finish_successfully(
+        function_name=function_name,
+        arguments="()",
+    )
+    _test_steps.assert_that_no_functions_are_in_snowflake()
+    # I'm not sure if it is expected behavior to leave staged files belonging to the removed function.
```
sfc-gh-astus, line 94: Will be removed after tests with database anyway

tests_integration/test_snowpark_function.py
```diff
@@ -0,0 +1,417 @@
...
+from tests_integration.testing_utils.naming_utils import ObjectNameProvider
+from tests_integration.testing_utils.sql_utils import SqlTestHelper
+
+
+@pytest.mark.integration
+def test_snowpark_function_flow(_test_steps):
+    _test_steps.assert_that_no_functions_are_in_snowflake()
+    _test_steps.assert_that_no_files_are_staged_in_snowflake()
+
+    _test_steps.snowpark_function_list_should_return_no_data()
```
sfc-gh-astus, line 39: We should test it if functions are in scope of database, if yes this check is not needed

tests_integration/test_snowpark_function.py
```diff
@@ -0,0 +1,417 @@
...
+
+    _test_steps.snowpark_function_init_should_initialize_files_with_default_content()
+    _test_steps.snowpark_function_package_should_zip_files()
+
+    function_name = _test_steps.snowpark_function_create_should_finish_successfully()
+    _test_steps.assert_that_only_these_functions_are_in_snowflake(
+        f"{function_name}() RETURN VARCHAR"
+    )
+    _test_steps.assert_that_only_these_files_are_staged_in_snowflake(
+        f"deployments/{function_name}/app.zip"
```
sfc-gh-astus, line 49: Is there a way to pass stage name? If not maybe we should add it?

tests_integration/testing_utils/assertions/test_file_assertions.py
sfc-gh-astus: I think this check is not needed, I can't find case when this content can change

sfc-gh-astus: We copy files with external lib 

```
def copytree(src, dst, symlinks=False, ignore=None, copy_function=copy2,
             ignore_dangling_symlinks=False, dirs_exist_ok=False):
    """Recursively copy a directory tree and return the destination directory.
```

I think is safe :)

tests_integration/test_snowpark_function.py
```diff
@@ -0,0 +1,417 @@
...
+            result=result, path=[0, "name"], expected_value=function_name.upper()
+        )
+        assert_that_result_is_successful_and_json_output_contains_value_at_path(
+            result=result,
+            path=[0, "arguments"],
+            expected_value=f"{function_name}{arguments} RETURN {result_type}".upper(),
+        )
+        assert_that_result_is_successful_and_json_output_contains_value_at_path(
+            result=result, path=[0, "language"], expected_value="PYTHON"
+        )
```
sfc-gh-astus, line 230: This method can be replaced with 

```
self._setup.runner.invoke_with_config_and_integration_connection(
            [
                "--format",
                "json",
                "snowpark",
                "function",
                "list",
                "--like",
                f"{self._setup.object_name_prefix}%",
            ]
        )

    assert contains_row_with(
        row_from_mock(mock_print),
        {
            "name": function_name.upper(),
            "arguments": f"{function_name}{arguments} RETURN {result_type}".upper(),
            "language": "PYTHON", 
        },
    )
```

WDYT?

tests_integration/test_snowpark_function.py
```diff
@@ -0,0 +1,417 @@
...
+    )
+    _test_steps.assert_that_only_these_files_are_staged_in_snowflake(
+        f"deployments/{function_name}/app.zip"
+    )
+
+    _test_steps.snowpark_function_list_should_return_list_containing_function_at_the_first_place(
+        function_name=function_name,
+        arguments="()",
+        result_type="NUMBER",
+    )
```
sfc-gh-astus, line 81: I think one test `function list` is enough

tests_integration/conftest.py
sfc-gh-pjob: * `snowpark function init`
* `snowpark function package`

tests_integration/testing_utils/assertions/test_file_assertions.py
sfc-gh-pjob: We copy files using external lib, but we keep files in our repo (we can change files content by mistake). Maybe integration tests are not the best option for such check but what is the cost of having such check here?

tests_integration/test_snowpark_function.py
sfc-gh-pjob: I'm not sure I understand this question. What if they use the same stage?

tests_integration/conftest.py
sfc-gh-turbaszek: we should add this option if it's necessary for those commands

sfc-gh-pjob: These commands work offline, they don't need any connection.

tests_integration/test_warehouse.py
```diff
@@ -7,7 +7,7 @@
 @pytest.mark.integration
 @mock.patch("snowcli.cli.warehouse.print_db_cursor")
 def test_warehouse_status_query(mock_print, runner, snowflake_session):
-    runner.invoke_with_config(["warehouse", "status"])
+    runner.invoke_with_config_and_integration_connection(["warehouse", "status"])
```
sfc-gh-turbaszek, line 10: Maybe we should introduce `integration_runner` to avoid mixing responsibilities in single fixture?

tests_integration/testing_utils/file_utils.py
sfc-gh-pjob: I think it won't work.

* `w+` won't work because it will truncate the file (https://stackoverflow.com/a/30931305)
* `r+` also won't work because it will move the file pointer while reading
* current solution is like this one: https://stackoverflow.com/a/17141572

tests_integration/test_snowpark_function.py
sfc-gh-pjob: I don't think I understand.

1. `LS` seems to be just an alias for `LS`
2. I still have to somehow check if the stage exists (snowflake will return an error if the stage doesn't exist).
3. I still have to filter files to get only files created in this test (to avoid issues with debugging in case of bugs in other tests or in cleanups).

tests_integration/test_snowpark_function.py
sfc-gh-pjob: This my comment in the code was about the lack of removing files from a stage after removing a function. I'm removing the comment because I know now that the behaviour is correct.

tests_integration/test_snowpark_function.py
sfc-gh-pjob: Can you explain what do you mean by `if functions are in scope of database`?

Current assertions in the test case check exactly the same what I check manually when testing the tool:
* empty start at the beginning (both presented by Snowflake SQL output and presented by our tool)
    * I can agree that maybe I don't have to execute SQLs here, but still IMO it's worth to check correct output of `function list`.
* successful execution of commands
* correct state of local files after commands which can change it
* correct state in Snowflake after commands which can change it (checked using SQL)
* correct presentation and usage of functions by our tool (list, describe, execute) after commands which can change functions state

tests_integration/test_snowpark_function.py
sfc-gh-pjob: `deployments` stage is currently hardcoded (`src/snowcli/utils.py`). We can add a task for adding an option to pass custom stage name.

tests_integration/test_snowpark_function.py
sfc-gh-pjob: I don't like using mocks if we can execute an assertion on a real output. IMO it makes future refactors of printing more difficult. But as I wrote in another comment, I think we should talk about it.

tests_integration/test_snowpark_function.py
sfc-gh-pjob: I described in **[this comment](https://github.com/Snowflake-Labs/snowcli/pull/245#discussion_r1273658415)** why there are few invocations of `function list`. But IMO we should talk about it. I understand it may cost us execution time.

tests_integration/test_warehouse.py
sfc-gh-pjob: I'm not sure if it resolves any problem. We have two ways of execution commands in integration tests (with and without `-c integration` flag) but both are correct and both are used by integration tests.

Adding another "runner" fixture will make you think which runner to use instead of which method to use. I don't see any benefit here, only complications.

sfc-gh-pjob:
I rebased the branch and added two commits with fixes. Unfortunately, I did sth wrong (probably changed commit hash) and my original commit is not visible in [github commits history for this PR](https://github.com/Snowflake-Labs/snowcli/pull/245/commits) but it still visible correctly in git log (8eab8e5):
```
git log --oneline
cf57462 [SNOW-851220] Fixed invocation of CLI runner in other integration tests
b89f994 [SNOW-851220] Review fixes
8eab8e5 [SNOW-851220] Integration tests of snowpark function commands
7b26ca0 (tag: v1.0.1-rc1, origin/main, origin/HEAD, main) Bump version to 1.0.1 (#264)

```

@sfc-gh-astus , @sfc-gh-turbaszek 
Please look at my fixes and my responses to your comments.

tests_integration/conftest.py
sfc-gh-astus: We can define default connection now, this invoke will not be needed soon. Leave as it is

tests_integration/testing_utils/assertions/test_result_assertions.py
sfc-gh-pjob: removed

tests_integration/snowflake_connector.py
```diff
@@ -12,7 +12,7 @@
 
 @pytest.fixture(scope="session")
 def create_database():
-    database_name = f"db_{uuid.uuid4().hex}"
+    database_name = f"snowcli_integration_tests_db_{uuid.uuid4().hex}"
```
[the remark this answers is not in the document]
sfc-gh-pjob: done

sfc-gh-astus (approved):

PR #248, opened 2023-07-19
adding additional options to avoid errors when reading options from c…

sfc-gh-mrojas:
…onfig file

### Pre-review checklist
   * [ x] I've confirmed that my changes are up-to-date with the target branch.
   * [ x] I've described my changes in the section below.

### Changes description
Currently when running the snow CLI with some snowsql configurations with commands like 
snow streamlit list I get errors when options like warehouse / shcema / role / db are set

### DIFF
--- a/src/snowcli/connection_params.py
+++ b/src/snowcli/connection_params.py
@@ -6,8 +6,12 @@ class ConnectionParams:
     user: str
     password: str
     account: str
-    host: str = "snowflakecomputing.com"
+    host: str = 'snowflakecomputing.com'
     port: int = 443
-    protocol: str = "https"
+    protocol: str = 'https'
+    warehouse:str = None
+    role:str = None
+    schema:str = None
+    db:str = None
     # add a new parameter to support registry login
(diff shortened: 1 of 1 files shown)


sfc-gh-pjob:
Hello, I think I don't understand what and how this PR fixes.

The changes are only in `ConnectionParams` class in `snowcli/connection_params.py`. This class is an unused leftover from the merge of snowpark containers commands into the main branch (1295af519d31cecfaf9a703d0ac4a0bd939a70f0). I've just removed it in https://github.com/Snowflake-Labs/snowcli/pull/250.

Additionally, not all github checks have passed and there is no test added to verify that the solution really works.

If you see any errors, can you paste what command you're executing and what is the CLI's output (please include `--debug` option, for example `snow --debug streamlit list`).

sfc-gh-mrojas:
Ok. So If I update I should not see this class anymore. The reason I posted the change is that when using containers / streamlits and reading my settings from the .snowsql/config file I was getting exceptions because some properties were not recognized. Maybe I was just running an old version. I will update and re-check and you can dis-regard this PR then.

PR #254, opened 2023-07-21
Unit and integration tests for procedure

sfc-gh-jsikorski:
### Pre-review checklist

### Changes description
* Added integration and unit tests for snowpark procedure
*  Separated and reformatted integration tests steps for both snowpark function and procedure,
* Minor fixes for test_utils.py


tests/test_procedure.py
```diff
@@ -0,0 +1,50 @@
...
+    DIR_INITIAL_CONTENTS = {
+        "requirements.txt",
+        "local_connection.py",
+        ".gitignore",
+        "app.py",
+        "config.toml",
+    }
+
+    def test_procedure_init(self, tmp_dir_for_procedure_tests):
+        procedure.procedure_init()
```
sfc-gh-turbaszek, line 25: I think in other unit tests we use `runner` to invoke the command instead calling the method directly. I think we should decided witch approach we want to go.

tests_integration/test_procedure.py
```diff
@@ -0,0 +1,110 @@
...
+)
+from tests_integration.testing_utils.sql_utils import sql_test_helper
+from tests_integration.testing_utils.working_directory_utils import (
+    temporary_working_directory,
+)
+
+
+@pytest.mark.integration
+def test_snowpark_procedure_flow(_test_steps):
+    _test_steps.assert_that_no_entities_are_in_snowflake()
```
sfc-gh-turbaszek, line 20: what entities? There will be entities like accounts, roles. Should we be more explicite?

tests_integration/test_procedure.py
```diff
@@ -0,0 +1,110 @@
...
+    _test_steps.assert_that_no_files_are_staged_in_snowflake()
```
sfc-gh-turbaszek, line 21: Staged anywhere or on particular stage?

sfc-gh-jsikorski: I changed "snowflake" to "test_db" - hope this will be clearer

tests/test_procedure.py
sfc-gh-jsikorski: Done

tests/test_procedure.py
```diff
@@ -0,0 +1,53 @@
...
+    TEMP_DIRECTORY_NAME = "test_procedure_tmp"
+    DIR_INITIAL_CONTENTS = {
+        "requirements.txt",
+        "local_connection.py",
+        ".gitignore",
+        "app.py",
+        "config.toml",
+    }
+
+    def test_procedure_init(self, tmp_dir_for_procedure_tests, runner):
```
sfc-gh-astus, line 24: This method is used anywhere?

tests/test_procedure.py
```diff
@@ -0,0 +1,53 @@
...
+from zipfile import ZipFile
+
+import pytest
+
+import snowcli.utils
+from snowcli.cli.snowpark import procedure
+from snowcli.utils import SplitRequirements
+
+
+class TestProcedure:
```
sfc-gh-astus, line 14: I'm not sure about this class, usually we write tests without class

tests_integration/test_procedure.py
sfc-gh-turbaszek: I would suggest `_assert_no_procedures_in_test_schema`

sfc-gh-turbaszek (approved):

tests_integration/test_procedure.py
sfc-gh-jsikorski: What about functions then? This test step is used for both

PR #260, opened 2023-07-24
Switch test_utils to tempfile

sfc-gh-jsikorski:
### Pre-review checklist

### Changes description
Change from manually creating directories and files for unit tests for utils.py to files generated by tempfile
+ minor refactor

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -64,9 +64,7 @@ class TestUtils:
         temp_directory_for_app_zip: str,
     ):
         result = utils.prepare_app_zip(app_zip, temp_directory_for_app_zip)
-        assert result == os.path.join(
-            temp_directory_for_app_zip, Path(app_zip).name
-        )
+        assert result == os.path.join(temp_directory_for_app_zip, Path(app_zip).name)
 
     def test_prepare_app_zip_if_exception_is_raised_if_no_source(
         self, temp_directory_for_app_zip
@@ -144,9 +142,7 @@ class TestUtils:
         self, correct_requirements_txt, temp_test_directory: str
     ):
         os.chdir(temp_test_directory)
-        result = utils.generate_streamlit_environment_file(
-            [], correct_requirements_txt
-        )
+        result = utils.generate_streamlit_environment_file([], correct_requirements_txt)
         os.chdir("..")
 
         assert result == PosixPath("environment.yml")
@@ -202,9 +198,7 @@ class TestUtils:
                 in coverage_file.read()
             )
 
-    def test_add_file_to_existing_zip(
-        self, app_zip, correct_requirements_txt: str
-    ):
+    def test_add_file_to_existing_zip(self, app_zip, correct_requirements_txt: str):
         utils.add_file_to_existing_zip(app_zip, correct_requirements_txt)
         zip_file = ZipFile(app_zip)
 
@@ -214,7 +208,7 @@ class TestUtils:
         self,
         temp_test_directory: str,
         txt_file_in_a_subdir: str,
-            temp_file_in_other_directory,
+        temp_file_in_other_directory,
     ):
         zip_file_path = os.path.join(temp_test_directory, "packed.zip")
 
@@ -252,7 +246,9 @@ class TestUtils:
         )
         assert str(Path(temp_file_in_other_directory).name) in zip_file.namelist()
 
-    def test_standard_zip_dir(self, temp_test_directory: str, txt_file_in_a_subdir: str):
+    def test_standard_zip_dir(
+        self, temp_test_directory: str, txt_file_in_a_subdir: str
+    ):
         zip_file_path = os.path.join(temp_test_directory, "packed.zip")
         utils.standard_zip_dir(zip_file_path)
         zip_file = ZipFile(zip_file_path)
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -65,8 +65,10 @@ class TestUtils:
         temp_directory_for_app_zip: str,
     ):
         result = utils.prepare_app_zip(correct_app_zip, temp_directory_for_app_zip)
-
-        assert result == temp_directory_for_app_zip + "/app.zip"
+        print(result)
+        assert result == os.path.join(
+            temp_directory_for_app_zip, Path(correct_app_zip).name
+        )
 
     def test_prepare_app_zip_if_exception_is_raised_if_no_source(
         self, temp_directory_for_app_zip
@@ -135,14 +137,18 @@ class TestUtils:
             result = utils.parse_anaconda_packages(test_data.packages)
 
     def test_generate_streamlit_environment_file_with_no_requirements(self):
-        result = utils.generate_streamlit_environment_file([])
+        result = utils.generate_streamlit_environment_file(
+            [],
+        )
         assert result is None
 
     def test_generate_streamlit_file(
         self, streamlit_requirements_txt, temp_test_directory: str
     ):
         os.chdir(temp_test_directory)
-        result = utils.generate_streamlit_environment_file([])
+        result = utils.generate_streamlit_environment_file(
+            [], streamlit_requirements_txt
+        )
         os.chdir("..")
 
         assert result == PosixPath("environment.yml")
@@ -153,7 +159,7 @@ class TestUtils:
     ):
         os.chdir(temp_test_directory)
         result = utils.generate_streamlit_environment_file(
-            test_data.excluded_anaconda_deps
+            test_data.excluded_anaconda_deps, streamlit_requirements_txt
         )
         os.chdir("..")
         env_file = os.path.join(temp_test_directory, "environment.yml")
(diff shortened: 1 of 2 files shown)


tests/test_utils.py
```diff
@@ -456,7 +449,9 @@ def include_paths_env_variable(self, other_directory: str) -> Generator:
         os.environ.pop("SNOWCLI_INCLUDE_PATHS")
 
     @staticmethod
-    def create_file(filepath: str, contents: List[str]) -> None:
-        with open(filepath, "w") as new_file:
-            for line in contents:
-                new_file.write(line + "\n")
+    def create_file(suffix: str, dir: str, contents: List[str]) -> str:
```
sfc-gh-pjob, line 452:
```suggestion
    def create_temp_file(suffix: str, dir: str, contents: List[str]) -> str:
```

tests/test_utils.py
```diff
@@ -388,66 +399,48 @@ def test_deduplicate_and_sort_reqs(self):
...
-        ).lower()
-        os.mkdir(path)
-        yield path
-        rmtree(path)
+        other_dir = tempfile.TemporaryDirectory()
+        yield other_dir.name
+        rmtree(other_dir.name)
 
     @pytest.fixture
     def file_in_other_directory(self, other_directory: str) -> Generator:
```
sfc-gh-pjob, line 442:
```suggestion
    def temp_file_in_other_directory(self, other_directory: str) -> Generator:
```

tests/test_utils.py
```diff
@@ -388,66 +399,48 @@ def test_deduplicate_and_sort_reqs(self):
...
-        current_path = Path(os.getcwd())
-        path = os.path.join(
-            current_path.parent.absolute(), self.SECOND_TEST_DIRECTORY
-        ).lower()
-        os.mkdir(path)
-        yield path
-        rmtree(path)
+        other_dir = tempfile.TemporaryDirectory()
+        yield other_dir.name
+        rmtree(other_dir.name)
```
sfc-gh-pjob, line 439: Do we have to remove directory on our own? Can't we do sth like the following?:

```
with TemporaryDirectory() as tmp_dir:
        yield tmp_dir
```

tests/test_utils.py
```diff
@@ -456,7 +449,9 @@ def include_paths_env_variable(self, other_directory: str) -> Generator:
...
-    def create_file(filepath: str, contents: List[str]) -> None:
-        with open(filepath, "w") as new_file:
-            for line in contents:
-                new_file.write(line + "\n")
+    def create_file(suffix: str, dir: str, contents: List[str]) -> str:
+        with tempfile.NamedTemporaryFile(suffix=suffix, dir=dir, delete=False) as tmp:
+            with open(tmp.name, "w") as new_file:
+                for line in contents:
+                    new_file.write(line + "\n")
+        return tmp.name
```
sfc-gh-pjob, line 457: Does it return a path or only a name?

If it's only a name, then you won't know the full path when using functions like `file_in_a_subdir` which calls `create_file`.

tests/test_utils.py
```diff
@@ -388,66 +399,48 @@ def test_deduplicate_and_sort_reqs(self):
...
     def correct_metadata_file(self, temp_test_directory: str) -> Generator:
-        path = os.path.join(temp_test_directory, self.CORRECT_METADATA)
-        self.create_file(path, test_data.correct_package_metadata)
-        yield path
+        yield self.create_file(
+            ".yaml", temp_test_directory, test_data.correct_package_metadata
+        )
 
     @pytest.fixture
     def file_in_a_subdir(self, temp_test_directory: str) -> Generator:
```
sfc-gh-pjob, line 431:
```suggestion
    def txt_file_in_a_subdir(self, temp_test_directory: str) -> Generator:
```

tests/test_utils.py
```diff
@@ -388,66 +399,48 @@ def test_deduplicate_and_sort_reqs(self):
...
-        self.create_file(path, test_data.requirements)
-        yield path
+        yield self.create_file(".txt", temp_test_directory, test_data.requirements)
 
     @pytest.fixture
     def streamlit_requirements_txt(self, temp_test_directory: str) -> Generator:
-        path = os.path.join(temp_test_directory, self.REQUIREMENTS_SNOWFLAKE)
-        self.create_file(path, test_data.requirements)
-        yield path
+        yield self.create_file(".txt", temp_test_directory, test_data.requirements)
```
sfc-gh-pjob, line 422: It looks the same as `correct_requirements_txt`

tests/test_utils.py
```diff
@@ -388,66 +399,48 @@ def test_deduplicate_and_sort_reqs(self):
...
-        yield path
+        temp_dir = tempfile.TemporaryDirectory(dir=temp_test_directory)
+        yield temp_dir.name
 
     @pytest.fixture
     def correct_app_zip(self, temp_test_directory: str) -> Generator:
-        path = os.path.join(temp_test_directory, self.APP_ZIP)
-        self.create_file(path, [])
-        yield path
+        yield self.create_file(".zip", temp_test_directory, [])
```
sfc-gh-pjob, line 414: I wonder if `correct_app_zip` is a good name for function creating an empty file only with `.zip` suffix.

tests/test_utils.py
```diff
@@ -388,66 +399,48 @@ def test_deduplicate_and_sort_reqs(self):
 
     @pytest.fixture
     def temp_test_directory(self) -> Generator:
-        path = os.path.join(os.getcwd(), self.TEMP_TEST_DIRECTORY)
-        os.mkdir(path)
-        yield path
-        rmtree(self.TEMP_TEST_DIRECTORY)  # We delete whole directory in teardown -
+        temp_dir = tempfile.TemporaryDirectory()
+        yield temp_dir.name
+        rmtree(temp_dir.name)  # We delete whole directory in teardown -
```
sfc-gh-pjob, line 404: The same comment about directory removing as to `other_directory` function.

tests/test_utils.py
```diff
@@ -274,20 +281,24 @@ def test_standard_zip_dir_with_env_variable(
...
         assert result == test_data.requirements
 
-    def test_get_snowflake_packages_delta(self, streamlit_requirements_txt):
+    def test_get_snowflake_packages_delta(
+        self, temp_test_directory: str, streamlit_requirements_txt
+    ):
         anaconda_package = test_data.requirements[-1]
-        os.chdir(".tests")
+        os.chdir(temp_test_directory)
+        os.rename(streamlit_requirements_txt, self.REQUIREMENTS_SNOWFLAKE)
```
sfc-gh-pjob, line 300: Why do we firstly create file with random name to change it later to a specific name?

tests/test_utils.py
```diff
@@ -274,20 +281,24 @@ def test_standard_zip_dir_with_env_variable(
...
 
         assert result == test_data.requirements
 
-    def test_get_snowflake_packages_delta(self, streamlit_requirements_txt):
+    def test_get_snowflake_packages_delta(
+        self, temp_test_directory: str, streamlit_requirements_txt
+    ):
         anaconda_package = test_data.requirements[-1]
-        os.chdir(".tests")
+        os.chdir(temp_test_directory)
```
sfc-gh-pjob, line 299: Is it safe to stay in a temp dir after a test?

The same question to other usages of `chdir`.

tests/test_utils.py
```diff
@@ -274,20 +281,24 @@ def test_standard_zip_dir_with_env_variable(
             os.path.join("subdir", os.path.basename(file_in_a_subdir))
             not in zip_file.namelist()
         )
-        assert os.path.join(self.FILE_IN_SECOND_TEST_DIRECTORY) in zip_file.namelist()
+        assert Path(file_in_other_directory).name in zip_file.namelist()
 
-    def test_get_snowflake_packages(self, streamlit_requirements_txt):
-        os.chdir(".tests")
+    def test_get_snowflake_packages(
```
sfc-gh-pjob, line 286: The same 2 comments as to `test_get_snowflake_packages_delta`

tests/test_utils.py
sfc-gh-jsikorski: Contrary to Path.name (that returns only filename) tempfile .name returns an absolute path

tests/test_utils.py
sfc-gh-pjob: If you need a file with a specific name and if you create it in temp directory (what takes care of cleaning), then IMO it's better to create a file with the target name right up front.

tests/test_utils.py
sfc-gh-pjob: Looks good, but I would wrap it in a fixture.

I have such util in my PR with integration tests (https://github.com/Snowflake-Labs/snowcli/pull/245/files#diff-9ef616b95f50c54f5ca3b20f74dab7bdb579c6abd97a1c9cb8d624449bb6da8dR20).

I don't know if we want to share utils between tests and integration tests, so if you want to, you may simply copy-paste it. WDYT?

```python
class WorkingDirectoryChanger:
    def __init__(self):
        self._initial_working_directory = os.getcwd()

    @staticmethod
    def change_working_directory_to(directory: str):
        os.chdir(directory)

    def restore_initial_working_directory(self):
        self.change_working_directory_to(self._initial_working_directory)


@pytest.fixture
def temporary_working_directory():
    working_directory_changer = WorkingDirectoryChanger()
    with TemporaryDirectory() as tmp_dir:
        working_directory_changer.change_working_directory_to(tmp_dir)
        yield tmp_dir
        working_directory_changer.restore_initial_working_directory()

```

sfc-gh-jsikorski: I tried few different approaches (yours, pure fixture, decorator), and i chose using separate fixture. It`s not as versatile as your Changer, but i only need some tests to enter temp directory, and this way i avoid duplicating large chunks of code

sfc-gh-pjob (approved):

PR #268, opened 2023-07-27
Create new database for each integration test

sfc-gh-astus:
### Pre-review checklist

### Changes description
* For each integration test new database is created
* Removed second connection to Snowflake

### DIFF
--- a/tests_integration/snowflake_connector.py
+++ b/tests_integration/snowflake_connector.py
@@ -10,42 +10,29 @@ from snowcli.exception import EnvironmentVariableNotFoundError
 _ENV_PARAMETER_PREFIX = "SNOWFLAKE_CONNECTIONS_INTEGRATION"
 
 
-@pytest.fixture(scope="session")
-def create_database():
+@pytest.fixture(scope="function")
+def test_database(snowflake_session):
     database_name = f"db_{uuid.uuid4().hex}"
-
-    config = {
-        "application": "INTEGRATION_TEST",
-        "account": _get_from_env("ACCOUNT", True),
-        "user": _get_from_env("USER", True),
-        "password": _get_from_env("PASSWORD", True),
-        "schema": _get_from_env("SCHEMA", False, "PUBLIC"),
-        "host": _get_from_env("HOST", False),
-        "role": _get_from_env("ROLE", False),
-    }
-    config = {k: v for k, v in config.items() if v is not None}
-    connection = connector.connect(**config)
-
-    connection.execute_string(f"create database {database_name}")
+    snowflake_session.execute_string(
+        f"create database {database_name}; use database {database_name}; use schema public;"
+    )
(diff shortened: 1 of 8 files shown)


tests_integration/snowflake_connector.py
```diff
@@ -10,46 +10,34 @@
...
-@pytest.fixture(scope="session")
-def create_database():
-    database_name = f"db_{uuid.uuid4().hex}"
+@pytest.fixture(scope="function")
+def database_name():
+    yield f"db_{uuid.uuid4().hex}"
 
+
+@pytest.fixture(scope="function")
+def snowflake_session(database_name):
```
sfc-gh-pjob, line 19: Is it possible to have two fixtures like following:

* `test_database` - scope "function" - responsible for creating a database, setting "use database" and "use schema", setting env variables, dropping the database and removing the environment variables
* `snowflake_session` - scope "session" - responsible for creating a connection to snowflake before all tests and closing it after all tests. The connection would be configured without database level settings (database, schema).

IMO it could allow to have much less connections. WDYT? Is it possible?

tests_integration/test_cp.py
```diff
@@ -1,7 +1,7 @@
 import pytest
 
 from unittest import mock
-from tests_integration.snowflake_connector import create_database, snowflake_session
+from tests_integration.snowflake_connector import database_name, snowflake_session
```
sfc-gh-pjob, line 4: IMO it may be confusing that you have to use "snowflake_session" fixture in your test case even if you don't want to execute SQLs in your test (you want to only use SnowCLI command requiring some database).

tests_integration/snowflake_connector.py
sfc-gh-astus: Good idea, but it requires to add `database_name` to all tests

sfc-gh-astus: Nvm it is a great idea :)

tests_integration/snowflake_connector.py
```diff
@@ -10,39 +10,28 @@
...
+def snowflake_session():
     config = {
         "application": "INTEGRATION_TEST",
         "account": _get_from_env("ACCOUNT", True),
         "user": _get_from_env("USER", True),
         "password": _get_from_env("PASSWORD", True),
         "host": _get_from_env("HOST", True),
-        "database": _get_from_env("DATABASE", True),
         "role": _get_from_env("ROLE", True),
         "schema": _get_from_env("SCHEMA", False),
```
sfc-gh-pjob, line 36: should schema be defined for connection (scope = session) if we create separate databases per test case?

sfc-gh-astus: I think we should keep `public` as a default until we will have tests that require other schema

sfc-gh-turbaszek (dismissed):

sfc-gh-pjob (dismissed):

sfc-gh-turbaszek (approved):

PR #284, opened 2023-08-01
Added toml to the requirements for local dev

sfc-gh-cconner:
### Pre-review checklist
   * [ * ] I've confirmed that instructions included in README.md are still correct after my changes in the codebase.
   * [ * ] I've confirmed that my changes are up-to-date with the target branch.
   * [ * ] I've described my changes in the section below.

### Changes description
Fixing:

https://github.com/Snowflake-Labs/snowcli/issues/282

Added toml to the requirements.txt for procedure and function for local development purposes.

### DIFF
--- a/src/templates/default_function/requirements.txt
+++ b/src/templates/default_function/requirements.txt
@@ -1,1 +1,2 @@
 snowflake-snowpark-python
+toml # For local development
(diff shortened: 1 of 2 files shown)


src/templates/default_function/requirements.txt
```diff
@@ -1 +1,2 @@
 snowflake-snowpark-python
+toml # For local development
```
sfc-gh-turbaszek, line 2: I don't think this is the right fix. The right think would be to updated `snowflake-snowpark-python` to use latest `snowflke-python-connector` which requires toml.

Knowing this will take time I would suggest adding explicite `snowflake-connector-python>=3.1.0`.

sfc-gh-turbaszek:
Related to #283

PR #294, opened 2023-08-03
Handling results of multiple queries

sfc-gh-astus:
### Pre-review checklist

### Changes description
Added print_db_cursors function to handle many responses in single command

### DIFF
--- a/src/snowcli/cli/sql.py
+++ b/src/snowcli/cli/sql.py
@@ -7,7 +7,7 @@ from click import UsageError
 
 from snowcli.cli.common.snow_cli_global_context import snow_cli_global_context_manager
 from snowcli.cli.common.decorators import global_options
-from snowcli.output.printing import print_db_cursor
+from snowcli.output.printing import print_output, OutputData
 
 
 @global_options
@@ -62,5 +62,7 @@ def execute_sql(
         sql_text=sql,
         remove_comments=True,
     )
+    output_data = OutputData()
     for result in results:
-        print_db_cursor(result)
+        output_data.add_data(result)
+    print_output(output_data)
--- a/src/snowcli/output/printing.py
+++ b/src/snowcli/output/printing.py
@@ -1,13 +1,13 @@
 from __future__ import annotations
 
+import click
+from collections.abc import Iterable
 from datetime import datetime
 from json import JSONEncoder
 from pathlib import Path
-from typing import List, Optional, Dict
-
 from rich import box, print, print_json
-import click
 from snowflake.connector.cursor import SnowflakeCursor
+from typing import List, Optional, Dict, Union
 
 from snowcli.output.formats import OutputFormat
 
@@ -23,6 +23,69 @@ class CustomJSONEncoder(JSONEncoder):
         return super().default(o)
 
 
+class OutputData:
+    def __init__(self, format: Optional[OutputFormat] = None) -> None:
+        self.output: List[Union[SnowflakeCursor, str, List[Dict]]] = []
+        self.counter = 0
+        self._format = format
+
+    @property
+    def format(self):
+        if not self._format:
+            self._format = _get_format_type()
+        return self._format
+
+    def add_data(self, data: Union[str, List[Dict], SnowflakeCursor]):
+        if (
+            isinstance(data, str)
+            or isinstance(data, list)
+            or isinstance(data, SnowflakeCursor)
+        ):
+            self.output.append(data)
+            self.counter += 1
+            return self
+        else:
+            raise TypeError(f"unsupported data type {type(data)}")
+
+    def get_data(self) -> Union[str, Dict, Iterable[List[Dict]]]:
+        for output in self.output:
+            if isinstance(output, SnowflakeCursor):
+                yield _get_data_from_cursor(output)
+            elif isinstance(output, str):
+                if self.format != OutputFormat.JSON:
+                    yield output
+                else:
+                    yield {"result": output}
+            elif isinstance(output, list):
+                yield output
+
+    def size(self) -> int:
+        return self.counter
+
+
+def print_output(output_data: OutputData) -> None:
+    if output_data.size() == 0:
+        print("Done")
+        return
+
+    if output_data.format == OutputFormat.TABLE:
+        for data in output_data.get_data():
+            _print_table(data)
+    elif output_data.format == OutputFormat.JSON:
+        _print_json(output_data)
+    else:
+        raise Exception(f"Unknown {output_data.format} format option")
+
+
+def _print_json(output_data: OutputData) -> None:
+    import json
+
+    data = list(output_data.get_data())
+    if len(data) == 1:
+        data = data[0]
+    print_json(json.dumps(data, cls=CustomJSONEncoder))
+
+
 def print_db_cursor(
     cursor: SnowflakeCursor,
     columns: Optional[List[str]] = None,
@@ -35,19 +98,24 @@ def print_db_cursor(
         not provided then all columns are returned
     :return:
     """
+    data = _get_data_from_cursor(cursor, columns)
+    _print_formatted(data)
+
+
+def _get_data_from_cursor(
+    cursor: SnowflakeCursor, columns: Optional[List[str]] = None
+) -> List[Dict]:
     result = cursor.fetchall()
     column_names = [col.name for col in cursor.description]
     columns_to_include = columns or column_names
 
-    data = [
+    return [
         {k: v for k, v in zip(column_names, row) if k in columns_to_include}
         for row in result
     ]
 
-    _print_formatted(data, columns_to_include)
 
-
-def _get_format_type():
+def _get_format_type() -> OutputFormat:
     context = click.get_current_context()
     format_from_ctx = context.find_root().params.get("output_format")
     if format_from_ctx:
@@ -55,10 +123,10 @@ def _get_format_type():
     return OutputFormat.TABLE
 
 
-def _print_formatted(data: List[Dict], columns: Optional[List[str]] = None):
+def _print_formatted(data: List[Dict]) -> None:
     output_format = _get_format_type()
     if output_format == OutputFormat.TABLE:
-        _print_table(data, columns)
+        _print_table(data)
     elif output_format == OutputFormat.JSON:
         import json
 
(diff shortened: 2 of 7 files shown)


src/snowcli/output/printing.py
```diff
@@ -23,6 +23,28 @@ def default(self, o):
...
+    if output_format == OutputFormat.TABLE:
+        for cursor in cursors:
+            print_db_cursor(cursor)
+    elif output_format == OutputFormat.JSON:
+        import json
+
+        data = [_get_data_from_cursor(c) for c in cursors]
+        print_json(json.dumps(data, cls=CustomJSONEncoder))
+    else:
+        raise Exception(f"Unknown {output_format} format option")
```
sfc-gh-astus, line 45: Maybe someone will have better idea with this format parameter check. I do it twice for table output :/

tests_integration/test_streamlit.py
```diff
@@ -43,16 +44,16 @@ def test_streamlit_create_and_deploy(
         row_from_mock(mock_print), row_from_snowflake_session(expect)[0]
     )
 
-    runner.invoke_with_config_and_integration_connection(
-        ["streamlit", "describe", streamlit_name]
+    result = runner.invoke_with_config_and_integration_connection(
+        ["--format", "JSON", "streamlit", "describe", streamlit_name]
```
sfc-gh-astus, line 48: I will change format to json in all integration tests in other PR

src/snowcli/cli/sql.py
```diff
@@ -62,5 +62,7 @@ def execute_sql(
         sql_text=sql,
         remove_comments=True,
     )
-    for result in results:
-        print_db_cursor(result)
+    if len(results) == 1:
```
sfc-gh-jsikorski, line 65: Is this check necessary? What would happen if we would  pass the one-item-long list to print_db_cursors?

src/snowcli/output/printing.py
sfc-gh-jsikorski: First thing that comes to my mind is keyword argument, with default value set to None, both in `print_db_cursor` and `print_formatted` - so in the latter, it would be `output_format= output_format if output_format else _get_output_format`

src/snowcli/cli/sql.py
```diff
@@ -62,5 +62,4 @@ def execute_sql(
         sql_text=sql,
         remove_comments=True,
     )
-    for result in results:
-        print_db_cursor(result)
+    print_output(OutputData(*results))
```
sfc-gh-astus, line 65: Allow pass list of cursors

sfc-gh-turbaszek: I'm wondering about something more informative
```py
output = OutputData()
for cursors in cursors:
  output.add_cursor(cursor)
```

The result is probably the same but the code is more expressive. WDYT?

tests/output/test_printing.py
```diff
@@ -2,72 +2,91 @@
...
-    assert captured.out == dedent(
+    assert _get_output(capsys) == dedent(
         """\
     +---------------------------------------------------------------------+
     | string | number | array     | object          | date                |
     |--------+--------+-----------+-----------------+---------------------|
     | string | 42     | ['array'] | {'k': 'object'} | 2022-03-21 00:00:00 |
     | string | 43     | ['array'] | {'k': 'object'} | 2022-03-21 00:00:00 |
     +---------------------------------------------------------------------+
+    +---------------------------------------------------------------------+
```
sfc-gh-turbaszek, line 50: Should we print query between tables? It will help users to navigate the output. WDYT?

tests/output/test_printing.py
```diff
@@ -2,72 +2,91 @@
...
+    | string | 43     | ['array'] | {'k': 'object'} | 2022-03-21 00:00:00 |
+    +---------------------------------------------------------------------+
+    Command done
+    +--------+
+    | key    |
+    |--------|
+    | value  |
+    +--------+
+    """
+    )
```
sfc-gh-turbaszek, line 83: I think this is fine in both table and json format.

tests/output/test_printing.py
```diff
@@ -85,18 +104,134 @@ def test_print_db_cursor_json(context_mock, capsys):
...
+            },
+            {
+                "string": "string",
+                "number": 43,
+                "array": ["array"],
+                "object": {"k": "object"},
+                "date": "2022-03-21T00:00:00",
+            },
+        ],
+        [{"result": "Command done"}],
```
sfc-gh-turbaszek, line 176: Should this be array? We know it's a single result.

src/snowcli/output/printing.py
```diff
@@ -23,6 +23,75 @@ def default(self, o):
...
+        string_output: Optional[str] = None,
+        data: Optional[List[Dict]] = None,
+    ) -> None:
+        self.output: List[Union[SnowflakeCursor, str, List[Dict]]] = []
+        self.counter = 0
+
+        if format:
+            self.format = format
+        else:
+            self.format = _get_format_type()
```
sfc-gh-turbaszek, line 40:
```suggestion
        self._format = format
        
     @property
     def format(self):
            if not self._format:
                  self.format = _get_format_type()
            return self._format
```

src/snowcli/output/printing.py
```diff
@@ -23,6 +23,75 @@ def default(self, o):
...
+
+        if cursor:
+            self.add_cursor(cursor)
+
+        if string_output:
+            self.add_string_output(string_output)
+
+        if data:
+            self.add_data(data)
```
sfc-gh-turbaszek, line 49: Instead of that I would consider builder pattern.
```
output = OutputData().add_string("ddd").add_cursor(curr).add_data([1,2,3])
```
what do you think?

src/snowcli/output/printing.py
```diff
@@ -23,6 +23,75 @@ def default(self, o):
...
+
+    def add_cursor(self, cursor: SnowflakeCursor):
+        self.output.append(cursor)
+        self.counter += 1
+
+    def add_string_output(self, string_output: str):
+        self.output.append(string_output)
+        self.counter += 1
```
sfc-gh-turbaszek, line 57: Should we make this a helper method?

src/snowcli/output/printing.py
```diff
@@ -23,6 +23,75 @@ def default(self, o):
...
+            self.add_cursor(cursor)
+
+        if string_output:
+            self.add_string_output(string_output)
+
+        if data:
+            self.add_data(data)
+
+    def add_cursor(self, cursor: SnowflakeCursor):
+        self.output.append(cursor)
```
sfc-gh-turbaszek, line 52: Should we do type validation in every method? We don't want users to add cursors via `add_data`, right?

src/snowcli/output/printing.py
```diff
@@ -23,6 +23,75 @@ def default(self, o):
         return super().default(o)
 
 
+class OutputData:
```
sfc-gh-turbaszek, line 26: Maybe we can reduce the output printing to `print(OutputData(...))`? This can be done by implementation of `__str__`. In this way we can hide the implementation from users.

sfc-gh-turbaszek (changes requested):
Few comments to consider but the direction is very good 👌

tests/output/test_printing.py
sfc-gh-astus: I like the idea, but it should be new task. I will create

src/snowcli/output/printing.py
sfc-gh-astus: changed

src/snowcli/output/printing.py
sfc-gh-astus: I like it, changed

src/snowcli/output/printing.py
sfc-gh-astus: That was not necessary, merged it into one method

src/snowcli/cli/sql.py
sfc-gh-astus: nvm, I changed it to your idea

tests/output/test_printing.py
sfc-gh-astus: Right, changed

src/snowcli/output/printing.py
sfc-gh-astus: I like the idea, but it is not that easy to do it. We are using print(table) and print_jason from rich :/

src/snowcli/cli/sql.py
sfc-gh-astus: outdated :)

src/snowcli/output/printing.py
```diff
@@ -23,6 +23,62 @@ def default(self, o):
...
+        if not self._format:
+            self._format = _get_format_type()
+        return self._format
+
+    def add_data(self, data: Union[str, SnowflakeCursor, List[Dict]]):
+        self.output.append(data)
+        self.counter += 1
+        return self
+
+    def get_data(self) -> Iterable:
```
sfc-gh-turbaszek, line 43: What is the type of items in this iterable?

sfc-gh-astus: `Iterator[List[Dict]]`

Changed

sfc-gh-astus: Changed to `def get_data(self) -> Union[str, Dict, Iterable[List[Dict]]]:`

src/snowcli/output/printing.py
sfc-gh-astus: I agree, but think it should be done in other PR

src/snowcli/output/printing.py
sfc-gh-turbaszek: But this is stil case right? What will happen if someone passes something that we do not accept?

src/snowcli/output/printing.py
```diff
@@ -23,6 +23,62 @@ def default(self, o):
...
+        self._format = format
+
+    @property
+    def format(self):
+        if not self._format:
+            self._format = _get_format_type()
+        return self._format
+
+    def add_data(self, data: Union[str, SnowflakeCursor, List[Dict]]):
+        self.output.append(data)
```
sfc-gh-turbaszek, line 39: Let's validate that data is of supported type, fast fail will be useful.

sfc-gh-turbaszek (approved):
Small comment but otherwise looks good!

src/snowcli/output/printing.py
sfc-gh-astus: You mean to check if data is one of type [str, List[Data], SnowflakeCursor] with isinstance?

sfc-gh-turbaszek (approved):

PR #297, opened 2023-08-04
Handle string as output

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
...

### DIFF
--- a/src/snowcli/output/printing.py
+++ b/src/snowcli/output/printing.py
@@ -1,9 +1,10 @@
 from __future__ import annotations
 
+from collections.abc import Iterable
 from datetime import datetime
 from json import JSONEncoder
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


src/snowcli/output/printing.py
```diff
@@ -67,24 +68,51 @@ def _print_formatted(data: List[Dict], columns: Optional[List[str]] = None):
...
         return
 
-    columns = columns or list(data[0].keys())
-
     table = Table(show_header=True, box=box.ASCII)
-    for column in columns:
-        table.add_column(column)
     for row in data:
+        if not columns:
+            columns = list(row.keys())
```
sfc-gh-astus, line 89: You will assign list of columns for each row, this list be the same for all rows.

PR #304, opened 2023-08-08
Refactor for snowpark package + fixes

sfc-gh-jsikorski:
### Pre-review checklist

### Changes description
* Fix for empty zip file after executing 'snow snowpark package create'
* Reformat for package commands
* Reformated pytest fixtures, to avoid duplicated code


src/snowcli/cli/snowpark/package/commands.py
```diff
@@ -0,0 +1,89 @@
...
+        help="Install packages that are not available on the Snowflake anaconda channel",
+    ),
+    **kwargs,
+):
+    """
+    Checks if a package is available on the Snowflake anaconda channel.
+    In install_packages flag is set to True, command will check all the dependencies of the packages
+    outside snowflake channel.
+    """
+    log.info(PackageManager().lookup(name, install_packages))
```
sfc-gh-turbaszek, line 37: I think we should return this information instead of log. This message is the result of this operation

src/snowcli/cli/snowpark/package/commands.py
```diff
@@ -0,0 +1,89 @@
...
+        "--overwrite",
+        "-o",
+        help="Overwrite the file if it already exists",
+    ),
+    **kwargs,
+):
+    """
+    Upload a python package zip file to a Snowflake stage, so it can be referenced in the imports of a procedure or function.
+    """
+    return PackageManager().upload(file, stage, overwrite)
```
sfc-gh-turbaszek, line 67:
```suggestion
    return PackageManager().upload(file=file, stage=stage, overwrite=overwrite)
```
Please use keywords. It's makes refactoring easier later on

src/snowcli/cli/snowpark/package/commands.py
```diff
@@ -0,0 +1,89 @@
...
+        "-y",
+        help="Install packages that are not available on the Snowflake anaconda channel",
+    ),
+    **kwargs,
+):
+    """
+    Create a python package as a zip file that can be uploaded to a stage and imported for a Snowpark python app.
+    """
+    PackageManager().lookup(name, install_packages, True)
+    log.info(PackageManager().create(name, install_packages))
```
sfc-gh-turbaszek, line 89: We should return here a message. Probably @sfc-gh-astus work will be essential here

src/snowcli/cli/snowpark/package/manager.py
```diff
@@ -0,0 +1,82 @@
...
+        if package_response.snowflake:
+            return f"Package {name} is available on the Snowflake Anaconda channel. You can just include it in your 'packages' declaration"
+        else:
+            if install_packages:
+                status, result = utils.install_packages(
+                    perform_anaconda_check=True, package_name=name, file_name=None
+                )
+                self._cleanup_after_install(_run_nested)
+                return self._determine_lookup_result(status, result, name)
+        return ""
```
sfc-gh-turbaszek, line 34: Let's return meaningful message

src/snowcli/cli/snowpark/package/manager.py
```diff
@@ -0,0 +1,82 @@
...
+            log.info(
+                "Package already exists on stage. Consider using --overwrite to overwrite the file."
+            )
+
+    def create(self, name: str, install_packages):
+        if os.path.exists(".packages"):
+
+            utils.recursive_zip_packages_dir(".packages", name + ".zip")
+            self._cleanup_after_install(False)
+        return f"Package {name}.zip created. You can now upload it to a stage (`snow package upload -f {name}.zip -s packages`) and reference it in your procedure or function."
```
sfc-gh-turbaszek, line 65: Should we differentiate the message for those two cases?
1. If .packages exists - we perform an operation
2. If id does not exist then we do nothin

We should inform user what exactly was done. For example if `.packages` does not exists we probably can create zip artifact, right? Should it be an error?

src/snowcli/cli/snowpark/package/manager.py
```diff
@@ -0,0 +1,82 @@
...
+            return f"""
+                The package {name} is supported, but does depend on the "
+                following Snowflake supported native libraries. You should "
+                include the following in your packages: {reqs.snowflake}"
+                """
+        return ""
+
+    @staticmethod
+    def _cleanup_after_install(run_nested: bool):
+        if not run_nested and os.path.exists(".packages"):
```
sfc-gh-turbaszek, line 81: I don't like this `run_nested` I think we need to get rid of that so we clearly separate  commands from helper functions. Let's think about what `run_nested` does - it's used to indicate some additional operations. Can we name those operations for example `do_cleanup`? Or maybe we can separate the logic in a way that removes the need for this flag altogether

src/snowcli/utils.py
```diff
@@ -503,9 +503,10 @@ def get_list_of_files_to_pack(
     files: List[File] = []
 
     def filenames_filter(filepath: Path) -> bool:
+
         return (
             not filepath.name.startswith(".")
-            and not str(filepath).startswith(".")
+            # and not str(filepath).startswith(".")
```
sfc-gh-turbaszek, line 509: To be removed?

sfc-gh-jsikorski: Yes

src/snowcli/cli/snowpark/package/commands.py
sfc-gh-jsikorski: Sure. That`s exactly what we talked about - what exactly we want to return. Original function sometimes logges, sometimes returned a string.
So, we should return a string and use @with_output?

src/snowcli/cli/snowpark/package/manager.py
sfc-gh-jsikorski: I also had this doubt. I`ll try calling cleanup from the command, not from the manager - so the command will be responsible for whole flow, and manager logic for single operations

src/snowcli/cli/snowpark/package/commands.py
sfc-gh-turbaszek: Yup. And I would consider using discussion from https://github.com/Snowflake-Labs/snowcli/pull/304#discussion_r1288076864 to implement the if here so `lookup` return result but the command decided what meesage should be returned

src/snowcli/cli/snowpark/package/manager.py
```diff
@@ -0,0 +1,82 @@
...
+    def _determine_lookup_result(
+        status: bool, reqs: SplitRequirements, name: str
+    ) -> str:
+        if status and reqs and reqs.snowflake:
+            return f"""
+                The package {name} is supported, but does depend on the "
+                following Snowflake supported native libraries. You should "
+                include the following in your packages: {reqs.snowflake}"
+                """
+        return ""
```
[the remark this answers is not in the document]
sfc-gh-turbaszek: Sounds good to me 👍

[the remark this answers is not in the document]
sfc-gh-jsikorski: Done

src/snowcli/cli/snowpark/package/manager.py
sfc-gh-jsikorski: Done

src/snowcli/cli/snowpark/package/commands.py
sfc-gh-jsikorski: Done

src/snowcli/cli/snowpark/package/commands.py
sfc-gh-jsikorski: Done

src/snowcli/cli/snowpark/package/commands.py
sfc-gh-jsikorski: Done

src/snowcli/cli/snowpark/package/manager.py
sfc-gh-jsikorski: Done

tests/test_package.py
```diff
@@ -114,16 +119,28 @@ def test_package_upload(self, mock_conn, package_file: str, runner) -> None:
         )
 
     @pytest.fixture
-    def packages_directory(self):
-        path = os.path.join(os.getcwd(), ".packages")
-        os.mkdir(path)
-        yield path
+    def temp_dir(self):
```
sfc-gh-astus, line 122: We have 3 fixtures that change directory already. Maybe you can reuse one of them? `execute_in_tmp_dir` looks similar

tests/test_package.py
```diff
@@ -114,16 +119,28 @@ def test_package_upload(self, mock_conn, package_file: str, runner) -> None:
...
+        os.chdir(initial_dir)
+        tmp.cleanup()
 
     @pytest.fixture
     def package_file(self):
         with tempfile.TemporaryDirectory() as tmp:
             yield create_named_file("app.zip", tmp, [])
 
+    @pytest.fixture
+    def dot_packages_directory(self, temp_dir):
```
sfc-gh-astus, line 136: 1. You can create directory with subdirectories https://stackoverflow.com/questions/273192/how-do-i-create-a-directory-and-any-missing-parent-directories
2. You don't have to chdir there
3. You use it in only one test, it has to be fixture?

src/snowcli/cli/snowpark/package/manager.py
sfc-gh-jsikorski: Done

tests/test_package.py
sfc-gh-jsikorski: 1. Done
2. Done
3. I`m thinking hard about adding another test, that would use it :)

tests/test_package.py
sfc-gh-jsikorski: Did the huge reformat of fixtures

tests/test_package.py
```diff
@@ -1,15 +1,20 @@
...
 import tempfile
 from requirements.requirement import Requirement
 from unittest.mock import ANY, MagicMock, patch
 
 from snowcli.cli.snowpark import package
+from snowcli.cli.snowpark.package.utils import NotInAnaconda
 from snowcli.utils import SplitRequirements
 from tests.test_data import test_data
 from tests.testing_utils.files_and_dirs import create_named_file
+from tests.testing_utils.fixtures import *
```
sfc-gh-astus, line 17: Do we need import everything everywhere?

tests/testing_utils/fixtures.py
```diff
@@ -0,0 +1,194 @@
...
+    os.chdir(tmp.name)
+    yield tmp.name
+    os.chdir(initial_dir)
+    tmp.cleanup()
+
+
+@pytest.fixture
+def temp_directory_for_app_zip(temp_dir) -> Generator:
+    temp_dir = tempfile.TemporaryDirectory(dir=temp_dir)
+    yield temp_dir.name
```
sfc-gh-astus, line 169: What is the difference between `temp_dir`, `temp_directory_for_app_zip` and `other_directory_with_chdir`?

sfc-gh-jsikorski: `temp_dir` is basic directory for all tests, `temp_directory_for_app_zip` is a subdir in `temp_dir`, and `other_directory_with_chdir` is a separate directory, for tests, that need two directories in different paths (but it seems most of them use only `other_directory`, so i'll delete it)

tests/test_package.py
sfc-gh-jsikorski: I thought about giving access to all fixtures at once - to keep code concise, but if you think it may cause any problems, i can be more specific

src/snowcli/cli/snowpark/package/commands.py
```diff
@@ -0,0 +1,131 @@
...
+    """
+    lookup_result = PackageManager().lookup(
+        name=name, install_packages=install_packages
+    )
+    PackageManager().cleanup_after_install()
+    log.info(
+        message := PackageManager().create_lookup_message(
+            lookup_result=lookup_result, name=name
+        )
+    )
```
sfc-gh-turbaszek, line 53: Should instantiate the `PackageManager` only once?

src/snowcli/cli/snowpark/package/commands.py
```diff
@@ -0,0 +1,131 @@
...
+    ),
+    **kwargs,
+) -> None:
+    """
+    Upload a python package zip file to a Snowflake stage, so it can be referenced in the imports of a procedure or function.
+    """
+    log.info(
+        message := PackageManager().upload(file=file, stage=stage, overwrite=overwrite)
+    )
+    print(message)
```
sfc-gh-turbaszek, line 88: I don't think there's need to log what is printed. WDYT?

src/snowcli/cli/snowpark/package/manager.py
```diff
@@ -0,0 +1,101 @@
...
+class PackageManager:
+    @staticmethod
+    def lookup(name: str, install_packages: bool) -> LookupResult:
+
+        package_response = utils.parse_anaconda_packages([Requirement.parse(name)])
+
+        if package_response.snowflake and not package_response.other:
+            return InAnaconda(package_response)
+        else:
+            if install_packages:
```
sfc-gh-turbaszek, line 36: This can be `elif`, right?

src/snowcli/cli/snowpark/package/manager.py
```diff
@@ -0,0 +1,101 @@
...
+            return f"Package {name} is available on the Snowflake anaconda channel."
+        elif type(lookup_result) == RequiresPackages:
+            return f"""The package {name} is supported, but does depend on the
+                    following Snowflake supported native libraries. You should
+                    include the following in your packages: {lookup_result.requirements.snowflake}"""
+        elif type(lookup_result) == NotInAnaconda:
+            return f"""The package {name} is avaiable through PIP. You can create a zip using:\n
+                    snow snowpark package create {name} -y"""
+        else:
+            return f"Lookup for package {name} resulted in some error. Please check the package name and try again"
```
sfc-gh-turbaszek, line 101: How about adding those messages to correspondind classes:
```
class InAnaconda(LookupResult):
     def __init__(self, package_name: str):
         self.package_name = package_name
   

     @property
     def message(self):
         return f"""The package {self.package_name} is supported, but does depend on the
                    following Snowflake supported native libraries. You should
                    include the following in your packages..."
```
WDYT?

src/snowcli/cli/snowpark/package/commands.py
sfc-gh-jsikorski: Actually, i decided to go for 0 times - as most methods didn`t need to be packed in an object, i turned them to functions.

src/snowcli/cli/snowpark/package/manager.py
```diff
@@ -0,0 +1,101 @@
...
+            if install_packages:
+                status, result = utils.install_packages(
+                    perform_anaconda_check=True, package_name=name, file_name=None
+                )
+
+                if status:
+                    if result.snowflake:
+                        return RequiresPackages(result)
+                    else:
+                        return NotInAnaconda(result)
```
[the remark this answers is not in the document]
sfc-gh-jsikorski: Done

src/snowcli/cli/snowpark/package/manager.py
sfc-gh-jsikorski: Done

src/snowcli/cli/snowpark/package/manager.py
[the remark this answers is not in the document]
sfc-gh-jsikorski: NotInAnaconda should be returned only if status is True and there are no results.snowflake - it shows, that the check was performed, yet no 
If status is false - we return NothingFound

src/snowcli/cli/snowpark/package/manager.py
sfc-gh-jsikorski: Done

src/snowcli/cli/snowpark/package/commands.py
sfc-gh-jsikorski: Done

sfc-gh-astus (approved):

tests_integration/test_package.py
```diff
@@ -10,7 +13,7 @@
 class TestPackage:
     STAGE_NAME = "PACKAGE_TEST"
 
-    @pytest.mark.integration
+    # @pytest.mark.integration
```
sfc-gh-astus, line 16: Comment

sfc-gh-jsikorski: Done

sfc-gh-astus (approved):

PR #305, opened 2023-08-09
Plugins PoC

sfc-gh-pjob:
### Pre-review checklist

#### Plugins PoC
Proof of concept of having "Plugins API" and using it to load both built-in and external plugins (with the help of pluggy library).

How to try this PoC:
1. Checkout the branch
2. Create a new clean python venv
3. Run `pip install -e core -e plugins/connection -e plugins/snowpark-extensions`
4. At the first run `snow --help` shouldn't show `connection` command group and `snow snowpark --help` shouldn't show `function`, `package` and `say-hello`. All these commands / command groups are external plugins in this PoC and have to be enabled manually.
5. To list all installed plugins run `snow plugin list`.
6. To enable a plugin run `snow enable plugin-name`.

What is contained in this PR and what is not (or: What this PR is and what it isn't):
* It is only PoC - it does not aim to be merged, so it can be out of sync with base branch (last rebase on Aug 4, 2023) and it is not adjusted to work with currently used github actions (mainly because of changes in directory tree)
* It proofs that:
    * [pluggy](https://pluggy.readthedocs.io/en/stable/) library can be used to introduce plugins mechanism. It allows us to have both built-in and external plugins and we have an idea how to do it.
    * We can use plugins to add both command groups and single commands
    * We can add plugin commands and command groups not only on top level (`snow plugin-name`) but we can also add them nested in existing command groups (for example: `snow snowpark plugin-name`).
    * We have an idea how we can manage plugins (enabled/disable/list) using our global SnowCLI config file.
    * We can load plugins safely, accepting failure of single plugins without influence on others.
    * We can verify plugin dependencies (including snowcli core) when loading a plugin and stop it if some dependecies are not met (for example if snowcli core does not match plugin's requirements).
    * We can validate if plugin is installed before enabling it by an user
* Logic of plugins loading and plugins management is not covered by automatic tests.
* Already existing tests (checked with unit tests) of command groups moved in this PoC to plugins are still working.

Changes:
* SnowCLI code moved to the new `core` directory (together with pyproject.toml)
* Created "plugin api" package in the core: `core/src/snowcli/plugin/api` containing API which needs to be implemented by plugins
* Created plugins loading logic: `core/src/snowcli/plugin` (used by `app.py` and `external_plugin_management.py`)
* Added logic of enabling / disabling / listing external plugins in SnowCLI config file (`core/src/snowcli/config.py`)
* Created commands helping with managing external plugins: `core/src/snowcli/cli/external_plugin_management.py`
* Modified default SnowCLI callback `core/src/snowcli/cli/app.py` to evaluate options (including config file) in order which will allow to use necessary options by plugins loading logic
* Created built-in plugin definition for `stage` command group (snowcli/cli/stage/stage_builtin_plugin_hook_impl.py)
* Created a new directory for simulation of external plugins (`plugins`) and following subdirectories in it:
  * `connection` (containing `connection` plugin) with own pyproject.toml
  * `snowpark-extensions` (containing `snowpark-function`, `snowpark-package`, `snowpark-say-hello` plugins) with own pyproject.toml
* Moved typer definitions of `connection`, `snowpark function` and `snowpark package` command groups to plugins subdirectories
* Added definition of a single command as plugin in `plugins/snowpark-extensions/snowcli_say_hello_plugin`
* Removed old registration `stage`, `connection`, `snowpark function` and `snowpark package` Typer instances, replaced with registration of built-in and external plugins (invocation of `load_and_register_plugins_in_typer from snowcli/plugin/plugins_registration.py` in `app.py`)


core/src/snowcli/config.py
```diff
@@ -74,11 +74,43 @@ def add_connection(self, name: str, parameters: dict):
         self._find_section("connections").add(name, parameters)
         self._dump_config()
 
+    # TODO: think about extracting plugin specific methods to another layer
```
sfc-gh-pjob, line 77: IMO we have to think about moving config helpers for plugin somewhere outside CliConfigManager.
The same about connections.

core/src/snowcli/__init__.py
sfc-gh-pjob, line 1: We have to discuss what we want to have in core.
Do we want to have everything in core and allow plugins to use everything (including built-in commands)?
Or maybe we just want to extract some commons (configuration, logging, connection) to core and have a separate package for snowcli as a tool which will depend on core and will have commands implementation as built-in plugins?

core/src/snowcli/plugin/api/__init__.py
```diff
@@ -0,0 +1,19 @@
+from dataclasses import dataclass
+from typing import List
+
+import pluggy
+from typer import Typer
+
+plugin_hook_spec = pluggy.HookspecMarker("snowcli.plugin")
+plugin_hook_impl = pluggy.HookimplMarker("snowcli.plugin")
```
sfc-gh-turbaszek, line 8: Let's name this marker as `snowcli.command` to avoid problems in future hooks for more than commands.

core/src/snowcli/cli/external_plugin_management.py
```diff
@@ -0,0 +1,123 @@
...
+            "command",
+            "package_name",
+            "package_version",
+            "verification_error",
+        ],
+    )
+
+
+@app.command(name="list")
+def list_plugins(
```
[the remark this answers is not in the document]
sfc-gh-turbaszek: Let's add `snow` keyword t command

core/src/snowcli/plugin/load_external_plugins.py
```diff
@@ -0,0 +1,83 @@
...
+    ExternalPluginsLoadingMode,
+    ExternalLoadedPlugin,
+    ExternalPluginPackageInfo,
+)
+from snowcli.plugin.api import plugin_hook_specs
+
+log = logging.getLogger(__name__)
+
+
+class LoadExternalPlugins:
```
sfc-gh-pjob, line 18: Comment from Barry:

We should think about possibility to configure the path where plugins will be loaded from to allow for installation of them in a specific location.

jamesweakley:
I have a use case for snowcli plugins and would be happy to try this out and provide feedback when the time is right

PR #314, opened 2023-08-17
Unify commands output

sfc-gh-astus:
### Pre-review checklist

### Changes description
...


src/snowcli/cli/streamlit/manager.py
```diff
@@ -99,13 +99,14 @@ def deploy(
...
-        else:
-            log.info(url)
+        # TODO fix it!
+        return base_url
+        # url = self._get_url(base_url, qualified_name)
+        #
+        # if open_in_browser:
+        #     typer.launch(url)
+        # else:
+        #     return url
```
sfc-gh-astus, line 109: I have no idea what to do here, `_get_url` throws error because in our host are only 3 parts

src/snowcli/output/decorators.py
```diff
@@ -1,14 +1,33 @@
...
 
 def with_output(func):
     @wraps(func)
     def wrapper(*args, **kwargs):
         result = func(*args, **kwargs)
-        print_db_cursor(result)
+        output_data = OutputData()
+
+        if result is None:
+            print_output(output_data)
```
sfc-gh-astus, line 15: Maybe make output_data optional?

tests/test_snow_connector.py
```diff
@@ -24,25 +21,22 @@ def __repr__(self):
...
-    mock_conn.return_value.execute_stream.return_value = [mock_cursor]
-    mock_conn.return_value.execute_string.return_value = [mock_cursor]
-    result = runner.invoke_with_config(cmd)
-    assert result.exit_code == 0, result.output
-    kwargs = mock_conn.call_args_list[-1][-1]
-    assert "application" in kwargs
-    assert kwargs["application"] == expected
+    SnowflakeConnector({})
+
+    mock_connector.connect.assert_called_once_with(application=expected)
```
sfc-gh-astus, line 39: This test was too complicated, I had to simplify it

src/snowcli/cli/streamlit/manager.py
sfc-gh-turbaszek: Let's fix it first before merging this PR. Wh we have this validation? Does it has to be always 4 part?

src/snowcli/output/decorators.py
```diff
@@ -1,14 +1,33 @@
...
 
     return wrapper
+
+
+def _is_list_of_results(result):
+    return (
+        isinstance(result, list)
+        and len(result) > 0
+        and (isinstance(result[0], list) or isinstance(result[0], SnowflakeCursor))
+    )
```
sfc-gh-turbaszek, line 33: I think we should support more general cases. The result should be iterable. But there's a lot of possibilites. So the question is - do we want to support generic types? Maybe we should always expect `OutputData`?

For example we can do
```py
return OutputData.from_cursor(result)

return OutputData.from_string("ok")

return OutputData.from_list([x for x in xs])

return OutputData.from_iterable(x for x in xs)
```

Each such implementation will have it's own checks and the only validation we do in `with_output` is to check if result is instance of `OutputData`.

Other thing to consider (in future work) is to extend OutputData  class with support for prompts. This would unify console output in 100%

src/snowcli/exception.py
```diff
@@ -23,3 +23,13 @@ def __init__(self, snowflake_err: Exception):
...
+
+
+class OutputDataTypeError(ClickException):
+    def __init__(self, got_type: type, expected_type: type):
+        super().__init__(f"Got {got_type} type but expected {expected_type}")
+
+
+class CommandReturnTypeError(ClickException):
+    def __init__(self, got_type: type):
+        super().__init__(f"Commads have to return OutputData type, but got {got_type}")
```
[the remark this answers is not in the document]
sfc-gh-turbaszek: I would even consider using RuntimeException to avoid handling this gracefully.

src/snowcli/cli/snowpark/jobs/manager.py
```diff
@@ -3,29 +3,34 @@
...
         spec=@{stage}/{stage_dir}/{spec_filename};
         """
         )
+        return OutputData.from_cursor(cursor)
 
-    def desc(self, job_name: str):
-        return self._execute_query(f"desc service {job_name}")
+    def desc(self, job_name: str) -> OutputData:
+        cursor = self._execute_query(f"desc service {job_name}")
+        return OutputData.from_cursor(cursor)
```
sfc-gh-turbaszek, line 25: I'm not sure about it. I would return cursor here and do output formating in the command. To me this would be a better separation of responsibilities. WDYT?

src/snowcli/cli/sql.py
```diff
@@ -1,15 +1,47 @@
...
+            sql = sys_input
+        else:
+            sql = query if query else file.read_text()  # type: ignore
+        result = self._execute_queries(sql)
+        if isinstance(result, List):
+            output_data = OutputData()
+            for single_result in result:
+                output_data.add_cursor(single_result)
+            return output_data
+        return OutputData.from_cursor(result)
```
sfc-gh-turbaszek, line 41: Imho this should be part of command not manager

src/snowcli/cli/snowpark/jobs/manager.py
sfc-gh-astus: It is good idea when there is only one result. I see problem when you would like to return few cursors and strings. I added methods for specific types, so you have to return type of results to properly add them in command

sfc-gh-astus: Or maybe it will not be that huge problem?

sfc-gh-turbaszek: I think that form test perspective and possible reusablity separating query executingand output constructing will be useful. 

To me this
```py
return OutputData.from_cursor(JobManager().desc(job_name))
```
is much more explicit and yields more information

sfc-gh-astus: Moved it to commands

src/snowcli/cli/snowpark/procedure_coverage/clear.py
```diff
@@ -44,8 +46,8 @@ def procedure_coverage_clear(
         ),
     )
     coverage_path = f"""{deploy_dict["directory"]}/coverage"""
-    results = StageManager(connection=conn).remove(
+    output_data = StageManager(connection=conn).remove(
         stage_name=deploy_dict["stage"], path=coverage_path
     )
     log.info("Deleted the following coverage results from the stage:")
-    print_db_cursor(results)
+    return output_data
```
sfc-gh-turbaszek, line 53:
```suggestion
    return OutputData.from_cursor(output_data)
```
?

src/snowcli/cli/sql.py
```diff
@@ -29,40 +58,17 @@ def execute_sql(
...
-
-    results = conn.ctx.execute_string(
-        sql_text=sql,
-        remove_comments=True,
-    )
-    output_data = OutputData()
-    for result in results:
-        output_data.add_data(result)
-    print_output(output_data)
+    cursors = SqlManager().execute(query, file)
```
sfc-gh-turbaszek, line 68: Doesn't it return iterator? Can't we always do
```py
for cursors in cursors:
   output_data.add_cursor(cursor)
```

src/snowcli/cli/stage/commands.py
```diff
@@ -21,15 +22,19 @@
...
     """
     manager = StageManager()
 
     if stage_name:
-        return manager.list(stage_name=stage_name)
-    return manager.show()
+        cursor = manager.list(stage_name=stage_name)
+    else:
+        cursor = manager.show()
+    return OutputData.from_cursor(cursor)
```
sfc-gh-turbaszek, line 37: Return cursors not OutputData

src/snowcli/cli/stage/commands.py
```diff
@@ -47,11 +52,12 @@ def stage_get(
...
     ),
     **options,
-):
+) -> OutputData:
     """
     Download all files from a stage to a local directory.
     """
-    return StageManager().get(stage_name=stage_name, dest_path=path)
+    cursor = StageManager().get(stage_name=stage_name, dest_path=path)
+    return OutputData.from_cursor(cursor)
```
sfc-gh-turbaszek, line 60: Same here

src/snowcli/cli/sql.py
sfc-gh-astus: Right

sfc-gh-turbaszek (approved):

PR #315, opened 2023-08-21
Added performance test

sfc-gh-astus:
### Pre-review checklist

### Changes description
...

### DIFF
--- a/tests/test_performance.py
+++ b/tests/test_performance.py
@@ -1,0 +1,17 @@
+import subprocess
+from timeit import default_timer as timer
+
+SAMPLE_AMOUNT = 20
+EXECUTION_TIME_THRESHOLD = 1.0
+
+
+def test_snow_help_performance():
+    results = []
+    for _ in range(SAMPLE_AMOUNT):
+        start = timer()
+        subprocess.run(["snow", "--help"], stdout=subprocess.DEVNULL)
+        end = timer()
+        results.append(end - start)
+
+    results.sort()
+    assert results[int(SAMPLE_AMOUNT * 0.9)] <= EXECUTION_TIME_THRESHOLD


sfc-gh-jsikorski (approved):

.github/workflows/performance_test.yaml
```diff
@@ -0,0 +1,30 @@
...
+          python-version: "3.11"
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -e .
+      - name: Run integration tests
+        env:
+          PYTHONPROFILEIMPORTTIME: "1"
+        run:
+          time snow --help
```
sfc-gh-turbaszek, line 30: I think this is more of a check than test, we don't compare with anything right? Maybe we should make it a normal pytest test? We can run snowcli via `subprocess` and measure the time. This gives us more elasticity in saving the results as well as comparing them

tests/teset_performance.py
```diff
@@ -0,0 +1,17 @@
+import subprocess
```
sfc-gh-turbaszek, line 1: There's typo in file name

tests/teset_performance.py
```diff
@@ -0,0 +1,17 @@
...
+from timeit import default_timer as timer
+
+SAMPLE_AMOUNT = 20
+EXECUTION_TIME_THRESHOLD = 0.55
+
+
+def test_snow_help_performance():
+    results = []
+    for i in range(0, SAMPLE_AMOUNT):
```
sfc-gh-turbaszek, line 10:
```suggestion
    for i_ in range(0, SAMPLE_AMOUNT):
```

As we don't use `i`

tests/teset_performance.py
sfc-gh-astus: fixed

tests/teset_performance.py
sfc-gh-astus: changed

sfc-gh-turbaszek (approved):

sfc-gh-turbaszek (approved):

PR #321, opened 2023-08-24
Unit tests for snowpark_shared

sfc-gh-jsikorski:
### Pre-review checklist

### Changes description
another portion of unit tests - this time for snowpark_shared module


tests/snowpark/test_snowpark_shared.py
```diff
@@ -0,0 +1,137 @@
...
+                input_parameters="()",
+                return_type="str",
+                replace=False,
+                install_coverage_wrapper=True,
+            )
+
+    assert (
+        "You cannot install a code coverage wrapper on a function, only a procedure."
+        in caplog.text
+    )
```
sfc-gh-astus, line 97: Instead of checking logs I would use mocks of connection, utils and check if proper functions are executed

tests/snowpark/test_snowpark_shared.py
```diff
@@ -0,0 +1,137 @@
...
+):
+
+    mock_parse.return_value = SplitRequirements(
+        [], [Requirement.parse("totally-awesome-package")]
+    )
+
+    mock_install.return_value = (True, None)
+
+    with caplog.at_level(logging.INFO, logger="snowcli.cli.snowpark_shared"):
+        result = shared.snowpark_package("yes", False, "yes")
```
sfc-gh-turbaszek, line 28: I'm personally any assertions on logs unless necessary. First I don't see logs as "business logic", second it's often suggest bad design. 

In this case I don't think the assertion is necessary. If we want to be sure something was execute then let's mock it and do assertion on the mock.

sfc-gh-jsikorski: Done

tests/snowpark/test_snowpark_shared.py
sfc-gh-jsikorski: Done

sfc-gh-pjob (approved):

sfc-gh-turbaszek (approved):

PR #322, opened 2023-08-24
Added format, verbose and debug to global options

sfc-gh-astus:
### Pre-review checklist

### Changes description
Moved format, verbose and debug options to global options

### DIFF
--- a/src/snowcli/cli/common/decorators.py
+++ b/src/snowcli/cli/common/decorators.py
@@ -2,8 +2,10 @@ from __future__ import annotations
 
 import inspect
 from functools import wraps
-from typing import Callable, Optional, get_type_hints
+from inspect import Signature
+from typing import Callable, Optional, get_type_hints, List
 
+from snowcli.cli import loggers
 from snowcli.cli.common.flags import (
     ConnectionOption,
     AccountOption,
@@ -13,7 +15,12 @@ from snowcli.cli.common.flags import (
     RoleOption,
     WarehouseOption,
     PasswordOption,
+    OutputFormatOption,
+    VerboseOption,
+    DebugOption,
 )
+from snowcli.cli.common.snow_cli_global_context import snow_cli_global_context_manager
+from snowcli.output.formats import OutputFormat
 
 
 def global_options(func: Callable):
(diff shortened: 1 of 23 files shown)


src/snowcli/cli/common/decorators.py
```diff
@@ -101,10 +150,16 @@ def _extend_signature_with_global_options(func):
...
             *GLOBAL_CONNECTION_OPTIONS,
+            *GLOBAL_OPTIONS,
+        ]
+    else:
+        parameters = [
+            *existing_parameters_with_evaluated_types,
+            *GLOBAL_OPTIONS,
         ]
-    )
+    sig = sig.replace(parameters=parameters)
```
sfc-gh-turbaszek, line 164: Instead of this if I would be in favor of passing array of prameters that should be added. WDYT?

src/snowcli/cli/common/flags.py
```diff
@@ -81,3 +83,26 @@
...
+    case_sensitive=False,
+    callback=update_callback("output_format"),
+)
+
+VerboseOption = typer.Option(
+    None,
+    "--verbose",
+    "-v",
+    help="Print logs from level info and higher",
+    callback=update_callback("verbose"),
```
sfc-gh-turbaszek, line 100: should use here `is_flag=True` to indicate this option is flag?

src/snowcli/cli/connection.py
```diff
@@ -39,7 +40,8 @@ def _mask_password(connection_params: dict):
 
 @app.command(name="list")
 @with_output
-def list_connections() -> OutputData:
+@global_options
+def list_connections(**kwargs) -> OutputData:
```
sfc-gh-turbaszek, line 44: Should we stick to convention of using `**options` instead of `**kwargs`?

src/snowcli/cli/common/decorators.py
sfc-gh-astus: It looks much better 👍

src/snowcli/cli/common/flags.py
sfc-gh-astus: I will add

src/snowcli/cli/connection.py
sfc-gh-astus: Do we have a convention? 😱

sfc-gh-astus: Changed and in few more places too

sfc-gh-turbaszek (approved):

PR #327, opened 2023-08-24
Project definition: snowflake.yml and snowflake.local.yml

sfc-gh-cgorrie:
### Pre-review checklist

### Changes description
Adds `snowflake.yml` and `snowflake.local.yml` validation as per the project definition spec. Looks up resolved role and warehouse so we can calculate certain optional values if they are not present in either file.

### TODO
* [x] test + actually use global connection / mock to get resolved warehouse + role


tests/project/test_config.py
```diff
@@ -0,0 +1,17 @@
+import os
+from pathlib import Path
+from tempfile import TemporaryDirectory
+from unittest import mock
+
+from snowcli.cli.project.config import load_project_config, load_local_config
+
+from tests.project.fixtures import *
+
+def test_project_1_config(with_project_1):
```
sfc-gh-bgoel, line 10: I assume we'll add the fixture as decorator with parameters for different test inputs later on, instead of passing it as an argument.

src/snowcli/cli/project/config.py
```diff
@@ -0,0 +1,50 @@
...
+    "artifacts": Seq(Path() | PathMapping)
+})
+
+project_schema = Map({
+    "native_app": app_schema,
+})
+
+local_schema = Map({
+    "package": Map({
+        "role": Regex(IDENTIFIER),
```
sfc-gh-bgoel, line 33: Why do we need a regex check here? I do understand that snowflake has a requirement around it: https://docs.snowflake.com/en/sql-reference/identifiers-syntax
But should we be performing this check ourselves instead of just passing it on to snowflake, and returning whatever error message (or lack of) that snowflake produces?

sfc-gh-bgoel (commented):
I like how this is coming along! I understand this is a PoC, but I've left some implementation-specific questions that will me better visualize your vision in the PoC 😃

tests/project/test_config.py
sfc-gh-cgorrie: Yeah, we can use the same `parameterize` trick later.

src/snowcli/cli/project/config.py
sfc-gh-cgorrie: I don't see these identifier requirements changing anytime, and this follows the maxim of failing fast.

pyproject.toml
```diff
@@ -18,6 +18,7 @@ dependencies = [
   "requests==2.31.0",
   "requirements-parser==0.5.0",
   "snowflake-connector-python[secure-local-storage]==3.1.0",
+  "strictyaml==1.7.3",
```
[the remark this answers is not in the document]
sfc-gh-cgorrie: Security updates, yes, but not major / minor versions.

sfc-gh-nkhalid (approved):

.gitignore
```diff
@@ -13,3 +13,4 @@ dist/
 ./.packages
 .coverage
 gen_docs/
+.vscode/settings.json
```
sfc-gh-cgorrie, line 16: Happy to take this out and commit the file, my settings:
```json
{
    "editor.formatOnSave": true,
    "[python]": {
        "editor.defaultFormatter": "ms-python.black-formatter"
    }
}
```

src/snowcli/cli/project/config.py
```diff
@@ -0,0 +1,123 @@
...
+    as_document,
+)
+from collections import OrderedDict
+
+IDENTIFIER = r'(?:("[^"]*(""[^"]*)*")|([A-Za-z_][\w$]{0,254}))'
+SCHEMA_AND_NAME = f"{IDENTIFIER}[.]{IDENTIFIER}"
+GLOB_REGEX = r"^[a-zA-Z0-9_\-./*?**\p{L}\p{N}]+$"
+RELATIVE_PATH = r"^[^/][\p{L}\p{N}_\-.][^/]*$"
+
+# TODO: use the above regexes to validate paths + globs
```
sfc-gh-cgorrie, line 23: Leaving this TODO in as actually resolving paths comes later; we can verify the regexes work then

src/snowcli/cli/project/config.py
```diff
@@ -0,0 +1,123 @@
...
+FilePath = Str
+Glob = Str
+
+
+class RelaxedMap(MapCombined):
+    """A version of a Map that allows any number of unknown key/value pairs."""
+
+    def __init__(self, map_validator):
+        super(RelaxedMap, self).__init__(map_validator, Str(), Bool() | Any())
```
sfc-gh-cgorrie, line 32: `Any()` was coercing unknown booleans to strings; this fixes that (if `Bool()` doesn't match, it re-attempts to validate with `Any()`).

.gitignore
sfc-gh-turbaszek: I would suggest to separate it from main changes. I'm fine with adding your settings to repo, it should not hurt.

pyproject.toml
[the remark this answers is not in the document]
sfc-gh-turbaszek: Why no major/minor updates?

tests/project/project_1/project.yml
```diff
@@ -0,0 +1,20 @@
+
+native_app:
```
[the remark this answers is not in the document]
sfc-gh-turbaszek: Few things to thinks about:
1. single vs many apps - if my project contains two apps should I have to ymls or one with two definitions?
2. native_app vs app - should we consider structure like this:
```yml
app:
  type: native_app
  ...
```

This brings all app-things like Streamlit, Native Apps and any future objects under one umbrella of "apps". Implementation-wise this shouldn't be any harder (schema of type one-of). Plus, if we decide to support "one app" per project then it makes the config more explicite. WDYT?

tests/project/project_1/project.yml
```diff
@@ -0,0 +1,22 @@
...
+  name: myapp
+
+  # the following shows how to properly use a quoted identifier in YAML
+  source_stage: >-
+    "MySourceSchema"."SRC_Stage"
```
sfc-gh-turbaszek, line 6: Why we limit our selves to only schema.object identifier instead of db.schema.object?

tests/project/project_1/project.yml
```diff
@@ -0,0 +1,22 @@
...
+        - java/deps/myDep.jar
+    - dest: ext_code/
+      src: python/build/my-package-0.0.0.wheel
+    - setup.sql
+    - manifest.yml
+    - app/README.md
+    - src: app/streamlit/*.py
+      dest: streamlit/
+
+  scripts:
```
sfc-gh-turbaszek, line 21: I'm not sure this is the best name. If we consider allowing "custom scripts" as in `npm run` then scripts key would the best one to use.
```yml
scripts:
   name: build
   cmd: gradle build
```

src/snowcli/cli/project/config.py
```diff
@@ -0,0 +1,123 @@
...
+def clean_identifier(input):
+    """
+    Removes characters that cannot be used in an unquoted identifier,
+    converting to lowercase as well.
+    """
+    return re.sub(r"[^a-z0-9_$]", "", f"{input}".lower())
+
+
+def generate_local_config(project: OrderedDict, conn: dict) -> OrderedDict:
+    user = clean_identifier(os.getenv("USER"))
```
sfc-gh-bgoel, line 106: It seems like `USER` may not be defined in every possible system, in which case we should search for alternatives. This post shows possible alternatives: https://unix.stackexchange.com/a/502361
We should probably also have a default value/fallback: `anonUser` or something along those lines.

tests/project/project_1/project.yml
sfc-gh-bgoel: We are uploading all artifacts to the schema within the application package db. We already know the name of the app package that will be, hence schema and stage are the unknowns.

tests/project/project_1/project.yml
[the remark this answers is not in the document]
sfc-gh-bgoel: IMO for Point 2 - @sfc-gh-turbaszek  what do you envision for bringing more products under the umbrella that are non-app (outside streamlit, native apps) such as SPCS? They would not fit under `app:type:`. And if we define a unified yml file for all these, would `spcs:` belong at the same level as `app:`?

sfc-gh-bgoel (commented):
Added few more comments + additional ones on our thread internally.

pyproject.toml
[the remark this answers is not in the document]
sfc-gh-cgorrie: I was assuming "us" meant NADE. I meant that we should be responsible for security concerns but that we should not commit to keeping `strictyaml` up-to-date, unless there is a good reason for it.

tests/project/project_1/project.yml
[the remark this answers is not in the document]
sfc-gh-cgorrie: In light of today's discussion I want to propose this way forward:
- `native_app:` stays as-is
- the value of "type" is used as the top-level keys, e.g. `streamlit:`

For multiple applications, they should live in different directories of the same repo, with their own `snowflake.yml` files. Use cases like that of the connectors team can be supported by using different overrides files; because they can change any aspect of the project config (as the schema is identical), apps that share most things can be deployed that way, e.g. `cd connectors/ && snow app run -e servicenow.yml`

.gitignore
sfc-gh-cgorrie: Will do!

tests/project/project_1/project.yml
sfc-gh-cgorrie: I think these would end up in the top-level of the file, not under the `native_app:` category. But we can definitely come up with a better name. These are SQL scripts that are executed after the app package is created, but before it is installed, to define shared content (i.e. schemas on the package that use reference_usage). Any suggestions?

sfc-gh-cgorrie: I ended up putting it under "package"; made some sense once we merged the schemas as we already had the key. But I'm also open to `package_scripts`, `package.exec_scripts`.

Could also do `package.shared`?

src/snowcli/cli/project/config.py
```diff
@@ -0,0 +1,76 @@
...
+
+    return config.data
+
+
+def generate_local_override_yml(project: dict | YAML) -> YAML:
+    user = clean_identifier(get_env_username() or DEFAULT_USERNAME)
+    role = snow_cli_global_context_manager.get_single_value("select current_role()")
+    warehouse = snow_cli_global_context_manager.get_single_value(
+        "select current_warehouse()"
+    )
```
sfc-gh-turbaszek, line 61: This can be probably extracted to more general functionality.  For example it can be useful when adding new connection to populate defaults fro warehouse and role.  WDYT?

src/snowcli/cli/common/snow_cli_global_context.py
```diff
@@ -88,8 +88,17 @@ def update_global_context(
     def get_connection(self):
         return self.get_global_context_copy().connection.build_connection()
 
+    def execute_string(self, *args, **kwargs):
+        return self.get_connection().ctx.execute_string(*args, **kwargs)
 
-def _create_snow_cli_global_context_manager_with_default_values() -> SnowCliGlobalContextManager:
+    def get_single_value(self, query: str):
+        cursor = self.execute_string(query)
+        return cursor.fetchone()[0]
```
sfc-gh-turbaszek, line 96: I think the direction is fine. But I would suggest to add those methods to `ConnectionDetails` class (and probably rename it?) so we have better separation of responsibilities. The usage would be:
```py
snow_cli_global_context_manager.connection.execute_string(...)
```

src/snowcli/cli/project/config.py
sfc-gh-cgorrie: Could you expand on that use case? Are you thinking about building a new connection profile based on an existing live connection here?

sfc-gh-bgoel (approved):
LGTM for all NADE centric code. Bunch of comments are good to resolve post the configuration discussion this last Monday. 
I defer to the snowCLI team's judgment ultimately before merge, especially for config related code.

src/snowcli/cli/common/snow_cli_global_context.py
sfc-gh-cgorrie: Thinking more about this, I think these are both the wrong solutions, because right now a new connection will be established with every `execute_string` call.

Does it make sense to have a global connection singleton? How are we preventing multiple connections from being made over the entire lifecycle of a SnowCLI invocation right now?

pyproject.toml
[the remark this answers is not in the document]
sfc-gh-bwarsaw: That's one of the problems with version pinning.  It helps in the short term, but causes increasing tech debt over time that is often very difficult or time consuming to pay down.

src/snowcli/cli/common/snow_cli_global_context.py
sfc-gh-bwarsaw: What about wrapping this function or some other in a `@cache` or `@cache_property` [decorator](https://docs.python.org/3/library/functools.html#functools.cache)?

src/snowcli/cli/project/schema.py
```diff
@@ -0,0 +1,96 @@
...
+Glob = Str
+
+
+class RelaxedMap(MapCombined):
+    """
+    A version of a Map that allows any number of unknown key/value pairs.
+    """
+
+    def __init__(self, map_validator):
+        super(RelaxedMap, self).__init__(
```
sfc-gh-bwarsaw, line 33: This can just be `super().__init__(`

sfc-gh-bwarsaw (commented):
Comments and suggestions

pyproject.toml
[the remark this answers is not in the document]
sfc-gh-cgorrie: Yep, definitely the age old debate. But considering this new requirement is doing the same as a lot of others, can I resolve this conversation? What would folks like to see changed?

src/snowcli/cli/common/snow_cli_global_context.py
sfc-gh-cgorrie: Definitely could do. The global context config never changes once it's set up then, I assume?

pyproject.toml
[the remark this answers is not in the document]
sfc-gh-bwarsaw: Definitely don't change anything in this PR. It is part of a larger conversation.

src/snowcli/cli/common/snow_cli_global_context.py
sfc-gh-bwarsaw: I think so, but I'm not 100% sure.  CC @sfc-gh-mkeller

sfc-gh-cgorrie: I saw an `update_global_context` function, so I made it clear the cached connection if that's ever called. Felt clearer to implement by hand rather than with `@cache_property` / `del self.get_connection`, lmk what you think!

sfc-gh-bwarsaw: Ah, so if I'm reading that correctly, you just cache it in an instance variable?  If so SGTM.

tests/project/fixtures.py
```diff
@@ -0,0 +1,70 @@
...
+    def test_my_project(project_config_files):
+        config = load_project_config(project_config_files)
+    """
+    project_dir = request.param
+    with snowflake_ymls(project_dir) as ymls:
+        yield ymls
+
+
+@pytest.fixture
+def mock_global_connection():
```
sfc-gh-cgorrie, line 57: Can I type this like ` -> Generator[Mock[SnowflakeConnector]]`? Would that work?

src/snowcli/cli/project/config.py
```diff
@@ -0,0 +1,80 @@
...
+
+def generate_local_override_yml(project: dict | YAML) -> YAML:
+    """
+    Generates defaults for optional keys in the same structure as the project
+    schema. The returned YAML object can be saved directly to a file, if desired.
+    A connection is made using global context to resolve current role and warehouse.
+    """
+    conn = snow_cli_global_context_manager.get_connection()
+    role = conn.get_single_value("select current_role()")
+    warehouse = conn.get_single_value("select current_warehouse()")
```
sfc-gh-turbaszek, line 64: I'm wondering if we need this. Won't we get those values from `conn.ctx.role` and `conn.ctx.warehouse`? I'm not sure I correctly understand `SnowfakeConnection` implementation from the connector but it seems those fields are being populated when establishing connection.

tests/project/fixtures.py
sfc-gh-turbaszek: Have you considered using `@mock.patch` instead of custom fixture?

sfc-gh-cgorrie: Yes, that's how I had it before. However, it seems like mocking out a connection and what gets return from executing SQL statements would be a very common use case, so I moved it into a fixture. I can always move it back, but I think we should be establishing some best practices for connection mocking at some point here.

src/snowcli/cli/project/config.py
sfc-gh-cgorrie: Yes, it does seem that way after reading through relevant source code -- I will use those instead.

sfc-gh-turbaszek (dismissed):

src/snowcli/cli/project/definition.py
```diff
@@ -0,0 +1,80 @@
...
+    project_schema,
+    project_override_schema,
+)
+
+from snowcli.cli.common.snow_cli_global_context import snow_cli_global_context_manager
+
+DEFAULT_USERNAME = "unknown_user"
+
+
+def merge_left(target: dict | YAML, source: dict | YAML) -> None:
```
sfc-gh-turbaszek, line 20:
```suggestion
def merge_left(target: Union[Dict, YAML], source: Union[Dict, YAML]) -> None:
```
Consider introducing a custom type for `Union[Dict, YAML]`

src/snowcli/cli/project/definition.py
```diff
@@ -0,0 +1,80 @@
...
+            overrides = load(override_yml.read(), project_override_schema)
+            merge_left(definition, overrides)
+
+        # TODO: how to show good error messages here?
+        definition.revalidate(project_schema)
+
+    return definition.data
+
+
+def generate_local_override_yml(project: dict | YAML) -> YAML:
```
sfc-gh-turbaszek, line 56: Same here

sfc-gh-turbaszek (approved):

PR #328, opened 2023-08-25
Changed asserts on commands output to json

sfc-gh-astus:
### Pre-review checklist

### Changes description
Changed asserts to check commands responses in JSON format, where it was possible

### DIFF
--- a/src/snowcli/cli/snowpark/procedure/commands.py
+++ b/src/snowcli/cli/snowpark/procedure/commands.py
@@ -36,11 +36,13 @@ app.add_typer(procedure_coverage_app)
 
 
 @app.command("init")
-def procedure_init() -> None:
+@with_output
+def procedure_init() -> OutputData:
     """
     Initialize this directory with a sample set of files to create a procedure.
     """
     create_project_template("default_procedure")
+    return OutputData()
 
 
 @app.command("create")
@@ -148,6 +150,7 @@ def procedure_create(
 
 
 @app.command("update")
+@with_output
 def procedure_update(
     environment: str = ConnectionOption,
     pypi_download: str = PyPiDownloadOption,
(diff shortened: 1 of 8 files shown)


sfc-gh-jsikorski (dismissed):

src/snowcli/cli/snowpark/procedure/commands.py
```diff
@@ -36,11 +36,13 @@
...
 
 @app.command("init")
-def procedure_init() -> None:
+@with_output
+def procedure_init() -> OutputData:
     """
     Initialize this directory with a sample set of files to create a procedure.
     """
     create_project_template("default_procedure")
+    return OutputData()
```
sfc-gh-turbaszek, line 45: Let's return some message, "Project initialised" or something

src/snowcli/cli/snowpark/procedure/commands.py
```diff
@@ -218,20 +221,23 @@ def procedure_update(
...
     package_native_libraries: str = PackageNativeLibrariesOption,
-):
+) -> OutputData:
     """Packages procedure code into zip file."""
     snowpark_package(
         pypi_download,  # type: ignore[arg-type]
         check_anaconda_for_pypi_deps,
         package_native_libraries,  # type: ignore[arg-type]
     )
+    return OutputData()
```
sfc-gh-turbaszek, line 240: How does it works? First we will do `print_db_cursor` and later empty output?

src/snowcli/cli/snowpark/procedure/commands.py
sfc-gh-astus: For `OutputData()` we print `Done`. Is it ok?

src/snowcli/cli/snowpark/procedure/commands.py
sfc-gh-astus: This command don't print cursor and `Output()` prints `Done`

Btw I removed `print_db_cursor`

sfc-gh-turbaszek (approved):
I'm ok with this change. However, the empt `OutpurData()` is bit ambiguous and we should really think about introducing some common messages.

PR #332, opened 2023-08-25
Allow glob-style paths for snow stage put

### LINKED ISSUE #331: SNOW-900602: Cannot put "folder/*.py" to a stage because it's not a valid Path
### Python version

Python 3.8.12

### Platform

macOS-13.5-arm64-arm-64bit

### What happened

I should be able to put `folder/*.py` to a stage, and have it only upload all the .py files from that folder to a stage.

```bash
snow stage put "shared_modules/*.py" snow://streamlit/streamlit.sis_apps.test_app_recursive/default_checkout/shared_modules
```

This should be perfectly valid https://docs.snowflake.com/en/sql-reference/sql/put#examples, but it fails because Typer is checking to see that the path is valid.

If you put `exists=False` it works fine, and uploads all of the .py files in that folder to the stage. If you put an invalid folder, you still get a File doesn't exist message, just from SQL rather than Typer.

### Console output

```Shell
╭─ Error ───────────────────────────────────────────────────────────────────────────────────────────╮
│ Invalid value for 'PATH': Path 'shared_modules/*.py' does not exist.                              │
╰───────────────────────────────────────────────────────────────────────────────────────────────────╯
```
```


### How to reproduce

Try to `snow put` a path with `*` in it

### PR DESCRIPTION
**Pull request by sfc-gh-zblackwood**
### Pre-review checklist

### Changes description

Fixes #331 by not requiring Typer to enforce that the path that is passed actually resolves to a valid local path, and allow glob-style paths to be passed directly.

### DIFF
--- a/src/snowcli/cli/stage/commands.py
+++ b/src/snowcli/cli/stage/commands.py
@@ -90,7 +90,7 @@ def stage_put(
     Upload files to a stage from a local client
     """
     manager = StageManager()
-    local_path = str(path) + "/*" if Path(path).is_dir() else str(path)
+    local_path = str(path) + "/*" if path.is_dir() else str(path)
 
     cursor = manager.put(
         local_path=local_path, stage_path=name, overwrite=overwrite, parallel=parallel
--- a/src/snowcli/cli/stage/commands.py
+++ b/src/snowcli/cli/stage/commands.py
@@ -66,6 +66,10 @@ def stage_put(
     path: Path = typer.Argument(
         ...,
         exists=False,
+        file_okay=True,
+        dir_okay=True,
+        writable=True,
+        resolve_path=True,
         help=(
             "File or directory to upload to stage, can include a * in the path, "
             "like folder/*.csv"
--- a/src/snowcli/cli/stage/commands.py
+++ b/src/snowcli/cli/stage/commands.py
@@ -72,7 +72,8 @@ def stage_put(
         resolve_path=True,
         help=(
             "File or directory to upload to stage, can include a * in the path, "
-            "like folder/*.csv"
+            'like "folder/*.csv". Make sure you put quotes around the path if it'
+            " includes a *. "
         ),
     ),
     name: str = StageNameOption,
--- a/src/snowcli/cli/stage/commands.py
+++ b/src/snowcli/cli/stage/commands.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 from pathlib import Path
 
 import typer
-
 from snowcli.cli.common.decorators import global_options_with_connection
 from snowcli.cli.common.flags import DEFAULT_CONTEXT_SETTINGS
 from snowcli.cli.stage.manager import StageManager
@@ -66,12 +65,11 @@ def stage_get(
 def stage_put(
     path: Path = typer.Argument(
         ...,
-        exists=True,
-        file_okay=True,
-        dir_okay=True,
-        writable=True,
-        resolve_path=True,
-        help="File or directory to upload to stage",
+        exists=False,
+        help=(
+            "File or directory to upload to stage, can include a * in the path, "
+            "like folder/*.csv"
+        ),
     ),
     name: str = StageNameOption,
     overwrite: bool = typer.Option(
@@ -88,7 +86,7 @@ def stage_put(
     Upload files to a stage from a local client
     """
     manager = StageManager()
-    local_path = str(path) + "/*" if path.is_dir() else str(path)
+    local_path = str(path) + "/*" if Path(path).is_dir() else str(path)
 
     cursor = manager.put(
         local_path=local_path, stage_path=name, overwrite=overwrite, parallel=parallel
--- a/tests/test_stage.py
+++ b/tests/test_stage.py
@@ -85,3 +85,26 @@ def test_stage_remove(mock_execute, runner, mock_cursor):
     )
     assert result.exit_code == 0, result.output
     mock_execute.assert_called_once_with("remove @stageName/my/file/foo.csv")
+
+
+@mock.patch(f"{STAGE_MANAGER}._execute_query")
+def test_stage_put_star(mock_execute, runner, mock_cursor):
+    mock_execute.return_value = mock_cursor(["row"], [])
+    with TemporaryDirectory() as tmp_dir:
+        result = runner.invoke_with_config(
+            [
+                "stage",
+                "put",
+                "-c",
+                "empty",
+                "--overwrite",
+                "--parallel",
+                42,
+                str(tmp_dir) + "/*.py",
+                "stageName",
+            ]
+        )
+    assert result.exit_code == 0, result.output
+    mock_execute.assert_called_once_with(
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/snowcli/cli/stage/commands.py [hunk 1]
```diff
@@ -66,12 +65,11 @@ def stage_get(
 def stage_put(
     path: Path = typer.Argument(
         ...,
-        exists=True,
-        file_okay=True,
-        dir_okay=True,
-        writable=True,
-        resolve_path=True,
-        help="File or directory to upload to stage",
+        exists=False,
```
sfc-gh-pjob, line 68: Hmmm, at the beginning I was wondering if it's a good solution. I thought that maybe we should accept a list of files and let a shell do glob expansion (snow `stage put ./*.py stageName` translated to list of paths in Typer arg). But then I looked at [SQL docs for PUT operation](https://docs.snowflake.com/en/sql-reference/sql/put) and I couldn't find any way to upload these multiple files without a glob pattern, what can mean that we cannot expand a pattern before passing it to SQL.

So, if I understand it correctly (please correct me if you see sth wrong), then:
* Your current solution will work for `snow stage put "./*.py" stageName` (path in quotes) on MacOS and Linux
* Your current solution may not work for `snow stage put "./*.py" stageName` (path in quotes) on Windows
   * I didn't test it but I found GitHub discussion and PR to Click (base framework for Typer) adding custom glob extension for Windows:
     * https://github.com/pallets/click/issues/1096
     * https://github.com/pallets/click/pull/1830
* Your current solution may not work for `snow stage put ./*.py stageName` (path not in quotes)
   * I didn't test it but I think it won't work because:
      * Linux and MacOS shells will do glob expansion before passing arguments to SnowCLI
      * Click may do some magic on Windows
* Previous implementation couldn't allow for glob patterns but worked for both quoted and unquoted paths.

IMO we should confirm how your implementation works for the following cases:
* Linux / MacOS + path with glob pattern and without quotes
* Windows + path with glob pattern and with quotes
* Windows  + path with glob pattern and without quotes

tests/test_stage.py
```diff
@@ -85,3 +85,26 @@ def test_stage_remove(mock_execute, runner, mock_cursor):
     )
     assert result.exit_code == 0, result.output
     mock_execute.assert_called_once_with("remove @stageName/my/file/foo.csv")
+
+
+@mock.patch(f"{STAGE_MANAGER}._execute_query")
+def test_stage_put_star(mock_execute, runner, mock_cursor):
```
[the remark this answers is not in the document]
sfc-gh-pjob: Sorry I meant just below `test_stage_put` :)

src/snowcli/cli/stage/commands.py
sfc-gh-pjob: I forgot to write that at the moment we need to test these cases manually because our tests use Click's CliRunner, so they don't execute real commands from real shells. It's good enough for most cases but in this case IMO we have to be sure how it really works in real shell.

sfc-gh-zblackwood: I can confirm that it does not work on mac without quotes due to the expansion you mentioned. I don't have a good way to test it on Windows, but agree that we should test if the behavior works on Windows. My guess is that it would work on Windows with or without quotes, since Windows doesn't expand glob-style paths by default, but I don't know for sure how Click handles paths like that.

My take on this is that, it would be great to support all of the cases, but at a minimum this doesn't remove any old functionality, and at last on Mac OS allows for new functionality (which I need for my use-case).

One reason why this is particularly important is because of the unexpected (to me) behavior of doing stage put with a folder name -- if the folder has other folders nested in it, it fails, because it tries to upload all of the nested folders as well as the files.

tests/test_stage.py
```diff
@@ -85,3 +85,26 @@ def test_stage_remove(mock_execute, runner, mock_cursor):
...
+                "empty",
+                "--overwrite",
+                "--parallel",
+                42,
+                str(tmp_dir) + "/*.py",
+                "stageName",
+            ]
+        )
+    assert result.exit_code == 0, result.output
+    mock_execute.assert_called_once_with(
```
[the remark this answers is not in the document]
sfc-gh-pjob: I don't think we have this issue with other tempfolder-based tests.

I have just tried to run `test_stage_put` on the main branch and it works there.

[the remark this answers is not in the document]
sfc-gh-zblackwood: I updated both put tests to drop the /private/

tests/test_stage.py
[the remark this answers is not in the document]
sfc-gh-zblackwood: Yes, moved

tests/test_stage.py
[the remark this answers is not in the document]
sfc-gh-pjob: IMO we should be sure in tests what are actual and expected paths and we shouldn't do this drop - especially if it worked before.

I found out that previously actual value was with `/private` (`/var/...` was being resolved to `/private/var/...` thanks to `resolve_path=True` in Typer's argument configuration). That is why the test case worked before.

Your changes removed `resolve_path=True` and now actual value is `/var/...` (not resolved to `/private/var/...`).
Is it necessary to remove `resolve_path=True` to have your changes working? I tried to add it back (combined with commenting out `_resolve_tmp_dir` in tests) and your new test still worked but maybe there is sth I don't know about?

src/snowcli/cli/stage/commands.py
sfc-gh-pjob: I understand your point of view, but I am afraid of adding a functionality which may not work on all platforms and which may be not so easy to use (and to document) because of possible different behaviours described above.

The first issue may be resolved by tests (probably manual tests at the moment). 

The second one is more difficult. Most CLI tools usually do not need quotes to use glob patterns for files. Even more - users cannot use quotes because shell won't expand the pattern if they do. 
Maybe a good help message could be a solution. Such which will mention that you to have use quotes when using a glob pattern for this argument. And such which will include info about Windows case if it turns out that glob patterns don't work on Windows.

@sfc-gh-turbaszek What do you think about it?

sfc-gh-pjob:
Note about integration tests:
We still don't have a solution for running integration tests from forks but I executed them locally and everything is ok.

src/snowcli/cli/stage/commands.py
sfc-gh-pjob: One more note about Windows:

Generally you shouldn't have to figure out how to test it on windows. It should be on us - because it won't be the last PR with changes which may not work on Windows and we have to take care of it. And we have planned a task to add support for Windows in automatic tests executed from GitHub actions but unfortunately we haven't taken care of it yet.

sfc-gh-pjob: @sfc-gh-zblackwood 

Ok, I talked with @sfc-gh-turbaszek and we decided to follow your take on this and merge the PR as it is, even without tests of glob patterns on Windows. Let's just change the help text for the argument to make it clear that users have to enclose the path in quotes when using glob patterns.

sfc-gh-zblackwood: Updated the note on the field -- how does that look?

sfc-gh-turbaszek (approved):

sfc-gh-pjob (approved):

PR #334, opened 2023-08-28
[SNOW-896865] Add Default Native Apps Project

sfc-gh-bgoel:
### Pre-review checklist
   * [ ] I've added new automated tests to verify correctness of my new code. (not required for this PR)
   * [x] I've confirmed that my changes are up-to-date with the target branch (last rebase Aug 28, 9AM EST)

### Changes description
This PR introduces the default native apps project, which will be added to a user's local filesystem when invoked with `snow app init <project_name>`. The implementation of `snow app init` to come in the following PRs. 
There are no tests needed for this PR, they will be a part of the `snow app init` command described above. 
Update: This PR will not be merged to SnowCLI main, and instead will reside under a separate repository. Keeping this PR as a placeholder to get reviews while we wait on certain permissions on the other side.


src/templates/default_nativeapp/app/setup_script.sql
```diff
@@ -0,0 +1,48 @@
+CREATE OR REPLACE APPLICATION ROLE app_instance_role;
```
sfc-gh-bgoel, line 1: The case of sql file is decided by examples taken from https://docs.snowflake.com/en/developer-guide/native-apps/creating-setup-script. We should stay consistent with the official snowflake documentation.

src/templates/default_nativeapp/app/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+For a deep dive into the world of Native Apps, please read the [official Snowflake documentation](https://docs.snowflake.com/en/developer-guide/native-apps/native-apps-about) which goes in depth about many additional functionalities of this framework.
+
+## Using the application after installation
+To interact with the application after it has successfully installed in your account, switch to the application owner role first.
+
+### Calling a stored procedure
+
+```
+CALL <application_name>.versioned_schema.area_of_square_sproc(2);
```
sfc-gh-turbaszek, line 13: Have you considered templating `<application_name>` so it works out of the box after generating a project?

src/templates/default_nativeapp/app/setup_script.sql
```diff
@@ -0,0 +1,48 @@
...
+        non_versioned_schema.id_table VALUES (1);
+
+    GRANT SELECT
+        ON non_versioned_schema.id_table
+        TO APPLICATION ROLE app_instance_role;
+
+
+CREATE OR ALTER VERSIONED SCHEMA versioned_schema;
+
+    CREATE OR REPLACE
```
sfc-gh-turbaszek, line 20: Is there any reason for different indentation?

src/templates/default_nativeapp/README.md
```diff
@@ -0,0 +1,32 @@
+## Introduction
+
+This is the default project template for a Snowflake Native Apps project. It contains minimal code meant to help you set up your first application instance in your account quickly.
+
+### Project Structure
+| File Name | Purpose |
+| --------- | ------- |
+| ./README.md | The current file you are looking at, meant to guide you through a native apps project. |
+| app/setup_script.sql | Contains SQL statements that are run when an account installs or upgrades an application. |
```
sfc-gh-turbaszek, line 9: I assume this is initial work. Do you plan to bootstrap python, java or scala project? Having a real "project structure" would be awesome.

sfc-gh-bgoel: There are two snowflake-provided templates for our MVP:
1. This PR aims to provide a barebones template, with no extension code, i.e. python/java/scala code. This shows all the code residing in the setup_script.sql. The reason is to provide a lean native app project that is one-click ready to be installed in the snowflake account. 
2. There will be an additional project template that contains a structure that shows the extension code files. This will be provided as part of a separate ticket under our epic.

src/templates/default_nativeapp/app/README.md
sfc-gh-bgoel: Updated, thanks!

src/templates/default_nativeapp/app/setup_script.sql
sfc-gh-bgoel: It is for ease of reading. Indentation exists for objects that belong to the same schema.

src/templates/default_nativeapp/README.md
```diff
@@ -0,0 +1,29 @@
...
+  role: <your_app_pkg_owner_role, resolved connection role* by default>
+  name: <name_of_app_pkg, project_name_pkg_$USER by default>
+
+application:
+  role: <your_app_owner_role, accountadmin by default>
+  name: <name_of_app, project_name_$USER by default>
+  debug: <true|false, true by default>
+
+```
+resolved connection role* - If you installed snowCLI correctly, you should contain a global [config.toml](https://docs.snowflake.com/LIMITEDACCESS/snowcli/connecting/connect#how-to-add-snowflake-credentials-using-a-configuration-file) file where you specify a connection, and a user for that connection. The connection role is derived from this user, or via a role override within the config.toml file.
```
sfc-gh-cgorrie, line 29: Suggestion: "If SnowCLI was installed correctly [...]"

src/templates/default_nativeapp/snowflake.yml.jinja
```diff
@@ -0,0 +1,9 @@
+native_app:
+  name: {{project_name}}
+
+  source_stage: app_src.stage
```
sfc-gh-cgorrie, line 4: We don't necessarily need this line as it's the default, but could be good for learning.

src/templates/default_nativeapp/app/manifest.yml
```diff
@@ -0,0 +1,7 @@
+artifacts:
```
sfc-gh-cgorrie, line 1: Suggest adding a `manifest_version` field here.

src/templates/default_nativeapp/.gitignore
```diff
@@ -0,0 +1 @@
+/snowflake.*.yml
```
sfc-gh-cgorrie, line 1: I think we should specifically gitignore just the local one.

sfc-gh-bgoel: I'm good with that, want to voice one thing: if user decides to name their file `snowflake.debug.yml` instead of local, then they might end up committing that. 
But I also take your point that we want to be able to commit override files that target specific connections. In which case, having those files is more important than having accidentally checked in a local/debug file, which can always be removed.

src/templates/default_nativeapp/snowflake.yml.jinja
sfc-gh-bgoel: +1, hence would like to keep this here.

src/templates/default_nativeapp/.gitignore
sfc-gh-cgorrie: Let's revisit after MVP. I have a feeling we'll likely be hardcoding the name `snowflake.local.yml`, and when we introduce e.g. an `-e` command line switch we can start to talk about what names make sense here.

src/templates/default_nativeapp/app/README.md.jinja
```diff
@@ -0,0 +1,22 @@
...
+
+For a deep dive into the world of Native Apps, please read the [official Snowflake documentation](https://docs.snowflake.com/en/developer-guide/native-apps/native-apps-about) which goes in depth about many additional functionalities of this framework.
+
+## Using the application after installation
+To interact with the application after it has successfully installed in your account, switch to the application owner role first.
+
+### Calling a stored procedure
+
+```
+CALL {{application_name}}.versioned_schema.area_of_square_sproc(2);
```
sfc-gh-cgorrie, line 13: So the way we will interpolate vars in will be by taking all keys (recursively) under `native_app:` and replacing `.`s with `_`s?

sfc-gh-turbaszek (approved):

PR #350, opened 2023-09-04
Output types

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
...

### DIFF
--- a/src/snowcli/output/types.py
+++ b/src/snowcli/output/types.py
@@ -1,0 +1,84 @@
+from __future__ import annotations
+import typing as t
+
+from snowflake.connector.cursor import SnowflakeCursor
+
+
+class CommandResult:
+    @property
+    def result(self):
+        raise NotImplemented()
+
+
+class ObjectResult(CommandResult):
+    def __init__(self, element: t.Dict):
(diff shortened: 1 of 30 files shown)


src/snowcli/output/types.py
```diff
@@ -0,0 +1,67 @@
+from __future__ import annotations
+import json
+import typing as t
+
+from snowflake.connector.cursor import SnowflakeCursor
+
+
+class Object:
```
sfc-gh-astus, line 8: Maybe all classes from this module should have prefix `Result`? WDYT?

I think Message and Stream are too general

sfc-gh-turbaszek: How about suffix instead of prefix? `MessageResult`, `StreamResult` or `SingleQueryResult` wdyt?

sfc-gh-astus: Yes, I meant suffix 🤦

sfc-gh-astus (approved):

sfc-gh-astus (approved):

sfc-gh-astus (approved):

PR #369, opened 2023-09-07
[NADE] Add Support for Project Definition Files + Native App Initialize Project

sfc-gh-cgorrie:
### Pre-review checklist
* [x]  I've confirmed that instructions included in README.md are still correct after my changes in the codebase. _**There are intentionally no changes to the README.md yet as our commands need to stay hidden while they are still in progress**_
* [x]  I've added new automated tests to verify correctness of my new code.
* [x]  I've confirmed that my changes are up-to-date with the target branch. [as of Sept 11, 1PM EST]
* [x]  I've described my changes in the section below.

### Changes description
This PR introduces two different, but related sets of changes:

1. Project Definition Files: Residing in `src/snowcli/cli/project/` and its corresponding unit test folder. These changes allow for a `snowflake.yml` to be present, read and validated by a `DefinitionManager`. There is also support for allowing overrides on this `snowflake.yml` based on other files, such as a `snowflake.local.yml`.
2. Native App Initialize Project: Residing in `src/snowcli/cli/nativeapp/` and its corresponding unit test folder. These changes allow for a user to use `snow app init <project-name>` command without specifying any template as an argument. The templates are currently under review in a separate public repository by the legal team. 

Couple of Notes:
 
1. The `snow app` sub-command needs to stay as `hidden=True` while the work on this is still in progress for this quarter.
2. While there are unit tests for every single change introduced, there are no integration tests yet for this one initialize command. They will be added as NADE makes more progress on the number of commands.
3. There is also a refactor on `src/snowcli/cli/render.py` which was introduced post discussion with @sfc-gh-turbaszek
 
All the PRs within NADE's fork have been internally reviewed by the team. This PR is to seek feedback of the main SnowCLI team before merging to their repository.


src/snowcli/cli/nativeapp/artifacts.py
```diff
@@ -0,0 +1,175 @@
...
+    def __str__(self):
+        return (
+            self.__doc__
+            + f"\ndest_path = {self.dest_path}"
+            + f"\ndeploy_root = {self.deploy_root}"
+        )
+
+
+@dataclass
+class SrcDestPair:
```
sfc-gh-turbaszek, line 79: Can we name it more meaningfully?  Without context I don't understand "A pair of paths, relative to some context."

src/snowcli/cli/nativeapp/artifacts.py
```diff
@@ -0,0 +1,175 @@
...
+            return SrcDestPair(item["src"], item["dest"])
+        else:
+            return SrcDestPair(item["src"], item["src"])
+
+    elif isinstance(item, str):
+        return SrcDestPair(item, item)
+
+    else:
+        # XXX: validation should have caught this
+        raise ArtifactError("Item is not a valid artifact!")
```
sfc-gh-turbaszek, line 132:
```suggestion
    # XXX: validation should have caught this
    raise ArtifactError("Item is not a valid artifact!")
```

src/snowcli/cli/nativeapp/artifacts.py
```diff
@@ -0,0 +1,175 @@
...
+    """
+    Prepares a local folder (deploy_root) with configured app artifacts.
+    This folder can then be uploaded to a stage.
+    """
+    if deploy_root.exists() and not deploy_root.is_dir():
+        raise ValueError(f"Deploy root {deploy_root} exists, but is not a directory!")
+
+    for artifact in artifacts:
+        # build the list of source files
+        source_paths: List[Path]
```
sfc-gh-turbaszek, line 145: What do you think about encapsulating this logic into a function? It's a quite long logic in this for loop

src/snowcli/cli/nativeapp/init.py
```diff
@@ -0,0 +1,299 @@
...
+        generic_render_template(
+            template_path=parent_to_snowflake_yml.joinpath(snowflake_yml_jinja),
+            data={"project_name": parent_to_snowflake_yml.name},
+            output_file_path=parent_to_snowflake_yml.joinpath("snowflake.yml"),
+        )
+        subprocess.run(
+            f"rm {snowflake_yml_jinja}",
+            shell=True,
+            cwd=str(parent_to_snowflake_yml),
+        )
```
sfc-gh-turbaszek, line 180: Why not use `os.remove`?

src/snowcli/cli/nativeapp/init.py
```diff
@@ -0,0 +1,299 @@
...
+            data={
+                "application_name": f"{default_application_name_prefix}_{default_application_name_suffix}"
+            },
+            output_file_path=parent_to_readme.joinpath("README.md"),
+        )
+        subprocess.run(
+            f"rm {readme_jinja}",
+            shell=True,
+            cwd=str(parent_to_readme),
+        )
```
sfc-gh-turbaszek, line 218: Same comment, why not `os.remove`?

src/snowcli/cli/project/definition_manager.py
```diff
@@ -0,0 +1,71 @@
...
+class DefinitionManager:
+    BASE_CONFIG_FILENAME = "snowflake.yml"
+    USER_CONFIG_FILENAME = "snowflake.local.yml"
+
+    project_root: Path
+    _project_config_paths: List[Path]
+
+    def __init__(self, project: str | None = None) -> None:
+        search_path = os.getcwd()
+        if project and len(project) > 0:
```
sfc-gh-turbaszek, line 21:
```suggestion
        if project:
```

src/snowcli/cli/project/definition_manager.py
```diff
@@ -0,0 +1,71 @@
...
+            search_path = os.path.abspath(project)
+        config_files = self._find_config_files(Path(search_path))
+        if not config_files:
+            raise MissingConfiguration(
+                f"Cannot find native app project configuration. Please provide a path to the project or run this command in a valid native app project directory."
+            )
+        self._project_config_paths = config_files
+
+    def _find_config_files(self, project_path: Path) -> Optional[List[Path]]:
```
sfc-gh-turbaszek, line 30: Can you describe the search logic? Where are we looking for definition files?

src/snowcli/cli/project/definition_manager.py
```diff
@@ -0,0 +1,71 @@
...
+                    return [
+                        base_config_file_path,
+                        user_config_file_path,
+                    ]
+                return [base_config_file_path]
+
+            parent_path = parent_path.parent
+        return None
+
+    def _config_if_available(
```
sfc-gh-turbaszek, line 55:
```suggestion
    def _definition_if_available(
```
Or `project_config`? Just to be sure we do not create confusion. Same for other methods

src/snowcli/output/decorators.py
```diff
@@ -25,3 +28,25 @@ def _is_list_of_results(result):
...
+        def wrapper(*args, **kwargs):
+            try:
+                return func(*args, **kwargs)
+            except exception_class as e:
+                print(message if message else str(e), file=stderr)
+                raise typer.Exit(code=exit_code)
+
+        return wrapper
+
+    return decorator
```
sfc-gh-turbaszek, line 52: The idea of throwing always the same error is reasonable. However, I'm not sure if we need decorator. We can probably implement this on app level. @sfc-gh-pjob wdyt?

src/snowcli/utils.py
```diff
@@ -582,3 +583,23 @@ def create_project_template(template_name: str):
...
+    """
+    output = subprocess.check_output(["git", "--version"], text=True)
+    pattern = r"git version (\d+\.\d+)"
+    match_found = re.search(pattern, output.strip())
+    if match_found:
+        git_version = match_found.group(1)
+        return git_version
+    else:
+        log.error("Could not find a version for git")
+        raise typer.Abort()
```
sfc-gh-turbaszek, line 605: Please use [GitPython](https://gitpython.readthedocs.io/en/stable/tutorial.html). This will make the code more readable andwill help in future.

```py
import git
return git.cmd.Git().version()
```

src/snowcli/output/decorators.py
sfc-gh-pjob: Currently we catch all unhandled exceptions on the app level (main typer level), we print information about unhandled exception and exit with "1" code. User also can use `--debug` option to have full traceback instead of it. To make this working the command has to be decorated with `@global_options` or `@global_options_with_connection`. If it is not decorated, then `--debug` mode is by default.

The message without `--debug` looks like the following:
```
An unexpected exception occurred. Use --debug option to see the traceback.
Exception message:
Test exception message
```

This is what we have on the app level.

---

As I can see this decorator tries to achieve sth different. 
It allows to catch errors about which you are sure that it is safe to catch and silent them.
As long as it is used to catch exceptions generated by our code, and for which we are sure we don't need tracebacks, it offers cleaner messages. 

You receive:
```
Message extracted from test exception
```
or:
```
Custom message
```
instead of:
```
An unexpected exception occurred. Use --debug option to see the traceback.
Exception message:
Message extracted from test exception
```

IMO this decorator is ok if it is used only to simplify returning errors generated by ourselves but it may cause issues with errors debugging if used to catch exceptions raised either not by us or in multiple places in our code.

---

But I'm not sure if we need such decorator and if we need to add anything on the app level.
There is `ClickException` which has a few child classes delivered by Click and next few implemented by us.
We usually use simply `raise XyzException("wrong path provided")`, where `XyzException` is extending `ClickException` or one of its children. Extending the `ClickException` allows Click to catch the Exception on the framework level, print it nicely with colours and exit with proper exit code. It all is being done before our global handling.

IMO we should extend and use Click exceptions as often as possible because it guarantees consistent formatting of our errors.

Please take a look at https://click.palletsprojects.com/en/7.x/api/#click.ClickException and at `exception.py` in our codebase. If you will want to add custom exceptions for native apps functionality, then keep it in your subpackage (as you're doing it now) and do not add them to `exception.py` because this module will be part of public API for plugins.

src/snowcli/cli/project/definition_manager.py
```diff
@@ -0,0 +1,71 @@
...
+    _project_config_paths: List[Path]
+
+    def __init__(self, project: str | None = None) -> None:
+        search_path = os.getcwd()
+        if project and len(project) > 0:
+            search_path = os.path.abspath(project)
+        config_files = self._find_config_files(Path(search_path))
+        if not config_files:
+            raise MissingConfiguration(
+                f"Cannot find native app project configuration. Please provide a path to the project or run this command in a valid native app project directory."
```
sfc-gh-pjob, line 26: If I understand correctly, then the `project` package is designed to be used not only by native apps, so exception messages shouldn't have native apps context (at least not hardcoded).

src/snowcli/output/decorators.py
sfc-gh-cgorrie: This is very helpful, thanks for the feedback! I think the two big differences are:
- ability to configure exit code
- removal of the traceback

Happy to just extend `ClickException` and remove this decorator. It would be nice to have the exit code configurable but it's not a hill I'm willing to die on :D

sfc-gh-pjob: IMO you can customise exit_code. Take a look at `UsageError` which is provided by Click and which extends `ClickException`. Default exit_code for `ClickException` is 1 but `UsageError` overrides it and uses exit_code = 2.

src/snowcli/cli/nativeapp/init.py
```diff
@@ -0,0 +1,291 @@
...
+    """
+
+    clone_command = (
+        f"git clone -n --depth=1 --filter=tree:0 {git_url} {target_parent_directory}"
+    )
+    sparse_checkout_command = f"""
+        cd {target_parent_directory} &&
+            git sparse-checkout set --no-cone {repo_sub_directory} &&
+                git checkout
+        """
```
sfc-gh-turbaszek, line 110: As we added Git Python as dependency, shouldn't we use https://gitpython.readthedocs.io/en/stable/reference.html#git.repo.base.Repo.clone?

src/snowcli/cli/render/commands.py
```diff
@@ -142,16 +142,35 @@ def render_template(
             key, value = _parse_key_value(key_value_str)
             data[key] = value
 
+    generic_render_template(
+        template_path=template_path, data=data, output_file_path=output_file_path
+    )
+
+
+def generic_render_template(
+    template_path: Path, data: dict, output_file_path: Optional[Path]
```
sfc-gh-turbaszek, line 151:
```suggestion
    template_path: Path, data: dict, output_file_path: Optional[Path] = None
```
Because the file is optional

src/snowcli/utils.py
```diff
@@ -11,6 +11,9 @@
 import shutil
 import subprocess
 import warnings
+import git
+
+# from git import Repo, GitCommandError
```
sfc-gh-turbaszek, line 16: Leftover?

src/snowcli/utils.py
```diff
@@ -582,3 +585,23 @@ def create_project_template(template_name: str):
...
+    Returns the major and minor version of git installed on a user's machine.
+    Args:
+        None
+
+    Returns:
+        str: the major and minor version of git, e.g. "2.25"
+    """
+    output = git.cmd.Git().version()
+    pattern = r"git version (\d+\.\d+)"
+    match_found = re.search(pattern, output.strip())
```
sfc-gh-turbaszek, line 601:
```
In [2]: git.cmd.Git().version()
Out[2]: 'git version 2.39.2 (Apple Git-143)'

In [4]: git.cmd.Git().version_info
Out[4]: (2, 39, 2)
```

The `version_info` command seem's to be offering what you need. It's also relatively easy to do version checks if necessary: 
```
In [7]: git.cmd.Git().version_info[:2] >= (2, 25)
Out[7]: True
```

tests/nativeapp/test_artifacts.py
```diff
@@ -0,0 +1,117 @@
...
+def test_outside_deploy_root_two_ways(project_definition_files):
+    project_root = project_definition_files[0].parent
+    with pytest.raises(OutsideDeployRootError):
+        build_bundle(
+            project_root,
+            deploy_root=Path(project_root, "deploy"),
+            artifacts=[ArtifactMapping("setup.sql", "..")],
+        )
+
+    with pytest.raises(OutsideDeployRootError):
```
sfc-gh-turbaszek, line 99: I would suggest to separate this as another unit test

src/snowcli/cli/nativeapp/commands.py
```diff
@@ -0,0 +1,69 @@
...
+
+from snowcli.output.types import (
+    CommandResult,
+    MessageResult,
+)
+
+app = typer.Typer(
+    context_settings=DEFAULT_CONTEXT_SETTINGS,
+    hidden=True,
+    name="app",
```
sfc-gh-pjob, line 20: One of discussions from our engineering review doc was about the name of this command: https://docs.google.com/document/d/1llIv8RTrvn4MA4brdh-mSOULegnjMm4nq-tWX3TUm0I/edit?disco=AAAAySr4GAI

Has the name of this command been approved by somebody? `snow app ...` is a very generic name.

Screenshot 2023-09-11 at 12 55 01

src/snowcli/cli/project/definition_manager.py
```diff
@@ -0,0 +1,80 @@
...
+            if (
+                project_path.is_mount() != starting_mount
+                or parent_path.parent == parent_path
+                or parent_path == Path.home()
+            ):
+                return None
+
+            base_config_file_path = self._base_definition_file_if_available(parent_path)
+            if base_config_file_path:
+                user_config_file_path = self._user_definition_file_if_available(
```
sfc-gh-pjob, line 48: Does it mean that `snowflake.local.yml` has to be located in the same directory as `snowflake.yml`?

Does it mean that following overrides won't work? (Running snowcli inside my-settings-1 directory my custom overrides won't be applied?)
```
.../some-path/project-x/
    - snowflake.yml
    - my-settings-1
        - snowflake.local.yml
    - my-settings-2
        - snowflake.local.yml
```

src/snowcli/cli/nativeapp/init.py
```diff
@@ -0,0 +1,291 @@
...
+        source_parent_directory (Path): The source parent directory of the sub directory.
+        target_parent_directory (Path): The target parent directory for the sub directory.
+        repo_sub_directory (str): The sub directory name within the cloned repository.
+        new_name (str): The new name to give to the sub directory after moving.
+
+    Returns:
+        None
+    """
+
+    # Move to target parent directory
```
sfc-gh-pjob, line 150: 1. Why this has to be done in two separate steps instead of moving the directory to target path (including target name) in just one step?
2. Why do we need custom function instead of `shutil.move`?

src/snowcli/cli/nativeapp/init.py
```diff
@@ -0,0 +1,291 @@
...
+
+    readme_jinja = "README.md.jinja"
+
+    default_application_name_prefix = clean_identifier(project_name)
+    default_application_name_suffix = clean_identifier(
+        get_env_username() or DEFAULT_USERNAME
+    )
+
+    try:
+        generic_render_template(
```
sfc-gh-pjob, line 203: This is a function from `render/commands.py`. 
I know that we already use logic of some commands in other commands, but we usually do it using "managers".
Maybe to be consistent, we should have `RenderManager`?
@sfc-gh-turbaszek WDYT?

sfc-gh-turbaszek: Personally I would move this function to common/utils module. This is something we may want to use for multiple usa-cases and keeping it command-specific does not make sense.

src/snowcli/cli/nativeapp/commands.py
sfc-gh-turbaszek: Let's proceed with `app` but let's bring this topic on Thursday

sfc-gh-turbaszek (approved):

PR #373, opened 2023-09-08
Updated options names and docs generation

sfc-gh-astus:
### Pre-review checklist

### Changes description
* Now always first option name is main one e.g. `--verbose`


src/snowcli/app/dev/docs/templates/usage.rst.jinja2
```diff
@@ -7,7 +7,7 @@ Syntax
 
   snow {% for part in path %}{{ part }}{{ " " if not loop.last }}{% endfor %}
 {%- for param in options if not param.hidden %}
-{{ "    " + param.opts[-1] }}{% if not param.is_flag %} <{{ param.name }}>{% endif %}
+{{ "    " + param.opts[0] }}{% if not param.is_flag %} <{{ param.name }}>{% endif %}
```
sfc-gh-pjob, line 10: Maybe we can find the first option starting with `--` and only if there is no such option then use the first option from all (`param.opts[0]`)? Probably it would be simpler to write this logic in generator.py, not in jinja template.

WDYT?

But IMO the convention assuming that the first variant is always the main variant is also ok.

sfc-gh-astus: It won't work. Connection has 2 options starting with `--`

```
ConnectionOption = typer.Option(
    None,
    "--connection",
    "-c",
    "--environment",
```

sfc-gh-pjob: Yeah and we will choose the first one.

sfc-gh-pjob (approved):

src/snowcli/app/dev/docs/templates/usage.rst.jinja2
sfc-gh-pjob: Agreed to change it in another PR.

PR #374, opened 2023-09-08
Avoiding unnecessary replace in function/procedure update

sfc-gh-jsikorski:
### Pre-review checklist

### Changes description
Currently, the procedure/function update will perform some unnecessary replace operations, as there is only an equality check between return types in python and snowflake. It works for most of the types, but in some cases, types are mapped to different ones, hence the equality check will fail.
To avoid it, simple mapping was added for the types that are changed.


src/snowcli/utils.py
```diff
@@ -582,3 +582,15 @@ def create_project_template(template_name: str):
...
+
+
+def sql_to_python_return_type_mapper(resource_return_type: str) -> str:
+
+    mapping = {
+        "number(38,0)": "int",
+        "timestamp_ntz(9)": "datetime",
+        "timestamp_tz(9)": "datetime",
+        "varchar(16777216)": "string",
+    }
```
sfc-gh-turbaszek, line 594: Can we add link to the source of this information for future references?

src/snowcli/cli/snowpark/function/commands.py
```diff
@@ -221,7 +222,8 @@ def function_update(
             replace = True
         elif (
             resource_json["handler"].lower() != handler.lower()
-            or resource_json["returns"].lower() != return_type.lower()
+            or sql_to_python_return_type_mapper(resource_json["returns"])
```
sfc-gh-pjob, line 225: IMO we shouldn't assume that the util function will always return lowercased type name.

```suggestion
            or sql_to_python_return_type_mapper(resource_json["returns"]).lower()
```

sfc-gh-jsikorski: Oh, but it will. The return of mapping function is:
`return mapping.get(resource_return_type.lower(), resource_return_type.lower())`

src/snowcli/utils.py
sfc-gh-jsikorski: Added

src/snowcli/cli/snowpark/function/commands.py
sfc-gh-pjob: I know it does it at the moment.
But it's only because you know about it now.
What if I want to add a new mapping and I don't know about the convention that all values of dict should be in lowercase?
What if I decide to completely change implementation of `sql_to_python_return_type_mapper` and I don't know that it has extra responsibility of converting the result to lowercase?

sfc-gh-jsikorski: Ok, good point. Done

sfc-gh-pjob (approved):

PR #375, opened 2023-09-08
Changed options selection

sfc-gh-astus:
### Pre-review checklist

### Changes description
...

### DIFF
--- a/src/snowcli/app/dev/docs/templates/usage.rst.jinja2
+++ b/src/snowcli/app/dev/docs/templates/usage.rst.jinja2
@@ -7,7 +7,7 @@ Syntax
 
   snow {% for part in path %}{{ part }}{{ " " if not loop.last }}{% endfor %}
 {%- for param in options if not param.hidden %}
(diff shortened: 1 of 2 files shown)


src/snowcli/app/dev/docs/templates/usage.rst.jinja2
```diff
@@ -7,7 +7,7 @@ Syntax
 
   snow {% for part in path %}{{ part }}{{ " " if not loop.last }}{% endfor %}
 {%- for param in options if not param.hidden %}
-{{ "    " + param.opts[0] }}{% if not param.is_flag %} <{{ param.name }}>{% endif %}
+{{ "    " + (param.opts | sort | first) }}{% if not param.is_flag %} <{{ param.name }}>{% endif %}
```
sfc-gh-pjob, line 10: It's clever but I think it may raise issues when we have multiple variants starting with `--` (for example to provide backward compatibility). If the old flag starts with "a" it will be chosen instead of using the new one.

sfc-gh-turbaszek (approved):

PR #385, opened 2023-09-12
[SNOW-901392] Builtin plugins

sfc-gh-pjob:
### Pre-review checklist

### Changes description
* Added registration of commands from builtin plugins.
* Transformed all existing commands to builtin plugins.
* Introduced `snowcli.api` package which will contain public API which can be used by plugins.


src/snowcli/cli/warehouse/commands.py
```diff
@@ -15,6 +15,12 @@
 )
 
 
+# Empty callback is required to create command group with only one command
```
sfc-gh-pjob, line 18: I don't have a better idea at the moment. In the future we can think about an util function which will be responsible for creating a typer of specific type (for example: "typer for command group with only one command").

src/snowcli/app/cli_app.py
```diff
@@ -1,48 +1,86 @@
...
-    setup_pycharm_remote_debugger_if_provided,
-)
 
 app: SnowCliMainTyper = SnowCliMainTyper()
 log = logging.getLogger(__name__)
 
+_commands_registration = CommandsRegistrationWithCallbacks()
+
+
+def _do_not_execute_on_completion(callback):
```
sfc-gh-pjob, line 31: Previously our completion was broken. It raised errors when trying to use flags like `--version`, because callbacks for these flags were executed and output of this callbacks was mixed with completion.

`resilient_parsing` is a flag determining completion mode (found in Typer docs).

src/snowcli/app/cli_app.py
```diff
@@ -1,48 +1,86 @@
...
+        callback(value)
+
+    return enriched_callback
+
+
+def _commands_registration_callback(value: bool):
+    if value:
+        _commands_registration.register_commands_if_ready_and_not_registered_yet()
+    # required to make the tests working
+    _commands_registration.reset_running_instance_registration_state()
```
sfc-gh-pjob, line 44: I don't have a better idea how to do it better.

Main `app` instance seems to be reused in tests (for sure in parameterized tests), but command needs to be registered in each execution of this `app`.
Required callbacks gathering is done only once (during the import of the module) while commands registration is done multiple times (for every runner.invoke)

src/snowcli/cli/sql/commands.py
```diff
@@ -8,7 +8,11 @@
 from snowcli.output.decorators import with_output
 from snowcli.output.types import QueryResult, CommandResult, MultipleResults
 
+# simple Typer with defaults because it won't become a command group as it contains only one command
```
sfc-gh-pjob, line 11: I don't have a better idea at the moment. In the future we can think about an util function which will be responsible for creating a typer of specific type (for example: "typer for command group with only one command").

src/snowcli/app/cli_app.py
sfc-gh-turbaszek: Can we rebuild test instance every time? Or that won't solve the problem?

sfc-gh-pjob: Currently we create a new `runner` instance from fixture for each test (`scope=function`).
```python
@pytest.fixture(scope="function")
def runner(test_snowcli_config):
    from snowcli.app.cli_app import app
    return SnowCLIRunner(app, test_snowcli_config)
```
However imported `app` is cached.
I was able to do resolve caching issue with following code:
```python
@pytest.fixture(scope="function")
def runner(test_snowcli_config):
    from snowcli.app.cli_app import app
    yield SnowCLIRunner(app, test_snowcli_config)
    sys.modules.pop("snowcli.app.cli_app")
```
But still there are tests which do multiple calls to runner.invoke in the same test case and for such tests removing imported modules won't help.

I still don't see a better option than `reset_running_instance_registration_state`.

sfc-gh-pjob: I made the comment more verbose.

sfc-gh-turbaszek: How about reloading the module or invalidating cache?
https://docs.python.org/3/library/importlib.html#importlib.reload

src/snowcli/app/commands_registration/command_plugins_loader.py
```diff
@@ -0,0 +1,113 @@
...
+        for (plugin_name, plugin) in self._plugin_manager.list_name_plugin():
+            self._load_plugin(plugin_name, plugin)
+        return list(self._loaded_plugins.values())
+
+    def _load_plugin(self, plugin_name: str, plugin) -> Optional[LoadedCommandPlugin]:
+        already_loaded_plugin = self._loaded_plugins.get(plugin_name)
+        if already_loaded_plugin:
+            return already_loaded_plugin
+        else:
+            return self._load_new_plugin(plugin_name, plugin)
```
sfc-gh-turbaszek, line 52:
```suggestion
        return self._load_new_plugin(plugin_name, plugin)
```

src/snowcli/app/commands_registration/commands_registration_with_callbacks.py
```diff
@@ -0,0 +1,82 @@
...
+
+from snowcli.app.commands_registration.command_plugins_loader import (
+    load_only_builtin_command_plugins,
+)
+from snowcli.app.commands_registration.typer_registration import (
+    add_plugins_to_main_typer,
+)
+from snowcli.utils import ThreadsafeCounter
+
+log = logging.getLogger(__name__)
```
sfc-gh-turbaszek, line 13: Do we use it somewhere in the file?

src/snowcli/app/commands_registration/command_plugins_loader.py
sfc-gh-pjob: I'm not sure if your version is easier to read but I applied your suggestion ;)

src/snowcli/app/commands_registration/command_plugins_loader.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def _load_new_plugin(
+        self, plugin_name: str, plugin
+    ) -> Optional[LoadedCommandPlugin]:
+        loaded_plugin = self._load_plugin_spec(plugin_name, plugin)
+        if loaded_plugin:
```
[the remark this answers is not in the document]
sfc-gh-pjob: Done

src/snowcli/app/commands_registration/commands_registration_with_callbacks.py
sfc-gh-pjob: No, I removed it.

src/snowcli/app/commands_registration/typer_registration.py
```diff
@@ -0,0 +1,86 @@
...
+
+from snowcli.app.commands_registration import LoadedCommandPlugin
+from snowcli.api.plugin.command import CommandSpec
+from snowcli.cli.exception_logging import exception_logging
+
+log = logging.getLogger(__name__)
+log_exception = exception_logging(log)
+
+
+class AddPluginsToMainTyper:
```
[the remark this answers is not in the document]
sfc-gh-pjob: I changed the name of the class to `TyperCommandsRegistration` and method `__call__` to `register_commands`. Does it look better?

src/snowcli/cli/render/commands.py
```diff
@@ -13,6 +13,12 @@
 app = typer.Typer(context_settings=DEFAULT_CONTEXT_SETTINGS, hidden=True, name="render")
 
 
+# Empty callback is required to create command group with only one command
+@app.callback()
+def empty_callback():
+    pass
```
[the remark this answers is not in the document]
sfc-gh-pjob: I think I've already answered this question in our discussion in the google doc.
If you think you need some explanation here, please let me know.

sfc-gh-turbaszek (approved):

src/snowcli/cli/render/commands.py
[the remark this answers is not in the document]
sfc-gh-pjob: We consider adding "typer_type" to CommandSpec. This new field could be responsible for determining if plugin is a single command or a command group, but I'll do it in a separate PR.

src/snowcli/cli/warehouse/commands.py
sfc-gh-pjob: More comments about it: https://github.com/Snowflake-Labs/snowcli/pull/385#discussion_r1328888405

src/snowcli/cli/sql/commands.py
sfc-gh-pjob: More comments about it: https://github.com/Snowflake-Labs/snowcli/pull/385#discussion_r1328888405

PR #386, opened 2023-09-13
Fix for problem with windows shortened paths

sfc-gh-jsikorski:
### Pre-review checklist

### Changes description
Fix for reported problem with 8.3 paths.
In Windows, sometimes path to tmp_dir gets shortened - instead of `\long_username\...` we get `\long_u~1\...`. 
I caused problems, when uploading files to stage.


src/snowcli/utils.py
```diff
@@ -604,3 +605,24 @@ def sql_to_python_return_type_mapper(resource_return_type: str) -> str:
...
+
+def get_full_path(path_to_resolve: str):
+    from ctypes import create_unicode_buffer, windll
+
+    BUFFER_SIZE = 260
+    buffer = create_unicode_buffer(BUFFER_SIZE)
+    get_long_path_name = windll.kernel32.GetLongPathNameW
+    return_value = get_long_path_name(path_to_resolve, buffer, BUFFER_SIZE)
+
+    if return_value == 0 or return_value > 260:
```
sfc-gh-astus, line 625: `return_value > BUFFER_SIZE`

tests/test_utils.py
```diff
@@ -386,3 +386,20 @@ def test_deduplicate_and_sort_reqs():
...
+    "argument, expected",
+    [
+        ("C:\Something\Something Else", "C:\Something\Something Else"),
+        (
+            "/var/folders/k8/3sdqh3nn4gg7lpr5fz0fjlqw0000gn/T/tmpja15jymq",
+            "/var/folders/k8/3sdqh3nn4gg7lpr5fz0fjlqw0000gn/T/tmpja15jymq",
+        ),
+    ],
+)
+def test_path_resolver(mock_system, argument, expected):
```
sfc-gh-jsikorski, line 402: This test does not check, if the solution works - as it would need a valid win 8.3 path. It just checks if nothing gets broken, and are the normal paths left unchanged.

src/snowcli/utils.py
```diff
@@ -25,6 +27,9 @@
 from jinja2 import Environment, FileSystemLoader
 from snowflake.connector.cursor import SnowflakeCursor
 
+if sys.platform == "win32":
+    from ctypes import create_unicode_buffer, windll  # type: ignore
```
sfc-gh-astus, line 31: Maybe move import to if in path_resolver?

src/snowcli/utils.py
```diff
@@ -604,3 +611,14 @@ def sql_to_python_return_type_mapper(resource_return_type: str) -> str:
     }
 
     return mapping.get(resource_return_type.lower(), resource_return_type.lower())
+
+
+def path_resolver(path_to_file: str):
+    if platform.system() == "Windows" and "~1" in path_to_file:
```
sfc-gh-astus, line 617: Why you use 2 different conditions? `platform.system() == "Windows"` and `sys.platform == "win32"`

sfc-gh-jsikorski: Reformatted to single check

src/snowcli/utils.py
sfc-gh-jsikorski: Moved

sfc-gh-astus (approved):

PR #387, opened 2023-09-13
Connections improvements

sfc-gh-astus:
### Pre-review checklist

### Changes description
* when you add first connection, that connection become default
* selecting default connection flow:
** get default_connection from config.toml
** if in config.toml is only one connection take it
** if default_connection in config.toml is not set and there is more than 1 connection in config.toml use `dev` connection

### DIFF
--- a/src/snowcli/config.py
+++ b/src/snowcli/config.py
@@ -17,6 +17,11 @@ from snowflake.connector.constants import CONFIG_FILE
 from snowflake.connector.config_manager import ConfigManager
 
 log = logging.getLogger(__name__)
+_DEFAULT_CONNECTION_KEY = "default_connection"
+
+
+class Empty:
+    pass
 
 
 class CliConfigManager(ConfigManager):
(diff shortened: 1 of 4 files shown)


src/snowcli/config.py
```diff
@@ -70,10 +75,18 @@ def get_connection(self, connection_name: str) -> dict:
 
     def add_connection(self, name: str, parameters: dict):
         if not self.section_exists("connections"):
-            self._initialize_connection_section()
-        self._find_section("connections").add(name, parameters)
+            self._initialize_section("connections")
+        connections = self._find_section("connections")
+        connections.add(name, parameters)
+        if len(connections) == 1:
+            self._set_default_connection(name)
```
sfc-gh-turbaszek, line 82: I don't think we should be doing this way as this creates additional config entry. I think we should check for len = 1 when reading connections

sfc-gh-astus: Right, changed

sfc-gh-pjob (approved):

sfc-gh-pjob (approved):

PR #394, opened 2023-09-20
[NADE] Add diff functionality between local directory and Snowflake stage

sfc-gh-bgoel:
### Pre-review checklist
   * [x] I've confirmed that my changes are up-to-date with the target branch. [As of Sept 22, 08:00AM EST]

### Changes description
Author: All code changes are by @sfc-gh-cgorrie 
This PR creates a new stage command that allows checking contents on a given stage with a given folder on a user's filesystem. The use case is currently only within Native Apps, hence the command resides under native-apps module, and needs to stay hidden as it is only used internally, and we have not yet discussed exposing this command for all users.


src/snowcli/cli/nativeapp/commands.py
```diff
@@ -71,3 +75,24 @@ def app_bundle(
...
+) -> MessageResult:
+    """
+    Diffs a stage with a local folder
+    """
+    diff: DiffResult = stage_diff(Path(folder_name), stage_fqn)
+    output = f"""\
+        only local:  {', '.join(diff.only_local)}
+        only stage:  {', '.join(diff.only_on_stage)}
+        mod/unknown: {', '.join(diff.different)}
+        identical:   {', '.join(diff.identical)}"""
```
sfc-gh-turbaszek, line 96: Have you though about using `ObjectResult`? In this way the output will be easier to programmatically parse, and for human readable we will get a table.

sfc-gh-bgoel: Cam's currently OOO, so I'm making the changes to ack the feedback. 
I have updated with ObjectResult, and IIUC, Table is the default output format so it should print as a table.

src/snowcli/cli/nativeapp/commands.py
```diff
@@ -71,3 +75,18 @@ def app_bundle(
     manager = NativeAppManager(project_path)
     manager.build_bundle()
     return MessageResult(f"Bundle generated at {manager.deploy_root}")
+
+
+@app.command("diff", hidden=True)
+@with_output
+@global_options_with_connection
+def nativeapp_stage_diff(
```
sfc-gh-pjob, line 83: other functions in this file have names like `app_command`

is there any reason why this function does not follow this convention?

src/snowcli/cli/nativeapp/commands.py
```diff
@@ -71,3 +75,18 @@ def app_bundle(
     manager = NativeAppManager(project_path)
     manager.build_bundle()
     return MessageResult(f"Bundle generated at {manager.deploy_root}")
+
+
+@app.command("diff", hidden=True)
```
sfc-gh-pjob, line 80: From PR description:
> The use case is currently only within Native Apps, hence the command resides under native-apps module, and needs to stay hidden as it is only used internally, and we have not yet discussed exposing this command for all users.

Why can't we put it to `snow stage` and keep it hidden there?

tests/testing_utils/files_and_dirs.py
```diff
@@ -19,3 +21,22 @@ def _write_to_file(path: str, contents: List[str]) -> None:
     with open(path, "w") as new_file:
         for line in contents:
             new_file.write(line + "\n")
+
+
+@contextmanager
+def temp_local_dir(files: Dict[str, Union[str, bytes]]) -> Generator[Path, None, None]:
```
sfc-gh-pjob, line 27: Very nice util!

src/snowcli/cli/nativeapp/commands.py
```diff
@@ -1,17 +1,21 @@
...
+from pathlib import Path
+
 import logging
 import typer
 
-from snowcli.cli.common.decorators import global_options
+from snowcli.cli.common.decorators import global_options, global_options_with_connection
 from snowcli.cli.common.flags import DEFAULT_CONTEXT_SETTINGS
 from snowcli.output.decorators import with_output
+from snowcli.cli.stage.diff import DiffResult, stage_diff
```
sfc-gh-pjob, line 10: @sfc-gh-turbaszek 
After reading guidelines which you prepared, I wonder if it is ok to use `diff` logic directly, or should it be wrapped by some function in `StageManager`?

src/snowcli/cli/stage/diff.py
```diff
@@ -0,0 +1,151 @@
...
+    different: List[str] = field(default_factory=list)
+    "Files that may be different between the stage and the local directory"
+
+    only_local: List[str] = field(default_factory=list)
+    "Files that only exist in the local directory"
+
+    only_on_stage: List[str] = field(default_factory=list)
+    "Files that only exist on the stage"
+
+    def display_pretty_diff(self):
```
sfc-gh-pjob, line 34: I would avoid names suggesting anything about output format.
IMO logic should only produce a result and should not be aware how it will be formatted.

src/snowcli/cli/stage/diff.py
```diff
@@ -0,0 +1,151 @@
...
+    """Returns the given stage path without the stage name as the first part."""
+    return "/".join(path.split("/")[1:])
+
+
+def build_md5_map(list_stage_cursor: SnowflakeCursor) -> Dict[str, str]:
+    """
+    Returns a mapping of relative stage paths to their md5sums.
+    """
+    return {
+        # XXX: how can we get dicts back here?
```
sfc-gh-pjob, line 109: What about this comment?
Is it TODO?
Is it about sth what is required to have everything working correctly?

src/snowcli/cli/nativeapp/commands.py
```diff
@@ -1,8 +1,10 @@
 from typing import Optional
+from pathlib import Path
```
sfc-gh-pjob, line 2: As command was moved to `stage` subgroup, there should be no changes in this file. Added imports are now unused.

src/snowcli/cli/stage/diff.py
```diff
@@ -0,0 +1,141 @@
+import re
+import hashlib
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import List, Dict
+from snowcli.output.types import ObjectResult
```
sfc-gh-pjob, line 6: this import is not used anymore

src/snowcli/cli/stage/diff.py
sfc-gh-pjob: I'd delete this comment if we can't understand it.

IMO it's introducing only some confusion.

sfc-gh-pjob (approved):

sfc-gh-turbaszek (approved):

PR #395, opened 2023-09-20
[NADE] Snow App Run Part 1

sfc-gh-bgoel:
### Pre-review checklist
   * [x] I've confirmed that my changes are up-to-date with the target branch. [as of Sept 28, 9:30AM EST]

### Changes description
This PR is to introduce a `snow app run` command which is used to create an application package and upload files from a user's local filesystem to a snowflake stage. This is only a part of the implementation of the `snow app run` command, as it is spread over three tickets tracked internally due to the complexity of its behavior. 
Tests have been added keeping in mind this partial implementation for now.  

How to review this PR:
This PR has two sets of changes, one of which has already been approved from another PR https://github.com/Snowflake-Labs/snowcli/pull/400 and I will mark those changes as such so you do not need to re-review them.

### DIFF
--- a/src/snowcli/cli/nativeapp/commands.py
+++ b/src/snowcli/cli/nativeapp/commands.py
@@ -58,7 +58,7 @@ def app_init(
 
 @app.command("bundle", hidden=True)
 @with_output
-@global_options_with_connection
+@global_options
 def app_bundle(
     project_path: Optional[str] = ProjectArgument,
     **options,
--- a/src/snowcli/cli/nativeapp/commands.py
+++ b/src/snowcli/cli/nativeapp/commands.py
@@ -3,7 +3,10 @@ from typing import Optional
 import logging
 import typer
 
-from snowcli.cli.common.decorators import global_options, global_options_with_connection
+from snowcli.cli.common.decorators import (
+    global_options_with_connection,
+    global_options,
+)
 from snowcli.cli.common.flags import DEFAULT_CONTEXT_SETTINGS
 from snowcli.output.decorators import with_output
 
@@ -58,7 +61,7 @@ def app_init(
 
 @app.command("bundle", hidden=True)
 @with_output
-@global_options_with_connection
+@global_options
 def app_bundle(
     project_path: Optional[str] = ProjectArgument,
     **options,
@@ -80,6 +83,8 @@ def app_run(
 ) -> CommandResult:
     """
     Creates an application package in your Snowflake account and uploads code files to its stage.
+    As a note, this command does not accept role or warehouse overrides to your config.toml file,
+    because your native app definition in snowflake.yml/snowflake.local.yml is used for any overrides.
     """
     manager = NativeAppManager(project_path)
     manager.build_bundle()
@@ -98,6 +103,8 @@ def app_teardown(
 ) -> CommandResult:
     """
     Drops an application and an application package as defined in the project definition file.
+    As a note, this command does not accept role or warehouse overrides to your config.toml file,
+    because your native app definition in snowflake.yml/snowflake.local.yml is used for any overrides.
     """
     manager = NativeAppManager(project_path)
     manager.teardown()
--- a/src/snowcli/cli/nativeapp/commands.py
+++ b/src/snowcli/cli/nativeapp/commands.py
@@ -58,7 +58,7 @@ def app_init(
 
 @app.command("bundle", hidden=True)
 @with_output
-@global_options
+@global_options_with_connection
 def app_bundle(
     project_path: Optional[str] = ProjectArgument,
     **options,
@@ -73,7 +73,7 @@ def app_bundle(
 
 @app.command("run")
 @with_output
-@global_options
+@global_options_with_connection
 def app_run(
     project_path: Optional[str] = ProjectArgument,
     **options,
@@ -87,3 +87,18 @@ def app_run(
     return MessageResult(
         f"Application Package is now active in your Snowflake account!"
     )
+
+
+@app.command("teardown")
+@with_output
+@global_options_with_connection
+def app_teardown(
+    project_path: Optional[str] = ProjectArgument,
+    **options,
+) -> CommandResult:
+    """
+    Drops an application and an application package as defined in the project definition file.
+    """
+    manager = NativeAppManager(project_path)
+    manager.teardown()
+    return MessageResult(f"Teardown is now complete.")
--- a/src/snowcli/cli/nativeapp/commands.py
+++ b/src/snowcli/cli/nativeapp/commands.py
@@ -3,17 +3,14 @@ from typing import Optional
 import logging
 import typer
 
-from snowcli.cli.common.decorators import global_options
+from snowcli.cli.common.decorators import global_options, global_options_with_connection
 from snowcli.cli.common.flags import DEFAULT_CONTEXT_SETTINGS
 from snowcli.output.decorators import with_output
 
-from .init import nativeapp_init
-from .manager import NativeAppManager
+from snowcli.cli.nativeapp.init import nativeapp_init
+from snowcli.cli.nativeapp.manager import NativeAppManager
 
-from snowcli.output.types import (
-    CommandResult,
-    MessageResult,
-)
+from snowcli.output.types import CommandResult, MessageResult
 
 app = typer.Typer(
     context_settings=DEFAULT_CONTEXT_SETTINGS,
@@ -51,7 +48,7 @@ def app_init(
     **options,
 ) -> CommandResult:
     """
-    Initialize a Native Apps project, optionally with a --git-url and a --template.
+    Initializes a Native Apps project, optionally with a --git-url and a --template.
     """
     nativeapp_init(name, git_url, template)
     return MessageResult(
@@ -61,7 +58,7 @@ def app_init(
 
 @app.command("bundle", hidden=True)
 @with_output
-@global_options
+@global_options_with_connection
 def app_bundle(
     project_path: Optional[str] = ProjectArgument,
     **options,
@@ -72,3 +69,21 @@ def app_bundle(
     manager = NativeAppManager(project_path)
     manager.build_bundle()
     return MessageResult(f"Bundle generated at {manager.deploy_root}")
+
+
+@app.command("run")
+@with_output
+@global_options
+def app_run(
+    project_path: Optional[str] = ProjectArgument,
+    **options,
+) -> CommandResult:
+    """
+    Creates an application package in your Snowflake account and uploads code files to its stage.
+    """
+    manager = NativeAppManager(project_path)
+    manager.build_bundle()
+    manager.app_run()
+    return MessageResult(
+        f"Application Package is now active in your Snowflake account!"
+    )
--- a/src/snowcli/cli/nativeapp/manager.py
+++ b/src/snowcli/cli/nativeapp/manager.py
@@ -32,6 +32,7 @@ from snowflake.connector.cursor import DictCursor
 
 SPECIAL_COMMENT = "GENERATED_BY_SNOWCLI"
 COMMENT_COL = "comment"
+OWNER_COL = "owner"
 
 log = logging.getLogger(__name__)
 
@@ -50,6 +51,15 @@ class CouldNotDropObjectError(ClickException):
 
     def __init__(self, message: str):
         super().__init__(message=message)
+
+
+class UnexpectedOwnerError(ClickException):
+    """An operation is blocked becuase an object is owned by an unexpected role."""
+
+    def __init__(self, item: str, expected_owner: str, actual_owner: str):
+        super().__init__(
+            f"Cannot operate on {item}: owned by {actual_owner} (expected {expected_owner})"
+        )
 
 
 class NativeAppManager(SqlExecutionMixin):
@@ -150,6 +160,12 @@ class NativeAppManager(SqlExecutionMixin):
 
                 if row_comment != SPECIAL_COMMENT:
                     raise ApplicationPackageAlreadyExistsError(app_pkg)
+
+                actual_owner = show_app_row[OWNER_COL]
+                if actual_owner != self.app_pkg_role:
+                    raise UnexpectedOwnerError(
+                        self.app_name, self.app_role, actual_owner
+                    )
 
             # Upload files from deploy root local folder to the above stage
             self.sync_deploy_root_with_stage(app_pkg_role, app_pkg)
--- a/src/snowcli/cli/nativeapp/manager.py
+++ b/src/snowcli/cli/nativeapp/manager.py
@@ -11,9 +11,11 @@ from snowflake.connector.cursor import SnowflakeCursor
 
 from snowcli.cli.project.util import clean_identifier
 from snowcli.cli.common.sql_execution import SqlExecutionMixin
-from snowcli.cli.project.definition import default_app_package, default_role
-from snowcli.output.printing import print_result
-from snowcli.output.types import ObjectResult
+from snowcli.cli.project.definition import (
+    default_app_package,
+    default_role,
+    default_application,
+)
 from snowcli.cli.stage.diff import (
     DiffResult,
     stage_diff,
@@ -39,6 +41,15 @@ class ApplicationPackageAlreadyExistsError(ClickException):
         super().__init__(
             f"An Application Package {name} already exists in account that may have been created without snowCLI. "
         )
+
+
+class CouldNotDropObjectError(ClickException):
+    """
+    Could not successfully drop the required Snowflake object.
+    """
+
+    def __init__(self, message: str):
+        super().__init__(message=message)
 
 
 class NativeAppManager(SqlExecutionMixin):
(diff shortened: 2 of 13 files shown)


src/snowcli/cli/nativeapp/manager.py
```diff
@@ -35,3 +57,89 @@ def deploy_root(self) -> Path:
...
+
+        # Upload diff-ed files to app pkg stage
+        log.info(
+            f"Uploading diff-ed files from your local {deploy_root} directory to the Snowflake stage."
+        )
+        sync_local_diff_with_stage(
+            role=role, diff_result=diff, stage_path=f"{app_pkg}.{schema_and_stage}"
+        )
+
+    def app_run(self) -> SnowflakeCursor:
```
sfc-gh-bgoel, line 104: This method's implementation is quite complex and will be spread out over three tickets, and hence its appearance is likely to change as it evolves over these tickets. Hence we do not have a unit test for this method yet.

src/snowcli/cli/stage/manager.py
```diff
@@ -33,17 +33,48 @@ def put(
...
     ) -> SnowflakeCursor:
+        *_, put_cursor = self._put(
+            local_path=local_path,
+            stage_path=stage_path,
+            parallel=parallel,
+            overwrite=overwrite,
+        )
+        return put_cursor
+
+    def _put(
```
sfc-gh-astus, line 44: I'm not sure if it is ok to add `_put` and `_remove` to StageManager. I think it will be better to move it to NativeAppManager

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -35,3 +57,89 @@ def deploy_root(self) -> Path:
...
+
+        app_pkg_role = get_required_field_from_definition(
+            "package", "role", self.definition, generated_local_yml_overrides
+        )
+
+        show_app_cursor = self.execute_query_with_role(
+            app_pkg_role, f"show application packages like {app_pkg}"
+        )
+
+        show_app_results = list(QueryResult(show_app_cursor).result)
```
sfc-gh-astus, line 125: For me it is unexpected usage of `QueryResult`, our purpose was to return `*Results` objects to format output properly

sfc-gh-turbaszek: I agree, this is unnecessary.  We can get the length of the results using `show_app_cursor.rowcount` and to fetch one result we can use `fetchone` method.

src/snowcli/cli/stage/manager.py
sfc-gh-bgoel: Since `put` essentially calls `_put` but without any role, if I separate `put` and `_put` into different classes, it would essentially mean duplicating code across the two classes (with small differences). Would that be okay?

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -1,13 +1,35 @@
...
+    DiffResult,
+    stage_diff,
+    sync_local_diff_with_stage,
+)
+from snowcli.cli.project.schema import FilePath
+from snowcli.cli.project.definition_manager import DefinitionManager
 
 from .artifacts import build_bundle, translate_artifact, ArtifactMapping
-from ..project.definition_manager import DefinitionManager
+from .run import (
```
sfc-gh-turbaszek, line 23: Please use absolute imports

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -35,3 +57,101 @@ def deploy_root(self) -> Path:
...
+        if show_app_cursor.rowcount == 0:
+            # Create an app pkg, with distribution = internal to avoid triggering security scan
+            log.info(f"Creating new application package {app_pkg} in account.")
+            self.execute_query_with_role(
+                app_pkg_role,
+                f"""
+                create application package {app_pkg}
+                    comment = {SPECIAL_COMMENT}
+                    distribution = internal
+            """,
```
sfc-gh-astus, line 143: I just remembered that `use role` change role in session, so if you execute 2 queries like

```
_execute_query("use role public; select ...")
_execute_query("select ...")
```

this second query will be executer with public role too.

sfc-gh-turbaszek: For the role changing, my experience says this is pretty nasty for debugging. One thing to consider would be to remove chaning role in single query in favour of chaning role in context. For this this we can use context managers, for example:

```py
with use_role("my_new_role"):
   _execute_query("select 1")
   _execute_query("select 2")

 _execute_query("select 2")  # executed using previous role
```

WDYT?

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -35,3 +60,101 @@ def deploy_root(self) -> Path:
...
+                app_pkg_role,
+            )
+        else:
+            # There can only be one possible pre-existing app pkg with the same name
+            show_app_row = show_app_cursor.fetchone()
+            if isinstance(show_app_row, dict):
+                row_comment = show_app_row[COMMENT_COL]
+            else:  # is list otherwise
+                row_comment = show_app_row[6]
```
sfc-gh-turbaszek, line 154: How this can happen?

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -35,3 +60,101 @@ def deploy_root(self) -> Path:
...
+        )
+
+    def app_run(self) -> SnowflakeCursor:
+
+        # It is possible that snowflake.yml does not contain some of the optional fields,
+        # and there is no snowflake.local.yml present in the search_path.
+        # In this case, we need to use a generated local yml which provides those default values.
+        generated_local_yml_overrides = generate_local_override_yml(
+            self.definition_manager.project_definition
+        )
```
sfc-gh-cgorrie, line 123: I think overall I'd suggest that we remove `generate_local_override_yml` (as we decided to not automatically generate it, only picking it up if it exists) and instead split out the generated defaults into their own functions. The reason I suggest this is that we don't ever need to generate the file into the yml structure, so it might be easier to just generate individual defaults on-demand when we need them.

src/snowcli/cli/nativeapp/manager.py
sfc-gh-cgorrie: Like this idea a lot, we did the same thing in the system test suite (can probably even copy over that exact context manager if we don't have anything like it in SnowCLI yet)

src/snowcli/cli/nativeapp/manager.py
sfc-gh-cgorrie: This is a good question. Actually it was something we were discussing earlier; not sure how exactly to know if a query will return dicts or tuples. Do you know?

src/snowcli/cli/nativeapp/run.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class ApplicationPackageAlreadyExistsError(ClickException):
+    def __init__(self, name: str):
+        super().__init__(
+            f"An Application Package {name} already exists in account that may have been created without snowCLI. "
+        )
+
+
+class SnowflakeSQLExecutionError(ClickException):
```
sfc-gh-cgorrie, line 13: Does this belong in `nativeapp/` or is it more generic?

sfc-gh-cgorrie (commented):
The initial stuff that comes to mind. Happy to do a deeper dive tomorrow.

src/snowcli/cli/nativeapp/manager.py
sfc-gh-bgoel: I've left the original implementation of `generate_local_override_yml` but did away with calling this function.

src/snowcli/cli/nativeapp/manager.py
sfc-gh-bgoel: Ack to both - edited with context manager `use_role`. I was initially under the impression that session is discarded after each `_execute_query` and hence setting the role in one statement won't make it available the next time we do `_execute_query`.

src/snowcli/cli/nativeapp/manager.py
sfc-gh-bgoel: I saw a list implementation, in streamlit's manager under `deploy`:
```
base_url = query_result.fetchone()[0]
``` 
but the return type is a union of dict and list, so I added a case for either return type.

src/snowcli/cli/nativeapp/run.py
sfc-gh-bgoel: I have moved this out to a top level exceptions class, I think it makes sense there.

src/snowcli/cli/stage/diff.py
```diff
@@ -137,3 +141,52 @@ def stage_diff(local_path: Path, stage_fqn: str) -> DiffResult:
...
 
     return result
+
+
+def delete_only_on_stage_files(
+    stage_manager: StageManager,
+    stage_fqn: str,
+    only_on_stage: List[str],
+    role: Optional[str] = None,
+):
```
sfc-gh-cgorrie, line 151: Suggest that these three methods go in another file in this same folder, e.g. `stage/operations.py` or something.

sfc-gh-bgoel:
Tests currently fail when using python 3.8 because the operand `|` is not yet supported here: https://github.com/tiangolo/typer/issues/371
Will check how to achieve it in 3.8

src/snowcli/cli/nativeapp/commands.py
```diff
@@ -72,3 +69,21 @@ def app_bundle(
     manager = NativeAppManager(project_path)
     manager.build_bundle()
     return MessageResult(f"Bundle generated at {manager.deploy_root}")
+
+
+@app.command("run")
+@with_output
+@global_options
```
sfc-gh-bgoel, line 76: With the current implementation of this command, it only accepts `@global_options` and not `@global_options_with_connection`. But as I recall from our recent meeting, we want to support `@global_options_with_connection`, and provide priority to `-r role` if specified, over the role in `snowflake.yml`. Is that correct?

src/snowcli/cli/stage/diff.py
sfc-gh-bgoel: These are specific to our diff method only, and I'm not sure they will be used beyond the scope of diff. As such, I'll like to keep these two here within diff.

src/snowcli/cli/stage/manager.py
sfc-gh-bgoel: An update to this:
1. `put` and `_put`, and `remove` and `_remove` have been separated out into their own commands, instead of adding an overload.
2. But I would like to still keep them in this file as they are doing what `put` and `remove` are, just with additional role changing. Additionally, other contributors might discover these two and find them fitting for their own use case as well.

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -35,3 +66,86 @@ def deploy_root(self) -> Path:
...
+        print_result(
+            ObjectResult(
+                {
+                    "only on local": ", ".join(diff.only_local),
+                    "only on stage": ", ".join(diff.only_on_stage),
+                    "modified/unknown": ", ".join(diff.different),
+                    "identical": ", ".join(diff.identical),
+                }
+            )
+        )
```
sfc-gh-turbaszek, line 101: Please avoid printing outside the command. Returning an output should be limited to the commands so we have proper control about formatting and behaviour.

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -35,3 +66,86 @@ def deploy_root(self) -> Path:
...
+
+        # Upload diff-ed files to app pkg stage
+        log.info(
+            f"Uploading diff-ed files from your local {self.deploy_root} directory to the Snowflake stage."
+        )
```
sfc-gh-turbaszek, line 106: How about moving this log to `sync_local_diff_with_stage`?

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -35,3 +66,86 @@ def deploy_root(self) -> Path:
...
+        project_name = clean_identifier(self.definition["name"])
+        app_pkg = self.definition.get("package", {}).get(
+            "name", default_app_package(project_name)
+        )
+        app_pkg_role = self.definition.get("package", {}).get("role", default_role())
+
+        with self.use_role(app_pkg_role):
+            show_cursor = self._execute_query(
+                f"show application packages like '{app_pkg}'",
+                cursor_class=DictCursor,
```
sfc-gh-turbaszek, line 125: +1, I think we should be using this by default but that's out of scope for this PR

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -35,3 +66,86 @@ def deploy_root(self) -> Path:
...
+                # Create an app pkg, with distribution = internal to avoid triggering security scan
+                log.info(f"Creating new application package {app_pkg} in account.")
+                self._execute_query(
+                    f"""
+                    create application package {app_pkg}
+                        comment = {SPECIAL_COMMENT}
+                        distribution = internal
+                """
+                )
+            elif show_cursor.rowcount > 0:
```
sfc-gh-turbaszek, line 138:
```suggestion
            else:
```
Or can we get negative values?

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -35,3 +66,86 @@ def deploy_root(self) -> Path:
...
+                # There can only be one possible pre-existing app pkg with the same name
+                show_app_row = show_cursor.fetchone()
+                row_comment = show_app_row[
+                    COMMENT_COL
+                ]  # Because we use a DictCursor, we are guaranteed a dictionary instead of indexing through a list.
+
+                if row_comment != SPECIAL_COMMENT:
+                    raise ApplicationPackageAlreadyExistsError(app_pkg)
+            else:  # rowcount is None
+                raise SnowflakeSQLExecutionError()
```
sfc-gh-turbaszek, line 148: Let's handle this first
```
if show_cursor.rowcount is None:
   raise ..
  
if show_cursor.rowcount == 0:
  ...
else:
  ...
```


WDYT?

src/snowcli/cli/nativeapp/commands.py
sfc-gh-astus: Yes, `-r role` will override role from `config.toml`

tests/nativeapp/test_manager.py
```diff
@@ -0,0 +1,79 @@
...
+    mock_stage_diff.return_value = mock_diff_result
+    mock_local_diff_with_stage.return_value = None
+    current_working_directory = os.getcwd()
+    create_named_file(
+        file_name="snowflake.yml",
+        dir=current_working_directory,
+        contents=[demo_snowflake_yml_file],
+    )
+    # for deploy_root
+    # os.makedirs(current_working_directory + "output/deploy/")
```
sfc-gh-astus, line 54: Is this needed?

src/snowcli/cli/nativeapp/commands.py
sfc-gh-turbaszek: So if we need to alter connection configurtion we should use `global_options_with_connection`

src/snowcli/cli/project/definition.py
```diff
@@ -78,3 +78,21 @@ def generate_local_override_yml(project: Union[Dict, YAML]) -> YAML:
...
+    return f"{project_name}_pkg_{user}"
+
+
+def default_role():
+    conn = snow_cli_global_context_manager.get_connection()
+    return conn.role
+
+
+def default_application(*args):
+    project_name: str = args[0]
```
sfc-gh-cgorrie, line 95: Can this (and `default_app_package` above) just be regular functions that take `(project_name: str)`?

sfc-gh-bgoel: My bad, this is left over from a previous commit and I did not revisit them after that - will change in the next commit.

src/snowcli/cli/nativeapp/commands.py
sfc-gh-bgoel: From our team's conversation - for `run` we should not have `global_options_with_connection` especially `-role`. This is because `run` uses two different roles - one for app pkg and one for app, and we would not have clear distinction on which role to actually override, and it would also cause confusion for the user. 
Once we start introducing atomic commands, then we can include the scope there.

src/snowcli/cli/nativeapp/commands.py
```diff
@@ -61,7 +58,7 @@ def app_init(
 
 @app.command("bundle", hidden=True)
 @with_output
-@global_options
+@global_options_with_connection
```
sfc-gh-cgorrie, line 61: Why does bundle need a connection?

src/snowcli/cli/nativeapp/commands.py
sfc-gh-cgorrie: I think without `global_options_with_connection`, the default connection isn't set up! At least I couldn't get `snow app run` to work without changing to that. What's been your experience?

src/snowcli/cli/stage/manager.py
```diff
@@ -40,11 +41,42 @@ def put(
...
+        self,
+        local_path: Union[str, Path],
+        stage_path: str,
+        role: Optional[str] = None,
+        parallel: int = 4,
+        overwrite: bool = False,
+    ) -> SnowflakeCursor:
+        """
+        Internal only method, created for Native Apps use case.
```
sfc-gh-cgorrie, line 53: Would suggest documenting what the function does, rather than why it was created.

src/snowcli/cli/stage/diff.py
sfc-gh-cgorrie: Sure, if you feel strongly about it! My counter-argument: I wouldn't expect there to be methods that actually modify anything in a file called `diff`.

src/snowcli/cli/nativeapp/commands.py
sfc-gh-bgoel: Shouldn't the default connection be coming from `config.toml` ? `global_options_with_connection` is about overrides to an existing default, afaik.

src/snowcli/cli/nativeapp/commands.py
sfc-gh-bgoel: Good catch, will correct this.

src/snowcli/cli/stage/diff.py
sfc-gh-bgoel: The other option is `manager.py` whose methods modify a stage in snowflake. I could put these there if the snowCLI team is open to it, but I'd be hesitant to create a different `stage/operations` for it.

sfc-gh-cgorrie: Definitely up to you! I don't feel extremely strongly.

src/snowcli/cli/common/decorators.py
```diff
@@ -46,6 +47,18 @@ def global_options_with_connection(func: Callable):
...
+    Decorator providing default flags including connection flags for overriding
+    global parameters. Values are updated in global SnowCLI state.
+
+    To use this decorator your command needs to accept **options as last argument.
+    """
+    return _options_decorator_factory(
+        func, [*GLOBAL_CONNECTION_LIMITED_OPTIONS, *GLOBAL_OPTIONS]
+    )
+
+
```
sfc-gh-bgoel, line 61: This change is introduced while we do not have the decorator that @sfc-gh-turbaszek discussed. 
It removes role and warehouse flags as those won't work in the native apps context as we have two separate roles that become active during `snow app run`. 
Alternatively, I can leave this out completely and just use the regular `global_options_with_connection` with a note in the docstring.

src/snowcli/cli/nativeapp/commands.py
```diff
@@ -72,3 +72,40 @@ def app_bundle(
...
+    unless provided in your native app definition in snowflake.yml/snowflake.local.yml.
+    """
+    manager = NativeAppManager(project_path)
+    manager.build_bundle()
+    manager.app_run()
+    return MessageResult(
+        f"Application Package is now active in your Snowflake account!"
+    )
+
+
```
sfc-gh-bgoel, line 96: Can ignore the code below, from the other PR mentioned in the description.

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -35,3 +87,159 @@ def deploy_root(self) -> Path:
...
+
+                actual_owner = show_app_row[OWNER_COL]
+                if actual_owner != self.app_pkg_role:
+                    raise UnexpectedOwnerError(
+                        self.app_name, self.app_role, actual_owner
+                    )
+
+            # Upload files from deploy root local folder to the above stage
+            self.sync_deploy_root_with_stage(app_pkg_role, app_pkg)
+
```
sfc-gh-bgoel, line 172: Can ignore the code below, from the other PR mentioned in the description.

tests/nativeapp/test_manager.py
```diff
@@ -0,0 +1,218 @@
...
+        native_app_manager.deploy_root, "app_pkg.app_src.stage"
+    )
+    mock_local_diff_with_stage.assert_called_once_with(
+        role="new_role",
+        deploy_root_path=native_app_manager.deploy_root,
+        diff_result=mock_diff_result,
+        stage_path="app_pkg.app_src.stage",
+    )
+
+
```
sfc-gh-bgoel, line 92: Can ignore the code below, from the other PR mentioned in the description.

src/snowcli/cli/common/decorators.py
sfc-gh-bgoel: Removed in agreement with @sfc-gh-turbaszek

sfc-gh-turbaszek (approved):

PR #398, opened 2023-09-22
Updated threshold for performance tests

sfc-gh-jsikorski:
### Pre-review checklist
   * [] I've added new automated tests to verify correctness of my new code.

### Changes description
Change in performance tests threshold, due to github runner speed


tests/test_performance.py
```diff
@@ -4,7 +4,7 @@
 import pytest
 
 SAMPLE_AMOUNT = 20
-EXECUTION_TIME_THRESHOLD = 1.0
+EXECUTION_TIME_THRESHOLD = 1.5
```
sfc-gh-pjob, line 7: I'm not sure if this is the right way to deal with the problem. IMO it's like ignoring it completely.
I know that retrying failed jobs is also ignoring but at least we know that the test fails randomly.
With higher threshold we won't know if 1s criteria is not met randomly or always.

sfc-gh-turbaszek: We should try to understand when the problems started. Also do we need to increase by 0.5?

sfc-gh-pjob (approved):

tests/test_performance.py
sfc-gh-pjob: We agreed to increase the threshold but at the same time we're starting analysis of decreased performance.

PR #402, opened 2023-09-25
Added parameters to procedure coverage integration test

sfc-gh-astus:
### Pre-review checklist

### Changes description
Added parameters for procedure in procedure coverage integration test
Procedure and function have unit tests

### DIFF
--- a/src/snowcli/utils.py
+++ b/src/snowcli/utils.py
@@ -570,7 +570,8 @@ def convert_resource_details_to_dict(function_details: SnowflakeCursor) -> dict:
 
 def generate_deploy_stage_name(identifier: str) -> str:
     return (
-        identifier.replace(
+        identifier.replace("()", "")
+        .replace(
             "(",
             "_",
         )
--- a/tests_integration/test_procedure_coverage.py
+++ b/tests_integration/test_procedure_coverage.py
@@ -26,40 +26,46 @@ def test_procedure_coverage_flow(_test_steps):
     _test_steps.snowpark_list_should_return_no_data()
 
     _test_steps.snowpark_init_should_initialize_files_with_default_content()
+    _test_steps.add_parameters_to_procedure("a: int, b: str")
     _test_steps.add_requirements_to_requirements_txt(["coverage"])
     _test_steps.requirements_file_should_contain_coverage()
     _test_steps.snowpark_package_should_zip_files()
 
+    parameters = "(a int, b string)"
     procedure_name = (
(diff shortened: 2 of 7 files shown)


src/snowcli/utils.py
```diff
@@ -570,7 +570,8 @@ def convert_resource_details_to_dict(function_details: SnowflakeCursor) -> dict:
 
 def generate_deploy_stage_name(identifier: str) -> str:
     return (
-        identifier.replace(
+        identifier.replace("()", "")
```
sfc-gh-pjob, line 573: 1. Why this change is required?
2. Do I see correctly that this change will affect also functions and procedures, not only procedure coverages? If yes, then the release notes should be updated.

tests_integration/test_procedure_coverage.py
```diff
@@ -26,40 +26,46 @@ def test_procedure_coverage_flow(_test_steps):
...
     _test_steps.snowpark_init_should_initialize_files_with_default_content()
+    _test_steps.add_parameters_to_procedure("a: int, b: str")
     _test_steps.add_requirements_to_requirements_txt(["coverage"])
     _test_steps.requirements_file_should_contain_coverage()
     _test_steps.snowpark_package_should_zip_files()
 
+    parameters = "(a int, b string)"
     procedure_name = (
-        _test_steps.snowpark_create_with_coverage_wrapper_should_finish_succesfully()
+        _test_steps.snowpark_create_with_coverage_wrapper_should_finish_successfully(
```
sfc-gh-pjob, line 36: Do we know what will happen when an update will be with parameters `"(c int, d string)"`? Do we have any test case for such scenario?

src/snowcli/utils.py
sfc-gh-astus: 1. In previous PR, I changed replace `(` to `_`, but later I found that there is second function doing almost the same thing (one of them is lowering result). When I changed replace in second function, tests with no parameters began to fail.

tests_integration/test_procedure_coverage.py
sfc-gh-astus: I totally forgot about it

sfc-gh-turbaszek (approved):

tests_integration/test_procedure_coverage.py
sfc-gh-pjob: We agreed to create a separate task for fixing this case.

sfc-gh-pjob (approved):

PR #411, opened 2023-09-26
[NADE] Create application / apply package scripts

sfc-gh-cgorrie:
### Pre-review checklist

### Changes description
Implements the "create application" and "apply package scripts" parts of snow app run.

Additional changes:
- `MockCursor` now has a correct `rowcount` property
- removes the ability to pass glob patterns in for package scripts, as well as the default value
- `SqlExecutionMixin.use_role` now correctly unsets the role upon exception
- fixed a bug that prevented single files with directory `dest`inations from working the second time around

### DIFF
--- a/src/snowcli/cli/common/sql_execution.py
+++ b/src/snowcli/cli/common/sql_execution.py
@@ -36,9 +36,11 @@ class SqlExecutionMixin:
         is_different_role = new_role.lower() != prev_role.lower()
         if is_different_role:
             self._execute_query(f"use role {new_role}")
-        yield
-        if is_different_role:
-            self._execute_query(f"use role {prev_role}")
+        try:
+            yield
+        finally:
+            if is_different_role:
+                self._execute_query(f"use role {prev_role}")
 
     def _execute_schema_query(self, query: str):
         self.check_database_and_schema()
--- a/src/snowcli/cli/nativeapp/commands.py
+++ b/src/snowcli/cli/nativeapp/commands.py
@@ -82,16 +82,15 @@ def app_run(
     **options,
 ) -> CommandResult:
     """
-    Creates an application package in your Snowflake account and uploads code files to its stage.
-    As a note, this command does not accept role or warehouse overrides to your config.toml file,
-    because your native app definition in snowflake.yml/snowflake.local.yml is used for any overrides.
+    Creates an application package in your Snowflake account, uploads code files to its stage,
+    then creates a development-mode instance of that application. As a note, this command does not
+    accept role or warehouse overrides to your config.toml file, because your native app definition
+    in snowflake.yml/snowflake.local.yml is used for any overrides.
     """
     manager = NativeAppManager(project_path)
     manager.build_bundle()
     manager.app_run()
-    return MessageResult(
-        f"Application Package is now active in your Snowflake account!"
-    )
+    return MessageResult(f'Your application ("{manager.app_name}") is now live!')
 
 
 @app.command("teardown")
--- a/src/snowcli/cli/nativeapp/manager.py
+++ b/src/snowcli/cli/nativeapp/manager.py
@@ -7,15 +7,19 @@ from typing import List, Optional
 from click.exceptions import ClickException
 from snowcli.exception import SnowflakeSQLExecutionError
 
-from snowflake.connector.cursor import SnowflakeCursor
+from snowflake.connector.cursor import DictCursor
 
-from snowcli.cli.project.util import clean_identifier
+import jinja2
+
+from snowcli.cli.project.util import clean_identifier, extract_schema
 from snowcli.cli.common.sql_execution import SqlExecutionMixin
 from snowcli.cli.project.definition import (
     default_app_package,
-    default_role,
     default_application,
+    default_role,
 )
+from snowcli.output.printing import print_result
+from snowcli.output.types import ObjectResult
 from snowcli.cli.stage.diff import (
     DiffResult,
     stage_diff,
@@ -31,16 +35,30 @@ from snowcli.cli.nativeapp.artifacts import (
 from snowflake.connector.cursor import DictCursor
 
 SPECIAL_COMMENT = "GENERATED_BY_SNOWCLI"
+LOOSE_FILES_MAGIC_VERSION = "dev_stage"
+
 COMMENT_COL = "comment"
 OWNER_COL = "owner"
+VERSION_COL = "version"
 
 log = logging.getLogger(__name__)
 
 
 class ApplicationPackageAlreadyExistsError(ClickException):
+    """An application package not created by SnowCLI exists with the same name."""
+
     def __init__(self, name: str):
         super().__init__(
-            f"An Application Package {name} already exists in account that may have been created without snowCLI. "
+            f"An Application Package {name} already exists in account that may have been created without snowCLI."
+        )
+
+
+class ApplicationAlreadyExistsError(ClickException):
+    """An application not created by SnowCLI exists with the same name."""
+
+    def __init__(self, name: str):
+        super().__init__(
+            f'A non-dev application "{name}" already exists in the account.'
         )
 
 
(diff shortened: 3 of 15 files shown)


src/snowcli/cli/nativeapp/manager.py
```diff
@@ -97,38 +194,128 @@ def sync_deploy_root_with_stage(self, role: str, app_pkg: str):
...
+                    comment = {SPECIAL_COMMENT}
+                """,
+            )
+
+    def app_run(self) -> None:
+        """
+        Implementation of the "snow app run" dev command.
+        """
+        app_pkg_role = (
+            self.definition.get("package", {}).get("role", None) or default_role()
```
sfc-gh-cgorrie, line 297: By moving `default_role()` out of the `.get(...)` call, we short-circuit and can avoid a round-trip to Snowflake when we've set the role in the project definition.

src/snowcli/cli/project/util.py
```diff
@@ -2,6 +2,12 @@
 import os
 from typing import Optional
 
+IDENTIFIER = r'((?:"[^"]*(?:""[^"]*)*")|(?:[A-Za-z_][\w$]{0,254}))'
```
sfc-gh-cgorrie, line 5: This was _slightly_ modified when moved from `schema.py`. Had to make sure all the interior groups were non-capturing, and that the over-arching group _was_ capturing (i.e. no `?:`)

tests/nativeapp/test_manager.py
```diff
@@ -28,8 +34,16 @@
...
+                name: myapp
+                role: app_role
+                warehouse: app_warehouse
+                debug: true
+
             package:
-                scripts: package/*.sql
+                name: app_pkg
+                scripts:
+                    - shared_content.sql
```
sfc-gh-cgorrie, line 46: Even though package scripts are tested elsewhere, I wanted to make sure the syntax was correct (we're no longer accepting globs).

tests/nativeapp/test_manager.py
```diff
@@ -41,7 +55,7 @@ def test_sync_deploy_root_with_stage(
     mock_local_diff_with_stage, mock_stage_diff, mock_execute, temp_dir, mock_cursor
 ):
     mock_execute.return_value = mock_cursor([{"CURRENT_ROLE()": "old_role"}], [])
-    mock_diff_result = DiffResult()
+    mock_diff_result = DiffResult(different="setup.sql")
```
sfc-gh-cgorrie, line 58: To ensure that the diff logic is actually called due to the new `has_changes()` guard.

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -97,38 +194,128 @@ def sync_deploy_root_with_stage(self, role: str, app_pkg: str):
...
+                show_app_row: dict = show_app_cursor.fetchone()
+                if (
+                    show_app_row[COMMENT_COL] != SPECIAL_COMMENT
+                    or show_app_row[VERSION_COL] != LOOSE_FILES_MAGIC_VERSION
+                ):
+                    raise ApplicationAlreadyExistsError(self.app_name)
+
+                actual_owner = show_app_row[OWNER_COL]
+                if actual_owner != app_role:
+                    raise UnexpectedOwnerError(self.app_name, app_role, actual_owner)
```
sfc-gh-bgoel, line 267: Will this ever happen with the current implementation?
We are using the context manager `with self.use_role(app_role):` and hence it will query `show applications` using this `app_role`. Even if the app exists but is not owned by `app_role`, it will not show up here. 

For `show` we should use `accountadmin` in order to actually verify if an object exists and its owner. 
(I need to make this change in my own PR too)

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -64,30 +116,75 @@ def artifacts(self) -> List[ArtifactMapping]:
...
+    def package_name(self) -> str:
+        return self.definition.get("package", {}).get(
+            "name", default_app_package(self.project_identifier)
+        )
+
+    @cached_property
+    def app_name(self) -> str:
+        return self.definition.get("application", {}).get(
+            "name", default_application(self.project_identifier)
+        )
```
sfc-gh-bgoel, line 152: +1 to the property methods introduced above.

src/snowcli/cli/common/sql_execution.py
```diff
@@ -35,6 +35,8 @@ def use_role(self, new_role: str):
         is_different_role = new_role.lower() != prev_role.lower()
         if is_different_role:
             self._execute_query(f"use role {new_role}")
-        yield
-        if is_different_role:
-            self._execute_query(f"use role {prev_role}")
+        try:
```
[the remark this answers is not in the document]
sfc-gh-cgorrie: No, it's to catch any exceptions that happen within the `with` clause when this is used as a context manager. So that if your code raises an exception the original role is applied again.

src/snowcli/cli/nativeapp/manager.py
sfc-gh-cgorrie: I'm against using `accountadmin` because they may only have specialized roles and no access to that one. It's a good point re: ownership but if I'm not mistaken, any apps owned by roles where that role has been granted to the current "app role" will also show up here.

Just tested, and yes, you can see apps owned by other roles if those roles are granted to your role. So I think we should keep this logic so we're guaranteed the app is owned by the role you expect it to be.

src/snowcli/cli/project/schema.py
```diff
@@ -14,12 +14,9 @@
...
 )
 
-IDENTIFIER = r'(?:("[^"]*(""[^"]*)*")|([A-Za-z_][\w$]{0,254}))'
-SCHEMA_AND_NAME = f"{IDENTIFIER}[.]{IDENTIFIER}"
-GLOB_REGEX = r"^[a-zA-Z0-9_\-./*?**\p{L}\p{N}]+$"
-RELATIVE_PATH = r"^[^/][\p{L}\p{N}_\-.][^/]*$"
+from .util import SCHEMA_AND_NAME, IDENTIFIER
 
-# TODO: use the above regexes to validate paths + globs
+# TODO: use the util regexes to validate paths + globs
```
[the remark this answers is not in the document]
sfc-gh-cgorrie: Still yet to be addressed. I don't see it as critical because the actual commands will fail if improper values are given here.

src/snowcli/cli/nativeapp/commands.py
```diff
@@ -73,7 +73,7 @@ def app_bundle(
 
 @app.command("run")
 @with_output
-@global_options
+@global_options_with_connection
```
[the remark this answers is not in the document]
sfc-gh-bgoel: Can ignore this, resolved offline.

src/snowcli/cli/nativeapp/manager.py
sfc-gh-bgoel: >  any apps owned by roles where that role has been granted to the current "app role" will also show up here.
I understand this, but I meant a case where owners are just two separate roles with no interaction with each other. Even if not accountadmin (where I take your point, you may not have access to it), but if an older `roleA` exists which owns an `appA`, and you're trying to query for an `application like 'appA'` from a `roleB`, and `roleA` is not granted to `roleB`, then it would not show up here in the results. But the creation would fail because `appA` still exists, regardless the ownership.

sfc-gh-cgorrie: Ohh I see your point now! Just tested, and in this case, we get this error:

> Application '<app_name>' already exists, but current role has no privileges on it. If this is unexpected and you cannot resolve this problem, contact your system administrator. ACCOUNTADMIN role may be required to manage the privileges on the object.

I think that's good enough for users to be able to action on, what do you think?

src/snowcli/cli/nativeapp/manager.py
sfc-gh-turbaszek: As I'm working on snowpark changes I'm thinking if we should introduce `NativeApp`, `Function` or `Procedure` objects that would be DTOs. Then the managers are real managers. This change make manager responsibility non-obvious

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -85,30 +125,85 @@ def artifacts(self) -> List[ArtifactMapping]:
...
+
+    @cached_property
+    def app_role(self) -> str:
+        return (
+            self.definition.get("application", {}).get("role", None) or default_role()
+        )
+
+    @cached_property
+    def debug_mode(self) -> bool:
+        return self.definition.get("application", {}).get("debug", True)
```
sfc-gh-turbaszek, line 175: Should we have a generic method for this? Like `self.definition.get("application", "debug", default=True)

src/snowcli/cli/nativeapp/manager.py
sfc-gh-cgorrie: Sure, open to that. Can we treat it as an enhancement?

src/snowcli/cli/nativeapp/manager.py
sfc-gh-cgorrie: I think it goes hand-in-hand with the DTO question you had above. Putting these methods closer to `project/schema.py` would be a win for discovery.

sfc-gh-turbaszek (approved):

PR #414, opened 2023-09-27
[NADE] Edits to snow app init command for Native Apps

sfc-gh-bgoel:
### Pre-review checklist
   * [x] I've confirmed that my changes are up-to-date with the target branch. [as of Sept 28, 1PM EST]
   * [ ] I've described my changes in the release notes. [No, because we do not have release notes for Native Apps just yet]

### Changes description
We would like to make some changes to `snow app init` in the following categories:
1. Move out the templates from residing in the snowCLI to our own repository
2. Rename `init` options to be more clear of their intent. 
3. Remove unnecessary functions. 
4. Add some more comments to explain the flow.


src/snowcli/cli/nativeapp/init.py
```diff
@@ -220,9 +188,13 @@ def _init_with_url_and_no_template(
         # Remove all git history
         rmtree(target_directory.joinpath(".git").resolve())
 
-        # Non-Snowflake git URLs should not have jinja files in their directory structure.
-        # If they do, snowCLI is not responsible for rendering them during init.
-        # If the SNOWFLAKELABS_GITHUB_URL is provided here, rendering is additionally skipped as there is no jinja file at the root.
+        # While non-Snowflake git URLs may have jinja files in their directory structure,
+        # snowCLI is not responsible for rendering them during init because they might have variables other than {{project_name}}.
```
sfc-gh-cgorrie, line 192: One thing that can be done is passing `undefined=jinja2.StrictUndefined` which will fail early in the case where there are undefined variables. Might be a path to having other git repos use jinja2?

src/snowcli/cli/nativeapp/commands.py
```diff
@@ -40,20 +40,20 @@ def app_init(
     name: str = typer.Argument(
         ..., help="Name of the Native Apps project to be initiated."
     ),
-    git_url: str = typer.Option(
+    template_repo: str = typer.Option(
```
sfc-gh-bgoel, line 43: One thing I discovered with cloning here:
If you provide an `https` url to a private repo instead of using the ssh url, the command line asks for your github username and password. If you enter your password, you get:
```
remote: Support for password authentication was removed on August 13, 2021.
remote: Please see https://docs.github.com/en/get-started/getting-started-with-git/about-remote-repositories#cloning-with-https-urls for information on currently recommended modes of authentication.
```
which says you need to enter your personal access token in the password prompt instead of your password. 
Should we mention this here in our CLI, or leave the logs as they are since they are descriptive enough to lead the user to the answer?

sfc-gh-astus: I think it would be better to mention it in help

sfc-gh-cgorrie (approved):
LGTM. Not templating the README makes sense to me.

sfc-gh-turbaszek (approved):

PR #415, opened 2023-09-27
[SNOW-901393] External plugins loaded from sys.path

sfc-gh-pjob:
### Pre-review checklist

### Changes description
* Added loading of external plugins installable by simple `pip install` command (plugins need to be available at SnowCLI's `sys.path`)
* Added `snowcli.api.Api` with a goal to make it a class gathering all interfaces which will be implemented in `snowcli.app` package. Plugins should use interfaces, not their implementations.
* Added `snowcli.api.plugin.plugin_config` with `PluginConfig` and `PluginConfigProvider` interface allowing plugins to access their configuration.
* Changed plugin entrypoint group namespace from `snowcli.command` to `snowflake.cli.plugin.command`
* Added new steps required to execute before running integration tests locally (installation of external plugins used in tests)

### Other notes
* **Plugins API is still not ready to use. We need to restructure packages before saying it's ready.**
* I'm not adding release notes in this PR because I'm planning to do it in a PR making changes in the packages structure.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,7 +17,7 @@ a wrapper for virtual environments like [pyenv](https://github.com/pyenv/pyenv).
 Once you created a dedicated virtual environment you can install SnowCLI in editable mode with all required dependencies:
 
 ```bash
-pip install -e ".[dev]"
+pip install -e ".[dev,local-integration-tests]"
 ```
 
 ## Integration tests
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,7 +17,7 @@ a wrapper for virtual environments like [pyenv](https://github.com/pyenv/pyenv).
 Once you created a dedicated virtual environment you can install SnowCLI in editable mode with all required dependencies:
 
 ```bash
-pip install -e ".[dev,local-integration-tests]"
+pip install -e ".[dev]"
 ```
 
 ## Integration tests
@@ -25,6 +25,14 @@ pip install -e ".[dev,local-integration-tests]"
 Every integration test should have `integration` mark. By default, integration tests are not execute when running `pytest`.
 
 To execute only integration tests run `pytest -m integration`.
+
+### Installation of external plugins used in integration tests
+
+To run integration tests you have to install external plugins which are part of the repository:
+```bash
+pip install -e test_external_plugins/snowpark_hello_single_command
+pip install -e test_external_plugins/multilingual_hello_command_group
+```
 
 ### Connection parameters in `config.toml`
 
--- a/src/snowcli/api/__init__.py
+++ b/src/snowcli/api/__init__.py
@@ -1,0 +1,32 @@
+from typing import Optional
+
+from snowcli.api.plugin.plugin_config import PluginConfigProvider
+
+
+class Api:
+    def __init__(
+        self,
+        plugin_config_provider: PluginConfigProvider,
+    ):
+        self.plugin_config_provider = plugin_config_provider
+
+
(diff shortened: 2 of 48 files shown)


CONTRIBUTING.md
```diff
@@ -26,6 +26,14 @@ Every integration test should have `integration` mark. By default, integration t
 
 To execute only integration tests run `pytest -m integration`.
 
+### Installation of external plugins used in integration tests
+
+To run integration tests you have to install external plugins which are part of the repository:
+```bash
+pip install -e test_external_plugins/snowpark_hello_single_command
+pip install -e test_external_plugins/multilingual_hello_command_group
+```
```
sfc-gh-astus, line 35: It sounds like these 2 plugins are required for integration tests? Are they?

src/snowcli/api/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+
+class Api:
+    def __init__(
+        self,
+        plugin_config_provider: PluginConfigProvider,
+    ):
+        self.plugin_config_provider = plugin_config_provider
+
+
+class ApiNotInitializedError(RuntimeError):
```
sfc-gh-astus, line 14: Maybe move it to `snowcli/exception.py`?

src/snowcli/app/api_impl/plugin/plugin_config_provider_impl.py
```diff
@@ -0,0 +1,30 @@
+from typing import List
+
+from snowcli.api import PluginConfigProvider
+from snowcli.api.plugin.plugin_config import PluginConfig
+from snowcli.config import cli_config
+
+
+class PluginConfigProviderImpl(PluginConfigProvider):
+    def get_enabled_plugin_names(self) -> List[str]:
+        config_path = ["snowcli", "plugins"]
```
sfc-gh-astus, line 10: Maybe change it to class constant?

tests_integration/test_external_plugins.py
```diff
@@ -0,0 +1,76 @@
...
+from tests_integration.testing_utils.assertions.test_result_assertions import (
+    assert_that_result_is_successful,
+    assert_that_result_is_successful_and_output_json_contains,
+)
+from tests_integration.snowflake_connector import snowflake_session
+
+
+@pytest.mark.integration
+def test_loading_of_installed_plugins_if_all_plugins_enabled(runner, snowflake_session):
+    runner.use_config("config_with_enabled_all_external_plugins.toml")
```
sfc-gh-astus, line 12: Maybe instead of overriding default config, create new `invoke` where you can pass path to config?

sfc-gh-pjob: I started with that approach but then I noticed that I am limited to only one version of `invoke` and I have to duplicate code.

The way implemented it allows me to use every version of `invoke` and makes the usage less complitated.

src/snowcli/api/__init__.py
sfc-gh-pjob: `api` package shouldn't use classes/functions placed outside of it

CONTRIBUTING.md
sfc-gh-pjob: They are required (as we discussed today).

sfc-gh-astus (approved):

src/snowcli/app/api_impl/plugin/plugin_config_provider_impl.py
sfc-gh-pjob: Discussed and agreed that current version is ok.

PR #428, opened 2023-09-29
Bump version to 1.1.1

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
...

### DIFF
--- a/src/snowcli/__about__.py
+++ b/src/snowcli/__about__.py
@@ -1,3 +1,3 @@
 from __future__ import annotations
 
-VERSION = "1.1.1-dev"
+VERSION = "1.1.1"


src/snowcli/__about__.py
```diff
@@ -1,3 +1,3 @@
 from __future__ import annotations
 
-VERSION = "1.1.1-dev"
+VERSION = "1.1.1"
```
sfc-gh-pjob, line 3: What will happen if users will install it, then we'll introduce some fixes and release the final 1.1.1?

I think, user's pip will see that they already have snowcli 1.1.1 and won't install the version with fixes.

sfc-gh-turbaszek: They will have to force-reinstall. Adding suffix will require change in already tested version which is not a good practice.

sfc-gh-astus (approved):

sfc-gh-pjob (approved):

PR #436, opened 2023-10-03
Access to context in tests

sfc-gh-pjob:
### Pre-review checklist

### Changes description
* Added a handle to app context to have access to it in tests. We can't create a separate new context in tests, because plugin command registration is done only in the context of the running command.

### DIFF
--- a/src/snowcli/app/cli_app.py
+++ b/src/snowcli/app/cli_app.py
@@ -1,11 +1,13 @@
 from __future__ import annotations
 
 import logging
+from dataclasses import dataclass
 from pathlib import Path
 from typing import Optional
 
 import click
 import typer
+from click import Context
 
 from snowcli import __about__
 from snowcli.api import api_provider, Api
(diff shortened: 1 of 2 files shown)


src/snowcli/app/cli_app.py
```diff
@@ -35,6 +37,15 @@
 _commands_registration = CommandsRegistrationWithCallbacks(_api.plugin_config_provider)
 
 
+@dataclass
+class AppContextHolder:
+    # needed to access the context from tests
+    app_context: Optional[Context] = None
```
sfc-gh-astus, line 43: I'm not sure if we should add something to code just for testing

sfc-gh-pjob: That's a good point, but I don't see any other alternative.

sfc-gh-jsikorski: The alternative is to add callback option. So maybe we should wait for @sfc-gh-turbaszek decision - which is better?

sfc-gh-jsikorski: Yes, it does - we get the list of params and commands, so it`s easy to check, if help is present

sfc-gh-pjob: Ok but, am I wrong thinking that the callback is also only for purpose of these tests?

sfc-gh-jsikorski: Yes it is

sfc-gh-jsikorski (approved):

src/snowcli/app/cli_app.py
sfc-gh-pjob: During our today's meeting we agreed that we accept my solution.

PR #452, opened 2023-10-05
Avoiding options duplicates [No-callback approach]

sfc-gh-jsikorski:
### Pre-review checklist

### Changes description
Adding automated tests for duplicated options - this time without adding callback

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -5,7 +5,7 @@ import json
 import typing as t
 
 from click import Command
-from typer.core import TyperArgument
+from typer.core import TyperArgument, TyperOption
 
 from snowcli.__about__ import VERSION
 from snowcli.app.cli_app import app_context_holder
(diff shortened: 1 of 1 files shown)


tests/test_main.py
```diff
@@ -89,3 +89,37 @@ def _check(command: Command, path: t.Optional[t.List] = None):
...
+
+        if duplicated_params := check_options_for_duplicates(command.params):
+            duplicates[" ".join(path)] = duplicated_params
+
+        if hasattr(command, "commands"):
+            for command_name, command_info in command.commands.items():
+                _check(command_info, [*path, command_name])
+
+    def check_options_for_duplicates(params: t.List[TyperOption]) -> t.Set[str]:
+        RESERVED_FLAGS = ["-h", "--help"]  # noqa: N806
```
sfc-gh-pjob, line 112: I'm not sure if `-h` should be reserved. 

It looks like Typer or click is handling the case well and removes `-h` from `--help` if a command requires `-h`.
Are you able to find any command which really has `-h` conflict?

However I see that having `-h` sometimes for `--help` and sometimes for sth else makes our options not consistent.

sfc-gh-jsikorski: Done

sfc-gh-pjob (approved):

PR #453, opened 2023-10-05
Support temporary connection in test command

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
...

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -16,6 +16,7 @@
    * `snow spcs service` for services
  * `snow sql` works now with `snowflake.yml` file. The variables defined in the new `env` section
        of `snowflake.yml` will be used to expand templates.
+ * `snow connection test` command accepts all connection parameters including support for temporary connection.
 
 ## Fixes and improvements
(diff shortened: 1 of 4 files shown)


src/snowcli/cli/connection/commands.py
```diff
@@ -188,11 +188,11 @@ def add(
...
-@global_options
+@global_options_with_connection
 @with_output
-def test(connection: str = ConnectionOption, **options) -> CommandResult:
+def test(**options) -> CommandResult:
     """
     Tests the connection to Snowflake.
     """
-    connect_to_snowflake(connection_name=connection)
+    connect_to_snowflake(**options)
```
sfc-gh-pjob, line 197: 1. IMO we should rewrite `connect_to_snowflake` arguments manually.
I don't like introducing some silent contract between that function and the command options.

2. If you agree, then IMO we should also revert old `connection_name` function's argument name.

3. Can't we use `get_connection` from global context with `force`?

sfc-gh-jsikorski (approved):

PR #471, opened 2023-10-11
Proper identifier unquoting

sfc-gh-cgorrie:
### Pre-review checklist

### Changes description
- rename `identifier_as_part` to `unquote_identifier`
- use `unquote_identifier` for all LIKE queries + introspecting returned results so any identifier will work properly
- search in returned resultsets for the right row, rather than assuming it's the first one (due to LIKE being unable to restrict to exact case)
- fix tests so that unquoted identifiers are returned in UPPERCASE
- fix a bug in config merging uncovered by using quoted identifiers (`YAML` objects can be scalar)

### DIFF
--- a/src/snowcli/cli/nativeapp/manager.py
+++ b/src/snowcli/cli/nativeapp/manager.py
@@ -108,9 +108,7 @@ class MissingSchemaError(ClickException):
         super().__init__(f'Identifier missing a schema qualifier: "{identifier}"')
 
 
-def find_row(
-    cursor: SnowflakeCursor, predicate: Callable[[Union[dict, tuple]], bool]
-) -> Optional[Union[dict, tuple]]:
+def find_row(cursor: DictCursor, predicate: Callable[[dict], bool]) -> Optional[dict]:
     """Returns the first row that matches the predicate, or None."""
     return next(
         (row for row in cursor.fetchall() if predicate(row)),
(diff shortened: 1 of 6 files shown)


src/snowcli/cli/nativeapp/manager.py
```diff
@@ -319,23 +331,36 @@ def app_exists(self) -> bool:
...
         with self.use_role(self.app_role):
             show_app_cursor = self._execute_query(
-                f"show applications like '{self.app_name}'", cursor_class=DictCursor
+                f"show applications like '{unquote_identifier(self.app_name)}'",
+                cursor_class=DictCursor,
+            )
+            return (
+                find_row(
+                    show_app_cursor,
+                    lambda row: row[NAME_COL] == unquote_identifier(self.app_name),
```
sfc-gh-bgoel, line 340: Should we make this into a utility function instead of defining it twice in lines 291 and 340?

sfc-gh-bgoel (approved):
LGTM, minor comment.

src/snowcli/cli/nativeapp/manager.py
sfc-gh-cgorrie: I debated it. One thing I didn't want to do was hide the fact that we were unquoting the identifier. We could do something like:
```python
find_row_matching(show_app_cursor, name=unquote_identifier(self.app_name))
```
But maybe that's too magical (unpacking kwargs as things that need to match in the dict)? What do you think?

sfc-gh-cgorrie: I tried it out; it doesn't really help with readability and it makes it harder to use a constant for `NAME_COL`, so I think we should keep things as-is.

sfc-gh-turbaszek (approved):

PR #478, opened 2023-10-12
Added support for private key auth

sfc-gh-cconner:
### Pre-review checklist
   * [* ] I've confirmed that my changes are up-to-date with the target branch.

### Changes description
Added support for keypair authentication.


src/snowcli/cli/common/flags.py
```diff
@@ -67,6 +67,26 @@
...
+    help="Snowflake authenticator. Overrides the value specified for the connection.",
+    hide_input=True,
+    callback=ConnectionDetails.update_callback("authenticator"),
+    show_default=False,
+    rich_help_panel=_CONNECTION_SECTION,
+)
+
+PrivatekeypathOption = typer.Option(
+    None,
+    "--private_key_path",
```
sfc-gh-turbaszek, line 82:
```suggestion
    "--private-key-path",
```
To be consistent with other options

src/snowcli/snow_connector.py
```diff
@@ -35,15 +44,36 @@ def connect_to_snowflake(temporary_connection: bool = False, connection_name: Op
             {k: v for k, v in overrides.items() if v is not None}
         )
 
+    private_key = None
+    if "private_key_path" in connection_parameters:
+        if (
+            "authenticator" in connection_parameters
+            and connection_parameters["authenticator"] == "SNOWFLAKE_JWT"
+        ):
```
sfc-gh-turbaszek, line 52:
```suggestion
        if connection_parameters.get("authenticator") == "SNOWFLAKE_JWT":
```

src/snowcli/snow_connector.py
```diff
@@ -35,15 +44,36 @@ def connect_to_snowflake(temporary_connection: bool = False, connection_name: Op
...
+                return snowflake.connector.connect(
+                    application=_find_command_path(),
+                    **connection_parameters,
+                )
+            else:
+                return snowflake.connector.connect(
+                    application=_find_command_path(),
+                    private_key=private_key,
+                    **connection_parameters,
+                )
```
sfc-gh-turbaszek, line 76: I would suggest to do something like this:
```py
connection_parameters["private_key"] = private_key
with contextlib.redirect_stdout(None):
                return snowflake.connector.connect(
                    application=_find_command_path(),
                    **connection_parameters,
                )
```
WDYT?

src/snowcli/snow_connector.py
```diff
@@ -16,8 +17,16 @@
 from snowcli.config import cli_config, get_default_connection
 from snowcli.exception import SnowflakeConnectionError, InvalidConnectionConfiguration
 
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives.serialization import Encoding
+from cryptography.hazmat.primitives.serialization import NoEncryption
+from cryptography.hazmat.primitives.serialization import PrivateFormat
+from cryptography.hazmat.primitives.serialization import load_pem_private_key
```
sfc-gh-turbaszek, line 24: Please move the imports to `load_pem_to_der` so we don't import those modules every time

src/snowcli/snow_connector.py
```diff
@@ -56,3 +86,46 @@ def _find_command_path():
...
+
+    private_key_passphrase = os.getenv("PRIVATE_KEY_PASSPHRASE", None)
+    if (
+        private_key_pem.startswith(ENCRYPTED_PKCS8_PK_HEADER)
+        and private_key_passphrase is None
+    ):
+        raise Exception(
+            "Encrypted private key, you must provide the"
+            "passphrase in the environment variable PRIVATE_KEY_PASSPHRASE"
+        )
```
sfc-gh-turbaszek, line 107: Correct me if I'm wrong, but we can throw this exception if PRIVATE_KEY_PASSPHRASE variable is not set, so we can do this validation before reading the file to potentially fail faster

src/snowcli/cli/common/flags.py
```diff
@@ -67,6 +67,26 @@
...
+    help="Snowflake private key path. Overrides the value specified for the connection.",
+    hide_input=True,
+    callback=ConnectionDetails.update_callback("private_key_path"),
+    show_default=False,
+    rich_help_panel=_CONNECTION_SECTION,
```
sfc-gh-turbaszek, line 87:
```suggestion
    rich_help_panel=_CONNECTION_SECTION,
```
```suggestion
    rich_help_panel=_CONNECTION_SECTION,
    exists=True,
    file_okay=True,
    dir_okay=False,
```
So we fail early on if the file does not exist

src/snowcli/snow_connector.py
```diff
@@ -56,3 +86,46 @@ def _find_command_path():
...
+
+    if not private_key_pem.startswith(
+        ENCRYPTED_PKCS8_PK_HEADER
+    ) and not private_key_pem.startswith(UNENCRYPTED_PKCS8_PK_HEADER):
+        raise Exception(
+            "Private key provided is not in PKCS#8 format. " "Pleas use correct format."
```
sfc-gh-turbaszek, line 113:
```suggestion
            "Private key provided is not in PKCS#8 format. Please use correct format."
```

src/snowcli/snow_connector.py
sfc-gh-cconner: So, I tried this initially, but:

`connection_parameters["private_key"] = private_key`

Fails because private key is a bytes object.  Any thoughts on how I could fix that?

`_ConvertError: Invalid type <class 'bytes'>`

src/snowcli/snow_connector.py
sfc-gh-cconner: Users can use an unencrypted key, so we have to read the file to know if it requires the passphrase.

src/snowcli/snow_connector.py
```diff
@@ -35,15 +38,33 @@ def connect_to_snowflake(temporary_connection: bool = False, connection_name: Op
...
         )
 
+    private_key = None
+    if "private_key_path" in connection_parameters:
+        if connection_parameters.get("authenticator") == "SNOWFLAKE_JWT":
+            private_key_passphrase = os.getenv("PRIVATE_KEY_PASSPHRASE", None)
+            private_key = load_pem_to_der(connection_parameters["private_key_path"])
+            del connection_parameters["private_key_path"]
+        else:
+            raise Exception(
```
sfc-gh-pjob, line 48:
```suggestion
            raise ClickException(
```
We use `ClickException` to have nice handling and formatting.

src/snowcli/snow_connector.py
```diff
@@ -35,15 +38,33 @@ def connect_to_snowflake(temporary_connection: bool = False, connection_name: Op
             {k: v for k, v in overrides.items() if v is not None}
         )
 
+    private_key = None
+    if "private_key_path" in connection_parameters:
+        if connection_parameters.get("authenticator") == "SNOWFLAKE_JWT":
+            private_key_passphrase = os.getenv("PRIVATE_KEY_PASSPHRASE", None)
```
sfc-gh-pjob, line 44: `private_key_passphrase` is unused here.

src/snowcli/snow_connector.py
```diff
@@ -56,3 +77,52 @@ def _find_command_path():
...
+
+    with open(private_key_path, "rb") as f:
+        private_key_pem = f.read()
+
+    private_key_passphrase = os.getenv("PRIVATE_KEY_PASSPHRASE", None)
+    if (
+        private_key_pem.startswith(ENCRYPTED_PKCS8_PK_HEADER)
+        and private_key_passphrase is None
+    ):
+        raise Exception(
```
sfc-gh-pjob, line 101: Please use `ClickException`

src/snowcli/snow_connector.py
```diff
@@ -56,3 +77,52 @@ def _find_command_path():
...
+            "Encrypted private key, you must provide the"
+            "passphrase in the environment variable PRIVATE_KEY_PASSPHRASE"
+        )
+
+    if not private_key_pem.startswith(
+        ENCRYPTED_PKCS8_PK_HEADER
+    ) and not private_key_pem.startswith(UNENCRYPTED_PKCS8_PK_HEADER):
+        raise Exception(
```
sfc-gh-pjob, line 109: Please use `ClickException`

tests/test_connection.py
```diff
@@ -259,3 +259,63 @@ def test_temporary_connection(mock_conn, option, runner):
         schema="PUBLIC",
         warehouse="xsmall",
     )
+
+
+@mock.patch("snowflake.connector.connect")
+@pytest.mark.parametrize("option", ["--temporary-connection", "-x"])
```
sfc-gh-pjob, line 265: This test does not need this parameterization. `-x` is the abbreviated form of `--temporary-connection` and both variants of creating temporary connection are already tested in `test_temporary_connection`. You can always use `--temporary-connection` in runner invocation in your test case.

tests/test_connection.py
```diff
@@ -259,3 +259,63 @@ def test_temporary_connection(mock_conn, option, runner):
...
+def test_key_pair_authentication(mock_conn, option, runner):
```
sfc-gh-pjob, line 266: This test works but tests in `test_common_decorators` started to fail.
https://github.com/Snowflake-Labs/snowcli/actions/runs/6508833780/job/17737398464?pr=478

src/snowcli/snow_connector.py
```diff
@@ -56,3 +77,52 @@ def _find_command_path():
...
+    from cryptography.hazmat.backends import default_backend
+    from cryptography.hazmat.primitives.serialization import Encoding
+    from cryptography.hazmat.primitives.serialization import NoEncryption
+    from cryptography.hazmat.primitives.serialization import PrivateFormat
+    from cryptography.hazmat.primitives.serialization import load_pem_private_key
+
+    with open(private_key_path, "rb") as f:
+        private_key_pem = f.read()
+
+    private_key_passphrase = os.getenv("PRIVATE_KEY_PASSPHRASE", None)
```
sfc-gh-pjob, line 96: I wonder if `PRIVATE_KEY_PASSPHRASE` isn't too generic name. Maybe we should add some prefix to it. @sfc-gh-turbaszek WDYT?

sfc-gh-turbaszek: This is the name we use in documentation for python connector. Let's stick to it.

sfc-gh-cconner: That's what I was thinking, thanks!

tests/test_connection.py
```diff
@@ -259,3 +259,62 @@ def test_temporary_connection(mock_conn, option, runner):
...
+                "test_dv",
+                "--schema",
+                "PUBLIC",
+            ]
+        )
+
+    assert result.exit_code == 1
+    mock_conn.assert_called_once_with(
+        application="SNOWCLI.WAREHOUSE.STATUS",
+        private_key=mock.ANY,
```
sfc-gh-turbaszek, line 313: It should not be any. We should check if this is the exact key from the file we provided.

sfc-gh-turbaszek (approved):

PR #479, opened 2023-10-13
Replaced create/update commands with deploy

sfc-gh-astus:
### Pre-review checklist

### Changes description
Replaced create/update commands with deploy for functions and procedures


src/snowcli/cli/snowpark/function/commands.py
```diff
@@ -164,22 +105,30 @@ def function_update(
...
 
     try:
-        current_state = fm.describe(identifier)
-    except:
-        log.info(f"Function does not exists. Creating it from scratch.")
-        replace = True
-    else:
+        current_state = fm.describe(remove_parameter_names(identifier))
+    except ProgrammingError as ex:
+        if ex.msg.__contains__("does not exist or not authorized"):
```
sfc-gh-pjob, line 122: Maybe we can find other way for determining if a function exists? Maybe we can use `SHOW FUNCTIONS LIKE <pattern>` and then filter the results by matching arguments list?

Why?
1) IMO it's dangerous to base our logic on a message of exception from Snowflake.
2) "does not exist or not authorized" means an ambigous state - we don't know if the function really does not exist.

src/snowcli/cli/snowpark/function/commands.py
```diff
@@ -164,22 +105,30 @@ def function_update(
...
     snowpark_package(
         pypi_download,  # type: ignore[arg-type]
         check_anaconda_for_pypi_deps,
         package_native_libraries,  # type: ignore[arg-type]
     )
 
     fm = FunctionManager()
-    sm = StageManager()
+    function_exists = True
+    deploy = False
```
sfc-gh-pjob, line 117:
```suggestion
    create_function_definition = False
```

src/snowcli/cli/snowpark/function/commands.py
```diff
@@ -293,3 +241,21 @@ def function_drop(
     """Deletes a function from a specified environment."""
     cursor = FunctionManager().drop(identifier=identifier)
     return SingleQueryResult(cursor)
+
+
+def _upload_snowpark_artifact(function_manager, function_identifier, file, overwrite):
```
sfc-gh-pjob, line 246: 1) Here: `overwrite` is now always `True`, so probably we don't need this arg anymore
2) `FunctionManager # create`: the same about `overwrite` - maybe we don't need this arg anymore. Maybe we should rename it to `FunctionManager # create_or_replace`?

src/snowcli/cli/snowpark/procedure/commands.py
```diff
@@ -216,22 +168,29 @@ def procedure_update(
...
+    deploy = False
     try:
-        current_state = pm.describe(identifier)
-    except:
-        log.info(f"Procedure does not exists. Creating it from scratch.")
-        replace = True
-    else:
+        current_state = pm.describe(remove_parameter_names(identifier))
+    except ProgrammingError as ex:
+        if ex.msg.__contains__("does not exist or not authorized"):
```
sfc-gh-pjob, line 184: All the same comments as for `function deploy` (not only about this line)

RELEASE-NOTES.md
```diff
@@ -13,8 +13,10 @@
 * Update function or procedure will upload function/procedure code to new path on stage. Previous code will remain under old path on stage.
 * Snowpark command `compute-pool` and its alias `cp` were replaced by `pool` command.
 * `snow snowpark registry` was replaced with `snow registry` command.
-* Removed `snow streamlit create` command. Streamlit can be deployd using `snow streamlit deploy`
+* Removed `snow streamlit create` command. Streamlit can be deployed using `snow streamlit deploy`
 * `snow connection test` now outputs all connection details (except for the password), along with connection status
+* Removed `snow snowpark function creat` and `snow snowpark function update`. Procedures can be deployed using `snow snowpark function deploy`
+* Removed `snow snowpark procedure creat` and `snow snowpark procedure update`. Procedures can be deployed using `snow snowpark procedure deploy`
```
sfc-gh-pjob, line 19:
```suggestion
* Removed `snow snowpark function create` and `snow snowpark function update`. Functions can be deployed using `snow snowpark function deploy`.
* Removed `snow snowpark procedure create` and `snow snowpark procedure update`. Procedures can be deployed using `snow snowpark procedure deploy`.
```

sfc-gh-astus: I added missing `.` in other points too

src/snowcli/cli/snowpark/function/commands.py
sfc-gh-astus: You are right, changed

src/snowcli/cli/snowpark/function/commands.py
sfc-gh-astus: Maybe `create_or_replace`?

sfc-gh-pjob: I like that.

src/snowcli/cli/snowpark/function/commands.py
sfc-gh-pjob: Discussed and agreed that it's not so easy to use `show function like`. Let's stay with the current solution and maybe we'll revisit it in the future.

sfc-gh-pjob (approved):

RELEASE-NOTES.md
```diff
@@ -36,18 +38,18 @@
...
-  * Removing `-c` short option for `--compute-pool` in `snow snowpark jobs create` (it was conflicting with short version of global `--connection` option)
-  * Removing `-c` short option for `--container-name` in `snow snowpark jobs logs` (it was conflicting with short version of global `--connection` option)
-* Fixed parsing of specs yaml in `snow snowpark services create` command
+  * Removing `-p` short option for `--password` option for all commands (backward incompatibility affecting all the commands using a connection) (it was conflicting with various options in a few commands).
+  * Removing `-a` short option for `--replace-always` in `snow snowpark function update` command (it was conflicting with short version of `--check-anaconda-for-pypi-deps`).
+  * Removing `-c` short option for `--compute-pool` in `snow snowpark jobs create` (it was conflicting with short version of global `--connection` option).
+  * Removing `-c` short option for `--container-name` in `snow snowpark jobs logs` (it was conflicting with short version of global `--connection` option).
+* Fixed parsing of specs yaml in `snow snowpark services create` command.
```
sfc-gh-turbaszek, line 55: Let's try to separate such changes, in case of cp or searching git log it will create confusion

src/snowcli/cli/snowpark/common.py
```diff
@@ -43,6 +43,15 @@ def print_log_lines(file: TextIO, name, id, logs):
         print(_prefix_line(prefix, log + "\n"), file=file, end="", flush=True)
 
 
+def remove_parameter_names(identifier: str):
```
sfc-gh-turbaszek, line 46: Can you add one-liner docstring explaining what and why we do?

src/snowcli/cli/snowpark/function/commands.py
```diff
@@ -163,22 +104,30 @@ def function_update(
...
     snowpark_package(
         pypi_download,  # type: ignore[arg-type]
         check_anaconda_for_pypi_deps,
         package_native_libraries,  # type: ignore[arg-type]
     )
 
     fm = FunctionManager()
-    sm = StageManager()
+    function_exists = True
+    create_or_replace = False
```
sfc-gh-turbaszek, line 116: Why is is different from `replace`?

src/snowcli/cli/snowpark/function/commands.py
```diff
@@ -163,22 +104,30 @@ def function_update(
...
+        current_state = fm.describe(remove_parameter_names(identifier))
+    except ProgrammingError as ex:
+        if ex.msg.__contains__("does not exist or not authorized"):
+            function_exists = False
+            create_or_replace = True
+        else:
+            raise ex
+
+    if function_exists and not replace:
+        raise FunctionAlreadyExistsError()
```
sfc-gh-turbaszek, line 128: Let's abstract this to `ObjectAlreadyExistsError(object_type: str, name: str)`, WDYT?

src/snowcli/cli/snowpark/function/commands.py
```diff
@@ -201,25 +150,22 @@ def function_update(
             log.info(
                 "Return type or handler types do not match. Replacing the function."
             )
-            replace = True
+            create_or_replace = True
```
sfc-gh-turbaszek, line 153: WDYT about  encapsulating this into helper method?

src/snowcli/cli/snowpark/function/commands.py
```diff
@@ -201,25 +150,22 @@ def function_update(
...
+    artifact_file = _upload_snowpark_artifact(
         function_manager=fm,
-        stage_manager=sm,
         function_identifier=identifier,
         file=file,
-        overwrite=True,
     )
 
-    if replace:
+    if create_or_replace:
```
sfc-gh-turbaszek, line 161: I think we need to think about a bit longer.

1. Function does not exists, we upload artifacts and create the function
2. Function exists and has same signature - we upload only the artefact
3. Function exists but has different signature - we create a new function (so we and up with having two functions)

src/snowcli/cli/snowpark/common.py
sfc-gh-astus: Added

src/snowcli/cli/snowpark/function/commands.py
sfc-gh-astus: `create_or_replace` tells if we need to execute query

`replace` is a protection against unconscious function/procedure update

src/snowcli/cli/snowpark/function/commands.py
sfc-gh-astus: Good idea, done

src/snowcli/cli/snowpark/function/commands.py
sfc-gh-astus: I left logic as it was before. 

Before:
* create was putting the package and was creating a function/procedure
* update was always putting package and only if there was difference in packages or there was a change with handler or return type

After:
* deploy works as update

src/snowcli/cli/snowpark/common.py
```diff
@@ -43,6 +44,119 @@ def print_log_lines(file: TextIO, name, id, logs):
...
+        ]:
+            updated_package_list.append(coverage_package)
+
+    if updated_package_list:
+        diff = len(updated_package_list) - len(anaconda_packages)
+        log.info(
+            f"Found difference of {diff} packages. Replacing the {object_type.value}."
+        )
+        return True
+    elif (
```
sfc-gh-turbaszek, line 99: As we do return earlier this can be `if` not `elif`

src/snowcli/cli/snowpark/common.py
```diff
@@ -43,6 +44,119 @@ def print_log_lines(file: TextIO, name, id, logs):
...
+
+def _convert_resource_details_to_dict(function_details: SnowflakeCursor) -> dict:
+    import json
+
+    function_dict = {}
+    json_properties = ["packages", "installed_packages"]
+    for function in function_details:
+        if function[0] in json_properties:
+            function_dict[function[0]] = json.loads(
+                function[1].replace("'", '"'),
```
sfc-gh-turbaszek, line 119: Hm, maybe it would be worth to introduce a DTO so we heave easier access than `[0]` or `[1]` WDYT? Maybe we can use `DictCursors` when retrieving the results from snowflake?

sfc-gh-astus: I only moved this code 😢

sfc-gh-turbaszek (dismissed):

sfc-gh-turbaszek (approved):

PR #480, opened 2023-10-13
Native App Integration tests

sfc-gh-cgorrie:
### Pre-review checklist

### Changes description
- Adds integration tests for Native Applications
- Allows setting `SNOWFLAKE_CONNECTIONS_INTEGRATION_HOST`, but also leaving it blank
- widened types from `dict[str, str]` to `dict[str, Any]` for cursors being compared in integration tests (can return e.g. numbers back)
- centralize test projects in `tests/test_data/projects`

### TODO
- [x] get template repository open-sourced
- [x] re-enable init test


tests/project/integration/app/setup.sql
```diff
@@ -0,0 +1,19 @@
+create application role app_public;
+create or alter versioned schema core;
+
+    create or replace procedure core.echo(inp varchar)
```
sfc-gh-bdufour, line 4: is the indentation here intentional?

tests_integration/test_nativeapp.py
```diff
@@ -0,0 +1,149 @@
...
+def pushd(dir: Path):
+    cwd = os.getcwd()
+    os.chdir(dir)
+    try:
+        yield dir
+    finally:
+        os.chdir(cwd)
+
+
+@pytest.mark.skip
```
sfc-gh-bdufour, line 31: ?

tests/project/integration/app/README.md
```diff
@@ -0,0 +1,3 @@
+# README
+
+This is the app/README.md file.
```
sfc-gh-bdufour, line 3: nit: consider saying this is a simple native app used for integ tests or something like that

tests/project/integration/app/setup.sql
sfc-gh-cgorrie: Yep, it's my preferred way of making the different sections in a large setup script easy to read (each schema is a section).

tests_integration/test_nativeapp.py
sfc-gh-bdufour: Ah, thanks, that's what I was missing. It's a token issue.

sfc-gh-bdufour (approved):

src/snowcli/cli/project/util.py
```diff
@@ -60,3 +61,15 @@ def first_set_env(*keys: str):
...
+def is_mount(path: Path):
+    """
+    A wrapper around Path.is_mount() that does not throw an exception
+    on Windows. Instead, we treat Windows as if it has no mount points.
+    On POSIX platforms, behaviour is identical to Path.is_mount().
+    """
+    try:
+        return path.is_mount()
+    except NotImplementedError:
+        return False
```
sfc-gh-turbaszek, line 75: Why not assume positive scenario?

tests/project/integration/app/README.md
```diff
@@ -0,0 +1,4 @@
+# README
```
sfc-gh-turbaszek, line 1: We have project already under `tests/test_data/projects`

src/snowcli/cli/project/util.py
sfc-gh-cgorrie: It doesn't really matter for our use case; we just use it as a stopping condition if it changes from the value of the initial directory. If you'd rather it default to `True` when not implemented, I'm fine with that.

I looked quickly to see if I could find the source code to understand how the Windows one works but my quick timebox didn't lead to the sourcecode :)

tests/project/integration/app/README.md
sfc-gh-cgorrie: Hmm, I had put projects in this directory many weeks ago (for the initial project def implementation), is this the new home? If this is where you'd rather have all of them, would you like them moved in this PR?

sfc-gh-turbaszek: Yes, please let's move all projects under `tests_data`, we are using this place for long time. Sorry we haven't spotted this earlier

sfc-gh-turbaszek (approved):

tests_integration/test_nativeapp.py
```diff
@@ -0,0 +1,142 @@
...
+    temp_dir,
+):
+    project_name = "myapp"
+    result = runner.invoke_with_config(
+        ["app", "init", project_name],
+        env=TEST_ENV,
+    )
+    assert result.exit_code == 0
+
+    with pushd(Path(os.getcwd(), project_name)):
```
sfc-gh-astus, line 44: @sfc-gh-turbaszek introduced project_file context manager https://github.com/Snowflake-Labs/snowcli/commit/3b59d1cb8f6b03ba5358d65ca21e7f2fc8623d44#diff-ad6b0ce118624337f6faf4e75af2aac6cc563a8632cbd95e699a04c1786f185dR208-R216

Logic is almost the same, but we will be consistent

sfc-gh-cgorrie: Do you mind if we refactor in a different ticket? This isn't the only place it's used (also in non-integration tests for definitions).

sfc-gh-astus: ok

sfc-gh-astus (approved):

PR #488, opened 2023-10-20
Add --experimental flag for deploying to embedded stage

sfc-gh-zblackwood:
### Pre-review checklist

### Changes description
Add --experimental flag for deploying streamlit with embedded stage


src/snowcli/cli/streamlit/commands.py
```diff
@@ -169,6 +166,12 @@ def streamlit_deploy(
     open_: bool = typer.Option(
         False, "--open", help="Whether to open Streamlit in a browser.", is_flag=True
     ),
+    experimental: bool = typer.Option(
+        False,
+        "--experimental",
+        help="Use experimental new embedded stage behavior",
+        is_flag=True,
+    ),
```
sfc-gh-astus, line 174: @sfc-gh-pjob created decorator for experimental flag, here is the commit https://github.com/Snowflake-Labs/snowcli/commit/e43eff364e7c8c3932e59ca365e65d789e234625.

You can use `@with_experimental_behaviour()` on command

src/snowcli/cli/streamlit/manager.py
```diff
@@ -47,9 +88,21 @@ def deploy(
...
+            return self._deploy_experimental(
+                streamlit_name,
+                main_file,
+                environment_file,
+                pages_dir,
+                stage_name,
+                warehouse,
+                replace,
+            )
+
```
sfc-gh-turbaszek, line 105: I would say that this part should be in command implemtation so we can explicitly see what parts differs

src/snowcli/cli/streamlit/manager.py
```diff
@@ -38,6 +37,48 @@ def share(self, streamlit_name: str, to_role: str) -> SnowflakeCursor:
...
+        stage_manager.put(main_file, root_location, 4, True)
+
+        if environment_file and environment_file.exists():
+            stage_manager.put(environment_file, root_location, 4, True)
+
+        if pages_dir and pages_dir.exists():
+            stage_manager.put(pages_dir / "*.py", f"{root_location}/pages", 4, True)
+
+        return self.get_url(streamlit_name)
```
sfc-gh-turbaszek, line 80: We can probably extract this to command implementation to have:
```py
def deploy(...)
   manager = StreamlitManager(...)
   if experimental:
      manager.deploy_experimental(...)
   else :
      manager.deploy(...)
    
    url = manager.get_url(streamlit_name)
    return MessageResult(...)
```

WDYT?

src/snowcli/cli/streamlit/manager.py
```diff
@@ -38,6 +37,48 @@ def share(self, streamlit_name: str, to_role: str) -> SnowflakeCursor:
...
         return self._execute_query(f"drop streamlit {streamlit_name}")
 
+    def _deploy_experimental(
+        self,
+        streamlit_name: str,
+        main_file: Path,
+        environment_file: Optional[Path] = None,
+        pages_dir: Optional[Path] = None,
+        stage_name: Optional[str] = None,
+        warehouse: Optional[str] = None,
```
sfc-gh-turbaszek, line 47:
```suggestion
        query_warehouse: Optional[str] = None,
```
To communicate the difference between query warehouse and connection warehouse

src/snowcli/cli/streamlit/manager.py
sfc-gh-zblackwood: Yeah, that's what I originally sought to do, but the challenge is that the order is inherently different -- in the normal case you create the stage, upload to it, and then create the streamlit. In the experimental version, you create the streamlit then upload to the embedded stage. I suppose if I refactor out some pieces I could simplify it down to:

```
if experimental:
   create_streamlit()
   upload_stuff()
else:
   create_stage()
   upload_stuff()
   create_streamlit()
```

Do you feel like that would be clearer?

src/snowcli/cli/streamlit/manager.py
sfc-gh-zblackwood: Yeah, that's not a bad idea. I'll take a swing at refactoring to something like that.

src/snowcli/cli/streamlit/commands.py
sfc-gh-zblackwood: Sweet -- I'll refactor to use that!

sfc-gh-zblackwood: @sfc-gh-pjob I tried using that, but for some reason the flag doesn't seem to be getting passed through, either in the pytest function or when I test it locally. Any suggestions for what I might have done wrong?

src/snowcli/cli/streamlit/manager.py
sfc-gh-zblackwood: I refactored a bit to add more shared functionality between the different deploy methods

src/snowcli/cli/streamlit/manager.py
sfc-gh-zblackwood: Let me know what you think of the refactoring I did above

src/snowcli/cli/streamlit/commands.py
sfc-gh-zblackwood: Fixed!

src/snowcli/cli/streamlit/commands.py
```diff
@@ -185,7 +187,8 @@ def streamlit_deploy(
         stage_name=stage,
         main_file=file,
         replace=replace,
-        warehouse=query_warehouse,
+        query_warehouse=query_warehouse,
```
sfc-gh-turbaszek, line 190: thanks for doing this change!

src/snowcli/cli/streamlit/manager.py
```diff
@@ -67,17 +55,89 @@ def deploy(
...
+            """
+            1. Create streamlit object
+            2. Upload files to embedded stage
+            """
+            # TODO: Support from_stage
+            # from_stage_stmt = f"FROM_STAGE = '{stage_name}'" if stage_name else ""
+            self._create_streamlit(streamlit_name, main_file, replace, query_warehouse)
+
+            database = self._conn.database.upper()  # type: ignore[attr-defined]
+            schema = self._conn.schema.upper()  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
sfc-gh-zblackwood: Good call! I switch to using StageManager to generate the dabase.schema.streamlit_name piece, so it handles a missing database and/or schema the same way that the standard deploy does (Error and defaulting to public)

sfc-gh-turbaszek (approved):

src/snowcli/cli/streamlit/manager.py
sfc-gh-turbaszek: Look awesome now! Thanks

PR #498, opened 2023-10-25
Upgrade apps instead of recreating them

sfc-gh-cgorrie:
### Pre-review checklist

### Changes description
When we detect a change in stage files but the app already exists on the user's account, issue an `alter application ... upgrade using @stage` instead of dropping and re-creating the application.

### DIFF
--- a/src/snowcli/cli/nativeapp/manager.py
+++ b/src/snowcli/cli/nativeapp/manager.py
@@ -304,16 +304,18 @@ class NativeAppManager(SqlExecutionMixin):
                         self.app_name, self.app_role, actual_owner
                     )
 
-                if not diff.has_changes():
-                    # the app already exists and is up-to-date
-                    # ensure debug_mode is up-to-date
+                if diff.has_changes():
+                    # the app needs to be upgraded
+                    log.info(f"Upgrading existing application {self.app_name}.")
                     self._execute_query(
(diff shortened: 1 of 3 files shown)


src/snowcli/cli/nativeapp/manager.py
```diff
@@ -304,16 +304,18 @@ def _create_dev_app(self, diff: DiffResult) -> None:
...
 
-                if not diff.has_changes():
-                    # the app already exists and is up-to-date
-                    # ensure debug_mode is up-to-date
+                if diff.has_changes():
+                    # the app needs to be upgraded
+                    log.info(f"Upgrading existing application {self.app_name}.")
                     self._execute_query(
-                        f"alter application {self.app_name} set debug_mode = {self.debug_mode}"
+                        f"alter application {self.app_name} upgrade using @{self.stage_fqn}"
```
sfc-gh-bgoel, line 311: In addition to `using @stage` , should we keep `set debug_mode` from before to obey `# ensure debug_mode is up-to-date`?

sfc-gh-cgorrie: Can you do both in one command? We no longer `return`, so it falls through and sets the debug mode afterwards (see the changed test).

sfc-gh-bdufour (approved):

sfc-gh-astus (approved):

PR #507, opened 2023-10-27
Fix: Add authenticator and private_key_path fields to ConnectionDetails dataclass

sfc-gh-pjob:
### Pre-review checklist

### Changes description
   * `tests/nativeapp/test_manager.py` was failing on try of running it as single test module (`pytest tests/nativeapp/test_manager.py`) but everything was ok when running it as part of all tests (just `pytest`)
      * The error occurred only in `test_get_snowsight_url` test case.
      * The error was indicating that `authenticator` attribute is not present in `ConnectionDetails` in the global context.
      * The error was correct because that class did not have `authenticator` and `private_key_path` fields.
      * So why it worked for all other tests, why it worked in regular CLI invocations, why `pytest` didn't fail?
        * `ConnectionDetails._connection_update` function sets attributes from CLI connection options. It added new attributes (not present as dataclass fields) to `ConnectionDetails` object for `--authenticator` and `--private-key-path` options.
        * So every in regular CLI invocations, these attributes were present but not as dataclass fields.
        * The same is about all other test cases which use connection -> they all use CLI runner, so there is CLI options handling included in each of them.
        * `pytest` did not fail for `tests/nativeapp/test_manager.py` because other tests invoking CLI runner in other test modules set the attribute before that test. 
      * So why it didn't work for `pytest tests/nativeapp/test_manager.py`?
        *  No test case in `tests/nativeapp/test_manager.py` runs actual CLI. This test module tests manager without invoking the CLI runner. So there was no CLI flags handling and no invocation of `ConnectionDetails._connection_update`.
   * Fixed by:
     * Adding `conftest.py # reset_global_context_after_each_test` autouse fixture to both `tests` and `tests_integrations` to guarantee clear global context for each test case and make `tests/nativeapp/test_manager.py` fail also when running just `pytest`.
     * Adding `authenticator` and `private_key_path` fields to `ConnectionDetails` dataclass.
     * Adding `pytest-randomly` plugin as a dev dependency to introduce random order of tests execution.
     * Fixing other issues found after adding `pytest-randomly`.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -18,10 +18,11 @@ class SnowCLIRunner(CliRunner):
 
     @functools.wraps(CliRunner.invoke)
     def invoke(self, *a, **kw):
+        return self.invoke_with_config_file(self.test_snowcli_config, *a, **kw)
+
+    def invoke_with_config_file(self, config_file: str, *a, **kw):
         kw.update(catch_exceptions=False)
-        return super().invoke(
-            self.app, ["--config-file", self.test_snowcli_config, *a[0]], **kw
-        )
+        return super().invoke(self.app, ["--config-file", config_file, *a[0]], **kw)
 
 
 @pytest.fixture(autouse=True)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,44 +1,39 @@
 from __future__ import annotations
 
-import functools
 import logging
 
 import pytest
-from typer import Typer
-from typer.testing import CliRunner
 
 from snowcli.cli.common import snow_cli_global_context
-
-
-class SnowCLIRunner(CliRunner):
-    def __init__(self, app: Typer, test_snowcli_config: str):
-        super().__init__()
-        self.app = app
-        self.test_snowcli_config = test_snowcli_config
-
-    @functools.wraps(CliRunner.invoke)
-    def invoke(self, *a, **kw):
-        return self.invoke_with_config_file(self.test_snowcli_config, *a, **kw)
-
-    def invoke_with_config_file(self, config_file: str, *a, **kw):
-        kw.update(catch_exceptions=False)
-        return super().invoke(self.app, ["--config-file", config_file, *a[0]], **kw)
+from snowcli.config import config_init
+from tests.testing_utils.fixtures import test_snowcli_config
 
 
 @pytest.fixture(autouse=True)
+# Global context reset is required.
+# Without it, context state from previous tests is visible in following tests.
 def reset_global_context_after_each_test(request):
     snow_cli_global_context.reset_global_context()
     yield
 
 
-# This cleanup function is required to avoid random breaking of logging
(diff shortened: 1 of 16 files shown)


tests/conftest.py
```diff
@@ -22,3 +25,9 @@ def invoke_with_config(self, *args, **kwargs):
             ["--config-file", self.test_snowcli_config, *args[0]],
             **kwargs,
         )
+
+
+@pytest.fixture(autouse=True)
+def reset_global_context_after_each_test(request):
+    snow_cli_global_context.reset_global_context()
```
sfc-gh-turbaszek, line 32: Personally I would consider doing here `snow_cli_global_context_manager.update_global_context(modifications)` instead of introducing test-only helper in production code. WDYT?

sfc-gh-pjob: A) I understand the argument about test-only helper in production code, but your approach would require knowing what are default values of context fields in test code and I don't think it's a good idea. Of course I could make `_create_snow_cli_global_context_manager_with_default_values` a public function but then it's the same problem - making a function public only for test purposes. And then this test-only helper is less verbose.

B) I tried a solution with module reload but then a lot of tests start failing because they import the module, use some objects from that module and then module reloading creates new instances different than these which are already imported and used in test code.

---

I'd stay with my solution.

tests/conftest.py
```diff
@@ -15,10 +19,25 @@ def __init__(self, app: Typer, test_snowcli_config: str):
...
+    yield
+
+
+# This cleanup function is required to avoid random breaking of logging
+# in one test caused by presence of capsys in other test.
+# See similar issues: https://github.com/pytest-dev/pytest/issues/5502
+@pytest.fixture(autouse=True)
+def clean_logging_handlers(request):
```
sfc-gh-pjob, line 38: Is has to be added to fix logging issues started to being visible after introducing random test order.

tests/conftest.py
```diff
@@ -14,11 +18,27 @@ def __init__(self, app: Typer, test_snowcli_config: str):
 
     @functools.wraps(CliRunner.invoke)
     def invoke(self, *a, **kw):
+        return self.invoke_with_config_file(self.test_snowcli_config, *a, **kw)
```
sfc-gh-pjob, line 21: Previously:
- a lot of tests were using `invoke` which didn't use test.conf from resources so either local config file was used or config file loaded in some other test (incorrect behaviour)
- some tests were using `invoke_with_config` which used test.conf from resources (correct behaviour)

Now:
- `invoke` uses test.conf from resources by default
- Almost all the tests use `invoke`
- A few tests with custom config use new `invoke_with_config_file` function

sfc-gh-pjob: + I moved `SnowCliRunner` to `fixtures.py` to resolve circular import

sfc-gh-astus (approved):

PR #527, opened 2023-11-07
[Do Not Merge][NADE] Fix Bugs with Native Apps CLI

sfc-gh-bgoel:
### Pre-review checklist
   * [x] I've confirmed that my changes are up-to-date with the target branch. [as of Nov 8, 10:30AM EST]
   * [ ] I've described my changes in the release notes. - N/A

### Changes description
1. There is a bug with the existing implementation of `_apply_package_scripts` method within `NativeAppManager` where it tries to use a consumer's warehouse for running provider scripts. This should not be the case. Provider scripts should use provider warehouse, if one is specified in project definition file. If not specified there, it will pick the resolved value from the connection information. 
2. There is a bug with the existing implementation of `_create_dev_app` method within `NativeAppManager`  where if the consumer and provider roles are separate, the consumer role is not able to create an application in loose files mode from the app pkg's stage because it does not have access to it. Extra permissions need to be granted for it to work.

### Note
Do not merge/auto-merge if you are the approver. I need to get reviews from both NADE and SnowCLI team, and only then I will manually merge this in.


src/snowcli/cli/nativeapp/manager.py
```diff
@@ -270,8 +274,22 @@ def _apply_package_scripts(self) -> None:
...
+                self._execute_query(f"use warehouse {self._conn.warehouse}")
+            else:
+                raise MissingWarehouseError(
+                    f"""
+                    No warehouse specified for provider role to use.
+                    Please add a warehouse for your provider role in your config.toml file, or
+                    project definition file or via command line.
+                    """
+                )
+        except Exception as err:
```
sfc-gh-bgoel, line 290: For those reviewing: 
The reason why we raise an error but also catch it and only `log.info` is it accommodate the case where a provider can argue they do not want to specify a warehouse because none of their package scripts require use of warehouse. In that case, we do not want to force them to have something here.
But at least printing this will allow for debugging for providers that should have provided a warehouse.

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -270,8 +274,22 @@ def _apply_package_scripts(self) -> None:
                 raise InvalidPackageScriptError(relpath, e)
 
         # once we're sure all the templates expanded correctly, execute all of them
-        if self.warehouse:
-            self._execute_query(f"use warehouse {self.warehouse}")
+        try:
+            if self.package_warehouse:
+                self._execute_query(f"use warehouse {self.package_warehouse}")
+            elif self._conn.warehouse:
+                self._execute_query(f"use warehouse {self._conn.warehouse}")
```
sfc-gh-melnacouzi, line 281: based on our offline discussion, it might be cleaner to deal with only one provider side warehouse, and do this if/else logic above.

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -270,8 +274,20 @@ def _apply_package_scripts(self) -> None:
                 raise InvalidPackageScriptError(relpath, e)
 
         # once we're sure all the templates expanded correctly, execute all of them
-        if self.warehouse:
-            self._execute_query(f"use warehouse {self.warehouse}")
+        try:
+            if self.package_warehouse:
+                self._execute_query(f"use warehouse {self.package_warehouse}")
+            else:
+                raise MissingWarehouseError(
```
sfc-gh-cgorrie, line 281: What about package scripts that don't require a warehouse to excecute? IMO, this should be raised after introspecting the error and seeing it's due to a missing warehouse.

sfc-gh-melnacouzi (dismissed):

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -152,7 +152,11 @@ def stage_schema(self) -> Optional[str]:
         return extract_schema(self.stage_fqn)
 
     @cached_property
-    def warehouse(self) -> Optional[str]:
+    def package_warehouse(self) -> Optional[str]:
+        return self.definition.get("package", {}).get("warehouse", self._conn.warehouse)
```
sfc-gh-cgorrie, line 156: Either both `package_warehouse` and `application_warehouse` should default to `self._conn.warehouse`, or both should default to `None`. I'm for the latter, personally

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -270,8 +274,20 @@ def _apply_package_scripts(self) -> None:
...
+                raise MissingWarehouseError(
+                    f"""
+                    No warehouse specified for provider role to use.
+                    Please add a warehouse for your provider role in your config.toml file, or
+                    project definition file or via command line.
+                    """
+                )
+        except Exception as err:
+            # Directly executing "use warehouse ..." can also result in an error if the session role does not usage privilege on the warehouse
+            log.info(err)
```
sfc-gh-cgorrie, line 290: `log.info` won't be shown unless `--debug` was passed I think?

sfc-gh-cgorrie (changes requested):

src/snowcli/cli/nativeapp/manager.py
sfc-gh-bgoel: Resolved offline.

src/snowcli/cli/nativeapp/manager.py
sfc-gh-bgoel: Updated.

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -270,20 +277,33 @@ def _apply_package_scripts(self) -> None:
...
+
+        try:
+            for i, queries in enumerate(queued_queries):
+                log.info(f"Applying package script: {self.package_scripts[i]}")
+                self._execute_queries(queries)
+        except ProgrammingError as err:
+            if str(err.errno) == "000606" or err.msg.__contains__(
+                "No active warehouse selected in the current session"
+            ):
+                raise MissingWarehouseError(
```
sfc-gh-cgorrie, line 291: Can we interpolate the original error number + message into this error's message? Want to make sure we don't obscure any extra information the backend might send.

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -270,20 +277,33 @@ def _apply_package_scripts(self) -> None:
...
-            log.info(f"Applying package script: {self.package_scripts[i]}")
-            self._execute_queries(queries)
+        if self.package_warehouse:
+            self._execute_query(f"use warehouse {self.package_warehouse}")
+
+        try:
+            for i, queries in enumerate(queued_queries):
+                log.info(f"Applying package script: {self.package_scripts[i]}")
+                self._execute_queries(queries)
+        except ProgrammingError as err:
```
sfc-gh-cgorrie, line 287: I think we should follow this same paradigm for application installation, WDYT?

sfc-gh-astus (approved):

sfc-gh-melnacouzi (dismissed):

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -309,18 +326,21 @@ def _create_dev_app(self, diff: DiffResult) -> None:
...
                     )
-
-                # ensure debug_mode is up-to-date
-                self._execute_query(
-                    f"alter application {self.app_name} set debug_mode = {self.debug_mode}"
-                )
-                return
+                    return
+                except ProgrammingError as err:
+                    self._raise_missing_warehouse_err(err)
```
sfc-gh-astus, line 343: At this moment you don't know if it is `missing warehouse error`, I would suggest a more general name such as `_raise_err`

sfc-gh-astus (dismissed):

sfc-gh-astus (approved):

sfc-gh-cgorrie (approved):

PR #540, opened 2023-11-14
[SNOW-938503] Refactoring global context

sfc-gh-pjob:
### Pre-review checklist

### Changes description
   * Refactored global context


src/snowcli/cli/common/cli_global_context.py
```diff
@@ -0,0 +1,146 @@
...
+class _GlobalContextManager:
+    _connection_details = ConnectionDetails()
+    _cached_connection: Optional[SnowflakeConnection] = None
+
+    enable_tracebacks = DEFAULT_ENABLE_TRACEBACKS
+    output_format = DEFAULT_OUTPUT_FORMAT
+    verbose = DEFAULT_VERBOSE
+    experimental = DEFAULT_EXPERIMENTAL
+
+    def set_default_values(self):
```
sfc-gh-astus, line 60: What do you think about `reset_context`?

src/snowcli/cli/common/cli_global_context.py
```diff
@@ -0,0 +1,146 @@
...
+    def connection_details(self) -> ConnectionDetails:
+        return self._connection_details
+
+    @connection_details.setter
+    def connection_details(self, connection_details: ConnectionDetails):
+        self._connection_details = connection_details
+        self._cached_connection = None
+
+    @property
+    def cacheable_connection(self) -> SnowflakeConnection:
```
sfc-gh-turbaszek, line 77: Do we plan to have other connection? Maybe we can drop the `cachable` prefix?

src/snowcli/cli/common/cli_global_context.py
```diff
@@ -0,0 +1,146 @@
...
+    @property
+    def verbose(self) -> bool:
+        return self._manager.verbose
+
+    @property
+    def experimental(self) -> bool:
+        return self._manager.experimental
+
+
+global_context_manager: _GlobalContextManager = _GlobalContextManager()
```
sfc-gh-turbaszek, line 119: Do we use it somewhere outside this module?

sfc-gh-pjob: Currently only in `conftest.py` files in unit and integration tests.

But I think it should be a public field because you may want to change the global context from the code in other places than callbacks to global options.

`global_context_manager` is separated from `global_context`, so if you use `global_context_manager` which has "write" access, you do it on purpose. If you just need "read" access, then you import `global_context`. IMO both should be public and we should suggest to use `global_context` and only mention that you have a possibility to change the context via `global_context_manager`.

src/snowcli/cli/common/cli_global_context.py
sfc-gh-pjob: Done

src/snowcli/cli/common/cli_global_context.py
sfc-gh-pjob: Done

src/snowcli/cli/common/cli_global_context.py
sfc-gh-turbaszek: Sounds good to me. The only concern that I still have is the naming, python has already concept of "context manager" 😢  not sure if we can do something about it

sfc-gh-astus: `snowcli_context_manager`? 😄

sfc-gh-pjob: Names changed to `cli_context` and `cli_context_manager`

src/snowcli/cli/common/cli_global_context.py
```diff
@@ -0,0 +1,211 @@
...
+        self._password: Optional[str] = None
+        self._authenticator: Optional[str] = None
+        self._private_key_path: Optional[str] = None
+        self._warehouse: Optional[str] = None
+        self._temporary_connection: bool = False
+
+    def __setattr__(self, key, value):
+        super.__setattr__(self, key, value)
+        if key is not "_cached_connection":
+            self._cached_connection = None
```
sfc-gh-turbaszek, line 27: Why it's needed?

src/snowcli/cli/common/flags.py
```diff
@@ -1,28 +1,35 @@
...
 from snowcli.output.formats import OutputFormat
 
 DEFAULT_CONTEXT_SETTINGS = {"help_option_names": ["--help", "-h"]}
 
-
 _CONNECTION_SECTION = "Connection configuration"
 _CLI_BEHAVIOUR = "Global configuration"
 
 
+def _callback(provide_setter: Callable[[], Callable[[Any], Any]]):
```
sfc-gh-turbaszek, line 15:
```suggestion
def _callback(setter: Callable[[], Callable[[Any], Any]]):
```
Won't it be enough?

src/snowcli/cli/common/cli_global_context.py
sfc-gh-pjob: To invalidate cache after connection parameters changes

src/snowcli/cli/common/flags.py
```diff
@@ -1,28 +1,35 @@
...
 ConnectionOption = typer.Option(
     None,
     "--connection",
     "-c",
     "--environment",
     help=f"Name of the connection, as defined in your `config.toml`. Default: `dev`.",
-    callback=ConnectionDetails.update_callback("connection_name"),
+    callback=_callback(
+        lambda: cli_context_manager.connection_context.set_connection_name
+    ),
```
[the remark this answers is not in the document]
sfc-gh-pjob: It won't work correctly because it will use connection_context instance from the moment of flags initialization which can be different (and it is different) than connection_context from the moment of callback execution.

However I can make `ConnectionContext` to be a singleton and add `reset` method to it.

src/snowcli/cli/common/flags.py
sfc-gh-pjob: IMO `provide_setter` sounds like a function which returns a  function, while just `setter` sounds like a function which is a setter it self

src/snowcli/cli/common/cli_global_context.py
sfc-gh-turbaszek: Let's add comment about it

sfc-gh-pjob: Done

src/snowcli/cli/common/flags.py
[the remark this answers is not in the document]
sfc-gh-pjob: But IMO lambda should be used here because we shouldn't depend on implementation of cli_context or connection_context.

sfc-gh-turbaszek (approved):

PR #562, opened 2023-11-23
Query the regionless URL parameter for URL generation

sfc-gh-cgorrie:
### Pre-review checklist

### Changes description
Use the UI bundle to query for regionless redirect, rather than assuming it's always False for internal deployments and always True for production ones.


src/snowcli/cli/connection/util.py
```diff
@@ -19,15 +19,40 @@ def __init__(self, conn: SnowflakeConnection):
...
-    information to determine the deployment, we return the organization
-    name instead, as this can be used in production Snowsight.
+    Determines if the deployment this connection refers to uses
+    regionless URLs in Snowsight (/orgname/account) or regional URLs
+    (/region/account). If we cannot determine the correct value
     """
-    if not conn.host:
-        raise MissingConnectionHostError(conn)
+    try:
+        *_, cursor = conn.execute_string(
```
sfc-gh-bgoel, line 29: Nit: we could be violating the abstraction barrier here since the parent class `SqlExecutionMixin` only exposes `_execute_qeury` and `_execute_queries`. 
Nit because it is still a valid way to do things and avoids the extra call to `_execute_query`.

sfc-gh-bgoel (approved):
LGTM, leaving  the addressing of the nit to your judgement.

src/snowcli/cli/connection/util.py
sfc-gh-cgorrie: We're not in the context of a class with the `SqlExecutionMixin` anymore, we're directly using a `SnowflakeConnection` object. Rather than restrict these utility functions to requiring a class with this mixin, I decided to open them up to anything that has a valid connection.

sfc-gh-bgoel (dismissed):
lgtm, resolved open comment.

sfc-gh-bgoel (approved):

sfc-gh-astus (approved):

PR #564, opened 2023-11-23
Proof-of-concept: only upgrade when necessary

sfc-gh-cgorrie:
### Pre-review checklist

### Changes description
Introspects the manifest and setup scripts to spider out and find the full set of scripts used to install the application (counting `EXECUTE IMMEDIATE FROM <file>`). When re-launching the app with `snow app run`, we only issue an `UPGRADE` command if the manifest or setup scripts have been changed.

### TODO
* [ ] Write unit introspection tests
* [ ] Integration test: change app artifacts 2 ways, one triggers upgrade, one doesn't
* [x] Figure out how Snowflake allows us to escape single quotes

### DIFF
--- a/src/snowcli/cli/nativeapp/introspection.py
+++ b/src/snowcli/cli/nativeapp/introspection.py
@@ -17,7 +17,7 @@ class ManifestNotFoundError(ClickException):
     """
 
     def __init__(self):
-        super().__init__(self.__doc__.strip())
+        super().__init__(self.__doc__)
 
 
 class SetupScriptNotFoundError(ClickException):
@@ -26,7 +26,7 @@ class SetupScriptNotFoundError(ClickException):
     """
 
     def __init__(self, file: Union[Path, str]):
-        super().__init__(f"{self.__doc__.strip()}\nExpected path: {file}")
+        super().__init__(f"{self.__doc__}\nExpected path: {file}")
 
 
 def load_manifest(deploy_root: Path) -> dict:
--- a/src/snowcli/cli/nativeapp/introspection.py
+++ b/src/snowcli/cli/nativeapp/introspection.py
@@ -1,3 +1,4 @@
+import io
 import re
 import codecs
 
@@ -5,6 +6,7 @@ from pathlib import Path
 from typing import List, Union
 
 from click import ClickException
+from snowflake.connector.util_text import split_statements
 from snowcli.cli.nativeapp.artifacts import NotInDeployRootError, resolve_without_follow
 from strictyaml import YAML, load
 
@@ -58,9 +60,11 @@ def apply_single_quote_escapes(literal: str) -> str:
 
 
 def extract_execute_immediate_relpaths(sqltext: str) -> List[str]:
+    raw_statements = split_statements(io.StringIO(sqltext), remove_comments=True)
+    statements = "\n".join([stmt for (stmt, _is_put_or_get) in raw_statements])
     return [
         apply_single_quote_escapes(relpath)
-        for relpath in EXECUTE_IMMEDIATE_FROM_REGEX.findall(sqltext)
+        for relpath in EXECUTE_IMMEDIATE_FROM_REGEX.findall(statements)
     ]
 
 
--- a/src/snowcli/cli/nativeapp/introspection.py
+++ b/src/snowcli/cli/nativeapp/introspection.py
@@ -1,4 +1,6 @@
 import re
+import codecs
+
 from pathlib import Path
 from typing import List, Union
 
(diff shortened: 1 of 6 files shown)


src/snowcli/cli/nativeapp/introspection.py
```diff
@@ -0,0 +1,74 @@
...
+
+    all_scripts: List[Path] = [base_script]
+    queued_scripts = list(all_scripts)
+
+    while len(queued_scripts) > 0:
+        script = queued_scripts.pop()
+        with open(script, "r") as f:
+            script_content = f.read()
+            for relpath in EXECUTE_IMMEDIATE_FROM_REGEX.findall(script_content):
+                referenced_script = resolve_without_follow(script.parent / relpath)
```
sfc-gh-cgorrie, line 65: TODO: make sure absolute path is within deploy root

src/snowcli/cli/nativeapp/introspection.py
```diff
@@ -0,0 +1,77 @@
+import re
+from pathlib import Path
+from typing import List, Union
+
+from click import ClickException
+from snowcli.cli.nativeapp.artifacts import resolve_without_follow, NotInDeployRootError
+from strictyaml import YAML, load
+
+EXECUTE_IMMEDIATE_FROM_REGEX = re.compile(
+    r"execute\s+immediate\s+from\s+'([^']+?)'", re.IGNORECASE
```
sfc-gh-melnacouzi, line 10: could the file name include escaped single quote?

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -228,6 +229,15 @@ def app_role(self) -> str:
     def debug_mode(self) -> bool:
         return self.definition.get("application", {}).get("debug", True)
 
+    @cached_property
```
sfc-gh-melnacouzi, line 232: what does caching do in this case? does any file change invalidate the cache?

sfc-gh-cgorrie: Yes, any file change would invalidate the cache. But the lifetime of a SnowCLI execution is only the command, so we're not interested in picking up changes that might happen during the few seconds we're executing.

src/snowcli/cli/nativeapp/introspection.py
sfc-gh-cgorrie: Good idea. TIL that we also support block comments (`/* ... */`) so this will get pretty complicated quickly. I think the move is to pre-process out all comments and then only run the regex on the comment-less code.

src/snowcli/cli/nativeapp/introspection.py
```diff
@@ -0,0 +1,112 @@
...
+
+from click import ClickException
+from snowcli.cli.nativeapp.artifacts import NotInDeployRootError, resolve_without_follow
+from strictyaml import YAML, load
+
+SINGLE_QUOTED_STRING_REGEX = r"'((?:\\.|[^'\n])+?)'"
+
+EXECUTE_IMMEDIATE_FROM_REGEX = re.compile(
+    rf"execute\s+immediate\s+from\s+{SINGLE_QUOTED_STRING_REGEX}", re.IGNORECASE
+)
```
sfc-gh-sgao, line 15: I didn't try this myself, but I was told this should work (or something similar):

`execute immediate 'execute immediate from ''' || prefix || suffix || '''`

That is, the name of the file to be executed is constructed dynamically. I wonder how this will be handled. It'll probably trigger `SetupScriptNotFoundError` and fail the command? Or does it fallback to "I don't know what's going on; I'll upgrade"? The latter is probably more desirable.

A crazy person could also write

`execute immediate 'execute ' || 'immediate ' || 'from ' || ...`

This makes it impossible to detect reliably. Then the question becomes: does being useful 99% of the time justify being incorrect for the remaining 1%?

There's also the interaction with the "manual reload - setup script" work. If a delta has been applied manually, then the file-content based diff could also become inaccurate. (Files may look the same, but the intention may be to revert the manually applied delta.) (Maybe we can require that the delta has to be one of the setup script files, and we upload it whenever we apply it, to keep app and stage in sync.)

This is very interesting, and thanks for experimenting with it. I just want to make sure we consider other relevant scenarios to avoid any unintended consequences.

sfc-gh-cgorrie: Good points. I know it's impossible to detect all of it without actually running through the real interpreter. With this proof-of-concept, the application will fail to upgrade properly because the regex is pretty restrictive: it won't find the secondary file, and as such it won't know it needs to upgrade. Alternative designs to get around this:

1. Mark each artifact that is / could be a setup script in the project definition so we don't have to scan.
2. Treat any `.sql` file in the deploy root as if it could be a setup script; always upgrade when any are changed. This is a simple solution, but I think we should start supporting a SnowGit-compatible approach where the `artifacts:` section is omitted and the deploy root == the directory where `snowflake.yml` exists (minus that file, of course). In this case, any utility scripts would cause an unnecessary upgrade.
3. Detect dynamic `execute immediate` code in the setup script, and always upgrade if it is present.

Considering that a nested `execute immediate ... execute immediate from` is the only way to conditionally include another setup script (I think?), we need to support some sort of approach if we want users to be able to do things like feature flagging to build different versions of app packages. Of course, this also needs some amount of templating functionality (what is the setup script making decisions based on otherwise?).

As far as interaction with manual setup script stuff, I think the right approach here is to split the two into very different workflows. We need to flag that the app has been "tainted" by manual patches and simply always upgrade if it has been, then clear the flag. This could be done by modifying the comment, potentially. Tags would be nice, but they're enterprise-only. This keeps the semantic that `snow app run` is always about syncing your local copy and updating the app from that.

sfc-gh-cgorrie: Agreed that false negatives cannot be allowed. I think the safest and easiest change is the 2nd one I proposed. It only contains false positives for the project-root-as-deploy-root case (which we do not support right now anyway).

1. Has `manifest.yml` changed, or are there any `.sql` changes in the deploy root?
   a. If so, issue an `UPGRADE`.
   b. Otherwise, just let stage dev mode capture the change.

sfc-gh-bdufour: As discussed offline, this works if there is no way to externally influence decisions made in the setup script (e.g. by changing a data file). Let's circle back after we get answers regarding whether that's feasible now (and if not, whether it could be in the future).

PR #565, opened 2023-11-27
Improve result streaming

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
Improve streaming results in case of multiple queries execution.

### DIFF
--- a/src/snowcli/cli/common/sql_execution.py
+++ b/src/snowcli/cli/common/sql_execution.py
@@ -38,7 +38,7 @@ class SqlExecutionMixin:
         instead of list. In case of executing multiple queries are executed one by one. This mean we can
         access result of previous queries while evaluating next one. For example, we can print the results.
         """
-        self._log.info("Executing %s", sql_text)
+        self._log.debug("Executing %s", sql_text)
         stream = StringIO(sql_text)
         stream_generator = self._conn.execute_stream(
             stream, remove_comments=remove_comments, cursor_class=cursor_class, **kwargs
--- a/src/snowcli/cli/common/sql_execution.py
+++ b/src/snowcli/cli/common/sql_execution.py
@@ -4,25 +4,13 @@ import logging
 from contextlib import contextmanager
 from functools import cached_property
 from io import StringIO
-from textwrap import dedent, indent
+from textwrap import dedent
 from typing import Iterable
 
 from click import ClickException
 from snowcli.cli.common.cli_global_context import cli_context
 from snowflake.connector.cursor import DictCursor, SnowflakeCursor
 from snowflake.connector.errors import ProgrammingError
-
-
-class _SnowcliCursor(SnowflakeCursor):
-    def execute(self, *args, **kwargs):
-        """This allows us to show failing query to users."""
-        # TODO: Use exception annotation once we support Python 3.11
-        try:
-            super().execute(*args, **kwargs)
-        except Exception:
-            print("Query:")
-            print(indent(args[0], "    "))
-            raise
 
 
 class SqlExecutionMixin:
@@ -37,12 +25,12 @@ class SqlExecutionMixin:
     def _log(self):
         return logging.getLogger(__name__)
 
-    def execute_string(
+    def _execute_string(
         self,
         sql_text: str,
         remove_comments: bool = False,
         return_cursors: bool = True,
-        cursor_class: SnowflakeCursor = _SnowcliCursor,
+        cursor_class: SnowflakeCursor = SnowflakeCursor,
         **kwargs,
     ) -> Iterable[SnowflakeCursor]:
         """
@@ -50,6 +38,7 @@ class SqlExecutionMixin:
         instead of list. In case of executing multiple queries are executed one by one. This mean we can
         access result of previous queries while evaluating next one. For example, we can print the results.
         """
+        self._log.info("Executing %s", sql_text)
         stream = StringIO(sql_text)
         stream_generator = self._conn.execute_stream(
             stream, remove_comments=remove_comments, cursor_class=cursor_class, **kwargs
@@ -61,7 +50,7 @@ class SqlExecutionMixin:
         return last_result
 
     def _execute_queries(self, queries: str, **kwargs):
-        return list(self._conn.execute_string(dedent(queries), **kwargs))
+        return list(self._execute_string(dedent(queries), **kwargs))
 
     @contextmanager
     def use_role(self, new_role: str):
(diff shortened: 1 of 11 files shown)


sfc-gh-astus (approved):

src/snowcli/cli/common/sql_execution.py
```diff
@@ -3,14 +3,28 @@
...
 
 
+class _SnowcliCursor(SnowflakeCursor):
+    def execute(self, *args, **kwargs):
+        """This allows us to show failing query to users."""
+        # TODO: Use exception annotation once we support Python 3.11
+        try:
+            super().execute(*args, **kwargs)
+        except Exception:
+            print("Query:")
```
sfc-gh-pjob, line 23: It looks like python's `print`, not click's `print`. Did you consider click's `print` and check it? Maybe we can get some nice formatting. I am just curious, not asking to change it.

src/snowcli/cli/common/sql_execution.py
```diff
@@ -3,14 +3,28 @@
...
+            print(indent(args[0], "    "))
```
sfc-gh-pjob, line 24: Do you think we should think about security of printing the query always, even if CLI is not in `verbose` / `debug` mode?

sfc-gh-turbaszek: Interesting point, there's definitely is some security concern. Using verbose/debug already enables query logging.

src/snowcli/cli/common/sql_execution.py
```diff
@@ -23,13 +37,31 @@ def _conn(self):
...
+        return stream_generator if return_cursors else list()
+
     def _execute_query(self, query: str, **kwargs):
         *_, last_result = self._execute_queries(query, **kwargs)
         return last_result
 
     def _execute_queries(self, queries: str, **kwargs):
-        self._log.debug("Executing %s", queries)
-        return self._conn.execute_string(dedent(queries), **kwargs)
+        return list(self._conn.execute_string(dedent(queries), **kwargs))
```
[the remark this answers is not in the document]
sfc-gh-turbaszek: Good point, fixed

src/snowcli/cli/common/sql_execution.py
```diff
@@ -23,13 +37,31 @@ def _conn(self):
     def _log(self):
         return logging.getLogger(__name__)
 
+    def execute_string(
```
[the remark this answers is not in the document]
sfc-gh-turbaszek: Good point, changed to private

tests_integration/test_sql.py
```diff
@@ -36,3 +39,43 @@ def _round_values(results):
...
+    )
+
+    add_row = mock_get_table().add_row
+    results = [p.args[0] for p in add_row.mock_calls]
+    assert len(results) == 3
+    start, wait, end = results
+    assert wait == "waited 10 seconds"
+    start_ts = datetime.strptime(start[:8], "%H:%M:%S")
+    end_ts = datetime.strptime(end[:8], "%H:%M:%S")
+    duration = (end_ts - start_ts).total_seconds()
```
sfc-gh-pjob, line 80: I'm not sure if this test is really able to verify that results are streamed. 
Checking duration between timestamps generated on the server side won't check that the results are really printed on client side in the right time.

src/snowcli/cli/common/sql_execution.py
```diff
@@ -23,13 +25,32 @@ def _conn(self):
...
+        return_cursors: bool = True,
+        cursor_class: SnowflakeCursor = SnowflakeCursor,
+        **kwargs,
+    ) -> Iterable[SnowflakeCursor]:
+        """
+        This is a custom implementation of SnowflakeConnection.execute_string that returns generator
+        instead of list. In case of executing multiple queries are executed one by one. This mean we can
+        access result of previous queries while evaluating next one. For example, we can print the results.
+        """
+        self._log.info("Executing %s", sql_text)
```
sfc-gh-pjob, line 41: Was it intentional to change log level (debug -> info)?

tests_integration/test_sql.py
```diff
@@ -49,3 +55,36 @@ def _round_values(results):
...
+            "-q",
+            "select 13; select system$wait(10);",
+        ],
+    )
+
+    assert len(log) == 2
+    (time_0, query_0), (time_1, query_1) = log
+
+    assert query_0 == "13"
+    assert time_1 - time_0 >= 10.0
```
sfc-gh-pjob, line 89:
```suggestion
    assert time_1 - time_0 >= 7.5
```

IMO setting the same exact value as in `wait` can cause flaky tests. If the first response from Snowflake will be slower than the second one, then this test may fail.

Example:
a. Time 0.0: Request sent to Snowflake
b. Time 0.1:  Request received in Snowflake
c. Time 0.2: Response prepared in Snowflake and transfer started to CLI
d. Time 0.2: `Wait(10)` started
e. Time 0.7: Response received in SnowCLI (time in `log`: 0.7), response transfer took 0.5
f. Time 10.2: `Wait(10)` completed and transfer of response to CLI started
g. Time 10.5: Response received in SnowCLI (time in `log`: 10.5), response transfer took 0.3

```
g - e = 10.5 - 0.7 = 9.8
9.8 < 10
assertion fails
```

src/snowcli/cli/common/sql_execution.py
sfc-gh-turbaszek: I don't have strong arguments for this change. I'm fine with going back to debug.

sfc-gh-pjob: Then IMO we should go back to `debug` as this change wasn't in plan for this task.

tests_integration/test_sql.py
sfc-gh-turbaszek: One for the advantages of proposed solution is that ignores latency. We just check time of printing. Due to using iterator executing of wait query will occur only after printing results of first query (calling next). It means that previous latency does not matter. In fact we have a guarantee that the time will be bigger than 10 seconds:
t=0 print results of query 1
t=0.1 execute wait query
t=10.2 print wait query

sfc-gh-pjob: Ok, I thought that `execute_stream` sends all queries together and only reads responses as a stream. I didn't look at its implementation. Now I see what you mean and you're right - it will work correctly.

src/snowcli/cli/common/sql_execution.py
sfc-gh-turbaszek: Fixed

sfc-gh-pjob (approved):

PR #571, opened 2023-11-28
Changed version to 2.0.0-alpha

sfc-gh-astus:
### Pre-review checklist

### Changes description
...

### DIFF
--- a/src/snowcli/__about__.py
+++ b/src/snowcli/__about__.py
@@ -1,3 +1,3 @@
 from __future__ import annotations
 
-VERSION = "2.0.0-dev"
+VERSION = "2.0.0-alpha.1"
--- a/tests_e2e/conftest.py
+++ b/tests_e2e/conftest.py
@@ -3,6 +3,7 @@ import subprocess
 import tempfile
 from pathlib import Path
 from tempfile import TemporaryDirectory
+from typing import List
 
 import pytest
 from snowcli import __about__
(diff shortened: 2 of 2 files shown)


src/snowcli/__about__.py
```diff
@@ -1,3 +1,3 @@
 from __future__ import annotations
 
-VERSION = "2.0.0-dev"
+VERSION = "2.0.0-alpha"
```
sfc-gh-pjob, line 3: Maybe we should think about `2.0.0-alpha.1`? WDYT?
How do we want version releases with fixes?

sfc-gh-pjob: @sfc-gh-turbaszek WDYT about this?

tests_e2e/conftest.py
```diff
@@ -60,14 +61,14 @@ def _build_snowcli(venv_path: Path, test_root_path: Path) -> None:
 def _install_snowcli_with_external_plugin(
     venv_path: Path, test_root_path: Path
 ) -> None:
-    version = __about__.VERSION.replace("-", ".")
+    whl_file = _find_whl_file(os.listdir(test_root_path / f"../dist/"))
```
sfc-gh-pjob, line 64: I can approve it as temporary solution but I think we should consider hardcoding an expected file name here.

sfc-gh-astus: I will search for some solution after release

sfc-gh-pjob: Ok, I forgot that we build snowcli in a venv, so I'm not sure we need any better solution :)

PR #586, opened 2023-12-04
Idempotent + flexible app teardown

sfc-gh-cgorrie:
### Pre-review checklist

### Changes description
`snow app teardown` will no longer error out if either the application instance or package does not exist. As such, it can now drop the package in cases where the app instance has already been dropped. It will also exit successfully when neither the instance nor package exist, reflecting that the desired end-state has been reached (albeit without doing anything).

### DIFF
--- a/src/snowcli/cli/nativeapp/commands.py
+++ b/src/snowcli/cli/nativeapp/commands.py
@@ -135,7 +135,8 @@ def app_teardown(
     **options,
 ) -> CommandResult:
     """
-    Drops an application and an application package as defined in the project definition file.
+    Attempts to drop both the application and package as defined in the project definition file.
+    This command will succeed even if one or both of these objects do not exist.
     As a note, this command does not accept role or warehouse overrides to your `config.toml` file,
     because your native app definition in `snowflake.yml/snowflake.local.yml` is used for any overrides.
     """
--- a/tests/nativeapp/test_manager.py
+++ b/tests/nativeapp/test_manager.py
@@ -247,7 +247,10 @@ def test_drop_object_no_special_comment(mock_execute, temp_dir, mock_cursor):
                     ],
                     [],
                 ),
-                mock.call("show application packages like 'sample_package_name'"),
+                mock.call(
+                    "show application packages like 'SAMPLE_PACKAGE_NAME'",
+                    cursor_class=DictCursor,
+                ),
             ),
             (mock_cursor(["row"], []), mock.call("use role old_role")),
         ]
@@ -273,7 +276,8 @@ def test_drop_object_no_special_comment(mock_execute, temp_dir, mock_cursor):
                 "show": "show application packages like",
             },
         )
-        assert mock_execute.mock_calls == expected
+
+    assert mock_execute.mock_calls == expected
 
 
 @mock.patch(NATIVEAPP_MANAGER_EXECUTE)
--- a/tests/nativeapp/test_manager.py
+++ b/tests/nativeapp/test_manager.py
@@ -199,7 +199,10 @@ def test_drop_object_no_show_object(mock_execute, temp_dir, mock_cursor):
             (mock_cursor(["row"], []), mock.call("use role sample_package_role")),
             (
                 mock_cursor([], []),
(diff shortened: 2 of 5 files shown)


tests/nativeapp/test_manager.py
```diff
@@ -199,7 +199,10 @@ def test_drop_object_no_show_object(mock_execute, temp_dir, mock_cursor):
             (mock_cursor(["row"], []), mock.call("use role sample_package_role")),
             (
                 mock_cursor([], []),
-                mock.call("show application packages like 'SAMPLE_PACKAGE_NAME'"),
+                mock.call(
+                    "show application packages like 'SAMPLE_PACKAGE_NAME'",
+                    cursor_class=DictCursor,
```
sfc-gh-cgorrie, line 204: The assert call below actually never got reached, which is why this had to change (even though the underlying code did not).

tests/nativeapp/test_manager.py
```diff
@@ -1025,76 +1026,3 @@ def test_get_snowsight_url(
         native_app_manager.get_snowsight_url()
         == "https://host/organization/account/#/apps/application/MYAPP"
     )
-
-
-@mock.patch(NATIVEAPP_MANAGER_EXECUTE)
-def test_quoting_app_teardown(mock_execute, temp_dir, mock_cursor):
```
sfc-gh-cgorrie, line 1031: Split out into separate test file.

tests/nativeapp/test_manager_teardown.py
```diff
@@ -0,0 +1,289 @@
...
+        contents=[mock_snowflake_yml_file],
+    )
+
+    native_app_manager = NativeAppManager()
+    native_app_manager.teardown()
+    assert mock_execute.mock_calls == expected
+
+
+@mock.patch(NATIVEAPP_MANAGER_EXECUTE)
+def test_quoting_app_teardown(mock_execute, temp_dir, mock_cursor):
```
sfc-gh-cgorrie, line 220: Pre-existing test moved from `test_manager.py`.

tests/nativeapp/test_manager.py
```diff
@@ -246,7 +247,10 @@ def test_drop_object_no_special_comment(mock_execute, temp_dir, mock_cursor):
                     ],
                     [],
                 ),
-                mock.call("show application packages like 'sample_package_name'"),
+                mock.call(
+                    "show application packages like 'SAMPLE_PACKAGE_NAME'",
+                    cursor_class=DictCursor,
+                ),
```
sfc-gh-cgorrie, line 253: The assert call below actually never got reached, which is why this had to change (even though the underlying code did not).

tests/nativeapp/test_manager_teardown.py
```diff
@@ -0,0 +1,334 @@
...
+                            "version": "UNVERSIONED",
+                            "owner": "APP_ROLE",
+                        }
+                    ],
+                    [],
+                ),
+                mock.call("show applications like 'MYAPP'", cursor_class=DictCursor),
+            ),
+            (None, mock.call("use role old_role")),
+            # teardown package: never happens
```
sfc-gh-melnacouzi, line 245: How about the usecase where drop happens but fails? Do we have an existing test for that?

src/snowcli/cli/nativeapp/commands.py
```diff
@@ -135,7 +135,8 @@ def app_teardown(
     **options,
 ) -> CommandResult:
     """
-    Drops an application and an application package as defined in the project definition file.
+    Attempts to drop both the application and package as defined in the project definition file.
```
sfc-gh-melnacouzi, line 138: nit: any reason we change application package to package terminology?

sfc-gh-cgorrie: No particular reason outside of the fact that there's nothing else package could refer to, so I often shorten it.

sfc-gh-melnacouzi (approved):

sfc-gh-bdufour (approved):

sfc-gh-astus (approved):

PR #591, opened 2023-12-07
Check logs in external plugins integration tests

sfc-gh-astus:
### Pre-review checklist

### Changes description
...

### DIFF
--- a/tests_integration/testing_utils/assertions/test_result_assertions.py
+++ b/tests_integration/testing_utils/assertions/test_result_assertions.py
@@ -5,8 +5,10 @@ from tests_integration.conftest import CommandResult
 from tests_integration.test_utils import contains_row_with
 
 
-def assert_that_result_is_successful(result: CommandResult) -> None:
+def assert_that_result_is_successful(result: CommandResult, caplog=None) -> None:
     assert result.exit_code == 0, result.output
+    if caplog:
+        _assert_that_no_error_logs(caplog)
 
 
 def assert_that_result_is_error(result: CommandResult, expected_exit_code: int) -> None:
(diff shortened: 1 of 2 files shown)


tests_integration/testing_utils/assertions/test_result_assertions.py
```diff
@@ -5,8 +5,10 @@
 from tests_integration.test_utils import contains_row_with
 
 
-def assert_that_result_is_successful(result: CommandResult) -> None:
+def assert_that_result_is_successful(result: CommandResult, caplog=None) -> None:
```
sfc-gh-astus, line 8: Should I add that way? WDYT?

tests_integration/testing_utils/assertions/test_result_assertions.py
```diff
@@ -5,8 +5,10 @@
...
     assert result.exit_code == 0, result.output
+    if caplog:
+        _assert_that_no_error_logs(caplog)
```
sfc-gh-pjob, line 11: IMO it should not be a part of `assert_that_result_is_successful`. Presence of error logs does not mean that result is automatically a failure.

For example running `snow sql -q "select 1"` using CLI with broken external plugin will still finish successfully.

IMO `assert_that_no_error_logs` should be a separate assertion.

tests_integration/testing_utils/assertions/test_result_assertions.py
```diff
@@ -40,3 +42,11 @@ def assert_that_result_is_successful_and_done_is_on_output(
...
     assert result.output is not None
     assert json.loads(result.output) == {"message": "Done"}
+
+
+def _assert_that_no_error_logs(caplog):
+    error_logs = [
+        record.message for record in caplog.records if record.levelname == "ERROR"
+    ]
+    assert error_logs == []
+    caplog.clear()
```
sfc-gh-pjob, line 52: IMO we should change test's state as part of the assertion.

sfc-gh-astus: Removed

tests_integration/testing_utils/assertions/test_result_assertions.py
sfc-gh-astus: Right, removed

sfc-gh-pjob (approved):

PR #600, opened 2023-12-11
[NADE] Production Ready App Pkg Changes + Refactor

sfc-gh-bgoel:
### Pre-review checklist
   * [x] I've confirmed that my changes are up-to-date with the target branch - as of December 19, 11:30AM EST

### Changes description
There are a bunch of changes introduced with this PR. All of the changes are inter-dependent and hence they need to be part of the same PR. 

1. The giant `NativeAppManager` class has been split into itself, `NativeAppRunProcessor` and `NativeAppTeardownProcessor`. Their methods have further been refactored to be independent entities so they can be unit tested easily and mocked.
2. All the tests have been rewritten to fit the above change, in addition to the new tests introduced as a result of the behavior change described below. In order to test a method in isolation, there are numerous mocks in place. It also helps avoid rewriting the entire sql execution order and just focus on the one test. 
3. `snow app run` and `snow app teardown` have been modified to allow custom behavior for setting internal/external distribution as well as interacting with objects not created by the CLI.


src/snowcli/cli/nativeapp/run_processor.py
```diff
@@ -0,0 +1,241 @@
...
+    def __init__(self, search_path: Optional[str] = None):
+        super().__init__(search_path)
+
+    def create_app_package(self) -> None:
+        """
+        Creates the application package with our up-to-date stage if none exists.
+        """
+        with self.use_role(self.package_role):
+            show_cursor = self._execute_query(
+                f"show application packages like '{unquote_identifier(self.package_name)}'",
```
sfc-gh-bgoel, line 44: Will be abstracting out all the show objects (pkg and app) as well.

src/snowcli/cli/nativeapp/teardown_processor.py
```diff
@@ -0,0 +1,233 @@
...
+                self._execute_query(drop_query)
+            except:
+                raise SnowflakeSQLExecutionError(drop_query)
+
+            log.info(f"Dropped application package {self.package_name} successfully.")
+            return True
+
+    def process(self, auto_yes: bool = False, *args, **kwargs):
+
+        self.needs_confirm = True
```
sfc-gh-cgorrie, line 227: Because both drop procedures have independent `needs_confirm`, I think it's safest to push this down from being an instance property and just have it as a local variable in each method

src/snowcli/cli/nativeapp/teardown_processor.py
```diff
@@ -0,0 +1,233 @@
...
+                else:
+                    if self.needs_confirm and not auto_yes:
+                        log.warn(
+                            f"Application package {self.package_name} was not created by SnowCLI!"
+                        )
+            else:
+                if self.needs_confirm and not auto_yes:
+                    log.warn(
+                        f"Application package {self.package_name} in your Snowflake account has distribution property '{EXTERNAL_DISTRIBUTION}'"
+                    )
```
sfc-gh-cgorrie, line 189: I don't think this else clause is necessary. The log below will print out the package distribution anyway.

sfc-gh-cgorrie (changes requested):
Great refactor, and probably the only usage of OOP / inheritance that I've liked in a while. I added some review comments; I think overall more things could be moved to utils (maybe exception classes?) and on reflection I'm wondering if `--force` is better than `--yes` but we can always make the change later.

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -416,80 +268,37 @@ def app_exists(self) -> bool:
...
-        object_name: str,
-        object_role: str,
-        object_type: Literal["application", "package"],
-        query_dict: dict,
-    ) -> bool:
+    def does_app_pkg_exist(self):
         """
-        N.B. query_dict['show'] must be a like % clause
-        Returns True if the object was dropped, False if the object didn't exist.
+        TEMPORARY workaround - will be removed in the followup PR as part of a refactor.
```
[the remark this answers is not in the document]
sfc-gh-bgoel: This is cleaned up with my upcoming commit on abstracting out some sub-methods. This is an old carry over from my previous PR.

src/snowcli/cli/nativeapp/teardown_processor.py
```diff
@@ -0,0 +1,233 @@
...
+                )
+
+            # 5. If distribution is internal, check for special comment
+            row_comment = show_obj_row[COMMENT_COL]
+            if actual_distribution == INTERNAL_DISTRIBUTION:
+                if row_comment == SPECIAL_COMMENT:
+                    self.needs_confirm = False
+                else:
+                    if self.needs_confirm and not auto_yes:
+                        log.warn(
```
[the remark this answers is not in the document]
sfc-gh-bgoel: Whoops, yes thanks for the call out. `warn` is deprecated in favor of `warning`.

src/snowcli/cli/nativeapp/teardown_processor.py
sfc-gh-bgoel: True, but the log below does not tell what the problem/mismatch is despite all the info. It could be anything - owner, created_on etc. Keeping this log gives the user more information in a clear manner.

tests/nativeapp/test_run_processor.py
```diff
@@ -0,0 +1,850 @@
...
+    mock_execute.side_effect = side_effects
+
+    current_working_directory = os.getcwd()
+    create_named_file(
+        file_name="snowflake.yml",
+        dir=current_working_directory,
+        contents=[mock_snowflake_yml_file],
+    )
+
+    native_app_manager = _get_na_manager()
```
sfc-gh-bdufour, line 83: ditto

tests/nativeapp/utils.py
```diff
@@ -0,0 +1,88 @@
...
+
+mock_connection = mock.patch(
+    "snowcli.cli.common.cli_global_context._CliGlobalContextAccess.connection",
+    new_callable=PropertyMock,
+)
+
+mock_get_app_pkg_distribution_in_sf = mock.patch(
+    NATIVEAPP_MANAGER_APP_PKG_DISTRIBUTION_IN_SF,
+    new_callable=PropertyMock,
+)
```
sfc-gh-bdufour, line 83: Shouldn't this be on-demand? Any time this file is included, the mocks will be created. Individual tests may want to configure how this happens.

tests/test_data/projects/integration_external/app/README.md
```diff
@@ -0,0 +1,4 @@
+# README
+
+This directory contains an extremely simple application that is used for
```
sfc-gh-bdufour, line 3: nit: extremely simple is subjective, and might change in the future. How about something like "a native application for integ testing purposes", or something equally vague w.r.t scope?

sfc-gh-cgorrie (commented):
The main issue for me right now is `needs_confirm`. Personally I think it's just an implementation detail that should be left inside of the methods. But if you feel strongly it should live at class level, let's be consistent and move the force flag there too.

sfc-gh-bdufour:
> The main issue for me right now is `needs_confirm`. Personally I think it's just an implementation detail that should be left inside of the methods. But if you feel strongly it should live at class level, let's be consistent and move the force flag there too.

Yeah, I agree the current version is a bit awkward. Cam's suggestions make sense to me. Another option would be to pass in some kind of policy object (could be a lambda) that can be queried whenever we need to decide whether to ask the customer a question. Assuming we'll have to do this from a bunch of places down the line, it might be worth considering.

tests/nativeapp/utils.py
sfc-gh-bgoel: The test_*.py files where this file is being imported use `mock_get_app_pkg_distribution_in_sf` in most of the test cases. Instead of specifying this same thing again for every test, it is defined here. Individual test are able to customize the behavior of this mock by setting the `return_value` of the mock.

src/snowcli/cli/nativeapp/teardown_processor.py
sfc-gh-bgoel: Addressed in the next commit.

sfc-gh-bgoel:
I have addressed previous comments, and resolved the minor conversations.
Also addressed the `self.needs_confirm` and its use in `self.needs_confirmation`. 
@sfc-gh-bdufour I do not completely understand what you mean by a policy object which determines whether we ask the customer a question. Do you mean querying if the command is being run in an interactive mode, and therefore only then it makes sense to add prompts, whereas if it is part of a script then we need to figure out a different behavior?

tests/nativeapp/utils.py
sfc-gh-bdufour: Absolutely fine with creating helpers to test fixtures. However, the very fact that something like 'utils.py' is imported shouldn't, IMHO, create mocks as a side effect. It tends to create weird maintenance problems later on. Pytest has lazy fixture capabilities for things like this. If nothing else, please move this to something like fixture_helpers.py so it's contained.

sfc-gh-bdufour:
> @sfc-gh-bdufour I do not completely understand what you mean by a policy object which determines whether we ask the customer a question.

I meant instead of auto_yes = yes|no, we can pass it something that makes the decision as needed (a policy object), or even asks the question as needed. Think something like:

```
ASK_ALWAYS = lambda ctx: ask_for_confirmation_yes_no(ctx.prompt)
ASK_NEVER = lambda ctx: return False
def ask_yes_no_always(ctx):
    if ctx.data is None:
        answer = ask_for_confirmation_yes_no_always(ctx.prompt)
        if answer == 'always':
            ctx.data = True # remember that the user answered 'always' and short circuit subsequent confirmations
    return ctx.data

perform_task(policy=ASK_ALWAYS)
```
I was just trying to demonstrate what's possible with this approach, the full expressive power of policy objects likely doesn't apply in this specific case.

tests/nativeapp/utils.py
sfc-gh-bgoel: That makes sense to me - the latest update has the changes. Ty!

sfc-gh-bgoel:
Synced offline with the policy object suggestion by @sfc-gh-bdufour . This PR can proceed without the modification, but there is scope to implement this during the version drop/create and will be added to that separate PR.

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -141,10 +69,41 @@ def _generic_sql_error_handler(
...
+def ensure_correct_owner(row: dict, role: str, obj_name: str) -> None:
+    """
+    Check if an object has the right owner role
+    """
+    actual_owner = row[
+        OWNER_COL
+    ].upper()  # Because unquote_identifier() always returns uppercase str
+    if actual_owner != unquote_identifier(role):
+        raise UnexpectedOwnerError(obj_name, role, actual_owner)
+    return
```
sfc-gh-cgorrie, line 93: [nit] don't need explicit return

src/snowcli/cli/nativeapp/teardown_processor.py
```diff
@@ -0,0 +1,190 @@
...
+    NativeAppManager,
+    ensure_correct_owner,
+)
+from snowcli.exception import SnowflakeSQLExecutionError
+from snowflake.connector.cursor import DictCursor
+
+
+class NativeAppTeardownProcessor(NativeAppManager, NativeAppCommandProcessor):
+
+    needs_confirm: bool = True
```
sfc-gh-cgorrie, line 28: Should remove from class scope as `needs_confirmation` gets it passed in as a variable; this instance variable is hidden / aliased

src/snowcli/cli/nativeapp/teardown_processor.py
```diff
@@ -0,0 +1,190 @@
...
+            try:
+                self._execute_query(drop_query)
+            except:
+                raise SnowflakeSQLExecutionError(drop_query)
+
+            print(f"Dropped {object_type} {object_name} successfully.")
+
+    def needs_confirmation(self, needs_confirm: bool, auto_yes: bool) -> bool:
+        return needs_confirm and not auto_yes
+
```
sfc-gh-cgorrie, line 49: As it does not use `self`, let's extract it out into a utility function

sfc-gh-cgorrie (approved):
Great exhaustive test cases and flow. A few non-blocking suggestions about how to make the code slightly cleaner.

sfc-gh-cgorrie (dismissed):

tests/project/test_config.py
```diff
@@ -30,33 +28,6 @@ def test_na_minimal_project(project_definition_files: List[Path]):
     assert project["native_app"]["name"] == "minimal"
     assert project["native_app"]["artifacts"] == ["setup.sql", "README.md"]
 
-    from os import getenv as original_getenv
```
sfc-gh-bgoel, line 33: @sfc-gh-pjob  This section is removed because `generate_local_override_yml` method is not invoked anywhere in the code. Therefore it can be removed from tests.

sfc-gh-cgorrie (approved):

src/snowcli/cli/nativeapp/run_processor.py
```diff
@@ -0,0 +1,214 @@
...
+                    create application {self.app_name}
+                        from application package {self.package_name}
+                        using {stage_name}
+                        debug_mode = {self.debug_mode}
+                        comment = {SPECIAL_COMMENT}
+                    """
+                    )
+                )
+            except ProgrammingError as err:
+                _generic_sql_error_handler(err)
```
sfc-gh-mraba, line 198: This method is pretty long and consists of three distinct steps:
- validate warehouse existence
- application update
- application creation 

Extracting each step into separate private methods could improve readability.
Something like this:
- create dedicated methods for new app and updating existing.
- invoke methods based on result of `show_app_row = self.get_existing_app_info()`
- 
```python
def _create_new_app(): 
    ...
def _update_existing_app(): 
    ...

def _create_dev_app(self, ...):
    # warehouse checks
    ....
    show_app_row = self.get_existing_app_info()
    if show_app_row:
         self._update_existing_app(show_app_row, ...)
    else:
         self._create_new_app(...)
```

sfc-gh-bgoel: I agree that this method can be split out into separate private methods and is on my docket. However for scope of this PR, this method has just been moved from a different file to this file. As this PR blocks two other PRs from going in, I'd like to merge this in, if it's okay with you, and address this method in a follow up PR.

src/snowcli/cli/nativeapp/run_processor.py
```diff
@@ -0,0 +1,214 @@
...
+        if show_obj_row:
+            # 1. Check for the right owner role
+            ensure_correct_owner(
+                row=show_obj_row, role=self.package_role, obj_name=self.package_name
+            )
+
+            # 2. Check distribution of the existing app package
+            actual_distribution = self.get_app_pkg_distribution_in_snowflake
+            if not self.is_app_pkg_distribution_same_in_sf():
+                print(
```
sfc-gh-pjob, line 52: It will break output when using `--format=JSON`. But I understand requirement of progress messages and as our task for such messages (SNOW-986249) is still in "to do", I think it's ok for now to use just a `print` here.

However we should revisit these prints after implementation of SNOW-986249 to make JSON output working correctly for NADE commands.

sfc-gh-pjob (approved):

src/snowcli/cli/nativeapp/run_processor.py
sfc-gh-bgoel: Agreed, this a temporary solution that we plan to revisit once the above mentioned ticket is picked up.

sfc-gh-bdufour (approved):

PR #609, opened 2023-12-13
Adopt project definition decorator in native apps commands

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
Adopt `@with_project_definition` decorator in native apps commands.

### DIFF
--- a/src/snowcli/cli/nativeapp/commands.py
+++ b/src/snowcli/cli/nativeapp/commands.py
@@ -1,10 +1,11 @@
 import logging
-from typing import Optional
 
 import typer
+from snowcli.cli.common.cli_global_context import cli_context
 from snowcli.cli.common.decorators import (
     global_options,
     global_options_with_connection,
+    with_project_definition,
 )
 from snowcli.cli.common.flags import DEFAULT_CONTEXT_SETTINGS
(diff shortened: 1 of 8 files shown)


sfc-gh-bgoel (approved):
lgtm, thank you!

src/snowcli/cli/nativeapp/commands.py
```diff
@@ -71,24 +64,27 @@ def app_init(
 
 @app.command("bundle", hidden=True)
 @with_output
+@with_project_definition("native_app")
```
sfc-gh-cgorrie, line 67: Implicit in this is that no command will need to grab multiple top-level objects of the project definition. That makes sense to me for now; we can revisit if we find a compelling use case.

sfc-gh-cgorrie (approved):
LGTM. Changes make sense to me

src/snowcli/cli/nativeapp/commands.py
sfc-gh-turbaszek: Correct, in future we can make this argument optional so the command will get whole definition

sfc-gh-jsikorski (approved):

PR #618, opened 2023-12-20
[NADE] Snow app version create + version drop

sfc-gh-bgoel:
### Pre-review checklist
   * [x] I've confirmed that my changes are up-to-date with the target branch - as of Jan 4, 2:00PM EST

### Changes description
This PR introduces two NA CLI commands:
1. `snow app version create`
2. `snow app version drop`

### DIFF
--- a/src/snowcli/cli/nativeapp/artifacts.py
+++ b/src/snowcli/cli/nativeapp/artifacts.py
@@ -5,7 +5,7 @@ from pathlib import Path
 from typing import Any, List, Optional, Tuple, Union
 
 import strictyaml
-from click.exceptions import ClickException
+from click import ClickException
 
 
 class DeployRootError(ClickException):
@@ -46,20 +46,6 @@ class SourceNotFoundError(ClickException):
     def __init__(self, path: Path):
         super().__init__(f"{self.__doc__}\npath = {path}")
         self.path = path
-
-
-class ManifestYmlFileNotFoundError(ClickException):
-    """
-    Manifest.yml file not found in the deploy root of the native apps project.
-    """
-
-    def __init__(self, deploy_root: Path):
-        super().__init__(
-            f"""
-            {self.__doc__}
-            deploy_root = {deploy_root}
-            """
-        )
 
 
 class TooManyFilesError(ClickException):
@@ -261,7 +247,9 @@ def find_manifest_file(deploy_root: Path) -> Path:
             if file.lower() == "manifest.yml":
                 return Path(os.path.join(root, file))
 
-    raise ManifestYmlFileNotFoundError(deploy_root=resolved_root)
+    raise ClickException(
+        "Required manifest.yml file not found in the deploy root of the native apps project."
+    )
 
 
 def find_version_info_in_manifest_file(
--- a/src/snowcli/cli/nativeapp/artifacts.py
+++ b/src/snowcli/cli/nativeapp/artifacts.py
@@ -2,8 +2,9 @@ import os
 import shutil
 from dataclasses import dataclass
 from pathlib import Path
-from typing import List, Union
+from typing import Any, List, Optional, Tuple, Union
 
+import strictyaml
 from click.exceptions import ClickException
 
 
@@ -45,6 +46,20 @@ class SourceNotFoundError(ClickException):
     def __init__(self, path: Path):
         super().__init__(f"{self.__doc__}\npath = {path}")
         self.path = path
+
+
+class ManifestYmlFileNotFoundError(ClickException):
+    """
+    Manifest.yml file not found in the deploy root of the native apps project.
+    """
+
+    def __init__(self, deploy_root: Path):
+        super().__init__(
+            f"""
+            {self.__doc__}
+            deploy_root = {deploy_root}
+            """
+        )
 
 
 class TooManyFilesError(ClickException):
@@ -234,3 +249,42 @@ def build_bundle(
             else:
                 # refuse to map multiple source files to one destination (undefined behaviour)
                 raise TooManyFilesError(dest_path)
+
+
+def find_manifest_file(deploy_root: Path) -> Path:
+    """
+    Find manifest.yml file, if available, in the deploy_root of the native apps project.
+    """
+    resolved_root = deploy_root.resolve()
+    for root, _, files in os.walk(resolved_root):
+        for file in files:
+            if file.lower() == "manifest.yml":
+                return Path(os.path.join(root, file))
+
+    raise ManifestYmlFileNotFoundError(deploy_root=resolved_root)
+
+
+def find_version_info_in_manifest_file(
+    deploy_root: Path,
+) -> Tuple[Optional[str], Optional[str]]:
+    """
+    Find version and patch, if available, in the manifest.yml file.
+    """
+    version_field = "version"
+    name_field = "name"
+    patch_field = "patch"
+
+    manifest_file = find_manifest_file(deploy_root)
+    with open(manifest_file, "r") as file:
+        manifest_content = strictyaml.load(file.read())
+
+    version_name: Optional[str] = None
+    patch_name: Optional[str] = None
+
+    if version_field in manifest_content:
+        version_info = manifest_content[version_field]
+        version_name = version_info[name_field]
+        if patch_field in version_info:
+            patch_name = version_info[patch_field]
+
+    return version_name, patch_name
--- a/src/snowcli/cli/nativeapp/policy.py
+++ b/src/snowcli/cli/nativeapp/policy.py
@@ -1,31 +1,25 @@
-from dataclasses import dataclass
-
 import typer
-
-
-@dataclass
-class TaskContext:
-    user_prompt: str
-    is_interactive_mode: bool
 
 
 class PolicyBase:
     exit_code: int = 0
 
-    def should_proceed(self, ctx: TaskContext):
+    def should_proceed(self, user_prompt: str):
         pass
 
 
 class AllowAlwaysPolicy(PolicyBase):
-    def should_proceed(self, ctx: TaskContext):
+    def should_proceed(self, user_prompt: str):
         return True
 
 
+class DenyAlwaysPolicy(PolicyBase):
+    def should_proceed(self, user_prompt: str):
+        self.exit_code = 1
+        return False
+
+
 class AskAlwaysPolicy(PolicyBase):
-    def should_proceed(self, ctx: TaskContext):
-        if ctx.is_interactive_mode:
-            should_continue = typer.confirm(ctx.user_prompt)
-            return should_continue
-        else:
-            self.exit_code = 1
-            return False
+    def should_proceed(self, user_prompt: str):
+        should_continue = typer.confirm(user_prompt)
+        return should_continue
--- a/src/snowcli/cli/nativeapp/policy.py
+++ b/src/snowcli/cli/nativeapp/policy.py
@@ -21,7 +21,7 @@ class AllowAlwaysPolicy(PolicyBase):
         return True
 
 
-class AlwaysAskPolicy(PolicyBase):
+class AskAlwaysPolicy(PolicyBase):
     def should_proceed(self, ctx: TaskContext):
         if ctx.is_interactive_mode:
             should_continue = typer.confirm(ctx.user_prompt)
--- a/src/snowcli/cli/nativeapp/policy.py
+++ b/src/snowcli/cli/nativeapp/policy.py
@@ -1,0 +1,31 @@
+from dataclasses import dataclass
+
+import typer
+
+
+@dataclass
+class TaskContext:
+    user_prompt: str
+    is_interactive_mode: bool
+
+
+class PolicyBase:
+    exit_code: int = 0
+
+    def should_proceed(self, ctx: TaskContext):
+        pass
+
+
+class AllowAlwaysPolicy(PolicyBase):
+    def should_proceed(self, ctx: TaskContext):
+        return True
+
+
+class AlwaysAskPolicy(PolicyBase):
+    def should_proceed(self, ctx: TaskContext):
+        if ctx.is_interactive_mode:
+            should_continue = typer.confirm(ctx.user_prompt)
+            return should_continue
+        else:
+            self.exit_code = 1
+            return False
--- a/src/snowcli/cli/nativeapp/policy.py
+++ b/src/snowcli/cli/nativeapp/policy.py
@@ -1,25 +1,34 @@
+from abc import ABC, abstractmethod
+from typing import Optional
+
 import typer
 
 
-class PolicyBase:
-    exit_code: int = 0
+class PolicyBase(ABC):
+    """Abtract Class for various policies that govern if a Snowflake CLI command can continue execution when it asks for a decision."""
 
-    def should_proceed(self, user_prompt: str):
+    @abstractmethod
+    def should_proceed(self, user_prompt: Optional[str]) -> bool:
         pass
 
 
 class AllowAlwaysPolicy(PolicyBase):
-    def should_proceed(self, user_prompt: str):
+    """Always allow a Snowflake CLI command to continue execution."""
+
+    def should_proceed(self, user_prompt: Optional[str] = None):
         return True
 
 
 class DenyAlwaysPolicy(PolicyBase):
-    def should_proceed(self, user_prompt: str):
-        self.exit_code = 1
+    """Never allow a Snowflake CLI command to continue execution."""
+
+    def should_proceed(self, user_prompt: Optional[str] = None):
         return False
 
 
 class AskAlwaysPolicy(PolicyBase):
-    def should_proceed(self, user_prompt: str):
+    """Ask the user whether to continue execution of a Snowflake CLI command."""
+
+    def should_proceed(self, user_prompt: Optional[str]):
         should_continue = typer.confirm(user_prompt)
         return should_continue
--- a/src/snowcli/cli/nativeapp/utils.py
+++ b/src/snowcli/cli/nativeapp/utils.py
@@ -1,4 +1,3 @@
-from functools import cache
 from sys import stdin, stdout
 from typing import Callable, List, Optional
 
@@ -24,6 +23,5 @@ def needs_confirmation(needs_confirm: bool, auto_yes: bool) -> bool:
     return needs_confirm and not auto_yes
 
 
-@cache
-def is_user_in_interactive_mode():
+def is_interactive_mode():
     return stdin.isatty() and stdout.isatty()
--- a/src/snowcli/cli/nativeapp/utils.py
+++ b/src/snowcli/cli/nativeapp/utils.py
@@ -1,8 +1,7 @@
-from dataclasses import dataclass
+from functools import cache
 from sys import stdin, stdout
 from typing import Callable, List, Optional
 
-import typer
 from snowflake.connector.cursor import DictCursor
 
 
@@ -25,34 +24,6 @@ def needs_confirmation(needs_confirm: bool, auto_yes: bool) -> bool:
     return needs_confirm and not auto_yes
 
 
-class UserConfirmationPolicy:
-    def __init__(self, is_force, is_interactive_mode, confirm_with_user):
-        self.is_force = is_force
-        self.is_interactive_mode = is_interactive_mode and not self.is_force
-        self.confirm_with_user = confirm_with_user
-
-    def requires_user_confirmation(self):
-        return not self.is_force
-
-
-@dataclass
-class Prompts:
-    give_info_and_ask: str
-    do_not_continue_message: str
-    error_message: str
-
-
-def ask_user_confirmation(
-    user_confirmation_policy: UserConfirmationPolicy, prompts: Prompts
-):
-    if user_confirmation_policy.requires_user_confirmation():
-        if user_confirmation_policy.is_interactive_mode:
-            should_continue = user_confirmation_policy.confirm_with_user(
-                prompts.give_info_and_ask
-            )
-            if not should_continue:
-                print(prompts.do_not_continue_message)
-                raise typer.Exit()
-        else:
-            print(prompts.error_message)
-            raise typer.Exit(1)
+@cache
+def is_user_in_interactive_mode():
+    return stdin.isatty() and stdout.isatty()
--- a/src/snowcli/cli/nativeapp/utils.py
+++ b/src/snowcli/cli/nativeapp/utils.py
@@ -1,15 +1,58 @@
-from typing import Callable, Optional
+from dataclasses import dataclass
+from sys import stdin, stdout
+from typing import Callable, List, Optional
 
+import typer
 from snowflake.connector.cursor import DictCursor
 
 
-def find_row(cursor: DictCursor, predicate: Callable[[dict], bool]) -> Optional[dict]:
+def _rows_generator(cursor: DictCursor, predicate: Callable[[dict], bool]):
+    return (row for row in cursor.fetchall() if predicate(row))
+
+
+def find_all_rows(cursor: DictCursor, predicate: Callable[[dict], bool]) -> List[dict]:
+    return list(_rows_generator(cursor, predicate))
+
+
+def find_first_row(
+    cursor: DictCursor, predicate: Callable[[dict], bool]
+) -> Optional[dict]:
     """Returns the first row that matches the predicate, or None."""
-    return next(
-        (row for row in cursor.fetchall() if predicate(row)),
-        None,
-    )
+    return next(_rows_generator(cursor, predicate), None)
 
 
 def needs_confirmation(needs_confirm: bool, auto_yes: bool) -> bool:
     return needs_confirm and not auto_yes
+
+
+class UserConfirmationPolicy:
+    def __init__(self, is_force, is_interactive_mode, confirm_with_user):
+        self.is_force = is_force
+        self.is_interactive_mode = is_interactive_mode and not self.is_force
+        self.confirm_with_user = confirm_with_user
+
+    def requires_user_confirmation(self):
+        return not self.is_force
+
+
+@dataclass
+class Prompts:
+    give_info_and_ask: str
+    do_not_continue_message: str
+    error_message: str
+
+
+def ask_user_confirmation(
+    user_confirmation_policy: UserConfirmationPolicy, prompts: Prompts
+):
+    if user_confirmation_policy.requires_user_confirmation():
+        if user_confirmation_policy.is_interactive_mode:
+            should_continue = user_confirmation_policy.confirm_with_user(
+                prompts.give_info_and_ask
+            )
+            if not should_continue:
+                print(prompts.do_not_continue_message)
+                raise typer.Exit()
+        else:
+            print(prompts.error_message)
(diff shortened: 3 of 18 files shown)


tests/nativeapp/patch_utils.py
```diff
@@ -0,0 +1,18 @@
+from unittest import mock
```
sfc-gh-bgoel, line 1: Can ignore all test changes for now, those should be resolved once the upstream branch's PR has been merged in.
This comment will be resolved once that happens, and new tests are available for review.

src/snowcli/cli/nativeapp/version/version_processor.py
```diff
@@ -0,0 +1,311 @@
...
+            print("Changes detected in your git repository!")
+            # show differences between current files and last commit
+            print(repo.git.diff(repo.head.commit.tree))
+
+            ask_user_confirmation(
+                user_confirmation_policy=user_confirmation_policy,
+                prompts=Prompts(
+                    f"You have local changes in this repository that are not part of a previous commit. Do you still want to continue? Consider using 'git status' and/or 'git diff' before proceeding.",
+                    f"Did not create a new version {version} of application package {package_name}.",
+                    "Cannot create a new version without user input. Consider using --force if you would like to attempt forcing the creation of version.",
```
sfc-gh-cgorrie, line 61: Same suggestions as the other `Prompts` (love the dataclass btw)

src/snowcli/cli/nativeapp/version/version_processor.py
```diff
@@ -0,0 +1,311 @@
...
+    def __init__(self, project_definition: Dict, project_root: Path):
+        super().__init__(project_definition, project_root)
+
+    def get_existing_version_info(self, version: str) -> Optional[dict]:
+        """
+        Check for an existing version by the same name for an application package.
+        It executes a 'show versions like ... in application package' query and returns the result as single row, if one exists.
+        """
+        with self.use_role(self.package_role):
+            show_obj_query = f"show versions like '{unquote_identifier(version)} in application package {self.package_name}"
```
sfc-gh-cgorrie, line 107: Missing end quote. Also I cannot remember off-the-top of my head: do versions follow the same quoting rules as everything else?

src/snowcli/cli/nativeapp/version/version_processor.py
```diff
@@ -0,0 +1,311 @@
...
+
+        # We need build_bundle() to (optionally) find version in manifest.yml and create app package
+        self.build_bundle()
+
+        # Make sure version is not None before proceeding any further.
+        # This will raise an exception if version information is not found. Patch can be None.
+        if version is None:
+            version, patch = find_version_info_in_manifest_file(self.deploy_root)
+
+        # Git repository checks
```
sfc-gh-cgorrie, line 210: [nit] unnecessary comment

src/snowcli/cli/nativeapp/version/version_processor.py
```diff
@@ -0,0 +1,311 @@
...
+        check_index_changes_in_git_repo(
+            project_root=self.project_root,
+            version=version,
+            package_name=self.package_name,
+            user_confirmation_policy=user_confirmation_policy,
+        )
+
+        # Create application package
```
sfc-gh-cgorrie, line 218: [nit] unnecessary comment

src/snowcli/cli/nativeapp/version/version_processor.py
```diff
@@ -0,0 +1,311 @@
...
+                  """
+                )
+            )
+            self.build_bundle()
+            version, _ = find_version_info_in_manifest_file(self.deploy_root)
+
+        print(
+            dedent(
+                f"""\
+                    About to drop version {version} of application package {self.package_name}. Once dropped, this operation cannot be undone.
```
sfc-gh-cgorrie, line 285: Suggest you move the "Once dropped, ..." bit to ask user confirmation below. It's only relevant to the user if they can interactively stop it.

src/snowcli/cli/nativeapp/artifacts.py
```diff
@@ -47,6 +52,29 @@ def __init__(self, path: Path):
         self.path = path
 
 
+class ManifestYmlFileNotFoundError(ClickException):
```
sfc-gh-bdufour, line 55: Said this before, but IMHO we have way too many of these exceptions. They're oddly specific. We shouldn't need to add 2-3 exception types for each change.

src/snowcli/cli/nativeapp/artifacts.py
```diff
@@ -2,9 +2,14 @@
...
 from dataclasses import dataclass
 from pathlib import Path
-from typing import List, Union
+from typing import Any, List, Optional, Union
 
 from click.exceptions import ClickException
+from strictyaml import (
+    YAML,
+    as_document,
+    load,
```
sfc-gh-bdufour, line 11: nit: I generally prefer to stick to qualified access for names like this that are highly ambiguous, for readability. I know we do this in a lot of places, but I'd be in favour of phasing those out.

src/snowcli/cli/nativeapp/artifacts.py
```diff
@@ -234,3 +262,51 @@ def build_bundle(
...
+
+
+def find_manifest_file(deploy_root: Path) -> Path:
+    """
+    Find manifest.yml file, if available, in the deploy_root of the native apps project.
+    """
+    resolved_root = deploy_root.resolve()
+    for root, dirs, files in os.walk(resolved_root):
+        for file in files:
+            if file == "manifest.yml":
```
sfc-gh-bdufour, line 265: what happens on case-insensitive file systems?

src/snowcli/cli/nativeapp/artifacts.py
```diff
@@ -234,3 +262,51 @@ def build_bundle(
...
+                return Path(os.path.join(root, file))
+
+        for subdir in dirs:
+            # TODO: can the manifest.yml ever be in a subdir under deploy_root? It belongs at the root of the stage, so maybe not required.
```
sfc-gh-bdufour, line 269: nope, not allowed: https://docs.snowflake.com/en/developer-guide/native-apps/creating-manifest#about-the-manifest-file

src/snowcli/cli/nativeapp/artifacts.py
```diff
@@ -234,3 +262,51 @@ def build_bundle(
...
+            subdir_path = Path(os.path.join(root, subdir))
+            manifest_path = find_manifest_file(subdir_path)
+            if manifest_path:
+                return manifest_path
+
+    raise ManifestYmlFileNotFoundError(deploy_root=resolved_root)
+
+
+def find_version_info_in_manifest_file(deploy_root: Path):
```
sfc-gh-bdufour, line 278: declare the return type here

sfc-gh-cgorrie (commented):
Some general notes. I think overall improving the signal-to-noise on what is printed as well as being more consistent in naming (e.g. `check` vs `warn`) would be a good idea apart from these specifics as well.

src/snowcli/cli/nativeapp/utils.py
```diff
@@ -1,15 +1,58 @@
...
 
 def needs_confirmation(needs_confirm: bool, auto_yes: bool) -> bool:
     return needs_confirm and not auto_yes
+
+
+class UserConfirmationPolicy:
+    def __init__(self, is_force, is_interactive_mode, confirm_with_user):
+        self.is_force = is_force
+        self.is_interactive_mode = is_interactive_mode and not self.is_force
+        self.confirm_with_user = confirm_with_user
```
sfc-gh-bdufour, line 32: that's odd to me. Generally you'd have the policy be some kind of pure interface, and specialize it for some specific cases (like ask always/never ask/etc). This feels like we're trying to merge everything into a single policy class.

src/snowcli/cli/nativeapp/utils.py
```diff
@@ -1,15 +1,58 @@
...
+
+    def requires_user_confirmation(self):
+        return not self.is_force
+
+
+@dataclass
+class Prompts:
```
sfc-gh-bdufour, line 39: Not sure what this is and why it's there. Is it supposed to hold all possible prompts to ask input from the user? If so, why? Prompts tend to be better defined near their use so they can include the relevant context info.

src/snowcli/cli/nativeapp/utils.py
```diff
@@ -1,15 +1,58 @@
...
+    give_info_and_ask: str
+    do_not_continue_message: str
+    error_message: str
+
+
+def ask_user_confirmation(
+    user_confirmation_policy: UserConfirmationPolicy, prompts: Prompts
+):
+    if user_confirmation_policy.requires_user_confirmation():
```
sfc-gh-bdufour, line 48: Might be useful to see it as a policy that controls the execution of a task, not a policy that controls asking the user. Asking is a mechanism through which we allow a task to execute. The ultimate goal is task execution. If you see it this way, it becomes possible to have a policy that allows certain tasks to execute while confirming others (e.g. ok to overwrite files, but not to drop a database)

src/snowcli/cli/nativeapp/utils.py
```diff
@@ -1,15 +1,58 @@
...
+        if user_confirmation_policy.is_interactive_mode:
+            should_continue = user_confirmation_policy.confirm_with_user(
+                prompts.give_info_and_ask
+            )
+            if not should_continue:
+                print(prompts.do_not_continue_message)
+                raise typer.Exit()
```
sfc-gh-bdufour, line 55: I don't think "ask for user confirmation" should exit. Often an invoking task would want to take a follow up action, like printing some message at least, or cleaning up before terminating. We should separate these concepts.

src/snowcli/cli/nativeapp/version/commands.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+@app.command()
+@with_output
+@global_options_with_connection
+def create(
+    version: Optional[str] = typer.Argument(
+        None,
+        help=f"""The identifier or 'version string' of the version you would like to create a version and/or patch for.
+        Defaults to undefined, which means the CLI will use the version specified in the manifest.yml.""",
```
sfc-gh-bdufour, line 36: if present? The manifest may not have a version.

src/snowcli/cli/nativeapp/version/commands.py
```diff
@@ -0,0 +1,94 @@
...
+    ),
+    patch: Optional[str] = typer.Option(
+        None,
+        "--patch",
+        "-p",
+        help=f"""The patch number you would like to create for an existing version.
+        Defaults to undefined if it is not set, which means the CLI will either use the one specified in the manifest.yml,
+        or the framework will determine the patch number itself.""",
```
sfc-gh-bdufour, line 44: nit: the patch number will be auto-generated?

not sure "the framework" is known concept externally

src/snowcli/cli/nativeapp/version/commands.py
```diff
@@ -0,0 +1,94 @@
...
+    ),
+    force: Optional[bool] = typer.Option(
+        False,
+        "--force",
+        help="Defaults to unset (False). If set (True), we will implicitly respond “yes” to any prompts that come up.",
```
sfc-gh-bdufour, line 49: nit: what's the different between unset and false? Do users care about the semantic difference?

src/snowcli/cli/nativeapp/version/commands.py
```diff
@@ -0,0 +1,94 @@
...
+        is_flag=True,
+    ),
+    **options,
+) -> CommandResult:
+    """
+    Creates a new version for an application package, or adds a patch if the version already exists.
```
sfc-gh-bdufour, line 55: Doesn't it simply create the specified version and patch? I find that description a bit confusing.

src/snowcli/cli/nativeapp/version/commands.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    if version is None and patch is not None:
+        raise ClickException("Cannot provide patch without version!")
+
+    processor = NativeAppVersionCreateProcessor(
+        project_definition=cli_context.project_definition,
+        project_root=cli_context.project_root,
+    )
+    processor.process(version, patch, force)
+    return MessageResult(f"Version create is now complete.")
```
sfc-gh-bdufour, line 65: nit: not very informative. Do we output the specific version/patch that was created anywhere, at least?

src/snowcli/cli/nativeapp/version/commands.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+@app.command()
+@with_output
+@with_project_definition("native_app")
+@global_options_with_connection
+def drop(
```
sfc-gh-bdufour, line 72: Are we planning on adding `snow app version list` or similar soon? It's a bit odd to have create/drop without an ability to inspect the versions.

src/snowcli/cli/nativeapp/version/version_processor.py
sfc-gh-bgoel: Addressed. 
Re: quoting rules - I am not sure myself. I'll need to check this.

sfc-gh-bgoel:
Re: consistent naming for `check` vs `warn`. However, they are doing two different things imo and hence the difference.
E.g. `check_index_changes_in_git_repo` is checking changes in git and displaying the warning to the user/asking for input.
But `warn_user_about_existing_rd` is just doing some string manipulation and displaying the warning to the user/asking for input.
Possible consistent corrections can be: 
`check_index_changes_in_git_repo` -> `check_index_changes_in_git_repo_and_warn`
OR
Remove use of `warn_user_about_existing_rd` as it only consists of the join and then the call to the user policy, which can be added back to the parent method. 

Other methods called `get_...` have their comments modified now. The intent for these functions is never to interact with the user, but query snowflake and get results.

src/snowcli/cli/nativeapp/version/version_processor.py
```diff
@@ -0,0 +1,311 @@
...
+            # show differences between current files and last commit
+            print(repo.git.diff(repo.head.commit.tree))
+
+            ask_user_confirmation(
+                user_confirmation_policy=user_confirmation_policy,
+                prompts=Prompts(
+                    f"You have local changes in this repository that are not part of a previous commit. Do you still want to continue? Consider using 'git status' and/or 'git diff' before proceeding.",
+                    f"Did not create a new version {version} of application package {package_name}.",
+                    "Cannot create a new version without user input. Consider using --force if you would like to attempt forcing the creation of version.",
+                ),
```
sfc-gh-bdufour: I find this difficult to read & reason about. Let's consider alternatives that make the flow more explicit?

src/snowcli/cli/nativeapp/version/version_processor.py
```diff
@@ -0,0 +1,311 @@
...
+    version: str,
+    package_name: str,
+    user_confirmation_policy: UserConfirmationPolicy,
+) -> None:
+    """
+    Checks if the project root, i.e. the native apps project is a git repository. If it is a git repository,
+    it also checks if there any local changes to the directory that may not be on the app package stage.
+    """
+    try:
+        repo = Repo(project_root)
```
sfc-gh-bdufour, line 50: Not sure why we fold git repos into this, can you explain? Having local files that have never been synced is one thing, but not having files checked in is another. What if the repo is actually in a parent directory?

src/snowcli/cli/nativeapp/version/version_processor.py
```diff
@@ -0,0 +1,311 @@
...
+                    f"Did not create a new version {version} of application package {package_name}.",
+                    "Cannot create a new version without user input. Consider using --force if you would like to attempt forcing the creation of version.",
+                ),
+            )
+
+    except InvalidGitRepositoryError:
+        pass  # not a git repository, which is acceptable
+
+
+def warn_user_about_existing_rd(
```
sfc-gh-bdufour, line 71: nit: expand rd? I had to keep reading to know what it meant

src/snowcli/cli/nativeapp/artifacts.py
sfc-gh-bgoel: Changed it here, since `load` is indeed ambiguous.

src/snowcli/cli/nativeapp/version/version_processor.py
sfc-gh-bgoel: This is implemented from our functional specs. Let's have that discussion in our channel.

src/snowcli/cli/nativeapp/artifacts.py
sfc-gh-bdufour: I don't understand what "important enough to warrant its own exception" means. Can you explain? We have, in the native apps module, 18 exception types if I count correctly. I'm worried we're creating bloat without any benefit.

src/snowcli/cli/nativeapp/version/commands.py
sfc-gh-bdufour: I strongly believe we should revisit the plan.

sfc-gh-bgoel:
Edited the policy object after offline discussion with @sfc-gh-bdufour .
Will resolve older comments on this topic. Please create new comments for starting the conversation on the newly pushed changes.

src/snowcli/cli/nativeapp/policy.py
```diff
@@ -0,0 +1,31 @@
...
+        return True
+
+
+class AlwaysAskPolicy(PolicyBase):
+    def should_proceed(self, ctx: TaskContext):
+        if ctx.is_interactive_mode:
+            should_continue = typer.confirm(ctx.user_prompt)
+            return should_continue
+        else:
+            self.exit_code = 1
```
sfc-gh-bgoel, line 30: Other possible approaches:
1. Use a tuple return type (T/F, exit_code)
2. Use enum values as return type.
The above two are in case we want no side-effects, such as re-assigning `exit_code` to 1.

src/snowcli/cli/nativeapp/version/version_processor.py
```diff
@@ -0,0 +1,312 @@
...
+    find_first_row,
+    is_user_in_interactive_mode,
+)
+from snowcli.cli.project.util import unquote_identifier
+from snowcli.exception import SnowflakeSQLExecutionError
+from snowflake.connector import ProgrammingError
+from snowflake.connector.cursor import DictCursor
+
+# Custom lambdas to plug in to UserConfirmationPolicy
+is_force_lambda = lambda force: force
```
sfc-gh-bdufour, line 33: unused?

src/snowcli/cli/nativeapp/utils.py
```diff
@@ -1,15 +1,29 @@
...
-    )
+    return next(_rows_generator(cursor, predicate), None)
 
 
 def needs_confirmation(needs_confirm: bool, auto_yes: bool) -> bool:
     return needs_confirm and not auto_yes
+
+
+@cache
+def is_user_in_interactive_mode():
```
sfc-gh-bdufour, line 28: Feels like this isn't in the right place. It's not uncommon to provide CLI options to force one way or the other (--non-interactive/--force-interactive). IMHO it would be a lot easier to bundle this with policy selection at the entry point, e.g.:

```
if force:
    policy = AllowAll()
elif is_interactive():
    policy = AlwaysAsk()
else:
    policy = DenyAll()
```

It would then be trivial to introduce flags to change the default behaviour in the future, if needed.

src/snowcli/cli/nativeapp/version/version_processor.py
```diff
@@ -0,0 +1,312 @@
...
+    it also checks if there any local changes to the directory that may not be on the app package stage.
+    """
+    try:
+        repo = Repo(project_root)
+
+        # Check if the repo has any changes, including untracked files
+        if repo.is_dirty(untracked_files=True):
+            print("Changes detected in your git repository!")
+            # show differences between current files and last commit
+            print(repo.git.diff(repo.head.commit.tree))
```
sfc-gh-bdufour, line 55: This is a bit odd to me for a few reasons:
-  git status would be better at showing the diff, no? What does this look like for the user?
- the only way out of this is --force, which is rather dangerous. Should we have a git-specific flag to disable this check? Something like --ignore-git or similar.
- still somewhat worried about the fact that the project root might not be the git root. If we had a "no git" flag, maybe that's a moot point.

src/snowcli/cli/nativeapp/version/version_processor.py
```diff
@@ -0,0 +1,312 @@
...
+            )
+
+            if not policy.should_proceed(ctx):
+                if policy.exit_code:
+                    print(
+                        "Cannot create a new version non-interactively without --force"
+                    )
+                else:
+                    print("Not creating a new version.")
+                raise typer.Exit(policy.exit_code)
```
sfc-gh-bdufour, line 69: Not fully understanding this. We're exiting either. What's the difference/why do we have different messages when exit_code is set? It's a bit odd to me that the policy can make a "and you must now exit" recommendation anyway.

src/snowcli/cli/nativeapp/version/version_processor.py
sfc-gh-bgoel: Answering both above comments in this one:
1. On whether we should check for git repositories (dirty or not) - this is based off of the functional specs. We can always revisit, but as discussed offline, let's discuss on a different channel.
2. I am neutral about git diff vs git status, will change it in the next commit. 
3. This git check is not behind a `--force` check. It happens regardless the policy.
4. project root not being the git root is taken care of by `search_parent_directories=True` coming up in the next commit. If the cwd is in some child directory (or further down), it will still be able to detect if it is a git repo.

src/snowcli/cli/nativeapp/version/version_processor.py
sfc-gh-bgoel: Exit codes also based on the functional specs:
1. User choosing to abort the rest of execution is exit code zero, i.e. success because that's what the user wants - `"Not creating a new version."`
2. If the execution aborts because it needed user input but was not able to proceed, that is exit code one, i.e. failure because user did not get to achieve the end state they desired - `"Cannot create a new version non-interactively without --force"`

I don't necessarily see it as the policy making a "you must now exit" recommendation. It informs the exit (on the type of exit code to use if exiting), but the invoking method is ultimately deciding when to exit. Example: in case of allow_always policy, exit code is 0 by default. But it does not tell the invoking method to exit, and the method continues execution.

src/snowcli/cli/nativeapp/utils.py
sfc-gh-bdufour: deny by default is odd in an interactive shell, we should ask instead. This is how most dev tools behave.

To be clear, I wasn't asking to include interactive/non-interactive switches, I was just pointing out that the implementation would make those difficult to add later. Usually this, to me, suggests that responsibilities might need to be moved around a bit, which is what I suggested.

src/snowcli/cli/nativeapp/version/version_processor.py
sfc-gh-bdufour: See my comments in the func spec

src/snowcli/cli/nativeapp/version/version_processor.py
sfc-gh-bdufour: Aborting usually means exit code wouldn't be zero, even if the user chose to abort. It's still not a success. Most often the exit code is used to determine whether the next command can proceed. 

Also, the error message seems to imply that creating a version always requires --force to be provided in a CI/CD context. Is that really true?

src/snowcli/cli/nativeapp/utils.py
```diff
@@ -1,15 +1,27 @@
...
-        None,
-    )
+    return next(_rows_generator(cursor, predicate), None)
 
 
 def needs_confirmation(needs_confirm: bool, auto_yes: bool) -> bool:
     return needs_confirm and not auto_yes
+
+
+def is_interactive_mode():
```
sfc-gh-cgorrie, line 26: [nit] I'd rename this to `is_tty_interactive()` so developers reading the code don't mistake it for whether or not the interactive (`-i`/`--interactive`) flag was passed.

sfc-gh-cgorrie (commented):
It's looking good to me right now, just one nit I saw. Let's wait until Bruno comes back before merging so he gets a chance to respond.

src/snowcli/cli/nativeapp/policy.py
```diff
@@ -0,0 +1,25 @@
+import typer
+
+
+class PolicyBase:
+    exit_code: int = 0
+
+    def should_proceed(self, user_prompt: str):
+        pass
```
sfc-gh-mraba, line 8: ABC could be used to ensure interface implementation?

```suggestion
class PolicyBase(abc.ABC):
    exit_code: int = 0

    def should_proceed(self, user_prompt: str) -> bool:
        pass
```

sfc-gh-bgoel:
@sfc-gh-bdufour Adding a common comment here after discussion with @sfc-gh-cgorrie on some of the open questions:
1. The policy on deny by default has been addressed with your suggestion.
2. Git status check is kept in the code, though it can be omitted by a new flag
3. Exit codes continue to follow the pattern described in the functional specs. 
4. Executing any commands that have a potential for interactive mode to be invoked will require the use of `--force` everytime to skip them, such as in a CICD context.

src/snowcli/cli/nativeapp/policy.py
sfc-gh-bgoel: Sure thing, fixing in the upcoming commit. Thanks!

src/snowcli/cli/nativeapp/version/version_processor.py
```diff
@@ -0,0 +1,306 @@
...
+            user_prompt = (
+                "You have local changes in this repository that are not part of a previous commit. Do you still want to continue?",
+            )
+            if not policy.should_proceed(user_prompt):
+                if policy.exit_code:
+                    print(
+                        "Cannot create a new version non-interactively without --force."
+                    )
+                else:
+                    print("Not creating a new version.")
```
sfc-gh-bdufour, line 59: From a responsibility POV, this seems backwards. This logic knows whether being interactive is a dealbreaker. The policy shouldn't. Otherwise we get some unwanted coupling that make policies difficult to evolve/reuse in different contexts. I think this should read something like:

```
should_proceed = policy...
if not should_proceed:
    if is_interactive: # we'll need to carry this context around
        raise SomeError("Cannot create a new version ...")
    else:
        raise SomeError("Not creating a new version")
```

We can either thread the interactive state through the processors, or query the policies for it if you prefer. I'm fine either way.

src/snowcli/cli/nativeapp/version/version_processor.py
```diff
@@ -0,0 +1,306 @@
...
+                    Version {version} already exists for application package {self.package_name} and in release directive(s): {release_directive_names}.
+                """
+                )
+            )
+
+            user_prompt = (
+                f"Are you sure you want to create a new patch for version {version} of application package {self.package_name}? Once added, this operation cannot be undone.",
+            )
+            if not policy.should_proceed(user_prompt):
+                if policy.exit_code:  # is 1
```
sfc-gh-bdufour, line 223: See above

sfc-gh-bdufour (approved):

sfc-gh-astus (approved):

PR #636, opened 2024-01-08
Snow 986254 add file logger

sfc-gh-pczajka:
### Pre-review checklist

### Changes description
This PR introduces saving logs into a file.

By default logs are not saved, you can turn them on by adding `[logs]` section to config.toml (this section will appear in newly created config file):
Default values:
```
[logs]
save_logs = false
path = "<logs directory created next to default location of config file>"
level = "info"
```

#### Code changes summary:
* `snowcli/config.py` - add parsing "logs" section to config.toml
* `snowcli/loggers.py` - logic of checking whether logs config is correct + create rotating file log handler
* `tests/test_logs.py` - add tests
* `tests/conftest.py` - merge two autoused fixtures into one, as their order is important (`logger.create_loggers` cannot use previous tests' config - it needs to be reloaded before)


src/snowcli/app/main_typer.py
```diff
@@ -1,12 +1,18 @@
...
 import sys
 
 import click
 from snowcli.cli.common.cli_global_context import cli_context
 from snowcli.cli.common.flags import DEFAULT_CONTEXT_SETTINGS, DebugOption
+from snowcli.cli.loggers import remove_console_output_handler_from_logs
 from typer import Typer
 
 
 def _handle_exception(exception: Exception):
```
sfc-gh-pczajka, line 11: Notes for reviewer: I'd expect logs to contain traceback in case of exception, but I also wanted to preserve current behavior. I decided to:
    - Keep console output handler, so the users won't loose the output they've seen until now
    - delete console handler in the line below, as:
      - the program is about to end :)
      - I'm not changing/reimplementing the logic of visibility of traceback on console output

tests/conftest.py
```diff
@@ -13,10 +13,16 @@
...
+#
+# This automatically used setup fixture is required to use test.conf from resources
+# in unit tests which are not using "runner" fixture (tests which do not invoke CLI command).
+def reset_global_context_and_setup_config_and_logging_levels(
+    request, test_snowcli_config
+):
     cli_context_manager.reset()
     cli_context_manager.set_verbose(False)
     cli_context_manager.set_enable_tracebacks(False)
+    config_init(test_snowcli_config)
```
sfc-gh-pczajka, line 25: I needed to move this line above `loggers.create_loggers` - otherwise it would use config from the previous test, throwing "logs directory not found" error

src/snowcli/cli/loggers.py
```diff
@@ -1,32 +1,113 @@
...
+
+def add_console_output_handler_to_logs(
+    log_level: int, formatter: logging.Formatter
+) -> None:
+    logger = logging.getLogger("snowcli")
+    console = logging.StreamHandler()
+    console.set_name(_CONSOLE_OUTPUT)
+    console.setFormatter(formatter)
+    console.setLevel(log_level)
+    logger.addHandler(console)
```
sfc-gh-turbaszek, line 72: I'm wondering if we can utilise `dictConfig` to store more structured information about loggers
https://docs.python.org/3/library/logging.config.html#logging-config-dictschema 

One of the examples I know:
https://github.com/apache/airflow/blob/95a83102e8753c2f8caf5b0d5c847f4c7f254f67/airflow/config_templates/airflow_local_settings.py#L66-L140

Maybe it would be useful to define multiple configurations? What do you think?

src/snowcli/cli/loggers.py
```diff
@@ -1,32 +1,113 @@
...
+    logger = logging.getLogger("snowcli")
+    logger.setLevel(global_log_level)
+    logging.getLogger("snowflake").setLevel(global_log_level)
+
+    add_console_output_handler_to_logs(
+        console_log_level, formatter=logging.Formatter(console_log_format, DATE_FORMAT)
+    )
+
+    if config.save_logs:
+        filename = config.path / "snowcli.log"
```
sfc-gh-turbaszek, line 109: Should it be a constant? Should we encapsulate this into a method?

src/snowcli/cli/loggers.py
sfc-gh-pczajka: Nice one! Definitely increases readability - done

src/snowcli/cli/loggers.py
sfc-gh-pczajka: done :)
[named _DEFAULT_LOGS_FILENAME]

src/snowcli/app/loggers.py
```diff
@@ -1,32 +1,132 @@
...
+            "disable_existing_loggers": True,
+            "formatters": {
+                "console_formatter": {
+                    "class": "logging.Formatter",
+                    "format": console_log_format,
+                    "datefmt": DATE_FORMAT,
+                },
+                "file_formatter": {
+                    "class": "logging.Formatter",
+                    "format": FILE_LOG_FORMAT,
```
sfc-gh-turbaszek, line 116: Do we need this constant? Maybe we can inline this value here?

src/snowcli/app/loggers.py
```diff
@@ -1,32 +1,132 @@
...
+
+    logging.config.dictConfig(
+        {
+            "version": 1,
+            "disable_existing_loggers": True,
+            "formatters": {
+                "console_formatter": {
+                    "class": "logging.Formatter",
+                    "format": console_log_format,
+                    "datefmt": DATE_FORMAT,
```
sfc-gh-turbaszek, line 112: Do we need to specify it? Won't this work out of box?

src/snowcli/app/loggers.py
```diff
@@ -1,32 +1,132 @@
...
+    }
+    if config.save_logs:
+        enabled_handlers.append("file")
+        handlers_config["file"] = {
+            "class": "logging.handlers.TimedRotatingFileHandler",
+            "filename": config.filename,
+            "when": "midnight",
+            "formatter": "file_formatter",
+            "level": config.file_log_level,
+        }
```
sfc-gh-turbaszek, line 102: What would you say to approach this the other way around. Now we are building the configs which makes it non-trivial to understand full configuration.

The other option would be to define all possible handlers in dictConfig constant (with hardcoded levels etc). Then we remove handler from this config according to users configuration. WDYT?

src/snowcli/app/loggers.py
sfc-gh-pczajka: It would, but it would give you default formatter:
```
In [3]: logging.error("message")
ERROR:root:message
```
If we want to use "prettier" format, we need to override it here

src/snowcli/app/loggers.py
sfc-gh-pczajka: I wanted to keep all format definitions in one place - that's why I defined it as a global.
I also thought about creating "detailed_formatter" and "basic_formatter", and choose formatter in console handler, but no strong preference here.

sfc-gh-pczajka: EDIT: I moved these definitions to `DEFAULT_LOGGING_CONFIG` - default_formatter vs detailed_formatter made further logic (verbose and debug flags) more clear

src/snowcli/app/loggers.py
sfc-gh-pczajka: Neat. Done :)

src/snowcli/app/loggers.py
```diff
@@ -1,32 +1,136 @@
...
+            "class": "logging.StreamHandler",
+            "formatter": "default_formatter",
+            "level": logging.ERROR,
+        },
+        "file": {
+            "class": "logging.handlers.TimedRotatingFileHandler",
+            "filename": None,
+            "when": "midnight",
+            "formatter": "detailed_formatter",
+            "level": logging.INFO,
```
sfc-gh-turbaszek, line 43: Do you need log level on handler level? Doesn't logger level overrides handler level? If that's true the logic probably can be simplified a bit

sfc-gh-pczajka: Unfortunately. "Log level" works as a filter, simplified flow looks like this:
1. We call `logging.get_logger('snowcli.app.connector').info('message')`
2. Log message goes through loggers (`snowcli.app.connector` -> `snowcli.app` -> `snowcli` -> `root`) until it is filtered out by log level or finds one with handlers. (it is possible to send the message further after finding handlers with `propagate` setting)
3. Log message goes through all handlers and is printed by ones with appropriate level setting.
Details: https://docs.python.org/3/howto/logging.html#logging-flow 

So basically what I'm doing here is to set loggers to lowest possible level to pass through all interesting messages. 

Alternatively I could set loggers on `DEBUG` and let the handlers do the filtering. I didn't do that as for some reason someone took an effort to set level on `snowflake` and `snowcli` loggers, so I guessed there was some reasoning behind it. I might be wrong on this one tho, I'll rely on your opinion.

src/snowcli/api/config.py
```diff
@@ -53,6 +59,30 @@ def add_connection(name: str, parameters: dict):
     _dump_config(conf_file_cache)
 
 
+_DEFAULT_LOGS_CONFIG = {
```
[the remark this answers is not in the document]
sfc-gh-pczajka: Done.
IMO it made default config a bit harder to read, but it's a fair tradeoff for spelling errors safety.

src/snowcli/app/loggers.py
```diff
@@ -1,32 +1,158 @@
...
+        self.level: int = logging.getLevelName(config["level"].upper())
+        if debug:
+            self.level = logging.DEBUG
+
+        self._check_log_level(config)
+        if self.save_logs:
+            self._check_logs_directory_exists()
+
+    def _check_logs_directory_exists(self):
+        print(self.path.exists())
```
sfc-gh-turbaszek, line 92:
```suggestion

```

sfc-gh-turbaszek (dismissed):

PR #639, opened 2024-01-09
[NADE] Distribution Warnings and Bad Parameter Exception

sfc-gh-bgoel:
### Pre-review checklist
   * [x] I've confirmed that my changes are up-to-date with the target branch - Jan 9 2:30PM EST

### Changes description
This PR introduces two changes:
1. Perform app package distribution mismatch check in `snow app version drop`.
2. For `snow app version create`, if a user passes in both version and patch and the version does not exist, throw an error.

### DIFF
--- a/src/snowcli/cli/nativeapp/manager.py
+++ b/src/snowcli/cli/nativeapp/manager.py
@@ -219,16 +219,16 @@ class NativeAppManager(SqlExecutionMixin):
             )
         )
 
-    def check_app_pkg_distribution_against_resolved_definition(
-        self, sf_distribution: Optional[str] = None
+    def verify_project_distribution(
+        self, expected_distribution: Optional[str] = None
     ) -> bool:
         """
         Returns true if the 'distribution' attribute of an existing application package in snowflake
         is the same as the the attribute specified in project definition file.
         """
         actual_distribution = (
-            sf_distribution
-            if sf_distribution
+            expected_distribution
+            if expected_distribution
             else self.get_app_pkg_distribution_in_snowflake
         )
         project_def_distribution = self.package_distribution.lower()
--- a/src/snowcli/cli/nativeapp/manager.py
+++ b/src/snowcli/cli/nativeapp/manager.py
@@ -219,12 +219,18 @@ class NativeAppManager(SqlExecutionMixin):
             )
         )
 
-    def is_app_pkg_distribution_same_in_sf(self) -> bool:
+    def is_app_pkg_distribution_same_in_sf(
+        self, sf_distribution: Optional[str] = None
+    ) -> bool:
         """
         Returns true if the 'distribution' attribute of an existing application package in snowflake
         is the same as the the attribute specified in project definition file.
         """
-        actual_distribution = self.get_app_pkg_distribution_in_snowflake
+        actual_distribution = (
+            sf_distribution
+            if sf_distribution
+            else self.get_app_pkg_distribution_in_snowflake
+        )
         project_def_distribution = self.package_distribution.lower()
         if actual_distribution != project_def_distribution:
             log.warning(
--- a/src/snowcli/cli/nativeapp/manager.py
+++ b/src/snowcli/cli/nativeapp/manager.py
@@ -219,7 +219,7 @@ class NativeAppManager(SqlExecutionMixin):
             )
         )
 
-    def is_app_pkg_distribution_same_in_sf(
+    def check_app_pkg_distribution_against_resolved_definition(
         self, sf_distribution: Optional[str] = None
     ) -> bool:
         """
--- a/src/snowcli/cli/nativeapp/run_processor.py
+++ b/src/snowcli/cli/nativeapp/run_processor.py
@@ -58,7 +58,9 @@ class NativeAppRunProcessor(NativeAppManager, NativeAppCommandProcessor):
 
             # 2. Check distribution of the existing app package
             actual_distribution = self.get_app_pkg_distribution_in_snowflake
-            if not self.is_app_pkg_distribution_same_in_sf(actual_distribution):
+            if not self.check_app_pkg_distribution_against_resolved_definition(
+                actual_distribution
+            ):
                 print(
                     f"Continuing to execute `snow app run` on app pkg {self.package_name} with distribution '{actual_distribution}'."
                 )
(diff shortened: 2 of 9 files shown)


src/snowcli/cli/nativeapp/manager.py
```diff
@@ -219,12 +219,18 @@ def get_app_pkg_distribution_in_snowflake(self) -> str:
             )
         )
 
-    def is_app_pkg_distribution_same_in_sf(self) -> bool:
+    def is_app_pkg_distribution_same_in_sf(
```
sfc-gh-bdufour, line 222: Not a fan of having something called is_<foo> have a side effect that logs a warning. How about we call this check or verify instead? Also, now that it accepts a parameter, the "_in_sf" part isn't necessarily true.

src/snowcli/cli/nativeapp/run_processor.py
```diff
@@ -227,9 +228,16 @@ def get_existing_version_info(self, version: str) -> Optional[dict]:
         """
         with self.use_role(self.package_role):
             show_obj_query = f"show versions like '{unquote_identifier(version)}' in application package {self.package_name}"
```
sfc-gh-bdufour, line 230: Feels like we should have a utility to do this already. If we don't, can we create one?

tests/nativeapp/test_run_processor.py
```diff
@@ -1264,17 +1249,41 @@ def test_upgrade_app_recreate_app(
...
     "snowcli.cli.nativeapp.run_processor.NativeAppRunProcessor.get_existing_version_info",
     return_value=None,
 )
-@mock_connection()
 @pytest.mark.parametrize(
     "policy_param", [allow_always_policy, ask_always_policy, deny_always_policy]
 )
-def test_upgrade_app_from_version(mock_conn, mock_existing, policy_param, temp_dir):
-    mock_conn.return_value = MockConnectionCtx()
+def test_upgrade_app_from_version_throws_usage_error_one(
```
sfc-gh-bdufour, line 1260: what does _one mean here?

tests/nativeapp/test_version_create_processor.py
```diff
@@ -228,6 +229,63 @@ def test_process_no_version_from_user_no_version_in_manifest(
...
+
+# Test version create when user passed in a version and patch AND app package does not exist
+@mock.patch(
+    f"{VERSION_MODULE}.{CREATE_PROCESSOR}.get_existing_version_info",
+    side_effect=ApplicationPackageDoesNotExistError("app_pkg"),
+)
+@pytest.mark.parametrize(
+    "policy_param", [allow_always_policy, ask_always_policy, deny_always_policy]
+)
+def test_process_no_version_exists_throws_bad_option_exception_two(
```
sfc-gh-bdufour, line 268: ditto, what's _one and _two here?

src/snowcli/cli/nativeapp/run_processor.py
sfc-gh-bgoel: Not sure I follow. This query is part of `get_existing_version_info`, and this is the only query executed in this method.

tests/nativeapp/test_version_create_processor.py
sfc-gh-bgoel: Replying for similar comments in one:
When you call `get_existing_version_info`, `BadOptionUsage` can be thrown in two situations: no existing version, or no existing app package. Hence the `one` and `two` to keep test names relatively short. But that is why each test has a comment above (like 260) describing what it tests.

src/snowcli/cli/nativeapp/run_processor.py
sfc-gh-bdufour: I mean the "get all versions" part of get_existing_version_info. The first part of the utility function seems to just try to find all versions, and then we look for the desired one. Felt like maybe that's something that should be factored out?

sfc-gh-bdufour: For instance, right above there's "get_all_existing_versions", but if the app package doesn't exit, I think that function has a different behaviour w.r.t. raised exceptions. Seemed odd is all.

sfc-gh-bgoel: Oh right, I did try to unify the two: `get_all_existing_versions` and `get_existing_version_info` by only performing the filtering on the returned results of `get_all_existing_versions` in `get_existing_version_info`. However, the way that the results are returned in `SnowflakeCursor` is an odd one to handle. Here's why:

If I try to dictate the `_execute_query` method within `get_all_existing_versions` to use `DictCursor` in order to filter on version name in `get_existing_version_info`, the `_execute_query` ends up wiping the values in the returned results entirely, this was discovered when I ran the integration tests for the PR that introduced the change: https://github.com/Snowflake-Labs/snowcli/actions/runs/7416652530/job/20181962128

This is why `get_existing_version_info` performs `show versions like <version> ... ` whereas `get_all_existing_versions` does a `show versions in ... `because the latter does not require filtering. 

In addition, there are a few other methods that also execute a singular query, and perform a filter. Some examples are: `NativeAppManager.get_existing_app_info`, `NativeAppManager.get_existing_app_pkg_info` and `NativeAppVersionCreateProcessor.add_new_patch_to_version`. That would make `get_existing_version_info` consistent with how these are written to. 

Re: _right above there's "get_all_existing_versions", but if the app package doesn't exit, I think that function has a different behaviour w.r.t. raised exceptions_ - I feel like this is a separate point about making the exceptions behavior consistent between the two methods, at the very least. 

Happy to chat about this on slack.

sfc-gh-bdufour: Thanks for the detailed explanation. Overall I feel there's a bit of a lack of abstraction in our code, but that's a minor point here. You're right, what jumped out at me is that depending on the utility method, the behaviour seems slightly inconsistent, at least outside of the happy path. To me these two things are somewhat related, but that might be something to discuss as a team in the near future instead. Don't let this block this PR.

src/snowcli/cli/nativeapp/manager.py
```diff
@@ -219,12 +219,18 @@ def get_app_pkg_distribution_in_snowflake(self) -> str:
             )
         )
 
-    def is_app_pkg_distribution_same_in_sf(self) -> bool:
+    def check_app_pkg_distribution_against_resolved_definition(
```
sfc-gh-bdufour, line 222: how about simply: `verify_project_distribution(expected_distribution: Optional[str)`

Without an argument, it checks the project distribution against SF. With an arg, it verifies against that instead.

Not sure what a "resolved definition" means at first glance. I like verify here because it implies that matching is very much expected, so issuing a warning in this method makes sense.

sfc-gh-bgoel: I like that way of reasoning, updating now!

sfc-gh-bdufour (approved):

PR #640, opened 2024-01-10
Move utils to proper modules

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
Move top level / loose utils modules to `snowcli.api` or `snowcli.app`. In service-specific cases utiles were moved to snowpark or streamlit modules.

This limits imports from `snowli.app` in `snowcli.cli` to bare minimum (currently 1 occurrence).


sfc-gh-jsikorski (dismissed):

src/snowcli/app/printing.py
```diff
@@ -8,9 +8,9 @@
 from rich import box, get_console, print
```
sfc-gh-astus, line 8: I think printing is a core functionality and should be in `api`. WDYT?

sfc-gh-turbaszek: Currenlty we don't encourage users to do printing. Once we have established way of showing output other than `with_output` decorator we can put it in `api`. I would treat current `printing` module as internal implementation

src/snowcli/api/commands/decorators.py
```diff
@@ -253,3 +256,14 @@ def _evaluate_param_type(
...
+
+def with_output(func):
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        output_data = func(*args, **kwargs)
+        if not isinstance(output_data, CommandResult):
+            raise CommandReturnTypeError(type(output_data))
+        print_result(output_data)
+
+    return wrapper
```
sfc-gh-turbaszek, line 269: Moved here to keep all commands decorators in one place

sfc-gh-astus (approved):

src/snowcli/cli/connection/commands.py
```diff
@@ -219,6 +217,8 @@ def test(connection: str = ConnectionOption, **options) -> CommandResult:
     """
     Tests the connection to Snowflake.
     """
+    from snowcli.app.snow_connector import connect_to_snowflake
```
sfc-gh-turbaszek, line 220: Moved here to avoid top level import from `app`

src/snowcli/output/decorators.py
sfc-gh-turbaszek, line 1: Content moved to `api.commands.decorators`

src/snowcli/utils/file_utils.py
sfc-gh-turbaszek, line 1: Content moved to `cli.snowpark.package.utils`

tests/nativeapp/test_init.py
```diff
@@ -1,8 +1,6 @@
-from pathlib import Path
 from textwrap import dedent
-from unittest import mock
 
-import pytest
```
sfc-gh-bgoel, line 5: The reason why removing these imports does not cause the tests to fail in this file is because of an import on line 21 `from tests.testing_utils.fixtures import *` which happens to contain the three imports as well. 
Maybe it would be better to keep these imports explicit in the current file too since if we change any import in `testing_utils.fixtures`, `test_init.py` will also break.

sfc-gh-turbaszek: This is possible but I want to limit code changes (other than renames) to bare minimum. We can deal with * imports in a followup. I think https://github.com/Snowflake-Labs/snowcli/pull/611 addresses it

sfc-gh-bgoel: sgtm, thanks!

sfc-gh-bgoel (approved):
LGTM for native apps + project definition.

PR #659, opened 2024-01-16
SNOW-986249-console-output-tools: cli_console for intermediate output

sfc-gh-mraba:
SNOW-986249

### Pre-review checklist

### Changes description

- `snowflake.cli.api.output` package introduces `snow_cli` object for displaying console intermediate output
- two methods `phase` and `step` provide displaying output to console
- `step` method for displaying messages to users
- `warning` method for displaying important messages to users
- `phase` method for grouping messages in distinctive way


src/snowflake/cli/api/console/console.py
```diff
@@ -0,0 +1,43 @@
...
+        """Utility for displaying high level steps in commands.
+
+        Messages are never indented. Subsequent step messages will be indented."""
+        self.register_output(Output.PHASE)
+        self._print(message)
+
+    def step(self, message: str):
+        """Prints indented message.
+
+        Indentation is based on previous usage of `phase` method."""
```
sfc-gh-bdufour, line 32: You might want to clarify that step is allowed outside of a phase, as I don't think this statement makes it clear.

src/snowflake/cli/api/console/console.py
```diff
@@ -0,0 +1,43 @@
+from __future__ import annotations
+
+from snowflake.cli.api.cli_global_context import cli_context
+from snowflake.cli.api.console.abc import AbstractConsole
+from snowflake.cli.api.console.enum import Output
+
+
+class CliConsole(AbstractConsole):
```
sfc-gh-bdufour, line 8: the phase/step API is interesting. It looks like top-level steps are allowed, but as soon as a phase is used, then top level steps are no longer allowed. So something like:
- step 1
- long phase
  - step a
  - step b
- step 2

cannot be output, is that right?

Also, it looks like there is no way to indicate a success / error for a step. Is that something you've considered? It's rather common to use something like a spinner that resolves to some kind of OK/Error indicator when the task is complete, which visually a little more intuitive for a user. If you're doing something like a parallel upload to a stage, having the ability to indicate progress for each file might come in handy, for instance.

sfc-gh-bdufour (commented):
Good to see this feature implemented, it will greatly improve the usability of NADE tools.

src/snowflake/cli/api/cli_global_context.py
```diff
@@ -218,8 +220,13 @@ def silent(self) -> bool:
         return self._silent
 
     def set_silent(self, value: bool):
+        if self.output_format == OutputFormat.JSON:
+            value = True
         self._silent = value
```
sfc-gh-turbaszek, line 225: I I have feeling this is bit counter intuitive. You set something but you don't. I would personally move this into getter of `self._silent`, WDYT?

src/snowflake/cli/api/console/__init__.py
```diff
@@ -0,0 +1,3 @@
+from snowflake.cli.api.console.console import cli_console
+
+__all__ = ("cli_console",)
```
sfc-gh-turbaszek, line 3: +1

src/snowflake/cli/api/console/console.py
sfc-gh-melnacouzi: Any reason why we ended up with only phase and step? It seems a bit too restrictive to only allow one/2 indentations. Is there anything specific to being in a phase or in a step except the indentation? and if not, can't we just accept indentation value as a parameter?

sfc-gh-mraba: The goal of separate `phase` and `step` was to display extra information while commands are executed:
- `phase` to be used for main events 
- `step` to be used when some detailed information needed to be displayed for main event - indentation.

more like `heading 1` and  `hading 2` format.

The case @sfc-gh-bdufour introduced may suggest people will use this methods interchangeably and keeping track in mind which method is intended to produce what kind of output is an unnecessary complexity.

I think a solution with less overhead for dev would be a simple `print` like method with a param to indicate whether output should be indented or not.

consider this two examples:

```python
import cli_console as cc

cc.phase("Task A")
cc.phase("Task B")
cc.step("Task B: part 1")
cc.step("Task B: part 2")
cc.exit_phase()
cc.step("Some C")
```

vs
```python
import cli_console a cc

cc.print("Task A")
cc.print("Task B")
cc.print("Task B: part 1", indent=True)   # or cc.print_indented("Task B: part 1")
cc.print("Task B: part 2", indent=True)   # or cc.print_indented("Task B: part 2")
cc.print("Some C")
```

WDYT?

src/snowflake/cli/api/console/__init__.py
sfc-gh-mraba: What do you mean by +1?

src/snowflake/cli/api/cli_global_context.py
sfc-gh-mraba: Done

src/snowflake/cli/api/console/console.py
sfc-gh-turbaszek: The main reason to have phase/step instead of print is consistency. With growing number of developers working on snowcli we want to assure that if we show progress, then we show it in the same way across whole cli. 

Yes, `print` is a quick win but that's not our goal. We don't want to have a script with bunch of commands. We want to have nice cli tool.

src/snowflake/cli/api/console/__init__.py
sfc-gh-turbaszek: That I support this decision

src/snowflake/cli/api/console/console.py
sfc-gh-mraba: > Any reason why we ended up with only phase and step? It seems a bit too restrictive to only allow one/2 indentations. Is there anything specific to being in a phase or in a step except the indentation? and if not, can't we just accept indentation value as a parameter?

The goal is to establish an interface for unified output of intermediate progress. 
We want to:
- have a control over the way the output is presented to users,
- be able to easily detect cases where output could be corrupted or create side effects.

The idea behind `phase` and `step` is to was to create predefined formatted output.
`phase` is ment to be used for displaying unindented and `step` for indented output.

Indentation tracking feature brings confusion and it will be removed for now.
There is always a possibility of futher development of console output that will result in more features for displaying output.

For now we will provide:
- a `phase` method for unindented output with predefined color
- a `step` method for 2 spaces indented output with predefined color
- a generic `error` method for error messages with no indentation and predefined color

CONTRIBUTING.md
```diff
@@ -114,6 +114,35 @@ git checkout <your-branch>
...
+```python
+from snowflake.cli.api.console import cli_console
+
+cli_console.phase("foo")
+cli_console.step("bar")
+cli_console.step("bar")
+cli_console.phase("foo")
+cli_console.phase("foo")
+cli_console.step("bar")
+cli_console.error("Ops")
```
sfc-gh-turbaszek, line 136: How does it differe from raising en exception?

sfc-gh-bdufour: +1 I think the most useful feature here would be some kind of warning API instead of error (assuming we all consider error to be something you don't recover from). There are a few use cases for NADE however that would justify issuing a warning to the user, I have some listed on the original JIRA. For errors, exceptions make sense to me.

sfc-gh-mraba: Renamed `error` to `warning`

sfc-gh-turbaszek (dismissed):

src/snowflake/cli/api/console/console.py
sfc-gh-mraba: Done.

src/snowflake/cli/api/console/console.py
sfc-gh-bdufour: Thanks for this. I'm not sure I follow the mental model behind the mapping then, other than phase = not indented message, and step = indented message. There seems to be no real semantic difference between them in the sample code you wrote. That worries me a little. Would you be open to whiteboarding some ideas next week since we'll be physically in the same location?

sfc-gh-mraba: > Thanks for this. I'm not sure I follow the mental model behind the mapping then, other than phase = not indented message, and step = indented message. There seems to be no real semantic difference between them in the sample code you wrote. That worries me a little. Would you be open to whiteboarding some ideas next week since we'll be physically in the same location?

I'll be happy to discuss the ideas,
I think we could schedule something remotely this week.

src/snowflake/cli/api/console/abc.py
```diff
@@ -0,0 +1,59 @@
...
+        rich_print(text)
+
+    @contextmanager
+    @abstractmethod
+    def phase(
+        self,
+        enter_message: str,
+        exit_message: Optional[str] = None,
+    ) -> Iterator[Callable[[str], None]]:
+        """Displays not indented message."""
```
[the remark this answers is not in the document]
sfc-gh-mraba: I've updated the doc string for both abc and console

src/snowflake/cli/api/console/abc.py
```diff
@@ -0,0 +1,59 @@
...
+
+    @abstractmethod
+    def step(self, message: str):
+        """Displays indented message."""
```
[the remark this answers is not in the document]
sfc-gh-mraba: I've updated the doc string for both abc and console

src/snowflake/cli/api/console/abc.py
```diff
@@ -0,0 +1,59 @@
...
+
+    @abstractmethod
+    def warning(self, message: str):
+        """Displays message with distinct style.
```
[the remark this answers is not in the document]
sfc-gh-mraba: I've updated the doc string for both abc and console

src/snowflake/cli/api/console/console.py
```diff
@@ -0,0 +1,72 @@
...
+        text = Text(message, style=style)
+
+        if self.in_phase and output in {Output.STEP, Output.IMPORTANT}:
+            text.pad_left(self._indentation_level)
+
+        return text
+
+    @contextmanager
+    def phase(self, enter_message: str, exit_message: Optional[str] = None):
+        """Displays unindented message formatted with PHASE style."""
```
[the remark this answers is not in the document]
sfc-gh-mraba: Done

src/snowflake/cli/api/console/console.py
```diff
@@ -0,0 +1,72 @@
...
+        self._in_phase = True
+
+        yield self.step
+
+        self._in_phase = False
+        if exit_message:
+            self._print(self._format_message(exit_message, Output.PHASE))
+
+    def step(self, message: str):
+        """Displays 2 spaces indented message with STEP style."""
```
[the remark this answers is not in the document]
sfc-gh-mraba: I've updated the doc string for both abc and console

tests/api/console/test_cli_console_output.py
```diff
@@ -0,0 +1,66 @@
...
+
+def test_only_step_no_indent(cli_console, capsys):
+    cli_console.step("73")
+    assert_output_matches("73\n", capsys)
+
+
+def test_step_indented_in_phase(cli_console, capsys):
+    with cli_console.phase("42"):
+        cli_console.step("73")
+    assert_output_matches("42\n  73\n", capsys)
```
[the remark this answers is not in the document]
sfc-gh-mraba: IMO more complex ones should be handles with `syrupy` snapshots.

CONTRIBUTING.md
```diff
@@ -136,6 +136,76 @@ git checkout <your-branch>
...
+from snowflake.cli.api.console import cli_console as cc
+
+def my_command():
+    cc.step("First step...")
+    with cc.phase("Long sequence of actions"):
+        cc.step("Phased step 1")
+        cc.step("Phased step 2")
+        if something_important:
+            cc.warning("It's important")
+    cc.step("Final step")
```
sfc-gh-turbaszek, line 193: Let's add business logic to this example to show how it should relate with steps and phases. WDYT?

sfc-gh-turbaszek (approved):
LGTM! One small comment

PR #664, opened 2024-01-18
[NADE] Enable directory on stage + debug mode for versioned upgrades

sfc-gh-bgoel:
### Pre-review checklist
   * [x] I've confirmed that my changes are up-to-date with the target branch -  as of Jan 18, 4:40PM EST

### Changes description
This PR makes two changes:
1. Enables directory param for the stage created where we upload files. This way our customers can explore the contents of the stage when they run `snow app` commands, and they do not need an extra step of enabling it.
2. Enables debug mode for versioned installs/upgrades, to keep it consistent across all development mode instances, not just loose files.

### DIFF
--- a/src/snowflake/cli/plugins/nativeapp/commands.py
+++ b/src/snowflake/cli/plugins/nativeapp/commands.py
@@ -160,7 +160,7 @@ def app_open(
     **options,
 ) -> CommandResult:
     """
-    Opens the (development mode) application inside of your browser,
+    Opens the application inside of your browser,
     once it has been installed in your account.
     """
     manager = NativeAppManager(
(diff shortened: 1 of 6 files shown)


src/snowcli/plugins/nativeapp/run_processor.py
```diff
@@ -341,6 +347,12 @@ def upgrade_app(
                     """
                     )
                 )
+
+                # ensure debug_mode is up-to-date
```
sfc-gh-melnacouzi, line 351: could you please elaborate why this is needed? Is there anything in between that might be reset the debug_mode flag?

sfc-gh-bgoel: Yes, a user could have run `snow app run` initially, which creates the application in debug mode. Then they play around with it in Snowsight and may accidentally or intentionally turn it off. 
Now when they upgrade, it is on the user to keep the value consistent in project definition file, and we will follow that and keep the debug mode as such.

sfc-gh-melnacouzi (dismissed):

sfc-gh-cgorrie (approved):
These changes make sense!

sfc-gh-melnacouzi (approved):

PR #668, opened 2024-01-19
Require project root path, allow custom project files

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
Commands using project definition require now specifying of project root. Users can use different than default project files. 

The implementation can be extended in future to support multiple files.

### DIFF
--- a/src/snowflake/cli/api/commands/flags.py
+++ b/src/snowflake/cli/api/commands/flags.py
@@ -1,9 +1,13 @@
 from __future__ import annotations
 
+import os
+from pathlib import Path
 from typing import Any, Callable, Optional
 
+import click
 import typer
-from snowflake.cli.api.cli_global_context import cli_context_manager
+from click.core import ParameterSource  # type: ignore
+from snowflake.cli.api.cli_global_context import cli_context, cli_context_manager
 from snowflake.cli.api.output.formats import OutputFormat
 
 DEFAULT_CONTEXT_SETTINGS = {"help_option_names": ["--help", "-h"]}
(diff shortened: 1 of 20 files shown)


src/snowflake/cli/api/commands/flags.py
```diff
@@ -218,30 +222,52 @@ def execution_identifier_argument(sf_object: str, example: str) -> typer.Argumen
     )
 
 
+def project_root_option(project_name: str):
+    def _callback(project_root: Path):
+        resolved_path = Path(project_root).resolve()
+        cli_context_manager.set_project_root(resolved_path)
+        # Change working directory
+        os.chdir(resolved_path)
```
sfc-gh-astus, line 230: Why you change directory?

src/snowflake/cli/plugins/nativeapp/init.py
```diff
@@ -189,7 +189,7 @@ def _validate_and_update_snowflake_yml(target_directory: Path, project_identifie
     )
 
     # 3. Validate the Project Definition File(s)
-    definition_manager.project_definition
+    _ = definition_manager.project_definition
```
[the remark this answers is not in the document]
sfc-gh-turbaszek: To avoid unused statement, logic has not changed.

src/snowflake/cli/api/commands/flags.py
```diff
@@ -218,30 +222,52 @@ def execution_identifier_argument(sf_object: str, example: str) -> typer.Argumen
...
         return project_definition
 
     return typer.Option(
-        None,
-        "-p",
-        "--project",
-        help=f"Path where the {project_name.replace('_', ' ').capitalize()} project resides. "
-        f"Defaults to current working directory.",
+        "snowflake.yml",
+        "-f",
```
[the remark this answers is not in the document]
sfc-gh-turbaszek: Good point, will remove it

src/snowflake/cli/api/commands/flags.py
sfc-gh-turbaszek: Not necessary, fixed

src/snowflake/cli/api/project/definition_manager.py
```diff
@@ -25,18 +24,30 @@ class DefinitionManager:
...
+        if project_root is None:
+            project_root = Path.cwd()
 
         self.project_root = project_root
-        self._project_config_paths = self._find_definition_files(self.project_root)
+
+        if project_files is None:
+            project_files = self._find_definition_files(self.project_root)
+        else:
+            project_files = [self.project_root / p for p in project_files]
```
sfc-gh-melnacouzi, line 42: As a CLI user, this might be a bit confusing. When I am entering project file in the CLI, I use autocomplete to locate the file I would like to use as a project file. This would not work if I am NOT in the project root, and I am explicitly specifying --project in the command.  example: 
```
cd project-folder/src
snow app run --project ../ --proj-file ../config/project.yml
```
To add to this, we can see usecases where you are managing multiple snow CLI projects, but you also have a set of yml files for different situations/environments in a separate folder that apply to multiple projects, and you want to switch between them. With the above approach, you cannot specify an absolute project file, specially for overrides.

PR #675, opened 2024-01-23
Skip 'checkout' step if checkout has already been added

sfc-gh-zblackwood:
### Pre-review checklist

### Changes description

Currently, if you do `snow streamlit deploy --experimental` it works fine, but if you do it a second time, it fails, because the ALTER STREAMLIT <name> CHECKOUT has already been done once. This is analogous to the normal `snow streamlit deploy`, which creates a STAGE the first time you do it, and then simply skips it in the future without raising an error.

There is an open [ticket](https://snowflakecomputing.atlassian.net/browse/STREAMLIT-5019) to make this behavior different, but it may be a while.

### DIFF
--- a/src/snowflake/cli/plugins/streamlit/manager.py
+++ b/src/snowflake/cli/plugins/streamlit/manager.py
@@ -116,7 +116,8 @@ class StreamlitManager(SqlExecutionMixin):
             try:
                 self._execute_query(f"ALTER streamlit {streamlit_name} CHECKOUT")
             except ProgrammingError as e:
-                # Handle case that CHECKOUT has already been created
+                # If an error is raised because a CHECKOUT has already occured,
+                # simply skip it and continue
                 if "Checkout already exists" not in str(e):
                     raise
                 log.info("Checkout already exists, continuing")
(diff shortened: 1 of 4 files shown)


src/snowflake/cli/plugins/streamlit/manager.py
```diff
@@ -112,7 +113,13 @@ def deploy(
                 query_warehouse=query_warehouse,
                 experimental=True,
             )
-            self._execute_query(f"ALTER streamlit {streamlit_name} CHECKOUT")
+            try:
+                self._execute_query(f"ALTER streamlit {streamlit_name} CHECKOUT")
+            except ProgrammingError as e:
+                # Handle case that CHECKOUT has already been created
+                if "Checkout already exists" not in str(e):
+                    raise
```
sfc-gh-mraba, line 121: You could consider creating dedicated Error ie. `StreamlitCheckoutMissing(ProgrammingError)`
and re-raising this error `StreamlitCheckoutMissing(e)`

It will result in more detailed context for error.


Also a test case for this behaviour would be useful.

sfc-gh-zblackwood: Sorry if this isn't very clear, but I don't think a custom exception is actually what I want in this case, because I want to:
1. Skip any "Checkout already exists" errors, because that's really not an issue that I care about when re-deploying a streamlit app, and otherwise the first `deploy` succeeds and subsequent ones fail
2. Any other errors should be raised as expected

sfc-gh-zblackwood: I switched to if/else as I think this makes it more clear now

sfc-gh-turbaszek (approved):

PR #677, opened 2024-01-24
Download required packages inside venv

sfc-gh-jsikorski:
### Pre-review checklist

### Changes description
* Introduced `PackageInstaller` class, that runs pip in newly created venv, to avoid problems with garbage in current working venv
* Moved pip output to log.debug
* Made package create output a little bit more readable


src/snowflake/cli/plugins/snowpark/package/utils.py
```diff
@@ -63,3 +66,7 @@ def prepare_app_zip(file_path: Path, temp_dir: str) -> str:
     temp_path = temp_dir + "/" + file_name
     shutil.copy(file_path, temp_path)
     return temp_path
+
+
+def get_readable_list_of_requirements(reqs: List[Requirement]):
+    return "\n".join(list(map(lambda x: x.line, reqs)))
```
sfc-gh-mraba, line 72:
```suggestion
    return "\n".join((req.line for req in reqs))
```

sfc-gh-jsikorski: Yup, way more pythonic. Thx!

src/snowflake/cli/plugins/snowpark/package_installer.py
```diff
@@ -0,0 +1,53 @@
...
+        arguments = ["-r", name] if req_type == "file" else [name]
+        with self.temp_dir:
+            try:
+                process = subprocess.run(
+                    [self.python_path, "-m", "pip", "install", "-t", ".packages/"]
+                    + arguments,
+                    capture_output=True,
+                    text=True,
+                )
+                log.debug(process.stdout)
```
sfc-gh-mraba, line 45: For log messages lazy formatting should be used 

```python
PIP_FULL_ERROR = "Full error response: %s"
log.error(PIP_FULL_ERROR, e.stderr)
```

sfc-gh-mraba (dismissed):

src/snowflake/cli/plugins/snowpark/package_installer.py
```diff
@@ -0,0 +1,53 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class PackageInstaller:
+    PIP_ERROR_MESSAGE = "Running pip install for {0} {1} returned code {2}"
+    PIP_FULL_ERROR = "Full error response: {0}"
+    VENV_ERROR_MESSAGE = "Error creating venv. Full error trace: {0}"
+
+    def __init__(self, destination: Path = Path.cwd()):
```
sfc-gh-turbaszek, line 15: Let's add context manager `with venv()` that will handled proper clean up in case of any errors. Plus it will make it easier to use it  for other purposes as well.  Maybe we can do something like
```py
with venv(...) as v:
  v.pip_install(...)
```

Later we can exted this with an option to for example execute python within this venv. WDYt?

src/snowflake/cli/plugins/snowpark/venv.py
```diff
@@ -0,0 +1,44 @@
+import logging
+import subprocess
+import venv
+from pathlib import Path
+from tempfile import TemporaryDirectory
+
+log = logging.getLogger(__name__)
+
+
+class Venv(object):
```
sfc-gh-mraba, line 14: No need to inherit from `object`

```suggestion
class Venv:
```

sfc-gh-mraba (dismissed):

src/snowflake/cli/plugins/snowpark/venv.py
sfc-gh-jsikorski: Done

sfc-gh-mraba (approved):

PR #678, opened 2024-01-24
Overriding default connection by env variables

sfc-gh-astus:
### Pre-review checklist

### Changes description
...

### DIFF
--- a/src/snowflake/cli/api/config.py
+++ b/src/snowflake/cli/api/config.py
@@ -116,6 +116,11 @@ def get_connection(connection_name: str) -> dict:
         raise MissingConfiguration(f"Connection {connection_name} is not configured")
 
 
+def get_default_connection() -> dict:
+    def_connection_name = CONFIG_MANAGER["default_connection_name"]
+    return get_connection(def_connection_name)
+
+
 def get_config_section(*path) -> dict:
     section = _find_section(*path)
     if isinstance(section, Container):
--- a/src/snowflake/cli/app/snow_connector.py
+++ b/src/snowflake/cli/app/snow_connector.py
@@ -8,13 +8,12 @@ from typing import Dict, Optional
 import click
 import snowflake.connector
 from click.exceptions import ClickException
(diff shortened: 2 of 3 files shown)


src/snowflake/cli/api/config.py
```diff
@@ -116,6 +116,11 @@ def get_connection(connection_name: str) -> dict:
         raise MissingConfiguration(f"Connection {connection_name} is not configured")
 
 
+def get_default_connection() -> dict:
+    def_connection_name = CONFIG_MANAGER["default_connection_name"]
```
sfc-gh-turbaszek, line 120: Is this error-safe?

sfc-gh-turbaszek (approved):
One question but otherwise LGTM 🚀

src/snowflake/cli/api/config.py
sfc-gh-astus: Yes, I copied it from python connector

```
    def _get_default_connection_params() -> dict[str, Any]:
    def_connection_name = CONFIG_MANAGER["default_connection_name"]
```

```
CONFIG_MANAGER.add_option(
    name="default_connection_name",
    default="default",
)
```

src/snowflake/cli/app/snow_connector.py
```diff
@@ -26,11 +25,10 @@
 def connect_to_snowflake(temporary_connection: bool = False, connection_name: Optional[str] = None, **overrides) -> SnowflakeConnection:  # type: ignore
 
     if not temporary_connection:
-        connection_parameters = (
-            get_connection(connection_name)
-            if connection_name is not None
-            else _get_default_connection_params()
-        )
+        if connection_name is not None:
+            connection_parameters = get_connection(connection_name)
```
sfc-gh-mraba, line 29: It might be a little confusing, hence `get_connection` does not return connection nor session but connection parameters.

sfc-gh-mraba (approved):

PR #680, opened 2024-01-25
Add json output streaming in case of MultipleResults output

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
...

### DIFF
--- a/src/snowflake/cli/app/printing.py
+++ b/src/snowflake/cli/app/printing.py
@@ -1,11 +1,13 @@
 from __future__ import annotations
 
+import json
 import sys
 from datetime import datetime
 from json import JSONEncoder
 from pathlib import Path
 
-from rich import box, get_console, print
+from rich import box, get_console
+from rich import print as rich_print
 from rich.live import Live
 from rich.table import Table
 from snowflake.cli.api.cli_global_context import cli_context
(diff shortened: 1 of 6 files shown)


tests_integration/__snapshots__/test_sql.ambr
```diff
@@ -0,0 +1,16 @@
...
+      {
+          "1": 1
+      }
+  ],
+  [
+      {
+          "2": 2
+      }
+  ]
+  '''
```
sfc-gh-turbaszek, line 15:
```
➜ snow sql -c int -q "select 1; select 2; select foo; select 4" --format=json          
[
[
    {
        "1": 1
    }
],
[
    {
        "2": 2
    }
]

000904 (42000): SQL compilation error: error line 1 at position 7
invalid identifier 'FOO'
```

This is the actual output, for some reason the error is redacted when using output from runner. It is present as `result.exception`

src/snowflake/cli/app/printing.py
```diff
@@ -53,13 +55,13 @@ def _get_table():
...
-        print(obj.query)
+        rich_print(obj.query)
     items = obj.result
     try:
         first_item = next(items)
     except StopIteration:
-        print(NO_ITEMS_FOUND)
-        print()
+        rich_print(NO_ITEMS_FOUND)
+        rich_print()
```
sfc-gh-mraba, line 64:
```suggestion
        rich_print(NO_ITEMS_FOUND, end="\n\n")
```

sfc-gh-mraba (dismissed):

src/snowflake/cli/app/printing.py
```diff
@@ -78,19 +80,32 @@ def is_structured_format(output_format):
...
-    return json.dump(result, sys.stdout, cls=CustomJSONEncoder, indent=4)
+def _stream_json(result):
+    """Simple helper for streaming multiple results as a JSON."""
+    print("[")
+    results = result.result
+    res = next(results, None)
+    while res:
+        json.dump(res, sys.stdout, cls=CustomJSONEncoder, indent=4)
+        if res := next(results, None):
+            print(",")
```
sfc-gh-mraba, line 97: Shouldn't the content of `[]` also be intended ?

sfc-gh-mraba (approved):

PR #685, opened 2024-01-26
Import git repo only if it is needed in native apps

sfc-gh-astus:
### Pre-review checklist

### Changes description
...

### DIFF
--- a/src/snowcli/cli/nativeapp/init.py
+++ b/src/snowcli/cli/nativeapp/init.py
@@ -7,7 +7,6 @@ from pathlib import Path
 from tempfile import TemporaryDirectory
 from click.exceptions import ClickException
 from shutil import move, rmtree
-from git import Repo
 from strictyaml import load, as_document
 
 
@@ -210,6 +209,8 @@ def _init_with_url_and_no_template(
         target_directory.mkdir(parents=True, exist_ok=False)
 
         # Clone the repository with options.
+        from git import Repo
+
         Repo.clone_from(
             url=git_url,
             to_path=target_directory,
@@ -258,6 +259,8 @@ def _init_with_url_and_template(
     try:
         with TemporaryDirectory(dir=current_working_directory) as temp_dir:
             # Clone the repository in the temporary directory with options.
+            from git import Repo
+
             Repo.clone_from(
                 url=git_url,
                 to_path=temp_dir,
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -74,8 +74,9 @@ select = [
 ]
 
 [tool.pytest.ini_options]
-addopts = "-m 'not integration and not performance'"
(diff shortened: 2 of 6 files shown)


src/snowcli/cli/nativeapp/init.py
```diff
@@ -7,7 +7,6 @@
 from tempfile import TemporaryDirectory
 from click.exceptions import ClickException
 from shutil import move, rmtree
-from git import Repo
```
sfc-gh-pjob, line 10: We have similar imports on main branch in `version_processor.py` in native apps code.

I added `print("xxxx")` below these imports in `version_processor.py` and `xxxx` was printed when I executed `snow sql` command, so I think `git` module will be always imported on every command execution.

sfc-gh-astus: That is ok, but on 1.2.4 not. There should not be native apps :)

sfc-gh-pjob: IMO it's not ok also on `main` because `git` should not be imported if not needed (for example when running `snow sql`).

tests/test_cli.py
```diff
@@ -63,3 +64,12 @@ def check_options_for_duplicates(options: Dict[str, List[str]]) -> Set[str]:
     return set(
         [flag for flag in flags if (flags.count(flag) > 1 or flag in RESERVED_FLAGS)]
     )
+
+
+def test_loaded_modules(runner):
```
sfc-gh-pjob, line 69: This test won't work with random test execution order. IMO you have to execute this test in a separate python execution.

The following fails:
```
pytest tests/nativeapp/test_init.py tests/test_cli.py
```

tests/test_cli.py
```diff
@@ -63,3 +64,12 @@ def check_options_for_duplicates(options: Dict[str, List[str]]) -> Set[str]:
...
+    should_not_load = {"git"}
+
+    runner.invoke(["--help"])
```
sfc-gh-pjob, line 72: I would execute command like "sql" because `--help` in interpreted on very early step of command processing and CLI exits and because of that we might overlook some unwanted imports.

sfc-gh-astus: Changed to sql

tests/test_cli.py
sfc-gh-astus: Run as separate command in tox

sfc-gh-davwang:
Just wondering, why is it important to not import git.Repo unless necessary? Is it a large module that would have performance impact to import? Or is it so that users don't need to have git installed unless they plan to initialize a native app from a git repository?

sfc-gh-astus:
GitPython required installed git, we will have to change imports in 2.0

sfc-gh-pjob (approved):

PR #700, opened 2024-01-29
SNOW-1019480-sql-filter-nonsql-statements: strip comments from stream…

sfc-gh-mraba:
…d queries

### Pre-review checklist

### Changes description

Enabled removal of comments when using `split_statements`

### DIFF
--- a/tests_integration/test_sql.py
+++ b/tests_integration/test_sql.py
@@ -109,5 +109,5 @@ def test_trailing_comments_queries(runner):
     result = runner.invoke_with_connection_json(["sql", "-q", trailin_comment_query])
     assert result.exit_code == 0
     assert result.json == [
-        ["1"],
+        {"1": 1},
     ]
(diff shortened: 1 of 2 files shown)


tests_integration/test_sql.py
```diff
@@ -100,3 +100,15 @@ def _log(
...
+@pytest.mark.integration
+def test_trailing_comments_queries(runner, snowflake_session, test_root_path):
+    trailin_comment_query = "select 1;\n\n-- trailing comment\n"
+    result = runner.invoke_with_connection_json(["sql", "-q", trailin_comment_query])
+    assert result.exit_code == 0
+    assert result.json == [
+        [
+            {"1": 1},
+        ],
+    ]
```
sfc-gh-astus, line 114: Do we need integration test for it? We could check what query we passed it in unit test

sfc-gh-mraba: With unit test we mock the connection and response. 
Doing so will not test for server side error with invalid input.

sfc-gh-turbaszek (approved):

PR #708, opened 2024-01-31
SNOW-1011766: Added 'snow spcs image-repository url <repo_name>' command

sfc-gh-davwang:
### Pre-review checklist

### Changes description
Added a new convenience command of `snow spcs image-repository url` that returns the full URL given a repository name. Note that image repositories do not allow for quoted identifiers, so I have added checks that repository name inputs must be unquoted identifiers.

**Example Usage**
get_repo_url

### DIFF
--- a/src/snowflake/cli/api/project/util.py
+++ b/src/snowflake/cli/api/project/util.py
@@ -145,6 +145,13 @@ def validate_version(version: str):
         )
 
 
-def escape_like_pattern(pattern: str) -> str:
-    pattern = pattern.replace("%", r"\\%").replace("_", r"\\_")
+def escape_like_pattern(pattern: str, escape_sequence: str = r"\\") -> str:
+    """
+    When used with LIKE in Snowflake, '%' and '_' are wildcard characters and must be escaped to be used literally.
+    The escape character is '\\' when used in SHOW LIKE and must be specified when used with string matching using the
+    following syntax: <subject> LIKE <pattern> [ ESCAPE <escape> ].
+    """
+    pattern = pattern.replace("%", rf"{escape_sequence}%").replace(
+        "_", rf"{escape_sequence}_"
+    )
     return pattern
--- a/src/snowflake/cli/api/project/util.py
+++ b/src/snowflake/cli/api/project/util.py
@@ -146,5 +146,5 @@ def validate_version(version: str):
 
 
 def escape_like_pattern(pattern: str) -> str:
-    pattern = pattern.replace("%", r"\\}%").replace("_", r"\\_")
+    pattern = pattern.replace("%", r"\\%").replace("_", r"\\_")
     return pattern
--- a/src/snowflake/cli/api/project/util.py
+++ b/src/snowflake/cli/api/project/util.py
@@ -143,3 +143,8 @@ def validate_version(version: str):
         raise ValueError(
             f"Project definition version {version} is not supported by this version of Snowflake CLI. Supported versions: {SUPPORTED_VERSIONS}"
         )
+
+
+def escape_like_pattern(pattern: str) -> str:
+    pattern = pattern.replace("%", r"\\}%").replace("_", r"\\_")
+    return pattern
--- a/src/snowflake/cli/api/sql_execution.py
+++ b/src/snowflake/cli/api/sql_execution.py
@@ -72,9 +72,9 @@ class SqlExecutionMixin:
             if is_different_role:
                 self._execute_query(f"use role {prev_role}")
 
-    def _execute_schema_query(self, query: str):
+    def _execute_schema_query(self, query: str, **kwargs):
         self.check_database_and_schema()
-        return self._execute_query(query)
+        return self._execute_query(query, **kwargs)
 
     def check_database_and_schema(self) -> None:
         database = self._conn.database
--- a/src/snowflake/cli/plugins/spcs/image_repository/manager.py
+++ b/src/snowflake/cli/plugins/spcs/image_repository/manager.py
@@ -50,13 +50,16 @@ class ImageRepositoryManager(SqlExecutionMixin):
             )
         return results[0]
 
-    def get_repository_url(self, repo_name: str):
+    def get_repository_url(self, repo_name: str, with_scheme: bool = True):
         if not is_valid_unquoted_identifier(repo_name):
             raise ValueError(
                 f"repo_name '{repo_name}' is not a valid unquoted Snowflake identifier"
             )
         repo_row = self.get_repository_row(repo_name)
-        return f"https://{repo_row['repository_url']}"
+        if with_scheme:
+            return f"https://{repo_row['repository_url']}"
+        else:
+            return repo_row["repository_url"]
 
     def get_repository_api_url(self, repo_url):
         """
@@ -70,12 +73,3 @@ class ImageRepositoryManager(SqlExecutionMixin):
         path = parsed_url.path
 
         return f"{scheme}://{host}/v2{path}"
-
-    def _remove_scheme(self, url: str) -> str:
-        if not urlparse(url).scheme and not url.startswith("//"):
-            url = f"//{url}"
-        parsed_url = urlparse(url)
-        return f"{parsed_url.netloc}{parsed_url.path}"
-
-    def get_repository_url_strip_scheme(self, repo_name):
-        return self._remove_scheme(self.get_repository_url(repo_name))
--- a/src/snowflake/cli/plugins/spcs/image_repository/manager.py
+++ b/src/snowflake/cli/plugins/spcs/image_repository/manager.py
@@ -42,7 +42,7 @@ class ImageRepositoryManager(SqlExecutionMixin):
         colored_repo_name = click.style(f"'{repo_name}'", fg="green")
         if len(results) == 0:
             raise ClickException(
-                f"Image repository {colored_repo_name} does not exist or not authorized."
+                f"Image repository {colored_repo_name} does not exist in database {self.get_database()} and schema {self.get_schema()} or not authorized."
             )
         elif len(results) > 1:
             raise ClickException(
--- a/src/snowflake/cli/plugins/spcs/image_repository/manager.py
+++ b/src/snowflake/cli/plugins/spcs/image_repository/manager.py
@@ -1,7 +1,6 @@
 from typing import Dict
 from urllib.parse import urlparse
 
-from click import ClickException
 from snowflake.cli.api.project.util import (
     escape_like_pattern,
     is_valid_identifier,
@@ -28,10 +27,13 @@ class ImageRepositoryManager(SqlExecutionMixin):
             )
         database = self.get_database()
         schema = self.get_schema()
-        name_is_quoted = is_valid_quoted_identifier(repo_name)
 
-        # if repo_name is a quoted identifier, strip quotes before using as pattern
-        repo_name_raw = repo_name[1:-1] if name_is_quoted else repo_name.upper()
+        # Unquoted identifiers are resolved as all upper case in Snowflake while quoted identifiers are case-sensitive
+        repo_name_raw = (
+            repo_name[1:-1]
+            if is_valid_quoted_identifier(repo_name)
+            else repo_name.upper()
+        )
         repository_list_query = (
             f"show image repositories like '{escape_like_pattern(repo_name_raw)}'"
         )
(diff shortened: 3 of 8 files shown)


tests/spcs/test_image_repository.py
```diff
@@ -131,3 +131,43 @@ def test_list_tags(
...
+@mock.patch(
+    "snowflake.cli.plugins.spcs.image_repository.commands.ImageRepositoryManager._execute_query"
+)
+@mock.patch(
+    "snowflake.cli.plugins.spcs.image_repository.commands.ImageRepositoryManager._conn"
+)
+def test_get_repository_url_cli(mock_conn, mock_execute_query, runner, mock_cursor):
+    mock_conn.database = "DB"
+    mock_conn.schema = "SCHEMA"
+    mock_conn.role = "MY_ROLE"
```
sfc-gh-astus, line 145: Are database, schema and role needed?

tests_integration/spcs/test_image_repository.py
```diff
@@ -36,7 +38,6 @@ def _list_images(runner):
     )
 
 
-@pytest.mark.integration
```
sfc-gh-astus, line 39: Why you removed mark? Now it will be executed as unit test

sfc-gh-davwang: This mark is actually a typo that should have been removed previously. If you expand the rest of this integration test, you'll see that _list_images and _list_tags are run as part of the `test_list_images_tags` test which is marked with pytest.mark.integration

tests/spcs/test_image_repository.py
sfc-gh-davwang: good point, they were previously needed for get_repository_url_list but are no longer necessary

src/snowflake/cli/api/project/util.py
```diff
@@ -143,3 +143,8 @@ def validate_version(version: str):
         raise ValueError(
             f"Project definition version {version} is not supported by this version of Snowflake CLI. Supported versions: {SUPPORTED_VERSIONS}"
         )
+
+
+def escape_like_pattern(pattern: str) -> str:
+    pattern = pattern.replace("%", r"\\}%").replace("_", r"\\_")
```
sfc-gh-yiwang, line 149: I am curious why do we need this escape? Is that expected to replace all "%" to r"\\}%"? Do we have test case about this?

sfc-gh-davwang: Good point, the escape is supposed to be "%" -> "\\%" without the "}". I've added tests so that I don't miss these typos in the future 😅. In general, the escape is needed because "%" and "_" are wildcard characters in SHOW LIKE, meaning that something like "wildcard_name" would match "wildcardAname".

src/snowflake/cli/plugins/spcs/image_repository/manager.py
```diff
@@ -15,38 +21,43 @@ def get_schema(self):
...
-            show image repositories like '{repo_name}';
-            """
-
-        return self._execute_query(registry_query)
-
-    def get_repository_url(self, repo_name):
-        database = self.get_database()
-        schema = self.get_schema()
+        # if repo_name is a quoted identifier, strip quotes before using as pattern
+        repo_name_raw = repo_name[1:-1] if name_is_quoted else repo_name.upper()
```
sfc-gh-cgorrie, line 34: We have `unquote_identifier` already that can help with this -- it doesn't change `%` or `_` though like your `escape_like_pattern` function does. By the way, why is `_` needed to be escaped?

sfc-gh-cgorrie: Perhaps we can create an `identifier_for_like_clause` utility that combines `unquote_identifier` and the `%` escaping!

src/snowflake/cli/plugins/spcs/image_repository/manager.py
```diff
@@ -15,38 +21,43 @@ def get_schema(self):
...
-        else:
-            if len(results) > 1:
-                raise Exception(
-                    f"Found more than one repositories with name {repo_name}. This is unexpected."
-                )
+        elif len(results) > 1:
+            raise Exception(
+                f"Found more than one repositories with name {repo_name}. This is unexpected."
+            )
+        return results[0]
```
sfc-gh-cgorrie, line 52: This pattern is common in Native Apps too -- suggest you take a look at `cli/plugins/nativeapp/manager.py` to see if there's anything you want to re-use

src/snowflake/cli/plugins/spcs/image_repository/manager.py
sfc-gh-davwang: when used in a LIKE pattern, _ gets interpreted as a single character wildcard

sfc-gh-davwang: Good idea to combine the two.

One confusing thing is that the LIKE in SHOW LIKE ... and regular LIKE (https://docs.snowflake.com/en/sql-reference/functions/like) behave differently. SHOW LIKE is case insensitive and LIKE is not. We may have to specify something like `identifier_for_show_like_clause`.

src/snowflake/cli/plugins/spcs/image_repository/manager.py
sfc-gh-davwang: Would you mind clarifying which pattern you are referring to? Do you mean the pattern of checking that there is exactly one result returned by a query or do you mean the error that is thrown? I looked through the native app manager and the utils but the closest thing I could find was the generic_sql_error_handler.

src/snowflake/cli/plugins/spcs/image_repository/manager.py
sfc-gh-davwang: One more note is that in this case unquote_identifier actually doesn't work as it changes the identifier fully to uppercase. Later on in the function I need the case-sensitive version of the unquoted repo_name in order to compare against a list of candidate results while `unquote_identifier` casts it to all uppercase.

For example, if our repo name is "test_REPO", there may be both TEST_REPO and "test_REPO" existing in the same schema, both of which would be returned by "SHOW IMAGE REPOSITORIES LIKE 'test_REPO'. Then, I need the case sensitivity to determine which one to actually return.

tests/spcs/test_image_repository.py
```diff
@@ -131,3 +117,20 @@ def test_list_tags(
...
+)
+def test_get_repository_url_cli(mock_conn, mock_execute_query, runner, mock_cursor):
+    mock_execute_query.return_value = mock_cursor(
+        rows=MOCK_ROWS_DICT,
+        columns=MOCK_COLUMNS,
+    )
+
+    result = runner.invoke(["spcs", "image-repository", "url", "IMAGES"])
+    assert result.exit_code == 0, result.output
+    assert result.output.strip() == MOCK_ROWS_DICT[0]["repository_url"]
```
sfc-gh-davwang, line 136: add a test if provided with a repository name that doesn't exist (can borrow no repository found error from https://github.com/Snowflake-Labs/snowcli/pull/715)

src/snowflake/cli/plugins/spcs/image_repository/manager.py
```diff
@@ -15,38 +20,46 @@ def get_schema(self):
...
+        # because SHOW LIKE uses case-insensitive matching, results may return multiple rows
+        results = [r for r in results if r["name"] == repo_name_raw]
 
-        results = result_set.fetchall()
         if len(results) == 0:
-            raise ClickException(
+            raise ValueError(
                 f"Specified repository name {repo_name} not found in database {database} and schema {schema}"
             )
-        else:
```
sfc-gh-davwang, line 43: Copy SQL error format: Image Repository 'LADADA' does not exist or not authorized.

src/snowflake/cli/plugins/spcs/image_repository/manager.py
sfc-gh-cgorrie: The overall pattern of "I've got an identifier that I want to pipe through a `show ... like` construct and get the exact row that corresponds to my identifier". It feels like it should be a universal helper that everything calls.

src/snowflake/cli/plugins/spcs/image_repository/manager.py
```diff
@@ -15,38 +21,42 @@ def get_schema(self):
...
 
-    def get_repository_url(self, repo_name):
-        database = self.get_database()
-        schema = self.get_schema()
+        repo_name = repo_name.upper()
 
-        result_set = self.get_repository_url_list(repo_name=repo_name)
+        # because image repositories only support unquoted identifiers, SHOW LIKE should only return one or zero rows
+        repository_list_query = (
+            f"show image repositories like '{escape_like_pattern(repo_name)}'"
```
sfc-gh-yiwang, line 34: > If you use the backslash as an escape character, then you must specify escape the backslash in the ESCAPE clause.

https://docs.snowflake.com/en/sql-reference/functions/like#usage-notes Looks like we need to add `ESCAPE '\\';`

src/snowflake/cli/api/project/util.py
```diff
@@ -143,3 +143,8 @@ def validate_version(version: str):
         raise ValueError(
             f"Project definition version {version} is not supported by this version of Snowflake CLI. Supported versions: {SUPPORTED_VERSIONS}"
         )
+
+
+def escape_like_pattern(pattern: str) -> str:
```
sfc-gh-yiwang, line 148: Comments of the reason to escape like pattern would be appreciated.

src/snowflake/cli/plugins/spcs/image_repository/manager.py
```diff
@@ -60,3 +70,12 @@ def get_repository_api_url(self, repo_url):
...
 
         return f"{scheme}://{host}/v2{path}"
+
+    def _remove_scheme(self, url: str) -> str:
+        if not urlparse(url).scheme and not url.startswith("//"):
+            url = f"//{url}"
+        parsed_url = urlparse(url)
+        return f"{parsed_url.netloc}{parsed_url.path}"
+
+    def get_repository_url_strip_scheme(self, repo_name):
```
sfc-gh-yiwang, line 80: I am not sure what does strip_schema mean. Would you mind adding some comments of reasons and input url, output url examples? Why don't we use the `get_repository_row` result directly?

sfc-gh-yiwang (approved):
Overall LGTM. Left few minor comments

src/snowflake/cli/plugins/spcs/image_repository/manager.py
sfc-gh-davwang: I've implemented a mixin that can be added to any class inheriting SqlExecutionMixin in `nativeapp/manager` let me know if that's what you had in mind.

src/snowflake/cli/plugins/spcs/image_repository/manager.py
sfc-gh-davwang: it's a weird inconsistency but that only applies to LIKE with string matching. SHOW ... LIKE ... uses a different semantic with \\ as an escape character by default and also case-insensitive while LIKE is case-sensitive

https://docs.snowflake.com/en/sql-reference/sql/show-image-repositories

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -98,7 +99,30 @@ def process(self, *args, **kwargs):
         pass
 
 
-class NativeAppManager(SqlExecutionMixin):
+class FindObjectRowMixin:
+    def __init__(self):
+        if not isinstance(self, SqlExecutionMixin):
+            raise TypeError(
+                "FindObjectRowMixin can only be used with classes that also mix in SqlExecutionMixin."
```
sfc-gh-davwang, line 106: I will say I'm not entirely sure the best place to put this mixin. We could just include the implemented function (find_row_by_object_name) in SqlExecutionMixin itself as well though I'm not sure if the use cases justify it.

src/snowflake/cli/api/sql_execution.py
```diff
@@ -72,9 +72,9 @@ def use_role(self, new_role: str):
             if is_different_role:
                 self._execute_query(f"use role {prev_role}")
 
-    def _execute_schema_query(self, query: str):
+    def _execute_schema_query(self, query: str, **kwargs):
         self.check_database_and_schema()
-        return self._execute_query(query)
+        return self._execute_query(query, **kwargs)
```
sfc-gh-davwang, line 78: allows you to pass cursor_class with _execute_schema_query

src/snowflake/cli/plugins/spcs/image_repository/manager.py
```diff
@@ -15,38 +21,45 @@ def get_schema(self):
...
+        repository_list_query = (
+            f"show image repositories like '{escape_like_pattern(repo_name)}'"
+        )
 
+        result_set = self._execute_schema_query(
+            repository_list_query, cursor_class=DictCursor
+        )
         results = result_set.fetchall()
+
+        colored_repo_name = click.style(f"'{repo_name}'", fg="green")
```
sfc-gh-astus, line 42: Is this color needed? It only cause problems with equals check

sfc-gh-davwang: I think it has marginal improvement for UI but testability is probably more important. Good point.

sfc-gh-astus (approved):

PR #720, opened 2024-02-01
Snow 1017573 connections toml tests

sfc-gh-pczajka:
### Pre-review checklist
   
   ---
   * [x] I've created `connection.toml` in the default location and run tests manually 

### Changes description
* Add (ugly) fixture overriding default `connections.toml` location, so existence of connections.toml will not cause unittests and integration tests to fail
* Add unit tests checking that connectins.toml override connections config from config.toml

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,8 +4,11 @@ import logging
 
 import pytest
 from snowflake.cli.api.cli_global_context import cli_context_manager
-from snowflake.cli.api.config import config_init
+from snowflake.cli.api.config import config_init, CONFIG_MANAGER, CONNECTIONS_SECTION
 from snowflake.cli.app import loggers
+
+from snowflake.connector.config_manager import ConfigSlice
+from pathlib import Path
 
 pytest_plugins = ["tests.testing_utils.fixtures", "tests.project.fixtures"]
 
@@ -34,6 +37,26 @@ def reset_global_context_and_setup_config_and_logging_levels(
 def clean_logging_handlers_fixture(request):
     yield
     clean_logging_handlers()
+
+
+# This automatically used fixture changes location in which
+# ConfigManager looks up connections.toml file to non-existing location.
+# This causes it to behave as the file was not provided
+# Reason: if connections.toml is found, it automatically overrides "connections"
+# section in config.toml, which causes tests to fail.
+@pytest.fixture(autouse=True)
+def remove_connections_toml_from_config():
+    # HACK: as ConfigManager does not provide explicit way to
+    # modify submanagers, we need to explicitly override variable
+    # in which they're stored.
+    for i, slice in enumerate(CONFIG_MANAGER._slices):
+        if slice.section == CONNECTIONS_SECTION:
+            CONFIG_MANAGER._slices[i] = ConfigSlice(
+                path=Path("/this/file/does/not/exist"),
+                section=slice.section,
+                options=slice.options,
+            )
+    yield
 
 
 def clean_logging_handlers():
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,10 +4,10 @@ import logging
 
 import pytest
 from snowflake.cli.api.cli_global_context import cli_context_manager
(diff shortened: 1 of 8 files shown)


sfc-gh-pczajka:
e2e tests are still failing, as they're executing "independent instance" of snowcli, without possibility to override config. My ideas:
1. do nothing, as error command is suggestive ("connection 'integrations' is not configured") [current state], plus we need to set up connections for tests to work in the first place
2. create a fixture, which changes `connections.toml` to `connections.toml.backup` for duration of the tests (and reverts the change after)
3. create a hidden flag (`--override-connections-toml-location-for-testing`), which executes the same code that is used for testing to (uglily) override `connections.toml` location, and pass it to e2e tests

WDYT?

sfc-gh-pczajka (commented):
e2e tests are still failing, as they're executing "independent instance" of snowcli, without possibility to override config. My ideas:

1. do nothing, as error command is suggestive ("connection 'integrations' is not configured") [current state], plus we need to set up connections for tests to work in the first place
2. create a fixture, which changes connections.toml to connections.toml.backup for duration of the tests (and reverts the change after)
3. create a hidden flag (--override-connections-toml-location-for-testing), which executes the same code that is used for testing to (uglily) override connections.toml location, and pass it to e2e tests

WDYT?

tests/conftest.py
```diff
@@ -36,6 +39,17 @@ def clean_logging_handlers_fixture(request):
...
 
+# This automatically used fixture changes location in which
+# ConfigManager looks up connections.toml file to non-existing location.
+# This causes it to behave as the file was not provided
+# Reason: if connections.toml is found, it automatically overrides "connections"
+# section in config.toml, which causes tests to fail.
+@pytest.fixture(autouse=True)
+def remove_connections_toml_from_config():
+    change_connections_toml_path_in_config_manager(Path("this/path/does/not/exist"))
+    yield
```
sfc-gh-turbaszek, line 50: Can't we instead of that set SNOWFLAKE_HOME to temporary/dir before tests (loading snowcli)? For test purposes we can have a fixture like `change_snowflake_home` that would allow us to switching between different config/connection configurations? WDYT?

sfc-gh-pczajka: I was thinking about that, but unfortunately `CONFIG_MANAGER` is created with connector.constants (including snowflake home), which are determined during imports (which happens even before pytest manages to start up, at least according to multiple prints I've added here and there). We would have to set up `SNOWFLAKE_HOME` environment variable to some temporary directory before pytest command is run.
Fixtures happen too late for this to work :(
(PS I've tried mocking stuff as well, no success)

tests/test_config.py
```diff
@@ -168,3 +167,20 @@ def test_not_found_default_connection_from_evn_variable(test_root_path):
...
 
     assert ex.value.message == "Connection not_existed_connection is not configured"
+
+
+def test_connections_toml_override_config_toml(test_snowcli_config, snowflake_home):
+    from snowflake.cli.api.config import CONFIG_MANAGER
+
+    connections_toml = snowflake_home / "connections.toml"
+    connections_toml.write_text(
+        """[default]
```
sfc-gh-turbaszek, line 177:
```suggestion
        """[something_other_than_default]
```
My personal preference is to avoid something that can be default 🙈

tests/testing_utils/fixtures.py
```diff
@@ -237,3 +239,22 @@ def _temporary_project_directory(
...
+    snowflake_home = Path(temp_dir) / ".snowflake"
+    snowflake_home.mkdir()
+    monkeypatch.setenv("SNOWFLAKE_HOME", str(snowflake_home))
+
+    for module in [
+        sys.modules["snowflake.connector.constants"],
+        sys.modules["snowflake.connector.config_manager"],
+        sys.modules["snowflake.cli.api.config"],
+    ]:
+        importlib.reload(module)
```
sfc-gh-turbaszek, line 258: Would it be enough to reload `snowflake`?

sfc-gh-turbaszek (approved):

PR #731, opened 2024-02-05
Add SnowTyper together with telemetry

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
To add telemetry we need place to place a callback either before or post execute. This PR adds `SnowTyper` that implements this idea. This PR applies the new typer instance to `connection` and `sql` commands. In follow up PRs we should adjust remaining command.


src/snowflake/cli/app/telemetry.py
```diff
@@ -0,0 +1,119 @@
...
+    def _telemetry(self):
+        return self._ctx.connection._telemetry  # noqa
+
+    def send(self, payload: TelemetryDict):
+        if self._telemetry:
+            message = self.generate_telemetry_data_dict(payload)
+            telemetry_data = TelemetryData.from_telemetry_data_dict(
+                from_dict=message, timestamp=get_time_millis()
+            )
+            print_json(data=telemetry_data.to_dict())
```
sfc-gh-astus, line 102: Some debug leftover?

src/snowflake/cli/app/main_typer.py
```diff
@@ -1,9 +1,14 @@
...
+import logging
 import sys
 
+import typer
 from rich import print as rich_print
 from snowflake.cli.api.cli_global_context import cli_context
 from snowflake.cli.api.commands.flags import DEFAULT_CONTEXT_SETTINGS, DebugOption
-from typer import Typer
+
+log = logging.getLogger(__name__)
```
sfc-gh-pjob, line 11: unused?

src/snowflake/cli/api/utils/error_handling.py
```diff
@@ -0,0 +1,9 @@
+from contextlib import contextmanager
+
+
+@contextmanager
+def safe():
```
sfc-gh-pjob, line 5:
```suggestion
def ignore_exceptions():
```

src/snowflake/cli/app/telemetry.py
```diff
@@ -0,0 +1,117 @@
...
+    COMMAND_FLAGS = "command_flags"
+    COMMAND_OUTPUT_TYPE = "command_output_type"
+    # Information
+    EVENT = "event"
+    ERROR_MSG = "error_msg"
+    ERROR_TYPE = "error_type"
+
+
+class TelemetryEvent(Enum):
+    CMD_EXECUTION = "executing_command"
```
sfc-gh-pjob, line 39: Just asking: is this event type ok without mentioning that the event is about "snowcli command"? I know that we have `source: snowcli` but I just wonder if event type should be somehow connected to SnowCLI.

tests/test_connection.py
```diff
@@ -235,7 +235,7 @@ def test_connection_test(mock_connect, runner):
     assert "Host" in result.output
     assert "Password" not in result.output
     assert "password" not in result.output
-    mock_connect.assert_called_once_with(connection_name="full")
+    mock_connect.assert_called_with(connection_name="full")
```
sfc-gh-pjob, line 238: why did you changed it? was it required? did old assertion stop working?

tests/test_snow_typer.py
```diff
@@ -0,0 +1,140 @@
...
+def app_factory(
+    pre_execute=None,
+    result_handler=None,
+    exception_handler=None,
+    post_execute=None,
+):
+    class _CustomTyper(SnowTyper):
+        @staticmethod
+        def pre_execute_callback():
+            if pre_execute:
```
sfc-gh-pjob, line 22: shouldn't all these implementations call `super`? isn't it the preferred way to use `SnowTyper` when adding own callbacks?

Currently you test callbacks invocations but there is no test confirming that callbacks already defined in `SnowTyper` work.

tests/test_connection.py
sfc-gh-turbaszek: There are two calls instead of one because of sending telemetry. This is to be refactored as we discussed some time ago (explicit app module usage in plugins)

src/snowflake/cli/app/telemetry.py
sfc-gh-turbaszek: I took a look at the other implementations (snowpark and connecotr) and there's no strict relation.

tests/test_snow_typer.py
sfc-gh-turbaszek: Good point. I will add more tests for specific `SnowTyper` implementation.

src/snowflake/cli/api/commands/snow_typer.py
```diff
@@ -0,0 +1,99 @@
...
+                try:
+                    result = command_callable(*args, **kw)
+                    return self.process_result(result)
+                except Exception as err:
+                    self.exception_execute_callback(err)
+                    raise
+                finally:
+                    self.post_execute_callback()
+
+            return cls_super.command(name=name, **kwargs)(command_callable_decorator)
```
sfc-gh-mraba, line 64: Is there any reason not to invoke `super` in regular way?

```suggestion
            return super(SnowTyper, self).command(name=name, **kwargs)(command_callable_decorator)
```

src/snowflake/cli/plugins/connection/commands.py
```diff
@@ -210,9 +205,7 @@ def add(
     )
 
 
-@app.command()
-@global_options
-@with_output
+@app.command(requires_connection=False)
```
sfc-gh-mraba, line 208: Isn't `requires_connection` false by default?

src/snowflake/cli/api/commands/snow_typer.py
sfc-gh-turbaszek: Good one! I totally forgot about this syntax

src/snowflake/cli/api/commands/snow_typer.py
```diff
@@ -0,0 +1,99 @@
...
+
+        return custom_command
+
+    @staticmethod
+    def pre_execute_callback():
```
sfc-gh-mraba, line 70: Are these really callbacks? There are no functions passed as arguments?
Do we plan to add support later?

sfc-gh-mraba (approved):

src/snowflake/cli/plugins/connection/commands.py
sfc-gh-turbaszek: Yes, I just wanted to be explicit here

src/snowflake/cli/api/commands/snow_typer.py
sfc-gh-turbaszek: This is a good question, I'm happy to rename this to simply `pre_execute`. WDYT?

sfc-gh-mraba: Sounds good 👍  
If we do not have clarified plans to support callbacks, we could rename.

sfc-gh-mraba (approved):

sfc-gh-mraba (approved):

PR #741, opened 2024-02-06
SNOW-1011779: Changing SPCS services integration test to use persistent service instead of job.

sfc-gh-davwang:
### Pre-review checklist

### Changes description
Currently integration tests for SPCS services simply execute 'hello world' then terminate. This makes it difficult to test lifecycle methods such as `suspend` and `resume`. Furthermore, services are meant to be persistent, not single-execution (that's what jobs are for). In fact, our integration tests currently check that our service reaches a DONE state, which is not technically a supported state for services ([reference](https://docs.snowflake.com/en/sql-reference/functions/system_get_service_status?utm_source=snowscope&utm_medium=serp&utm_term=service+status)).

I propose that for our integration tests we deploy a service that echos whatever is sent to it. This is the service created in the [SPCS tutorial.](https://docs.snowflake.com/en/developer-guide/snowpark-container-services/tutorials/tutorial-1). Later on, this will also let us test functions like SHOW ENDPOINTS.

### DIFF
--- a/tests_integration/spcs/test_image_repository.py
+++ b/tests_integration/spcs/test_image_repository.py
@@ -11,7 +11,7 @@ INTEGRATION_REPOSITORY = "snowcli_repository"
 
 @pytest.mark.integration
 def test_list_images_tags(runner):
-    # test assumes the testing environment has been set up with /SNOWCLI_DB/PUBLIC/snowcli_repository/snowpark_test:1
+    # test assumes the testing environment has been set up with /SNOWCLI_DB/PUBLIC/snowcli_repository/snowpark_test_echo:1
     _list_images(runner)
     _list_tags(runner)
 
--- a/tests_integration/spcs/test_image_repository.py
+++ b/tests_integration/spcs/test_image_repository.py
@@ -33,7 +33,7 @@ def _list_images(runner):
     assert contains_row_with(
         result.json,
         {
-            "image": f"/{INTEGRATION_DATABASE}/{INTEGRATION_SCHEMA}/{INTEGRATION_REPOSITORY}/snowpark_test"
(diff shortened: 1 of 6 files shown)


tests_integration/spcs/test_image_repository.py
```diff
@@ -9,9 +9,9 @@
 INTEGRATION_REPOSITORY = "snowcli_repository"
 
 
-@pytest.mark.integration
+@pytest.mark.integration_disable
```
sfc-gh-turbaszek, line 12: Why not use `itegration` and `pytests.mark.skip` instead of custom marker?

sfc-gh-davwang: fair point, these changes were temporary but I will note for the future to use pytest.mark.skip instead

tests_integration/testing_utils/snowpark_services_utils.py
```diff
@@ -103,30 +97,51 @@ def drop_service(self, service_name: str) -> None:
...
-                status.json,
-                {
-                    "serviceName": service_name.upper(),
-                    "status": "DONE",
-                    "message": "Completed successfully",
-                },
-            ):
+            if contains_row_with(status.json, target_status):
                 return
+            elif contains_row_with(status.json, {"status": "FAILED"}):
```
sfc-gh-yiwang, line 121: We do see some transient failures happening in service creation. The service reached transient FAILED status at first time. Then K8s retried and the service was READY again. Could we keep waiting for READY status until timeout, not fail right way when getting FAILED status(this may make test flaky)?

sfc-gh-yiwang (approved):
Thanks for catching this! LGTM with a minor comment.

tests_integration/testing_utils/snowpark_services_utils.py
sfc-gh-davwang: sounds good, will make this change

sfc-gh-turbaszek (approved):

PR #744, opened 2024-02-06
Added missing plugin mechanism tests

sfc-gh-astus:
### Pre-review checklist

### Changes description
Added missing tests:
* Code from disabled plugins did not run
* Plugin don't override command


tests/test_data/configs/disabled_override_plugin_config.toml
```diff
@@ -0,0 +1,2 @@
+[connections.test]
```
sfc-gh-pczajka, line 1: Should we add the test for not-enabled plugin as well? That is:
```
[cli.plugin.something]
enabled = false
```

sfc-gh-astus: This is that case `not in config` == disabled

sfc-gh-pczajka (dismissed):

tests/plugin/test_broken_plugin.py
```diff
@@ -0,0 +1,30 @@
+from textwrap import dedent
+
+import pytest
+
+
+def test_broken_plugin(runner, test_root_path, _install_plugin, caplog):
```
sfc-gh-pjob, line 6:
```suggestion
def test_broken_command_path_plugin(runner, test_root_path, _install_plugin, caplog):
```

tests/plugin/test_broken_plugin.py
```diff
@@ -0,0 +1,30 @@
+from textwrap import dedent
+
+import pytest
+
+
```
sfc-gh-pjob, line 5: as discussed let's add test case `test_broken_implementation_plugin` which will try to load a plugin which raises some error during its module loading

tests/plugin/test_override_by_external_plugins.py
```diff
@@ -0,0 +1,55 @@
...
+        """+---------------------------------------+
+| connection_name | parameters          |
+|-----------------+---------------------|
+| test            | {'account': 'test'} |
++---------------------------------------+
+    """
+    )
+
+
+def test_disabled_plugin_is_not_executed(
```
sfc-gh-pjob, line 27: as discussed let's add one more test case confirming that no code from disabled plugin's module is executed

we need to check that not only command implementation is not executed but we need to be sure that no code is executed

tests/plugin/test_broken_plugin.py
sfc-gh-astus: Named it `failing_plugin`

sfc-gh-pjob (approved):

PR #748, opened 2024-02-07
Create a pattern for drilling down to a single "show objects like" row

sfc-gh-cgorrie:
### Pre-review checklist

### Changes description
Introduced `SqlExecutionMixin.show_specific_object` to share code between all the times we need to `show objects like` in order to get more information about a particular entity.

Original author: @sfc-gh-davwang


src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -98,7 +99,30 @@ def process(self, *args, **kwargs):
         pass
 
 
-class NativeAppManager(SqlExecutionMixin):
+class FindObjectRowMixin:
```
sfc-gh-bgoel, line 102: What is the benefit to making a new Mixin as opposed to introducing the main method `find_row_by_object_name` in `SqlExecutionMixin` ?  New managers would only need to inherit from the`SqlExecutionMixin`, and we do not need to remember this additional class. It is possible that those who forget to include this new class when defining a new manager end up creating their own such method, which could be avoided if it were already contained in `SqlExecutionMixin`

sfc-gh-davwang: FWIW, I think that if this can be included in SqlExecutionMixin then it would be most useful in there. I just wasn't initially sure whether SqlExecutionMixin is meant to be, as Cam says, "lower-level than these sort of helpers."

sfc-gh-bgoel: I see your point @sfc-gh-cgorrie and it makes sense. Will leave it up to you, happy to approve once you have decided however you want to proceed.

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -227,27 +225,20 @@ def get_existing_version_info(self, version: str) -> Optional[dict]:
...
-
             try:
-                show_obj_cursor = self._execute_query(
-                    show_obj_query, cursor_class=DictCursor
+                version_obj = self.show_specific_object(
+                    "versions",
+                    version,
+                    name_col=VERSION_COL,
+                    in_clause=f"in application package {self.package_name}",
                 )
```
sfc-gh-davwang, line 234: nit: could be shortened to `return self.show_specific_object(...)`, no need to wait until after the except block

tests/nativeapp/test_manager.py
```diff
@@ -14,7 +13,7 @@
 )
 from snowflake.cli.plugins.object.stage.diff import DiffResult
 from snowflake.cli.api.project.definition_manager import DefinitionManager
-from snowflake.connector import ProgrammingError
+from snowflake.cli.api.sql_execution import SqlExecutionMixin
 from snowflake.connector.cursor import DictCursor
```
sfc-gh-davwang, line 17: are Mock and SqlExecutionMixin imports used in this test file?

sfc-gh-davwang (approved):

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-cgorrie: You'd think so, but sadly the linter complains that there's no return (because the except block calls the generic error handler). I could add an explicit `return None` there, might feel cleaner?

sfc-gh-davwang: ah you're right about the except block, this probably works best then

sfc-gh-davwang:
if you're amenable to including it, here's a PR to use your new function in image_registry/manager: https://github.com/Snowflake-Labs/snowcli/pull/755

sfc-gh-astus (approved):

sfc-gh-bgoel (approved):
LGTM for both Project and NA

src/snowflake/cli/api/sql_execution.py
```diff
@@ -137,3 +143,27 @@ def to_fully_qualified_name(self, name: str):
...
+        show_obj_cursor = self._execute_query(  # type: ignore
+            show_obj_query, cursor_class=DictCursor
+        )
+        if show_obj_cursor.rowcount is None:
+            raise SnowflakeSQLExecutionError(show_obj_query)
+        show_obj_row = find_first_row(
+            show_obj_cursor,
+            lambda row: row[name_col] == unquote_identifier(unqualified_name),
+        )
+        return show_obj_row
```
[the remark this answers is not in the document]
sfc-gh-cgorrie: To me, that's a different layer. My mental model of layers:

- command "spec" / app layer (i.e. `commands.py`)
- command implementation layer (i.e. managers; e.g. `ObjectManager.py`)
- utility layer (e.g. `utils/*.py`, and `SqlExecutionMixin`)

Suppose we were to put it in `ObjectManager`. Other managers that want to use it would have to instantiate an `ObjectManager()`, then call this. Right now there's no args for the constructor but I'm not sure we can count on that going forward given the design (e.g. if we ever needed to support multiple concurrent connections). So this felt like the cleanest place to put it.

sfc-gh-turbaszek (approved):

PR #749, opened 2024-02-07
[NADE] Replace print statements in NA commands

sfc-gh-bgoel:
### Pre-review checklist
   * [x] I've confirmed that my changes are up-to-date with the target branch - as of Feb 7, 4:00PM EST

### Changes description
With the introduction of https://github.com/Snowflake-Labs/snowcli/pull/659 for displaying intermediate console output, NA commands need to be updated to remove existing `print` statements and instead add in `step`/`warning`/`phase` as needed.


src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -258,7 +256,7 @@ def sync_deploy_root_with_stage(self, role: str) -> DiffResult:
 
         # Does a stage already exist within the app pkg, or we need to create one?
         # Using "if not exists" should take care of either case.
-        log.info("Checking if stage exists, or creating a new one if none exists.")
+        cc.step("Checking if stage exists, or creating a new one if none exists.")
```
sfc-gh-melnacouzi, line 259: Do we need to use phases anywhere in our code? For example for application package updates vs application instance updates?

sfc-gh-melnacouzi (approved):

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-bdufour: Some of our complex workflows could use it, something like run could have a phase for the app package, and a phase for the instance, for example. It might be helpful context for a user to understand what our warnings apply to I guess.

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -271,24 +269,27 @@ def sync_deploy_root_with_stage(self, role: str) -> DiffResult:
...
         )
         diff: DiffResult = stage_diff(self.deploy_root, self.stage_fqn)
-        log.info("Listing results of diff:")
-        log.info("New files only on your local: %s", ",".join(diff.only_local))
-        log.info("New files only on the stage: %s", ",".join(diff.only_on_stage))
-        log.info(
-            "Existing files modified or status unknown: %s", ",".join(diff.different)
-        )
-        log.info("Existing files identical to the stage: %s", ",".join(diff.identical))
+        with cc.phase("Listing results of diff:"):
```
sfc-gh-bdufour, line 277: Not sure that's a phase. Conceptually, a phase should be a group of related steps. We need to be careful since phases can't be nested, so we most likely only want to create them directly within top level commands, and not in helper code. We should definitely avoid using phase/step to control indentation. Maybe the phase here is the deploy phase within run?

The console API is pretty minimal, I think we'll need to revisit it when we add `snow app status`. Until then, I'd be OK just performing any required indentation in each step directly.

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
```diff
@@ -86,7 +86,7 @@ def drop_application(self, auto_yes: bool):
                     )
                 )
                 if not should_drop_object:
-                    print(f"Did not drop application {self.app_name}.")
+                    cc.step(f"Did not drop application {self.app_name}.")
```
sfc-gh-bdufour, line 89: This is neither a step nor a warning. Maybe we need to consider adding some kind of message API for cases like this. For now it could be an alias for step, but conceptually it would keep the presentation decisions centralized in the SnowCLI core. Thoughts?

Alternately, we use warning for now.

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
```diff
@@ -141,17 +141,17 @@ def drop_package(self, auto_yes: bool):
                 needs_confirm = False
             else:
                 if needs_confirmation(needs_confirm, auto_yes):
-                    print(
+                    cc.warning(
                         f"Application package {self.package_name} was not created by SnowCLI!"
```
sfc-gh-bdufour, line 145: nit and subjective, can we remove the trailing '!' ? Feels ... odd.

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
```diff
@@ -141,17 +141,17 @@ def drop_package(self, auto_yes: bool):
...
                     )
         else:
             if needs_confirmation(needs_confirm, auto_yes):
-                print(
+                cc.warning(
```
sfc-gh-bdufour, line 149: nit: I know you didn't change it, but I'm not sure this message is clear on its own. Why is this a warning, if I'm a user?

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -271,24 +269,27 @@ def sync_deploy_root_with_stage(self, role: str) -> DiffResult:
...
+            cc.step("New files only on your local: %s" % ",".join(diff.only_local))
+            cc.step("New files only on the stage: %s" % ",".join(diff.only_on_stage))
+            cc.step(
+                "Existing files modified or status unknown: %s"
+                % ",".join(diff.different),
+            )
```
sfc-gh-bdufour, line 283: I know that will make a code a little odd, but IMHO that's a single step, i.e. nothing should be interleaved between each line. I'd prefer to assemble the output once, then print it as a unit.

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -321,11 +321,11 @@ def upgrade_app(
                     if err.errno not in UPGRADE_RESTRICTION_CODES:
                         generic_sql_error_handler(err=err)
                     else:  # The existing app was created from a different process.
-                        print(err.msg)
+                        cc.step(err.msg)
```
sfc-gh-bdufour, line 324: warning?

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -258,10 +258,10 @@ def drop_application_before_upgrade(self, policy: PolicyBase, is_interactive: bo
         )
         if not policy.should_proceed(user_prompt):
             if is_interactive:
-                print("Not upgrading the application.")
+                cc.step("Not upgrading the application.")
```
sfc-gh-bdufour, line 261: not sure that's a step...

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -258,10 +258,10 @@ def drop_application_before_upgrade(self, policy: PolicyBase, is_interactive: bo
...
                 raise typer.Exit(0)
             else:
-                print(
+                cc.step(
```
sfc-gh-bdufour, line 264: ditto, not a step. Warning maybe?

src/snowflake/cli/plugins/nativeapp/version/version_processor.py
```diff
@@ -46,18 +43,18 @@ def check_index_changes_in_git_repo(
 
         # Check if the repo has any changes, including untracked files
         if repo.is_dirty(untracked_files=True):
-            print("Changes detected in your git repository!")
+            cc.warning("Changes detected in your git repository!")
```
sfc-gh-bdufour, line 46: nit: maybe something like "Changes detected in the git repository (rerun with --skip-git-check to ignore)"? Would be nice to tell the user how to get around this check if they don't expect this. I do like how the git CLI essentially teaches you how to use it as you interact with the output.

src/snowflake/cli/plugins/nativeapp/version/version_processor.py
```diff
@@ -46,18 +43,18 @@ def check_index_changes_in_git_repo(
...
             repo.git.execute(["git", "status"])
 
             user_prompt = (
                 "You have local changes in this repository that are not part of a previous commit. Do you still want to continue?",
             )
             if not policy.should_proceed(user_prompt):
                 if is_interactive:
-                    print("Not creating a new version.")
+                    cc.step("Not creating a new version.")
```
sfc-gh-bdufour, line 54: Not sure that's a step

src/snowflake/cli/plugins/nativeapp/version/version_processor.py
```diff
@@ -46,18 +43,18 @@ def check_index_changes_in_git_repo(
...
                     raise typer.Exit(0)
                 else:
-                    print(
+                    cc.step(
```
sfc-gh-bdufour, line 57: not a step

src/snowflake/cli/plugins/nativeapp/version/version_processor.py
```diff
@@ -155,13 +152,10 @@ def process(
         # Make sure version is not None before proceeding any further.
         # This will raise an exception if version information is not found. Patch can be None.
         if not version:
-            log.info(
-                dedent(
-                    f"""\
-                        Version was not provided through the CLI. Checking version in the manifest.yml instead.
-                    """
-                )
+            cc.step(
```
sfc-gh-bdufour, line 155: another one that's not a step and not a warning.

src/snowflake/cli/plugins/nativeapp/version/version_processor.py
```diff
@@ -220,10 +214,10 @@ def process(
             )
             if not policy.should_proceed(user_prompt):
                 if is_interactive:
-                    print("Not creating a new patch.")
+                    cc.step("Not creating a new patch.")
```
sfc-gh-bdufour, line 217: ditto

src/snowflake/cli/plugins/nativeapp/version/version_processor.py
```diff
@@ -220,10 +214,10 @@ def process(
...
                     raise typer.Exit(0)
                 else:
-                    print(
+                    cc.step(
```
sfc-gh-bdufour, line 220: not a step

src/snowflake/cli/plugins/nativeapp/version/version_processor.py
```diff
@@ -301,10 +295,10 @@ def process(
         )
         if not policy.should_proceed(user_prompt):
             if is_interactive:
-                print("Not dropping version.")
+                cc.step("Not dropping version.")
                 raise typer.Exit(0)
             else:
-                print("Cannot drop version non-interactively without --force.")
+                cc.step("Cannot drop version non-interactively without --force.")
```
sfc-gh-bdufour, line 301: Not steps

tests/nativeapp/test_run_processor.py
```diff
@@ -158,7 +158,7 @@ def test_create_app_pkg_external_distribution(
     run_processor = _get_na_run_processor()
     run_processor.create_app_package()
     if not is_pkg_distribution_same:
-        mock_warning.assert_called_once_with(
+        mock_step.assert_called_once_with(
```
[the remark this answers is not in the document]
sfc-gh-bdufour: Fair

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-bgoel: I mentioned in the description the guide I am using to decide between `step` and `warning`, adding that here too:
```
How I have used each of these:

1. If there any internal checks done on the objects (such as checking its distribution property, or owner role) and a mismatch is found, then warning is used.
2. For regular progress, step is used.
3. If a user interacts with the CLI, then the expected outcome is also captured as a step.
phase is used only once where it made sense to me.
```
I believe that when a user accepts or rejects a prompt from the CLI, then what is output is just information to them because it is an expected outcome. It is not a warning that the user needs to be alarmed about. 

In general, I think we need something like `cc.info` because I do agree with the argument that not all information is a step.

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-bgoel: same as above.

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
sfc-gh-bgoel: Same as above, using `step` because the user rejected the drop and now we're just informing them that we did not drop. Though this also falls under the category of a simple information message imo, and is neither a step nor a warning (like you mentioned).

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
sfc-gh-bgoel: I can clarify the message, but the reason it is shown to the user here is because they are choosing to drop an application package with `external` distribution property, which means it was either before the CLI existed, or could be a "prod" application package that is currently associated with a listing on the marketplace.

src/snowflake/cli/plugins/nativeapp/version/version_processor.py
sfc-gh-bgoel: same as before.

src/snowflake/cli/plugins/nativeapp/version/version_processor.py
sfc-gh-bgoel: ditto.

src/snowflake/cli/plugins/nativeapp/version/version_processor.py
sfc-gh-bgoel: agreed, keeping step as a replacement to info.

src/snowflake/cli/plugins/nativeapp/version/version_processor.py
sfc-gh-bgoel: same as before.

src/snowflake/cli/plugins/nativeapp/version/version_processor.py
sfc-gh-bgoel: same as before.

src/snowflake/cli/plugins/nativeapp/version/version_processor.py
sfc-gh-bgoel: same as before.

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
sfc-gh-bdufour: Yeah I get the context, I was just suggesting we clarify the message for our users.

src/snowflake/cli/api/console/abc.py
```diff
@@ -54,6 +54,10 @@ def phase(
     def step(self, message: str):
         """Displays message to output."""
 
+    @abstractmethod
+    def info(self, message: str):
```
sfc-gh-bdufour, line 58: I think we agreed on `message`

src/snowflake/cli/api/console/console.py
```diff
@@ -69,6 +71,11 @@ def step(self, message: str):
         text = self._format_message(message, Output.STEP)
         self._print(text)
 
+    def info(self, message: str):
+        """Displays an informational messge to output."""
```
sfc-gh-bdufour: typo: message

src/snowflake/cli/api/console/console.py
```diff
@@ -10,6 +10,7 @@
 
 PHASE_STYLE: Style = Style(color="grey93", bold=True)
 STEP_STYLE: Style = Style(color="grey89", italic=True)
+INFO_STYLE: Style = Style(color="black")
```
sfc-gh-bdufour, line 13: Thought we agreed to alias step, although I'm not sure why step is italic TBH

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -271,24 +269,32 @@ def sync_deploy_root_with_stage(self, role: str) -> DiffResult:
...
         )
         diff: DiffResult = stage_diff(self.deploy_root, self.stage_fqn)
-        log.info("Listing results of diff:")
-        log.info("New files only on your local: %s", ",".join(diff.only_local))
-        log.info("New files only on the stage: %s", ",".join(diff.only_on_stage))
-        log.info(
-            "Existing files modified or status unknown: %s", ",".join(diff.different)
+        diff_results_info = dedent(
+            f"""\
+                New files only on your local: %s
```
sfc-gh-bdufour, line 279: Wouldn't one file per line be a lot easier to read?

src/snowflake/cli/api/console/console.py
sfc-gh-bgoel: I thought about it, and am not convinced that the style for step should apply to info. Hence did not go the alias route.

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -271,24 +269,27 @@ def sync_deploy_root_with_stage(self, role: str) -> DiffResult:
...
+            % self.deploy_root
         )
         diff: DiffResult = stage_diff(self.deploy_root, self.stage_fqn)
-        log.info("Listing results of diff:")
-        log.info("New files only on your local: %s", ",".join(diff.only_local))
-        log.info("New files only on the stage: %s", ",".join(diff.only_on_stage))
-        log.info(
-            "Existing files modified or status unknown: %s", ",".join(diff.different)
+        diff_results_info = (
+            "\nNew files only on your local: %s \nNew files only on the stage: %s \nExisting files modified or status unknown: %s \nExisting files identical to the stage: %s"
```
sfc-gh-bgoel, line 278: Had to manually add in `\n` and strip leading and trailing whitespaces as `dedent` is interestingly not working as expected. Tried it out as part of the tests and `dedent` is not trimming the leading whitespaces.

```
def method():
      diff_results_info = dedent("""
      New files only on your local: 
      %s
      New files only on the stage:
      %s
      Existing files modified or status unknown: 
      %s
      Existing files identical to the stage: 
      %s
      """ % (
      "\n".join(["A", "B", "C"]),
      "\n".join(["A", "B", "C"]),
      "\n".join(["A", "B", "C"]),
      "\n".join(["A", "B", "C"])))
```

gave an output
```
message('\n            New files only on your local: \n            A\nB\nC\n            New files only on the stage:\n            A\nB\nC\n            Existing files modified or status unknown: \n            A\nB\nC\n            Existing files identical to the stage: \n            A\nB\nC\n')
```

src/snowflake/cli/api/console/console.py
```diff
@@ -62,17 +64,24 @@ def phase(self, enter_message: str, exit_message: Optional[str] = None):
             self._print(self._format_message(exit_message, Output.PHASE))
 
     def step(self, message: str):
-        """Displays messge to output.
+        """Displays a message to output.
```
sfc-gh-bdufour, line 67: TY!

src/snowflake/cli/api/console/console.py
```diff
@@ -62,17 +64,24 @@ def phase(self, enter_message: str, exit_message: Optional[str] = None):
...
+
+        If called within a phase, the output will be indented."""
+        text = self._format_message(_message, Output.INFO)
+        self._print(text)
+
     def warning(self, message: str):
         """Displays message in a style that makes it visually stand out from other output.
 
-        Intended for diplaying messeges related to important messages."""
+        Intended for diplaying messages related to important messages."""
```
sfc-gh-bdufour, line 84: That sentence needs work... @sfc-gh-mraba PTAL

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -271,24 +269,27 @@ def sync_deploy_root_with_stage(self, role: str) -> DiffResult:
...
+            % (
+                "\n".join(["A", "B", "C"]),
+                "\n".join(["A", "B", "C"]),
+                "\n".join(["A", "B", "C"]),
+                "\n".join(["A", "B", "C"]),
```
sfc-gh-bdufour, line 283: ?

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-bdufour: Not following. I'd probably just create an array and keep appending lines to it. Then, we can just join + display at the very end

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-bgoel: My bad, fixing this

src/snowflake/cli/plugins/object/stage/diff.py
```diff
@@ -42,6 +42,47 @@ def has_changes(self) -> bool:
...
+        """
+        Method override for the standard behavior of string representation for this class.
+        """
+        components: List[
+            str
+        ] = (
+            []
+        )  # py3.8 does not support subscriptions for builtin list, hence using List
+
+        # The specific order of conditionals is for an aesthetically pleasing output and ease of readability.
```
sfc-gh-bgoel, line 55: + also why I did not do a simple `if-else` for each list attribute, because interleaving "There are no new files ..." with "here are the files..." is hard to read.

sfc-gh-bdufour (dismissed):

sfc-gh-mraba (dismissed):

src/snowflake/cli/api/console/console.py
```diff
@@ -64,22 +64,24 @@ def phase(self, enter_message: str, exit_message: Optional[str] = None):
...
+
+        If called within a phase, the output will be indented."""
         text = self._format_message(message, Output.INFO)
         self._print(text)
 
     def warning(self, message: str):
         """Displays message in a style that makes it visually stand out from other output.
 
-        Intended for diplaying messeges related to important messages."""
+        Intended for diplaying messages related to important messages."""
```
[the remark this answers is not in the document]
sfc-gh-bgoel: @sfc-gh-mraba thanks!

sfc-gh-melnacouzi (approved):

sfc-gh-mraba (approved):

PR #754, opened 2024-02-08
SNOW-1011767: Adding 'spcs image-registry login' command

sfc-gh-davwang:
### Pre-review checklist

### Changes description
Added `snow spcs image-registry login` command that automatically fetches your account token and registry URL and tries to login with `docker login`.

**Usage**:
login-succeeded

**Errors**:
no-repository
docker-login-error

### DIFF
--- a/src/snowflake/cli/plugins/spcs/__init__.py
+++ b/src/snowflake/cli/plugins/spcs/__init__.py
@@ -11,7 +11,7 @@ from snowflake.cli.plugins.spcs.services.commands import app as services_app
 
 app = SnowTyper(
     name="spcs",
-    help="Manages Snowpark services, pools, jobs, image registries, and image repositories.",
+    help="Manages Snowpark services, pools, image registries, and image repositories.",
 )
 
 app.add_typer(compute_pools_app)  # type: ignore
--- a/src/snowflake/cli/plugins/spcs/image_registry/commands.py
+++ b/src/snowflake/cli/plugins/spcs/image_registry/commands.py
@@ -1,11 +1,6 @@
-import subprocess
-
-import typer
-from click import ClickException
 from snowflake.cli.api.commands.snow_typer import SnowTyper
 from snowflake.cli.api.output.types import MessageResult, ObjectResult
 from snowflake.cli.plugins.spcs.image_registry.manager import (
-    NoImageRepositoriesFoundError,
     RegistryManager,
 )
 
@@ -30,22 +25,10 @@ def url(**options) -> MessageResult:
     Gets the image registry URL for the current account. Must be called from a
     role that can view at least one image repository in the image registry.
     """
-    try:
-        return MessageResult(RegistryManager().get_registry_url())
-    except NoImageRepositoriesFoundError:
-        raise ClickException(
-            "No image repository found. To get the registry url, please switch to a role with read access to at least one image repository or create a new image repository first."
-        )
+    return MessageResult(RegistryManager().get_registry_url())
 
 
 @app.command(requires_connection=True)
 def login(**options) -> MessageResult:
     """Logs in to the account image registry with the current user's credentials. Must be called from a role that can view at least one image repository in the image registry."""
-    try:
-        return MessageResult(RegistryManager().docker_registry_login().strip())
-    except NoImageRepositoriesFoundError:
-        raise ClickException(
-            "No image repository found. To login to your image registry, please switch to a role with read access to at least one image repository or create a new image repository first."
-        )
-    except subprocess.CalledProcessError as e:
-        raise ClickException(f"Login Failed: {e.stderr}".strip())
+    return MessageResult(RegistryManager().docker_registry_login().strip())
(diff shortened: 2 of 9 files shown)


src/snowflake/cli/plugins/spcs/image_registry/commands.py
```diff
@@ -39,3 +41,18 @@ def url(**options) -> MessageResult:
...
+
+
+@app.command()
+@with_output
+@global_options_with_connection
+def login(**options) -> MessageResult:
+    """Logs in to the account image registry with the current user's credentials. Must be called from a role that can view at least one image repository in the image registry."""
+    try:
+        return MessageResult(RegistryManager().docker_registry_login().strip())
+    except NoImageRepositoriesFoundError:
```
sfc-gh-astus, line 53: Maybe instead of catch and throw new error, extend ClickException by NoImageRepositoriesFoundError and add error message?

sfc-gh-davwang: I see your point, especially as that would reduce some of the code needed. 

My thought process was that we should have a separation of concerns here such that the manager class deals with the "backend" logic and throws errors related to that (data not found, not right format, can't query, etc.) which the commands definition can then catch and use to generate an informative error message to display on the frontend via ClickException.

Do you think that directly throwing a ClickException (or a subclass of ClickException) would be more efficient here?

sfc-gh-astus: In my opinion using subclass of ClickException is a better solution. Keeps everything in one place. In your approach, you raise an error, catch somewhere higher and reraise as other error. If you have few `NoImageRepositoriesFoundError` how could you know which one cause error? If you add more types of errors, list of excepts is required here.

sfc-gh-turbaszek: Using ClikcException is already commonly used in our code base

src/snowflake/cli/plugins/spcs/image_registry/manager.py
```diff
@@ -66,3 +66,20 @@ def get_registry_url(self):
...
+            "login",
+            "--username",
+            "0sessiontoken",
+            "--password-stdin",
+            registry_url,
+        ]
+        result = subprocess.run(
+            command, input=json.dumps(token), text=True, capture_output=True
+        )
+        result.check_returncode()
```
sfc-gh-turbaszek, line 84: Can we use `check_output`? Also do we need to pass the token via stdin? Can we pass it as "--password" flag?

tests/spcs/test_registry.py
```diff
@@ -1,10 +1,13 @@
 import json
+import subprocess
```
sfc-gh-turbaszek, line 2: Do we use it somewhere?

src/snowflake/cli/plugins/spcs/__init__.py
```diff
@@ -11,7 +11,7 @@
 
 app = SnowTyper(
     name="spcs",
-    help="Manages Snowpark services, pools, jobs, image registries, and image repositories.",
+    help="Manages Snowpark services, pools, image registries, and image repositories.",
```
sfc-gh-yiwang, line 14: nit: Snowpark Container Services services, compute pools, image registries, and image repositories

tests/__snapshots__/test_help_messages.ambr
```diff
@@ -1,30 +1,4 @@
 # serializer version: 1
-# name: test_help_messages[app.bundle]
-  '''
-                                                                                  
-   Usage: default app bundle [OPTIONS]
```
sfc-gh-yiwang, line 5: Is this expected change? Looks like it is not related to image-registry login

src/snowflake/cli/plugins/spcs/image_registry/manager.py
sfc-gh-davwang: I believe passing token in via stdin is better practice for keeping it out of any shell logs; Docker throws the following warning if you do pass in the password via the --password flag:
image

tests/spcs/test_registry.py
sfc-gh-davwang: I only use subprocess.CalledProcessError for some tests, I've changed the import to specifically import that instead.

src/snowflake/cli/plugins/spcs/__init__.py
sfc-gh-davwang: Noted, in the interest of keeping this PR focused, I'll make another PR to rename all other instances of Snowpark to Snowpark Container Services

tests/__snapshots__/test_help_messages.ambr
sfc-gh-davwang: Good point, I've retargeted this pull request on a separate branch that gets rid of the duplicate snapshots.

sfc-gh-turbaszek (approved):

PR #762, opened 2024-02-10
SNOW-1011780: `spcs service set/unset <properties...>` commands

sfc-gh-davwang:
### Pre-review checklist

### Changes description
Added commands for spcs compute-pool set --<property1> <value1> ... and spcs compute-pool unset <name> --<property1>  .... These commands allow you to set values or reset to default values for properties of an existing compute pool, equivalent to ALTER COMPUTE POOL <name> SET/UNSET .... Added unit and integration tests for these commands.

### DIFF
--- a/src/snowflake/cli/plugins/spcs/services/manager.py
+++ b/src/snowflake/cli/plugins/spcs/services/manager.py
@@ -117,7 +117,7 @@ class ServiceManager(SqlExecutionMixin):
         ]
         if all([value is None for property_name, value in property_pairs]):
             raise NoPropertiesProvidedError(
-                "at least one of the properties passed to 'set_property' must not be None"
+                f"No properties specified for service '{service_name}'. Please provide at least one property to set."
             )
         query: list[str] = [f"alter service {service_name} set"]
         for property_name, value in property_pairs:
@@ -143,7 +143,7 @@ class ServiceManager(SqlExecutionMixin):
         ]
         if not any([value for property_name, value in property_pairs]):
             raise NoPropertiesProvidedError(
-                "at least one of the properties passed to 'unset_property' must be True"
+                f"No properties specified for service '{service_name}'. Please provide at least one property to reset to its default value."
             )
         unset_list = [property_name for property_name, value in property_pairs if value]
         query = f"alter service {service_name} unset {','.join(unset_list)}"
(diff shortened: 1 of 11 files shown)


tests/__snapshots__/test_help_messages.ambr
```diff
@@ -3008,59 +3008,66 @@
...
+  │                                                             Integrations(EA… │
+  │                                                             that the service │
+  │                                                             can access. This │
+  │                                                             option may be    │
+  │                                                             specified        │
+  │                                                             multiple times   │
+  │                                                             for multiple     │
+  │                                                             EAIs.            │
+  │                                                             [default: None]  │
+  │    --query-wareho…                         TEXT             Warehouse to use │
```
sfc-gh-yiwang, line 3050: Is the `--query-wareho…`  expected?

src/snowflake/cli/plugins/spcs/services/manager.py
```diff
@@ -91,3 +92,53 @@ def upgrade_spec(self, service_name: str, spec_path: Path):
...
+        comment: bool,
+    ):
+        property_pairs = [
+            ("min_instances", min_instances),
+            ("max_instances", max_instances),
+            ("query_warehouse", query_warehouse),
+            ("auto_resume", auto_resume),
+            ("comment", comment),
+        ]
+        if not any([value for property_name, value in property_pairs]):
```
sfc-gh-yiwang, line 138: It's a bit tricky. Could you add a comment to indicate if the property is not specified, the value will be false?

tests/spcs/test_services.py
```diff
@@ -327,3 +328,177 @@ def test_list_endpoints_cli(mock_list_endpoints, mock_cursor, runner):
...
+        == f"No properties specified for service '{service_name}'. Please provide at least one property to reset to its default value."
+    )
+
+
+@patch("snowflake.cli.plugins.spcs.services.manager.ServiceManager.unset_property")
+def test_unset_property_cli(mock_unset, mock_statement_success, runner):
+    cursor = mock_statement_success()
+    mock_unset.return_value = cursor
+    service_name = "test_service"
+    result = runner.invoke(
```
sfc-gh-yiwang, line 463: Do we have a way to disallow customer setting property value in unset command?
Could we add a test to verify `["spcs", "service", "unset", service_name, "--min-instances", 1]` is not allowed?

tests/__snapshots__/test_help_messages.ambr
sfc-gh-davwang: yep in console works fine, I believe the test runner will just truncate names that are too long based on its window
image

sfc-gh-turbaszek (approved):

Issue #763, opened 2024-02-12
SNOW-1053803: Second-chance Anaconda run suffers from version lag relative to PyPi


jamesweakley:
### Description

Currently [in snowpark_shared.py](https://github.com/Snowflake-Labs/snowcli/blob/main/src/snowflake/cli/plugins/snowpark/snowpark_shared.py#L76), we do a second `pip install` of the `requirements.other.txt` file and look for packages included as dependencies that can be satisfied by Anaconda.

This approach works well in general, but something I'm running into more frequently is that the `pip install` pulls the latest version of the package from PyPi, and it then fails the [Anaconda version check](https://github.com/Snowflake-Labs/snowcli/blob/main/src/snowflake/cli/plugins/snowpark/package_utils.py#L315) because Anaconda typically lags behind PyPi.

As an experiment, I've monkey-patched snowpark_shared.py [here](https://github.com/Snowflake-Labs/snowcli/blob/main/src/snowflake/cli/plugins/snowpark/snowpark_shared.py#L65) so that when `requirements.other.txt` is written out, we include both the "other" requirements and the "snowflake" requirements. This means that in my original `requirements.txt` I can pin specific versions of packages that I know will be pulled in as dependencies, and I pin them to the latest version that's available on Anaconda. This means that they satisfy the Anaconda version check, are added to the "snowflake" requirements and its files are deleted.

I'll leave this in place for a while, and eventually open a PR if it proves to be a good solution.

Meanwhile I'll happily take suggestions on alternate ways to solve this.

### Context

So that the size of app.zip files are much smaller and don't include native code.

sfc-gh-jsikorski:
Hi, thanks for noticing the problem. Could you give me some example package names, to recreate the problem?

jamesweakley:
If you include `july` (PyPi only), it'll bring in `matplotlib` in the second pass, but a newer version than what's on Anaconda.

This will mean that matplotlib is downloaded and bundled in app.zip.

With my workaround, I can put `matplotlib==3.7.4` in requirements.txt which will keep it pinned during the second round.

Issue #122, opened 2023-05-12
Allow creating package zips from Git or HTTP locations through pip


sfc-gh-hachouraria:
pip accepts a limited range of URL syntax for package names.

For example you could type the following to directly install packages from sources:

```bash
~> pip install 'git+https://github.com/user/repo.git@branch_name'
```

The `snowcli package create [name]` command accepts such URLs as arguments:

```bash
~> snow package create 'git+https://github.com/user/repo.git@branch_name'
```

However, it assumes that the `name` argument is also the filename for `name.zip` and will attempt to create a file named `git+https://github.com/user/repo.git@branch_name.zip`, which is invalid and will result in a final error:

```bash
~> snow package create 'git+https://github.com/user/repo.git@branch_name'
Collecting git+https://github.com/user/repo.git@branch_name
Cloning https://github.com/user/repo.git (to revision branch_name) to /private/var/folders/dj/8bdd294947dbq_y72_whrqj80000gn/T/pip-req-build-lug984pr
  Running command git clone --filter=blob:none --quiet https://github.com/user/repo.git /private/var/folders/dj/8bdd294947dbq_y72_whrqj80000gn/T/pip-req-build-lug984pr
Resolved https://github.com/user/repo.git to commit b76324ac90d93dc3bca382d90841a9ab5f398b50f
[… snipped …]
FileNotFoundError: [Errno 2] No such file or directory: 
'git+https://github.com/user/repo.git@branch_name.zip'
```

This prevents convenient packaging latest or fixed branches of popular repositories when they are required.

Could the `name` in case of URLs please be derived from the downloaded package metadata, or accepted as an optional add-on string?

jeffhollan:
Good point. I like the idea of an optional flag to set package name. Something like `snow package create 'git+...' -n my_package`

sfc-gh-turbaszek:
Done via #947

PR #781, opened 2024-02-14
Fixes for package creation and pip installs

sfc-gh-jsikorski:
### Pre-review checklist

### Changes description
*Tweaked package creation mechanism in `snow snowpark package create` - now we check for minimal required version, not the latest installed by pip, to reduce the size of uploaded packages (and problems mentioned in #763)
* Improved the way we handle packages created from git repo 

Fixes: #763 , #122 , #116 (probably was fixed before, but now confirmed)


src/snowflake/cli/plugins/snowpark/package_utils.py
```diff
@@ -101,86 +103,10 @@ def _get_anaconda_channel_contents():
...
-            version = results.group(1)
-            requirement_line += f"=={version}"
-        return Requirement.parse(requirement_line)
-
-
 def install_packages(
     file_name: str | None,
     perform_anaconda_check: bool = True,
-    package_native_libraries: PypiOption = PypiOption.ASK,
+    package_native_libraries: PypiOption = PypiOption.YES,
```
sfc-gh-turbaszek, line 109: How does the option look like? I spotted some weir flag like:
```
 --yes   -y        Installs packages that are not available on the Snowflake anaconda channel.
```

I don't think this looks nice. The flag name should be more meaningful than `--yes`

src/snowflake/cli/plugins/snowpark/package_utils.py
```diff
@@ -316,11 +229,34 @@ def check_if_package_is_avaiable_in_conda(package: Requirement, packages: dict)
...
-                imported directly from Anaconda, and will be excluded from
-                the zip: {}"""
+def create_zip_name(name: str) -> str:
+    if name.startswith("git+"):
+        return (
+            name.replace("git+https://github.com/", "")
+            .replace(".git", "")
+            .split("/")[-1]
+            + ".zip"
+        )
```
sfc-gh-turbaszek, line 258: Should we consider using regexp? `(?:git\+https://github.com/)(?:\w+/)*(\w+)` Here's also nice tool for testing regexps https://regex101.com/

src/snowflake/cli/plugins/snowpark/venv.py
```diff
@@ -48,7 +56,100 @@ def _create_venv(self):
...
+    def _get_library_path(self) -> Path:
+        return [
+            lib for lib in (Path(self.directory.name) / "lib").glob("**/site-packages")
+        ][0]
+
+    def get_package_dependencies(
+        self, name: str, req_type: str
+    ) -> List[RequirementWithFilesAndDeps]:
+
+        if req_type == "package":
```
sfc-gh-turbaszek, line 73: Let's extract the types to Enum or constants

src/snowflake/cli/plugins/snowpark/venv.py
```diff
@@ -48,7 +56,100 @@ def _create_venv(self):
...
+
+    def _get_dependencies(
+        self, package: Requirement
+    ) -> List[RequirementWithFilesAndDeps]:
+        package_info = self.get_package_info(package)
+        result = [package_info]
+
+        for dependency in package_info.dependencies:
+            if dependency:
+                result += self._get_dependencies(Requirement.parse_line(dependency))
```
sfc-gh-turbaszek, line 94: Please use `result.extend` instead of list concatenation

src/snowflake/cli/plugins/snowpark/venv.py
sfc-gh-jsikorski: Done

src/snowflake/cli/plugins/snowpark/venv.py
sfc-gh-jsikorski: Done

src/snowflake/cli/plugins/snowpark/package_utils.py
sfc-gh-jsikorski: There were two different options. `--yes` was resposible for deciding, if we should try to install package from PyPi. I added `--install-from-pip` alias for it, but left `--yes` and `-y` not to break any pipelines using it.
Also added `--allow-native-libraries` to manage this choice

sfc-gh-turbaszek: Let's make the `yes` a hidden flag with goal to deprecate it. It looks stange.

src/snowflake/cli/plugins/snowpark/package_utils.py
sfc-gh-turbaszek: If there's no match we should consider this as "normal", non-git package.

src/snowflake/cli/plugins/snowpark/models.py
```diff
@@ -30,3 +47,34 @@ class RequirementWithFiles:
...
+                the zip: {}"""
+
+
+def get_package_name(name: str) -> str:
+    if name.startswith("git+"):
+        pattern = r"github\.com\/([^\/]+)\/([^\/.]+)(\.git)?"
+        if match := re.search(pattern, name):
+            return match.group(2)
+        else:
+            return "package"
```
sfc-gh-turbaszek, line 75: Does it mean we will try to download `package`?

sfc-gh-jsikorski: A bit of a placeholder currently. I'm not quite sure, what we should return here

RELEASE-NOTES.md
```diff
@@ -46,6 +49,7 @@
 ## Fixes and improvements
 * Restricted permissions of automatically created files
 * Fixed bug where `spcs service create` would not throw error if service with specified name already exists.
+* Improved package lookup, to avoid unnecessary uploads
```
sfc-gh-turbaszek, line 52: Can you please add more info?

src/snowflake/cli/plugins/snowpark/models.py
```diff
@@ -30,3 +55,34 @@ class RequirementWithFiles:
...
+        return RequirementWithFiles(self.requirement, self.files)
+
+
+pip_failed_msg = """pip failed with return code {}.
+            If pip is installed correctly, this may mean you`re trying to install a package
+            that isn't compatible with the host architecture -
+            and generally means it has native libraries."""
+second_chance_msg = """Good news! The following package dependencies can be
+                imported directly from Anaconda, and will be excluded from
+                the zip: {}"""
```
sfc-gh-turbaszek, line 74: How do those messages look like in cli? Ain't there too many blank lines on the lef?

src/snowflake/cli/plugins/snowpark/models.py
```diff
@@ -13,6 +14,30 @@ class PypiOption(Enum):
...
+
+class Requirement(requirement.Requirement):
+    @classmethod
+    def parse_line(cls, line: str) -> Requirement:
+        if len(extras := line.split(";")) > 1:
+            line = extras[0]
+        result = super().parse_line(line)
+
+        if len(extras) > 1:
+            result.extras = extras[1]
```
sfc-gh-mraba, line 30: Is there a possibility that there will be more than 1 item in extra part?
`hello;something;more`

Or it will always be `something; more`?

src/snowflake/cli/plugins/snowpark/models.py
```diff
@@ -30,3 +55,34 @@ class RequirementWithFiles:
...
+        pattern = r"github\.com\/([^\/]+)\/([^\/.]+)(\.git)?"
+        if match := re.search(pattern, name):
+            return match.group(2)
+        else:
+            return name
+
+    elif name.endswith(".zip"):
+        return name.replace(".zip", "")
+    else:
+        return name
```
sfc-gh-turbaszek, line 88: Is this a really package name? Will this work for snowcl? Will this work if someone uses an url with hash or tag?

src/snowflake/cli/plugins/snowpark/models.py
```diff
@@ -30,3 +55,34 @@ class RequirementWithFiles:
...
+            and generally means it has native libraries."""
+second_chance_msg = """Good news! The following package dependencies can be
+                imported directly from Anaconda, and will be excluded from
+                the zip: {}"""
+
+
+def get_package_name(name: str) -> str:
+    if name.startswith("git+"):
+        pattern = r"github\.com\/([^\/]+)\/([^\/.]+)(\.git)?"
+        if match := re.search(pattern, name):
```
sfc-gh-mraba, line 80: it's usually better to have pattern precompiled before using it.

```python
pattern = re.compile(r"something\.to\ match")

def some_func(name, pat=pattern):
    if match := pat.search(name)
```

src/snowflake/cli/plugins/snowpark/package/commands.py
```diff
@@ -105,7 +113,13 @@ def package_create(
...
+        type(
+            lookup_result := lookup(
+                name=name,
+                install_packages=install_packages,
+                allow_native_libraries=allow_native_libraries,
+            )
+        )
         in [
             NotInAnaconda,
             RequiresPackages,
```
sfc-gh-mraba, line 125: I believe that `isinstance(look_up, (NotInAnaconda, RequiredPackages))` suits better

sfc-gh-mraba: also this lookup_result could be assigned using regular = outside if.
it would be easier to read

src/snowflake/cli/plugins/snowpark/package/manager.py
```diff
@@ -64,7 +73,7 @@ def upload(file: Path, stage: str, overwrite: bool):
 
 
 def create(zip_name: str):
-    file_name = zip_name if zip_name.endswith(".zip") else f"{zip_name}.zip"
+    file_name = get_package_name(zip_name) + ".zip"
```
sfc-gh-mraba, line 76:
```suggestion
    file_name = f"{get_package_name(zip_name)}.zip"
```

src/snowflake/cli/plugins/snowpark/package_utils.py
```diff
@@ -75,6 +79,9 @@ def parse_anaconda_packages(packages: List[Requirement]) -> SplitRequirements:
     Returns a dict with two keys: 'snowflake' and 'other'.
     Each key contains a list of Requirement object.
 
+    As snowflake currently doesn't support extra syntax (ex. 'jinja2[diagrams`), if such
```
sfc-gh-mraba, line 82:
```suggestion
    As snowflake currently doesn't support extra syntax (ex. `jinja2[diagrams]`), if such
```

sfc-gh-mraba (commented):
Most are style related, just for fun.

RELEASE-NOTES.md
```diff
@@ -25,6 +25,9 @@
   * Added `create` command for `image-repository`.
   * Added `status`, `set (property)`, `unset (property)`, `suspend` and `resume` commands for `compute-pool`.
   * Added `set (property)`, `unset (property)`,`upgrade` and `list-endpoints` commands for `service`.
+* You can now use github repo link in `snow snowpark package create` to prepare your code for upload
+* Added `allow-native-libraries` option to `snow snowpark package create` command
+* Added alias `--install-from-pip` for `-y` option in `snow snowpark package create` command
```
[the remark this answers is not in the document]
sfc-gh-jsikorski: Currently they are two different flags - fixed in #811

src/snowflake/cli/plugins/snowpark/models.py
sfc-gh-jsikorski: Fixed

src/snowflake/cli/plugins/snowpark/models.py
sfc-gh-jsikorski: Done

src/snowflake/cli/plugins/snowpark/package/commands.py
sfc-gh-jsikorski: Changed

src/snowflake/cli/plugins/snowpark/package/manager.py
```diff
@@ -64,7 +73,7 @@ def upload(file: Path, stage: str, overwrite: bool):
 
 
 def create(zip_name: str):
-    file_name = zip_name if zip_name.endswith(".zip") else f"{zip_name}.zip"
+    file_name = f"{get_package_name(zip_name)}.zip"
```
sfc-gh-turbaszek, line 76: Should we moved the `.zip`  suffix to `get_package_name` function?

sfc-gh-jsikorski: `get_package_name` is used to extract name, that we will use to resolve dependencies etc. - so for example, takes the name from github repo url, zipname etc. - it seems way too early to add `.zip` suffix there

sfc-gh-turbaszek (approved):

PR #784, opened 2024-02-15
SNOW-1057620: Enforce that get-repository-url is called with a connection database and schema

sfc-gh-davwang:
### Pre-review checklist

### Changes description
[BUGFIX]
For schema level objects (image_repository, service) the connection database and schema must be specified otherwise unqualified names may be ambiguous about what object they refer to*. `get_repository_url` did not do this because it uses the show_specific_object function which does not check that connection database and schema are valid. This means that show_specific_object could potentially find multiple rows for the same repository name.

I have added an option to show_specific_object to check that database and schema are set when we are looking for schema level objects and updated corresponding unit tests. I have also done some cleanup on check_database_and_schema.

\* Otherwise there could be two objects with the same name **o** in different schemas. Operating on object **o** without a database and schema specified would be ambiguous.

**Example Usage (unchanged)**:
image

**Example Errors**:
image

### DIFF
--- a/src/snowflake/cli/api/exceptions.py
+++ b/src/snowflake/cli/api/exceptions.py
@@ -114,3 +114,25 @@ class ConfigFileTooWidePermissionsError(ClickException):
         super().__init__(
             f'Configuration file {path} has too wide permissions, run `chmod 0600 "{path}"`'
         )
+
+
+class DatabaseNotProvidedError(ClickException):
+    def __init__(self):
+        super().__init__(
+            """
+            Database not specified. Please update connection to add `DATABASE` parameter,
+            or re-run command using `--dbname` option.
+            Use `snow connection list` to list existing connections
+            """
+        )
+
+
+class SchemaNotProvidedError(ClickException):
+    def __init__(self):
+        super().__init__(
+            """
+            Schema not specified. Please update connection to add `SCHEMA` parameter,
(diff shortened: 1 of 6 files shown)


sfc-gh-davwang:
Note: In the future we will likely add support for fully qualified names for this function, in which case the check must change from 'database and schema are specified' to 'database and schema are specified' or 'database is specified and name is schema.name' or 'name is fully qualified'.

src/snowflake/cli/api/exceptions.py
```diff
@@ -114,3 +114,17 @@ def __init__(self, path: Path):
         super().__init__(
             f'Configuration file {path} has too wide permissions, run `chmod 0600 "{path}"`'
         )
+
+
+class DatabaseNotProvidedError(ClickException):
+    def __init__(self):
+        super().__init__(
+            "Database not specified. Please update connection to add `DATABASE` parameter, or re-run command using `--database` option. Use `snow connection list` to list existing connections."
```
sfc-gh-turbaszek, line 122:
```suggestion
            "Database not specified. Please update connection to add `database` parameter, or re-run command using `--database` option. Use `snow connection list` to list existing connections."
```
I think config entries have to be lowercase

sfc-gh-yiwang (approved):
LGTM

sfc-gh-astus (approved):

PR #785, opened 2024-02-15
SNOW-1054145-improve-cli-console-output-readability: use system colors for printed messaged

sfc-gh-mraba:
# SNOW-1054145

### Pre-review checklist

### Changes description
Styles used for formatting CliConsole text use system colours.

### before
dark_bg_before

### after
dark_bg_after
 
### before
light_bg_before

### after
light_bg_after

### DIFF
--- a/src/snowflake/cli/api/console/console.py
+++ b/src/snowflake/cli/api/console/console.py
@@ -8,10 +8,10 @@ from rich.text import Text
 from snowflake.cli.api.console.abc import AbstractConsole
 from snowflake.cli.api.console.enum import Output
 
-PHASE_STYLE: Style = Style(color="grey93", bold=True)
-STEP_STYLE: Style = Style(color="grey89", italic=True)
-INFO_STYLE: Style = Style(color="black")
-IMPORTANT_STYLE: Style = Style(color="red", bold=True, italic=True)
+PHASE_STYLE: Style = Style(bold=True)
+STEP_STYLE: Style = Style(italic=True)
+INFO_STYLE: Style = Style()
+IMPORTANT_STYLE: Style = Style(bold=True, italic=True)
 INDENTATION_LEVEL: int = 2
 
 


src/snowflake/cli/api/console/console.py
```diff
@@ -8,10 +8,10 @@
...
 from snowflake.cli.api.console.enum import Output
 
-PHASE_STYLE: Style = Style(color="grey93", bold=True)
-STEP_STYLE: Style = Style(color="grey89", italic=True)
-INFO_STYLE: Style = Style(color="black")
-IMPORTANT_STYLE: Style = Style(color="red", bold=True, italic=True)
+PHASE_STYLE: Style = Style(bold=True)
+STEP_STYLE: Style = Style(italic=True)
+INFO_STYLE: Style = Style()
+IMPORTANT_STYLE: Style = Style(bold=True, italic=True)
```
sfc-gh-turbaszek, line 14: Can you show an example?

sfc-gh-mraba: Examples with light and dark background before and after in description

sfc-gh-bdufour, line 14: I'm a little worried about the amount of styling applied here. Bold and italic are used for informational messages, which might lead to poor UX. For example, important messages won't really stand out as much if everything is styled already.

Also, we _can_ use color if we stick to the 16 ANSI-defined ones. AFAIK Rich has a pretty good support for defining the color system (I think we can even set the color system to make sure we don't accidentally set RGB values in the future). It might be useful to consider using a different color for important messages for example (maybe red is too error-like, I'm not sure)

sfc-gh-mraba: light_bg
Below are examples with basic colours.

dark_bg

Still this will look different on user terminals. It all depends on profiles, transparency settings. There is [no reliable way](https://stackoverflow.com/questions/69348880/is-it-possible-to-use-background-aware-color-choices/69352620#comment122592997_69352620) of detecting terminal colour scheme.

In future we might define our own theme for application to make it consistent.

sfc-gh-mraba: light_bg
Below are examples with basic colours.

dark_bg

Still this will look different on user terminals. It all depends on profiles, transparency settings. There is [no reliable way](https://stackoverflow.com/questions/69348880/is-it-possible-to-use-background-aware-color-choices/69352620#comment122592997_69352620) of detecting terminal colour scheme.

In future we might define our own theme for application to make it consistent.

sfc-gh-pczajka (approved):

PR #789, opened 2024-02-15
Add support for generic env varaibles in connection params

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
Add support for `SNOWFLAKE_X` env variables in connection parameters.


src/snowflake/cli/api/commands/flags.py
```diff
@@ -55,6 +67,8 @@ def callback(value):
     callback=_callback(lambda: cli_context_manager.connection_context.set_account),
     show_default=False,
     rich_help_panel=_CONNECTION_SECTION,
+    envvar=_snowflake_env_variable("ACCOUNT"),
+    show_envvar=False,
```
sfc-gh-mraba, line 71: I noticed that some params for connection options repeats. 
Maybe we could introduce a base for this option
```python 
_ConnectionOptionBase = typer.Option(
    show_default=False,
    rich_help_panel=_CONNECTION_SECTION, 
    show_env_var=False,
) 

AccountOption = _ConnectionOptionBase(
    callback=_callback(lambda: cli_context_manager.connection_context.set_account), 
    envvar=_snowflake_env_variable("ACCOUNT"),
)
```

sfc-gh-mraba (approved):

RELEASE-NOTES.md
```diff
@@ -13,6 +13,18 @@
...
   * Added `upgrade` and `list-endpoints` commands for `service`.
+* Connections parameters are also supported by generic environment variables:
+  * `SNOWFLAKE_ACCOUNT`
+  * `SNOWFLAKE_USER`
+  * `SNOWFLAKE_PASSWORD`
+  * `SNOWFLAKE_DATABASE`
+  * `SNOWFLAKE_SCHEMA`
+  * `SNOWFLAKE_ROLE`
+  * `SNOWFLAKE_WAREHOUSE`
+  * `SNOWFLAKE_MFA_PASSCODE`
```
sfc-gh-pjob, line 24: why did you not add envs for `private key path` and `authenticator`?

tests/test_connection.py
```diff
@@ -1,11 +1,15 @@
...
+from unittest import mock
 from unittest.mock import MagicMock
 
-from snowflake.cli.api.exceptions import SnowflakeConnectionError
+import pytest
 
-from tests.testing_utils.fixtures import *
+from snowflake.cli.api.config import config_init
+from snowflake.cli.api.exceptions import SnowflakeConnectionError
+from snowflake.cli.app.snow_connector import _get_connection_details
```
sfc-gh-pjob, line 12: unused imports?

tests/test_connection.py
```diff
@@ -420,3 +424,138 @@ def test_no_mfa_passcode(mock_connect, runner):
...
+            "password": "password_from_flag",
+            "role": "role_from_flag",
+        }
+
+
+@mock.patch.dict(
+    os.environ,
+    {
+        "SNOWFLAKE_CONNECTIONS_EMPTY_ACCOUNT": "account_from_flag",
+        "SNOWFLAKE_ACCOUNT": "account_from_flag",
```
sfc-gh-pjob, line 533: can we have different values for every source:
- `account_from_flag`
- `account_from_global_env`
- `account_from_connection_env`
- `account_from_connection_config`

now when you have the same value in three sources testing it kind of doesn't make sense

this comment is not only about this variable and this test case but all variables in all your test cases

tests/test_connection.py
```diff
@@ -420,3 +424,138 @@ def test_no_mfa_passcode(mock_connect, runner):
...
+            "SNOWFLAKE_DATABASE": "test_database",
+            "SNOWFLAKE_WAREHOUSE": "large",
+            "SNOWFLAKE_ROLE": "role",
+            "SNOWFLAKE_SCHEMA": "my_schema",
+            "SNOWFLAKE_PASSWORD": "dummy",
+        },
+    ],
+)
+@mock.patch("snowflake.connector.connect")
+def test_connection_details_are_resolved_using_also_generic_variables(
```
sfc-gh-pjob, line 451: generic variables values are the same as "empty" connection values, so IMO the test isn't confirming what its name  is suggesting

tests/test_config.py
```diff
@@ -1,5 +1,7 @@
 from tempfile import TemporaryDirectory
 
+import pytest
```
sfc-gh-pjob, line 3: unused import?

src/snowflake/cli/app/snow_connector.py
```diff
@@ -59,6 +56,40 @@ def connect_to_snowflake(temporary_connection: bool = False, mfa_passcode: Optio
...
+                get_connection(connection_name)
+                if connection_name
+                else get_default_connection()
+            )
+        else:
+            connection_parameters = get_default_connection()
+    else:
+        connection_parameters = {}
+    if overrides:
+        ctx = click.get_current_context(silent=True)
```
sfc-gh-pjob, line 76: I can approve this PR but I don't like that we're mixing logic layers.

IMO when we convert click parameters to `overrides` we should extract ParameterSource there. `overrides` dict does not have to be simple `key -> value` dict, it can be `key -> ParameterOverride(value, source)` dict. And here we should only use already extracted values and sources instead of reaching parameters parsing layer from code responsible for building a connection.

RELEASE-NOTES.md
sfc-gh-turbaszek: Good question, my thinking was that won't be necessary as there's no "security" issue with using those in command line

tests/test_connection.py
sfc-gh-turbaszek: Fixed

tests/test_connection.py
sfc-gh-turbaszek: Adjusted test name

tests/test_config.py
sfc-gh-turbaszek: Fixed

RELEASE-NOTES.md
sfc-gh-turbaszek: No longer valid

sfc-gh-pjob (approved):

RELEASE-NOTES.md
```diff
@@ -13,6 +13,20 @@
...
   * Added `set (property)`, `unset (property)`,`upgrade` and `list-endpoints` commands for `service`.
+* Connections parameters are also supported by generic environment variables:
+  * `SNOWFLAKE_ACCOUNT`
+  * `SNOWFLAKE_USER`
+  * `SNOWFLAKE_PASSWORD`
+  * `SNOWFLAKE_DATABASE`
+  * `SNOWFLAKE_SCHEMA`
+  * `SNOWFLAKE_ROLE`
+  * `SNOWFLAKE_WAREHOUSE`
+  * `SNOWFLAKE_MFA_PASSCODE`
```
sfc-gh-astus, line 24: I think we should remove mfa from here. I tested it in `test_connection_details_are_resolved_using_environment_variables` and didn't work as we expected.

sfc-gh-turbaszek: This won't be probably a common use case,  but there's probably nothing wrong in doing so

PR #804, opened 2024-02-21
Fix for password warning shown twice

### LINKED ISSUE #791: SNOW-1060471: `snow connection add` with password prints the warning twice
### SnowCLI version

 2.0.1rc0

### Python version

3.11.7

### Platform

macOS-14.3-arm64-arm-64bit

### What happened

**Update 2/16/24**: Actually this seems like a bug from Typer. See this [question](https://github.com/tiangolo/typer/discussions/700). 

A workaround could be to validate the password at the beginning of the function rather than as an input callback. In other words, if password is not None, then that means the password was passed either through prompt or command line option, in which case we display the appropriate warning. This would change the time when we display the warning from right after you input a password in the prompt to after all inputs have been entered.

**Another question is: If we offer the prompt option for password, do we want to display the warning? I know that if we pass in the password with the `--password` flag it is insecure because it can end up in shell logs but is that true if the user passes in their password with prompt?** 

**Previous Comment 2/16/24**

When you run or `snow connection add` and fill in the password with the prompt, a warning is printed saying `WARNING! Using --password via the CLI is insecure. Use environment variables instead.` However, this warning is printed twice which is not intended. Note that the warning correctly prints once if you pass the password with the command line option (e.g. `snow connection add ... --password pass`). 

image

It seems that when you pass in input via prompt instead of CLI option, it actually applies the callback to that input twice. I quickly spun up another example where the callback actually modifies the input and the input is modified twice.

image
image

### Console output

```Shell
(snowflake-cli-labs) (main) snowcli$ snow connection add
Name for this connection: test
Snowflake account name: test
Snowflake username: test
Snowflake password [optional]: 
WARNING! Using --password via the CLI is insecure. Use environment variables instead.
WARNING! Using --password via the CLI is insecure. Use environment variables instead.
```

### How to reproduce

Run `snow connection add` without any arguments, fill in connection details including the password. It should show two warnings pop up.

sfc-gh-jsikorski:
### Pre-review checklist

### Changes description
Fixes #791


src/snowflake/cli/plugins/connection/commands.py
```diff
@@ -73,9 +73,10 @@ def callback(value: str):
     return callback
 
 
-def _password_callback(value: str):
-    if value:
+def _password_callback(ctx: Context, value: str):
+    if value and not ctx.default_map:
```
sfc-gh-pczajka, line 77: This checks whether `default_map` is not None / is not empty. Is this intended?

```py
In [1]: 'a' if {} else 'b'
Out[1]: 'b'
```

```suggestion
    if value and not ctx.default_map.get("password_callback_shown", False):
```

tests/test_connection.py
```diff
@@ -434,6 +434,43 @@ def test_mfa_passcode(mock_connect, runner, command):
...
+            tmp_file.name,
+            [
+                "connection",
+                "add",
+            ],
+            input="connName\naccName\nuserName\npassword",
+        )
+
+    assert result.exit_code == 0
+    assert result.output.count("WARNING!") == 1
```
sfc-gh-pjob, line 449: Should we log warning if password is passed via prompt?

sfc-gh-jsikorski: You mean - update callback function, to log it?

src/snowflake/cli/plugins/connection/commands.py
sfc-gh-jsikorski: Yes. ctx.default_map is an Optional[Dict] - it`s not present in the first run, so this passes.
If callback i called second time, the dicts is initialised, and the if condition fails.

tests/test_connection.py
sfc-gh-pjob: This test case is for password provided from prompt. I'm just wondering if we should log a warning in such case? The warning message says only about insecure `--password` flag, it doesn't say anything about prompt: `WARNING! Using --password via the CLI is insecure. Use environment variables instead.`

sfc-gh-jsikorski: Well, in my understanding, we discourage the users from using open-text passwords. Like in the prompt or flag, because they will be stored in an openly avaiable `config.toml`. So in both cases we should warn users, but maybe i should change warning text to cover both cases

sfc-gh-pjob: My understanding was different. IMO the goal of this warning was to discourage the users from using plain-text password included as part of CLI command (so from using `--password` option), because executed CLI commands can be visible in operating system's logs.
IMO we can't discourage users from using plain-text passwords in `config.toml` because we don't give them a good alternative.

@sfc-gh-turbaszek Can you share your thought on this?

sfc-gh-turbaszek: Using `--password=XYZ` should cause a warning. Passing password via config or prompt is relatively safe. So if `--password` opens a prompt that's ok

sfc-gh-pjob (approved):

PR #826, opened 2024-02-24
SNOW-1011771: Added generic REPLACE, IF EXISTS, IF NOT EXISTS flags

sfc-gh-davwang:
### Pre-review checklist

### Changes description
Added generic flags for REPLACE. IF EXISTS and IF NOT EXISTS to flags.py. Added these flags to `spcs service create`, `spcs compute-pool create`, and `spcs image-repository create`. Also replaced existing usages of `--replace` with the new flag.

**Usage**:
image

02/26/24: Added validation that at most one create mode option is set to true
image


src/snowflake/cli/api/commands/flags.py
```diff
@@ -13,6 +13,34 @@
...
+        param_decls = kwargs.get("param_decls", self.param_decls)
+        if not isinstance(param_decls, list) and not isinstance(param_decls, tuple):
+            raise TypeError("param_decls must be a list or tuple")
+        passed_kwargs = self.kwargs.copy()
+        passed_kwargs.update(kwargs)
+        passed_kwargs.pop("default", None)
+        passed_kwargs.pop("param_decls", None)
+        return typer.Option(default, *param_decls, **passed_kwargs)
+
+
```
sfc-gh-davwang, line 43: this chunk is just moving OverrideableOption to the top of the file

src/snowflake/cli/plugins/spcs/image_repository/manager.py
```diff
@@ -51,8 +51,21 @@ def get_repository_api_url(self, repo_url):
...
+        name: str,
+        if_not_exists: bool,
+        replace: bool,
+    ):
+
+        create_statement = (
+            f"{'create or replace' if replace else 'create'} image repository"
+        )
+        if if_not_exists:
+            create_statement = f"{create_statement} if not exists"
```
sfc-gh-cgorrie, line 65: Is it possible to pass in both `--replace` and `--if-not-exists` right now? This is a bit of a drive-by review and I do not have time to check, but we should enforce maximum one of the three introduced `OverrideableOptions` if possible.

sfc-gh-cgorrie: (otherwise this will generate invalid SQL)

sfc-gh-davwang: good point, as far as I know IF EXISTS, IF NOT EXISTS, and OR REPLACE should be mutually exclusive for all commands so I'll add a callback to check that at most one is set to true.

src/snowflake/cli/api/commands/flags.py
```diff
@@ -210,6 +238,51 @@ def _password_callback(value: str):
 )
 
 
+# consideration: it may be useful to put these options in a separate help panel, as they are somewhat generic among objects.
+# use these as your parameter names for IfExistsOption, IfNotExistsOption, and ReplaceOption to ensure that validation works
```
sfc-gh-cgorrie, line 242: Is there some way to add a "mutual exclusion" flag to the `OverrideableOption` so that `_create_mode_callback` doesn't have to depend on the exact parameter names? Or maybe a way to enforce the parameter names and fail if they weren't named properly?

src/snowflake/cli/plugins/spcs/image_repository/manager.py
```diff
@@ -51,8 +51,26 @@ def get_repository_api_url(self, repo_url):
...
+    def create(
+        self,
+        name: str,
+        if_not_exists: bool,
+        replace: bool,
+    ):
+        if if_not_exists and replace:
+            raise ValueError(
+                "'replace' and 'if_not_exists' options are mutually exclusive for ImageRepositoryManager.create"
+            )
```
sfc-gh-cgorrie, line 63: Ideally we can figure out a way to resolve this earlier and just pass in some `CreateMode` enum (see also my previous comment). Maybe this isn't so possible in Typer...

sfc-gh-cgorrie (commented):
I would like to approve because it provides a lot of customer value, but am not doing so because NADE has no ownership over the affected files. Added some food-for-thought comments... if your time is limited I'm happy to take this PR over but first we need some feedback from the Snowflake CLI folks.

src/snowflake/cli/plugins/spcs/image_repository/manager.py
sfc-gh-davwang: you're right that on the manager interface side it makes more sense to just use an enum.

src/snowflake/cli/api/commands/flags.py
sfc-gh-davwang: It is certainly possible to make this generic, though I do wonder what the future use cases of mutually exclusive options might look like. One way to format this might be to have a field in OverrideableOption that takes in a list of names that it is incompatible with and then dynamically form _create_mode_callback based on the passed names.

sfc-gh-davwang: Added a mutually exclusive field to OverrideableOption

src/snowflake/cli/api/commands/flags.py
```diff
@@ -13,6 +14,106 @@
 _CLI_BEHAVIOUR = "Global configuration"
 
 
+class OverrideableOption:
```
sfc-gh-astus, line 17: This class needs more tests, has a lot of ifs which should be tested.

src/snowflake/cli/plugins/spcs/image_repository/manager.py
sfc-gh-cgorrie: This chunking strategy makes sense to me.

tests/spcs/test_image_repository.py
```diff
@@ -44,37 +44,76 @@
 ]
 
 
+@pytest.mark.parametrize(
+    "replace, if_not_exists, expected_query",
+    [
+        (False, False, "create image repository test_repo"),
+        (False, True, "create image repository if not exists test_repo"),
+        (True, False, "create or replace image repository test_repo"),
+        # (True, True) is an invalid case as OR REPLACE and IF NOT EXISTS are mutually exclusive.
```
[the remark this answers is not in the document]
sfc-gh-cgorrie: Moreover, shouldn't we test specifically that the `(True, True)` case fails?

tests/api/commands/test_flags.py
```diff
@@ -30,3 +27,66 @@ def _(password: str = PasswordOption):
...
+    app = Typer()
+
+    @app.command()
+    def _(option: int = OverrideableOption(default=2, help="new help")):
+        return "ok"
+
+    runner = CliRunner()
+    result = runner.invoke(app, ["--help"], catch_exceptions=False)
+    assert result.exit_code == 0
+    assert result.output == snapshot
```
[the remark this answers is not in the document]
sfc-gh-davwang: Ah you're right, the test is intended to test that the new option has different properties than original, but I made a typo and reused OverrideableOption instead of original. Rewrote the test.

src/snowflake/cli/api/commands/flags.py
sfc-gh-davwang: good point, I've added more tests to test_flags but let me know if there's any cases I haven't thought of

sfc-gh-astus (approved):

PR #844, opened 2024-02-28
[NADE] PMM Updates to NA Commands

sfc-gh-bgoel:
### Pre-review checklist
   * [x] I've confirmed that my changes are up-to-date with the target branch - as of Feb 28, 9AM EST

### Changes description
The PMM has mentioned certain rules to follow when naming objects in the Snowflake Native App world:
1. Snowflake Native App for the framework, and the application installed in a consumer account from a listing.
2. Application object for the same account installation of an application.
3. Application package for app pkg/application pkg etc.
4. Versions are "defined in" an application package.

The main changes are on docstrings, cli option help params and comments across the code. I have not changed the method names to keep main code changes to a minimum. 

Other changes suggested by the Snowflake CLI team:
1. Snowflake CLI in place of "SnowCLI" or "CLI" etc. 
Earlier, objects created by the Snowflake CLI were being tagged with "GENERATED_BY_SNOWCLI" comment, but following the update to the term, the implementation will detect both "GENERATED_BY_SNOWCLI" and "GENERATED_BY_SNOWFLAKECLI" as valid objects created by the CLI to account for any users who have already started using our commands in their outer loop. 
However, going forward, "GENERATED_BY_SNOWFLAKECLI" will be the only comment used when creating objects.  

How will these changes make it to the PuPr branch?
The Snowflake CLI team has agreed to cherry pick these changes into their PuPr release. 

Reviewers in addition to the NADE team:
1. Snowflake CLI team, incl @sfc-gh-turbaszek who is aware of the changes being made.
2. @sfc-gh-mheavin for the Snowflake CLI docs.
3. Not part of this org, but Steven Leslie for NA terminology sign off.


sfc-gh-mheavin (dismissed):
lgtm

sfc-gh-cgorrie (dismissed):
Changes LGTM

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -137,7 +137,7 @@ def app_run(
         is_interactive=is_interactive,
     )
     return MessageResult(
-        f"Your application ({processor.app_name}) is now live:\n"
+        f"Your application object ({processor.app_name}) is now live:\n"
```
sfc-gh-sleslie, line 140: I'm not 100% sure the context of this message, but the term "live" makes me thing that it's available to the world. In this case, it's really just that the application object is created in the provider's account, no? Maybe say, "now available"? Just a thought.

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -157,10 +157,10 @@ def app_open(
     )
     if manager.get_existing_app_info():
         typer.launch(manager.get_snowsight_url())
-        return MessageResult(f"Application opened in browser.")
+        return MessageResult(f"Application object opened in browser.")
```
sfc-gh-sleslie, line 160: See comment above.

src/snowflake/cli/plugins/nativeapp/exceptions.py
```diff
@@ -5,11 +5,11 @@
 
 
 class ApplicationPackageAlreadyExistsError(ClickException):
-    """An application package not created by SnowCLI exists with the same name."""
+    """An application package not created by Snowflake CLI exists with the same name."""
```
sfc-gh-sleslie, line 8: Is this the "official" name? I'll make sure to use that in the NA docs as well.

src/snowflake/cli/plugins/nativeapp/exceptions.py
```diff
@@ -23,16 +23,16 @@ def __init__(self, name: str):
 
 
 class ApplicationAlreadyExistsError(ClickException):
-    """An application not created by SnowCLI exists with the same name."""
+    """An application object not created by Snowflake CLI exists with the same name."""
```
sfc-gh-sleslie, line 26: Just curious. If I create an application object via SQL, can I alter it with SnowCLI? We discussed that the idea is that customers user one or the other, but I'm curious if there are any specific prohibitions.

src/snowflake/cli/plugins/nativeapp/exceptions.py
```diff
@@ -23,16 +23,16 @@ def __init__(self, name: str):
...
 
     def __init__(self, name: str):
         super().__init__(
-            f'A non-dev application "{name}" already exists in the account.'
+            f'A non-dev application object "{name}" already exists in the account.'
```
sfc-gh-sleslie, line 30: What does "non-dev" mean here? We have the "development mode" which allows providers to see the contents of the application object. Is that the case here?

src/snowflake/cli/plugins/nativeapp/exceptions.py
```diff
@@ -63,11 +63,11 @@ def __init__(self, identifier: str):
 
 
 class CouldNotFindDistributionForAppPackage(ClickException):
-    """The 'distribution' attribute for an app package could not be found."""
+    """The 'distribution' attribute for an application package could not be found."""
```
sfc-gh-sleslie, line 66: Maybe: "The 'distribution' attribute has not been set for an application package."

"Not found" makes me think that there was some other error (e.g. an internal error.). The action to resolve this would be to set the distribution property, no?

src/snowflake/cli/plugins/nativeapp/version/commands.py
```diff
@@ -111,14 +111,14 @@ def version_list(
...
-        help="Version of the app package that you want to drop. Defaults to the version specified in the `manifest.yml` file.",
+        help="Version defined in an application package that you want to drop. Defaults to the version specified in the `manifest.yml` file.",
     ),
     interactive: Optional[bool] = InteractiveOption,
     force: Optional[bool] = ForceOption,
     **options,
 ) -> CommandResult:
     """
-    Drops a version associated with your application package. Version can either be passed in as an argument to the command or read from the `manifest.yml` file.
+    Drops a version defined in your application package. Version can either be passed in as an argument to the command or read from the `manifest.yml` file.
```
sfc-gh-sleslie, line 121:
```suggestion
    Drops a version defined in your application package. Versions can either be passed in as an argument to the command or read from the `manifest.yml` file.
```

src/snowflake/cli/plugins/nativeapp/version/version_processor.py
```diff
@@ -170,13 +170,13 @@ def process(
...
 
-        # Check if --patch needs to throw a bad option error, either if app package does not exist or if version does not exist
+        # Check if --patch needs to throw a bad option error, either if application package does not exist or if version does not exist
         if patch:
             try:
                 if not self.get_existing_version_info(version):
                     raise BadOptionUsage(
                         option_name="patch",
-                        message=f"Cannot create a custom patch when version {version} does not exist for application package {self.package_name}. Try again without using --patch.",
+                        message=f"Cannot create a custom patch when version {version} is not defined in application package {self.package_name}. Try again without using --patch.",
```
sfc-gh-sleslie, line 179:
```suggestion
                        message=f"Cannot create a custom patch when a version {version} is not defined in the application package {self.package_name}. Try again without using --patch.",
```

sfc-gh-sleslie (approved):
LGTM! These changes are all consistent with terminology we use in the NA core docs.

I made a few suggestions, but they are minor.

src/snowflake/cli/plugins/nativeapp/commands.py
sfc-gh-bgoel: Updating this to "available" in the next commit.

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -148,7 +148,7 @@ def app_open(
     **options,
 ) -> CommandResult:
     """
-    Opens the application inside of your browser,
+    Opens the application object inside of your browser,
```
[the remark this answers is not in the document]
sfc-gh-bgoel: Here, the application is a same-account installation. Hence using "application object" term. The Open command will open up the Snowsight link to this application, not the worksheet. 
Regardless same account or cross account install, the application is always listed under Apps in Snowsight.
So should this be "application object" or "Snowflake Native App"?

src/snowflake/cli/plugins/nativeapp/exceptions.py
sfc-gh-bgoel: Yes, this is the official name.

src/snowflake/cli/plugins/nativeapp/exceptions.py
sfc-gh-bgoel: Yes, a user can interact with objects created via SQL using the Snowflake CLI, assuming they provide the correct names etc to a project definition file. The Snowflake CLI will generally warn the user in that case.

src/snowflake/cli/plugins/nativeapp/exceptions.py
sfc-gh-bgoel: Yes, but specifically we are referring to "using files on a named stage"/loose files mode within the development mode. Do you have a suggestion on how we can improve this?

src/snowflake/cli/plugins/nativeapp/exceptions.py
sfc-gh-bgoel: This exception is actually not being used anywhere in the code. I'll remove it from here.

src/snowflake/cli/plugins/nativeapp/commands.py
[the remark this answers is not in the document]
sfc-gh-sleslie: This is a problem with the terminology we were give (and something I'll address in my proposal). I think it should be Snowflake Native App because it's in the App tab under Snowsight (unless they changed the UI again).

src/snowflake/cli/plugins/nativeapp/exceptions.py
sfc-gh-sleslie: Maybe "An application object created in development mode." Loose file mode is kind of an outage in the docs, currently. I have a JIRA to add more information about that.

sfc-gh-turbaszek (dismissed):

sfc-gh-turbaszek (approved):

sfc-gh-cgorrie (approved):

PR #860, opened 2024-03-01
mraba/app-factory: create app with factory

sfc-gh-mraba:
### Pre-review checklist

### Changes description
created app_factory method for app creation and test usage


sfc-gh-astus (approved):

sfc-gh-pczajka (dismissed):

src/snowflake/cli/app/cli_app.py
```diff
@@ -126,93 +125,98 @@ def _info_callback(value: bool):
...
-        pycharm_debug_library_path=pycharm_debug_library_path,
-        pycharm_debug_server_host=pycharm_debug_server_host,
-        pycharm_debug_server_port=pycharm_debug_server_port,
-    )
+def app_factory() -> SnowCliMainTyper:
+    app = SnowCliMainTyper()
+
+    @app.callback()
+    def default(
+        version: bool = typer.Option(
```
sfc-gh-jsikorski, line 133: Maybe we could move all those typer.Option to separate variables, and then put them here like 
`version: bool = version_option,`
It would increase the readability of the function, WDYT?

sfc-gh-jsikorski: Ok, the body of the function doesn't do much, so maybe it's not that obfuscating :D

sfc-gh-mraba: Defining these options outside the function declaration definitely would make it easier to read and navigate.
I wanted to keep changes as minimal as possible.

sfc-gh-astus (approved):

PR #878, opened 2024-03-07
Refactor Github Actions

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
...


.github/workflows/build.yaml
```diff
@@ -15,90 +15,29 @@ concurrency:
...
-        uses: actions/setup-python@v5
-        with:
-          python-version: '3.x'
-      - name: Install pre-commit
-        run: python -m pip install pre-commit
-      - name: Run pre-commit
-        run: pre-commit run --all-files
-
   build:
     name: Build packages
```
sfc-gh-pjob, line 19:
```suggestion
    name: Build project
```

.github/workflows/test_performance_history_analysis.yaml
sfc-gh-pjob, line 1: I wouldn't say the new filename is better. What was wrong with the old one?

.github/workflows/test_integration.yaml
```diff
@@ -15,11 +15,11 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  test:
-    runs-on: ${{ matrix.os }}
+  integration-tests:
```
sfc-gh-pjob, line 18: Maybe we should change job name also for e2e tests?

.github/workflows/build.yaml
```diff
@@ -15,90 +15,29 @@ concurrency:
...
     runs-on: ${{ matrix.os }}
     needs: lint
     strategy:
       matrix:
-        include:
-          - os: macos-latest
-            TARGET: macos
-            CMD_BUILD: >
-              pyinstaller snowcli.spec &&
```
sfc-gh-pjob, line 45: Is `snowcli.spec` file still needed for sth else? Or can we remove it?

.github/workflows/test_integration.yaml
sfc-gh-turbaszek: Aligned everywhere to `tests`

.github/workflows/test_performance_history_analysis.yaml
sfc-gh-pjob: Yes but I'm not sure if it is a test. The workflow prepares a summary of SnowCLI execution times. Its not a test which can fail.

sfc-gh-turbaszek: Yes, but it's related to "tests" context, I simply like this list to be somehow organized

sfc-gh-pjob: I wouldn't say it's in "tests" context and IMO this new name can be misleading but I won't fight for the old name, so I'm approving the PR :)

PR #879, opened 2024-03-07
Add basic templating to snow sql

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
Add basic rendering of templates in SQL.


src/snowflake/cli/api/utils/rendering.py
```diff
@@ -105,17 +72,21 @@ def generic_render_template(
...
-    env = jinja2.Environment(
-        loader=jinja2.loaders.FileSystemLoader(template_path.parent),
-        keep_trailing_newline=True,
-        undefined=jinja2.StrictUndefined,
+    env = _env_bootstrap(
+        Environment(
+            loader=loaders.FileSystemLoader(template_path.parent),
+            keep_trailing_newline=True,
+            undefined=StrictUndefined,
+        )
```
sfc-gh-bgoel, line 80: This is being used in `snow app init`, which currently uses the regular jinja syntax `{{}}`. I think we can constrain it such that all client side rendering must have `&{}` only. I can change our templates to conform to that. WDYT?

src/snowflake/cli/plugins/sql/commands.py
```diff
@@ -10,6 +10,14 @@
 app = SnowTyper()
 
 
+def _parse_key_value(key_value_str: str):
```
sfc-gh-bgoel, line 13: Since we are going to introduce templating across commands, not just `snow sql`, this function might end up being in some utility file so every command can access.

RELEASE-NOTES.md
```diff
@@ -12,6 +12,7 @@
 * Added support for fully qualified image repository names in `spcs image-repository` commands.
 * Added `--if-not-exists` option to `create` commands for `service`, and `compute-pool`. Added `--replace` and `--if-not-exists` options for `image-repository create`.
 * Added support for python connector diagnostic report.
+* `snow sql` command supports now templating of queries.
```
sfc-gh-bgoel, line 15: Can we clarify this to: supports client-side templating of queries?

tests/api/utils/test_rendering.py
```diff
@@ -0,0 +1,57 @@
...
+        # Green path
+        ("&{ foo }", "bar"),
+        # Using $ as sf variable and basic jinja for server side
+        ("${{ foo }}", "${{ foo }}"),
+        ("$&{ foo }{{ var }}", "$bar{{ var }}"),
+        ("${{ &{ foo } }}", "${{ bar }}"),
+        # Using $ as sf variable and client side rendering
+        ("$&{ foo }", "$bar"),
+    ],
+)
```
sfc-gh-bgoel, line 22: This is helpful, thank you!

tests/api/utils/test_rendering.py
```diff
@@ -0,0 +1,57 @@
...
+    {% endif %}
+    """,
+    ],
+)
+def test_that_common_logic_block_are_ignored(text):
+    assert snowflake_cli_jinja_render(text) == text
+
+
+def test_that_common_comments_are_respected():
+    assert snowflake_cli_jinja_render("{# note a comment &{ foo } #}") == ""
```
sfc-gh-bgoel, line 48: To confirm, jinja comments are not shown in the final output after the render? And that's why this entire input renders as an empty string?

src/snowflake/cli/plugins/sql/manager.py
sfc-gh-bgoel, line 1: Once the templating proposal is finalized (I'll finish it this week), let's discuss if the sql commands could also use template variables that may have been defined in a snowflake.yml file.

src/snowflake/cli/plugins/sql/commands.py
sfc-gh-turbaszek: Correct, I believe we have it already somewhere after adding `stage execute` I will need to rebase.

tests/api/utils/test_rendering.py
sfc-gh-turbaszek: Correct, I think I can extend the example to cover some templating

src/snowflake/cli/plugins/sql/manager.py
sfc-gh-turbaszek: That's definitely the plan, I would just like to separate those two changes to have better commit history

RELEASE-NOTES.md
sfc-gh-turbaszek: Added

src/snowflake/cli/plugins/sql/commands.py
sfc-gh-turbaszek: Done

src/snowflake/cli/api/utils/rendering.py
sfc-gh-turbaszek: I don't agree, for example imagine creating a project for template that includes client side templating...

RELEASE-NOTES.md
```diff
@@ -48,6 +48,7 @@
   * `snow snowpark build`
   * `snow snowpark package lookup`
   * `snow snowpark package create`
+* `snow sql` command supports now client-side templating of queries.
```
sfc-gh-astus, line 51: It should be in 2.3.0

src/snowflake/cli/api/utils/rendering.py
```diff
@@ -105,17 +73,39 @@ def generic_render_template(
...
+
+def _add_project_context(data: Dict):
+    context_key = "ctx"
+    if context_key in data:
+        raise ValueError(f"{context_key} in user defined data")
+    context_data = {context_key: {"env": _AttrGetter(os.environ)}}
+    return {**data, **context_data}
+
+
+def snowflake_cli_jinja_render(content: str, data: Dict | None = None) -> str:
```
sfc-gh-astus, line 109: `Optional[Dict]`

src/snowflake/cli/plugins/sql/manager.py
```diff
@@ -1,18 +1,27 @@
...
 
 
 class SqlManager(SqlExecutionMixin):
     def execute(
-        self, query: Optional[str], file: Optional[Path], std_in: bool
+        self,
+        query: Optional[str],
+        file: Optional[Path],
+        std_in: bool,
+        data: Dict | None = None,
```
sfc-gh-astus, line 24: `Optional[Dict]`

src/snowflake/cli/api/utils/rendering.py
```diff
@@ -33,66 +34,33 @@ def procedure_from_js_file(env: jinja2.Environment, file_name: str):
...
-        SecurePath(file_name).absolute().read_text(file_size_limit_mb=UNLIMITED)
+def _env_bootstrap(env: Environment) -> Environment:
+    for custom_filter in _CUSTOM_FILTERS:
+        env.filters[custom_filter.__name__] = custom_filter
+
+    return env
+
+
+def get_snowflake_cli_jinja_env():
+    _random_block = "___very___unique___block___to___disable___logic___blocks___"
```
sfc-gh-astus, line 48: What is it for?

sfc-gh-turbaszek: A unique block to disable logic blocks. We don't want to support `{% for foo in bar %}` syntax.

src/snowflake/cli/plugins/sql/manager.py
sfc-gh-turbaszek: [PEP604](https://peps.python.org/pep-0604/) make both syntaxes compatible. We don't have a common pattern in our code base.

sfc-gh-astus: `Python-Version: 3.10` What with 3.8 and 3.9?

RELEASE-NOTES.md
sfc-gh-turbaszek: Fixed

src/snowflake/cli/plugins/sql/manager.py
sfc-gh-turbaszek: @sfc-gh-astus we already use it across the code base so this is nothing new. If we want to align to standard format (for example #1001 ) we should do so in other PR.

sfc-gh-astus (approved):

src/snowflake/cli/api/utils/rendering.py
sfc-gh-bdufour: Native app has custom client-side template logic now that uses the standard jinja2 syntax. We should likely also support the Snowflake CLI syntax, but we can't remove support for the current behaviour without risking breaking existing users. I do agree with Tomasz that the Snowflake CLI does make it easier for clients to mix their own client-side template with ours.

sfc-gh-bdufour (approved):

PR #887, opened 2024-03-08
[SNOW-1039218] feat: add hidden app list-templates command

sfc-gh-mchok:
### Pre-review checklist

### Changes description
Added a `snow app list-templates` hidden command to support dynamic fetching of official Native App templates inside our PyCharm plugin

For more information please see: https://snowflakecomputing.atlassian.net/browse/SNOW-1039218

### DIFF
--- a/src/snowflake/cli/plugins/nativeapp/commands.py
+++ b/src/snowflake/cli/plugins/nativeapp/commands.py
@@ -7,9 +7,17 @@ from snowflake.cli.api.commands.decorators import (
     with_project_definition,
 )
 from snowflake.cli.api.commands.snow_typer import SnowTyper
-from snowflake.cli.api.output.types import CommandResult, MessageResult
+from snowflake.cli.api.output.types import (
+    CollectionResult,
+    CommandResult,
+    MessageResult,
+)
+from snowflake.cli.api.secure_path import SecurePath
 from snowflake.cli.plugins.nativeapp.common_flags import ForceOption, InteractiveOption
-from snowflake.cli.plugins.nativeapp.init import nativeapp_init
+from snowflake.cli.plugins.nativeapp.init import (
+    OFFICIAL_TEMPLATES_GITHUB_URL,
+    nativeapp_init,
+)
 from snowflake.cli.plugins.nativeapp.manager import NativeAppManager
 from snowflake.cli.plugins.nativeapp.policy import (
     AllowAlwaysPolicy,
@@ -20,7 +28,10 @@ from snowflake.cli.plugins.nativeapp.run_processor import NativeAppRunProcessor
 from snowflake.cli.plugins.nativeapp.teardown_processor import (
     NativeAppTeardownProcessor,
 )
-from snowflake.cli.plugins.nativeapp.utils import is_tty_interactive
+from snowflake.cli.plugins.nativeapp.utils import (
+    get_first_paragraph_from_markdown_file,
+    is_tty_interactive,
+)
 from snowflake.cli.plugins.nativeapp.version.commands import app as versions_app
 
 app = SnowTyper(
@@ -67,6 +78,49 @@ def app_init(
     return MessageResult(
         f"Snowflake Native App project {project.name} has been created at: {path}"
     )
+
+
+@app.command("list-templates", hidden=True)
+def app_list_templates(**options) -> CommandResult:
+    """
+    Prints information regarding templates that can be used with snow app init.
+    """
+    with SecurePath.temporary_directory() as temp_path:
+        from git import Repo
+
+        repo = Repo.clone_from(
+            url=OFFICIAL_TEMPLATES_GITHUB_URL,
+            to_path=temp_path.path,
+            filter=["tree:0"],
+            depth=1,
+        )
+
+        # Close repo to avoid issues with permissions on Windows
+        repo.close()
+
+        # Mark a directory as a template if a project definition jinja template is inside
+        template_directories = [
+            entry.name
+            for entry in repo.head.commit.tree
+            if (temp_path / entry.name / "snowflake.yml.jinja").exists()
+        ]
+
+        # get the template descriptions from the README.md in its directory
+        template_descriptions = [
+            get_first_paragraph_from_markdown_file(
+                (temp_path / directory / "README.md").path
+            )
+            for directory in template_directories
+        ]
+
+        result = (
+            {"template": directory, "description": description}
+            for directory, description in zip(
+                template_directories, template_descriptions
+            )
+        )
+
+        return CollectionResult(result)
 
 
 @app.command("bundle", hidden=True)
--- a/src/snowflake/cli/plugins/nativeapp/utils.py
+++ b/src/snowflake/cli/plugins/nativeapp/utils.py
@@ -1,3 +1,4 @@
+from pathlib import Path
 from sys import stdin, stdout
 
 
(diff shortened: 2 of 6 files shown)


tests/nativeapp/test_utils.py
```diff
@@ -0,0 +1,32 @@
+import pytest
+from snowflake.cli.api.secure_path import SecurePath
+from snowflake.cli.plugins.nativeapp.utils import get_first_paragraph_from_markdown_file
+
+
+@pytest.mark.parametrize(
+    "file_content, expected_paragraph",
+    [
+        (
+            """
```
sfc-gh-cgorrie, line 10: Suggest using `dedent` here

tests/nativeapp/test_utils.py
```diff
@@ -0,0 +1,32 @@
...
+    ],
+)
+def test_get_first_paragraph_from_markdown_file(file_content, expected_paragraph):
+    with SecurePath.temporary_directory() as temp_path:
+        temp_readme_path = (temp_path / "README.md").path
+
+        with open(temp_readme_path, "w+") as temp_readme_file:
+            temp_readme_file.write(file_content)
+
+        actual_paragraph = get_first_paragraph_from_markdown_file(temp_readme_path)
```
sfc-gh-cgorrie, line 30: The bulk of this test should be within the `with` statement; the file can be deleted at any time after you exit its context manager

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -69,6 +80,49 @@ def app_init(
...
+    """
+    with SecurePath.temporary_directory() as temp_path:
+        from git import Repo
+
+        repo = Repo.clone_from(
+            url=OFFICIAL_TEMPLATES_GITHUB_URL,
+            to_path=temp_path.path,
+            filter=["tree:0"],
+            depth=1,
+        )
```
sfc-gh-bdufour, line 96: IMHO we should reuse the logic from the init command instead of reimplementing it here. You'll probably have to factor it out first though.

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -69,6 +80,49 @@ def app_init(
...
+        template_directories = [
+            entry.name
+            for entry in repo.head.commit.tree
+            if (temp_path / entry.name / "snowflake.yml.jinja").exists()
+        ]
+
+        # get the template descriptions from the README.md in its directory
+        template_descriptions = [
+            get_first_paragraph_from_markdown_file(
+                (temp_path / directory / "README.md").path
```
sfc-gh-cgorrie, line 111: Right now, this will fail if any templates are created that don't have top-level `README.md` files. I suggest we provide a fallback to `None` both in this case and when there is no paragraph text in the readme

sfc-gh-cgorrie (changes requested):
A few notes!

src/snowflake/cli/plugins/nativeapp/commands.py
sfc-gh-mchok: Good callout, I extracted a function to `src/snowflake/cli/plugins/nativeapp/utils.py`, let me know if that works for you!

tests/nativeapp/test_utils.py
sfc-gh-mchok: I tried and wasn't able to make it work smoothly since some of the whitespace is crucial in how the function works (e.g. first paragraph needs to have no whitespace in between), but I turned the multiline strings into multi line-strings to have a similar effect where the file isn't as long, let me know if that works for you or if I massively misunderstood the assignment 😅

tests/nativeapp/test_utils.py
sfc-gh-mchok: I actually had the opposite problem where if I put the two lines inside the context manager, then the tests fail because the function still reads the file as if it was empty:

image

I'm guessing Python doesn't flush the file changes until the `open` context manager is exited?

I would've thought that only after exiting the top-level context manager with the temporary directory that the temp files would be deleted 🤔

src/snowflake/cli/plugins/nativeapp/commands.py
sfc-gh-mchok: Good callout - added a test for it as well, thanks!

tests/nativeapp/test_utils.py
sfc-gh-cgorrie: You're right.

I misread the command, we had one that would scope the file's lifetime to the same `with` statement as opening it. But that code has since been refactored.

sfc-gh-cgorrie (approved):
LGTM!

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -69,6 +81,39 @@ def app_init(
     )
 
 
+@app.command("list-templates", hidden=True)
+def app_list_templates(**options) -> CommandResult:
+    """
+    Prints information regarding templates that can be used with snow app init.
```
sfc-gh-bgoel, line 87: Nit: "... regarding the official templates ...", so it is clear that only our Github templates are listed in this command.

src/snowflake/cli/plugins/nativeapp/utils.py
```diff
@@ -7,3 +10,58 @@ def needs_confirmation(needs_confirm: bool, auto_yes: bool) -> bool:
...
+
+    Returns:
+        Optional[str]: the first paragraph as a string, or None
+        if no paragraph could be found
+
+    Raises:
+        FileNotFoundError: if file_path to Markdown file does not exist
+    """
+    if not file_path.exists():
+        raise FileNotFoundError(file_path)
```
sfc-gh-bgoel, line 30: I see the `None` fallback for line 33. But in the case this file is not found, this will throw an error instead of the `None` from the disc above. Can we make both cases `None`?

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -69,6 +81,39 @@ def app_init(
...
+        # get the template descriptions from the README.md in its directory
+        template_descriptions = [
+            get_first_paragraph_from_markdown_file(
+                (temp_path / directory / "README.md").path
+            )
+            for directory in template_directories
+        ]
+
+        result = (
+            {"template": directory, "description": description}
```
sfc-gh-bgoel, line 108: I really like the idea of a little description along with template names!

sfc-gh-bgoel (approved):
LGTM once the small change can be made to the description.

sfc-gh-bgoel (approved):
lgtm, thank you!

sfc-gh-bgoel (approved):

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -69,6 +81,39 @@ def app_init(
     )
 
 
+@app.command("list-templates", hidden=True)
```
sfc-gh-turbaszek, line 84: Today we have feature flags. Have you consider using them?

sfc-gh-mchok: I believe this was clarified offline, please see the slack thread of the PR for the full discussion, but to summarize:
- For now, we intend to make it available for our IDE plugin, but not expose directly it to users
- We don't want users to have to enable a feature flag before the IDE plugin can work
- It doesn't fit the use case of an experimental flag since this command doesn't change the behaviour of any existing commands nor is it interacting with Snowflake

Please let me know if this works for you :)

sfc-gh-turbaszek (approved):

PR #888, opened 2024-03-11
Add --pattern flag to stage list

sfc-gh-pczajka:
### Pre-review checklist

### Changes description
Add `--pattern` flag to `snow object stage list`


src/snowflake/cli/api/commands/flags.py
```diff
@@ -378,6 +378,16 @@ def like_option(help_example: str):
     )
 
 
+def pattern_option(help_example: str) -> typer.Option:
+    return typer.Option(
+        ".*",
+        "--pattern",
+        "-p",
+        help="Regex pattern for filtering files by name. Git repository scope prefix is not included in regex matching."
```
sfc-gh-turbaszek, line 386: Do not reference git repository in generic flag

src/snowflake/cli/plugins/object/stage/commands.py
```diff
@@ -20,14 +21,19 @@
 )
 
 StageNameArgument = typer.Argument(..., help="Name of the stage.")
+PatternOption = pattern_option(
+    help_example='list --pattern=".*\.txt will list all files with .txt extension'
+)
```
sfc-gh-turbaszek, line 26: Any reason why not include this in flag definition? If this is a stage operation it should be generic for stage as well as repo?

src/snowflake/cli/api/commands/flags.py
```diff
@@ -378,6 +378,16 @@ def like_option(help_example: str):
     )
 
 
+def pattern_option(help_example: str) -> typer.Option:
+    return typer.Option(
+        ".*",
```
sfc-gh-turbaszek, line 383: Let's make it `None` to avoid any issues. If pattern is not provided don't add it to `ls` query

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -53,9 +53,12 @@ def _to_uri(self, local_path: str):
             return uri
         return to_string_literal(uri)
 
-    def list_files(self, stage_name: str) -> SnowflakeCursor:
+    def list_files(self, stage_name: str, pattern: str = ".*") -> SnowflakeCursor:
```
sfc-gh-turbaszek, line 56:
```suggestion
    def list_files(self, stage_name: str, pattern: str | None = None) -> SnowflakeCursor:
```

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -53,9 +53,12 @@ def _to_uri(self, local_path: str):
...
         stage_name = self.get_standard_stage_name(stage_name)
-        return self._execute_query(f"ls {self.quote_stage_name(stage_name)}")
+        query = f"ls {self.quote_stage_name(stage_name)}"
+        if pattern is not None:
+            query += f" pattern = '{pattern}'"
```
sfc-gh-pjob, line 60: What if pattern contains `'`?

I checked and the following does not work:
```
snow object stage list TEST_STAGE_1 -p ".*asd'\.txt"
```

IMO we should either escape it (if possible) or inform users that they have to do it by themselves (if possible, in other case that they cannot use `'`).

sfc-gh-pjob: I think it can be more generic problem in building SQLs in our tool and maybe we should think about it outside of the context of this param. 

But also maybe in other cases `'` char shouldn't be allowed to be used in a value being part of SQL query, so maybe it's not a problem there but here this char is a valid part of filenames:

```
snow object stage list test_stage_1
ls @test_stage_1
+-----------------------------------------------------------------------------------------------------------+
| name                            | size | md5                              | last_modified                 |
|---------------------------------+------+----------------------------------+-------------------------------|
| test_stage_1/'test 1'.txt       | 16   | 0d56227f012d6bf1881bffa3e14d55a6 | Thu, 14 Mar 2024 13:48:05 GMT |
```

tests/object/stage/test_stage.py
```diff
@@ -17,6 +17,16 @@ def test_stage_list(mock_execute, runner, mock_cursor):
     mock_execute.assert_called_once_with("ls @stageName")
 
 
+@mock.patch(f"{STAGE_MANAGER}._execute_query")
+def test_stage_list_pattern(mock_execute, runner, mock_cursor):
```
sfc-gh-pjob, line 21: Can we add one execution of `stage list --pattern` in the integration test?

src/snowflake/cli/plugins/object/stage/manager.py
sfc-gh-pjob: Personally, I think it's so uncommon char in filenames that mentioning the issue to users should be enough.

sfc-gh-pczajka: help message was getting very long, so I decided to add validation and readable error message:
```
╭─ Error ───────────────────────────────────────────────────╮
│ All "'" characters in PATTERN must be escaped: "\'"       │
╰───────────────────────────────────────────────────────────╯
```

src/snowflake/cli/api/commands/flags.py
```diff
@@ -378,15 +378,21 @@ def like_option(help_example: str):
     )
 
 
+def _pattern_option_callback(value):
+    if value.count("'") != value.count("\\'"):
```
sfc-gh-pjob, line 382:
```suggestion
    if value and value.count("'") != value.count("\\'"):
```

Currently the following does not work:
```
snow object stage list TEST_STAGE_1 
```

sfc-gh-pjob: and you have to return the value in the callback because now the option does not work

sfc-gh-pczajka: Fixed :)

PR #897, opened 2024-03-13
[NADE][Bugfix] Use Pyyaml instead of StrictYAML

sfc-gh-bgoel:
### Pre-review checklist

### Changes description
The `manifest.yml` file in the Snowflake Native App world can use non-quoted scalar values within square brackets, as yaml allows for this. As an example:
```
references:
  - egress_eai:
      label: "Egress Eai"
      description: "EAI for Egress from NA+SPCS"
      privileges: [USAGE]
```
In Yaml, a value in square brackets is interpreted as a flow sequence, i.e a list. As a result, StrictYaml expects for any values inside to be enclosed in quotes. It throws an error in the case above along the lines of "JSONesque flow mapping". 

On the other hand, pyyaml enforces no such restriction and its behavior aligns with the yaml specification. Since pyyaml is also being used elsewhere in the code, we can drop strictyaml here in favor of pyyaml.

### DIFF
--- a/src/snowflake/cli/plugins/nativeapp/artifacts.py
+++ b/src/snowflake/cli/plugins/nativeapp/artifacts.py
@@ -3,11 +3,11 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import List, Optional, Tuple, Union
 
-import strictyaml
 from click import ClickException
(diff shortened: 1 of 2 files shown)


sfc-gh-turbaszek (approved):

sfc-gh-melnacouzi (approved):
lgtm

PR #899, opened 2024-03-13
Add snowgit commands to snowcli

sfc-gh-pczajka:
This branch cumulates PRs from epic https://snowflakecomputing.atlassian.net/browse/SNOW-990010
All changes were verified separately


src/snowflake/cli/plugins/git/commands.py
```diff
@@ -0,0 +1,239 @@
...
+            " For example: @my_repo/branches/main/"
+        )
+    return path
+
+
+RepoNameArgument = identifier_argument(sf_object="git repository", example="my_repo")
+RepoPathArgument = typer.Argument(
+    metavar="REPOSITORY_PATH",
+    help=(
+        "Path to git repository stage with scope provided."
```
sfc-gh-turbaszek, line 40: What does it mean? Also is `git repository stage` something we use in SnowGit world? Is it fine to mix git repository and stage?

src/snowflake/cli/plugins/git/commands.py
```diff
@@ -0,0 +1,239 @@
...
+        " Path to the repository root must end with '/'."
+        " For example: @my_repo/branches/main/"
```
sfc-gh-turbaszek, line 42: Does it? Maybe we can have logic like below to be more user friendly?
```py
path = path if path.endswith("/") else path + "/"
```

src/snowflake/cli/plugins/git/commands.py
```diff
@@ -0,0 +1,239 @@
...
+
+def _object_exists(object_type, identifier):
+    try:
+        ObjectManager().describe(
+            object_type=object_type.value.cli_name,
+            name=identifier,
+        )
+        return True
+    except ProgrammingError:
+        return False
```
sfc-gh-turbaszek, line 59: Consider moving this to SqlManager as this is a generic utility that will be often handful.

src/snowflake/cli/plugins/git/commands.py
```diff
@@ -0,0 +1,239 @@
...
+        secret_name = typer.prompt(
+            "Secret identifier (will be created if not exists)", default=secret_name
+        )
+        secret = {"name": secret_name}
+        if _object_exists(ObjectType.SECRET, secret_name):
+            cli_console.step(f"Using existing secret '{secret_name}'")
+        else:
+            cli_console.step(f"Secret '{secret_name}' will be created")
+            secret["username"] = typer.prompt("username")
+            secret["password"] = typer.prompt("password/token", hide_input=True)
```
sfc-gh-turbaszek, line 108: Any particular reason to keep this in dict instead of variables like `username` and `password`?

sfc-gh-turbaszek: Explicit over implicit 😉

src/snowflake/cli/plugins/git/commands.py
sfc-gh-pczajka: I tried ;)
This causes problems when someone points a single file (for example to copy/download it). Snowgit finishes with a bug with unreadable message (some Java exception) - I've notified snowgit team

src/snowflake/cli/plugins/git/commands.py
sfc-gh-pczajka: "git repository" is also called "repository stage" in snowgit documentation. 
https://docs.snowflake.com/LIMITEDACCESS/git-integration/git-integration-overview
I think that "git repository stage" will be intuitive as "git repository" alone might be confused with URLs to origin

src/snowflake/cli/api/sql_execution.py
```diff
@@ -82,6 +82,33 @@ def use_role(self, new_role: str):
...
+    ) -> SnowflakeCursor:
+        query = dedent(
+            f"""
+            create secret {name}
+            type = password
+            username = '{username}'
+            password = '{password}'
+            """
+        )
+        return self._execute_query(query)
```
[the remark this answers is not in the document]
sfc-gh-pczajka: done

src/snowflake/cli/plugins/git/commands.py
sfc-gh-pczajka: done

src/snowflake/cli/plugins/git/commands.py
sfc-gh-pczajka: I moved it to ObjectManager - it strongly depends on "describe" operation, which has common parts with other object commands

src/snowflake/cli/api/sql_execution.py
```diff
@@ -82,6 +82,35 @@ def use_role(self, new_role: str):
             if is_different_role:
                 self._execute_query(f"use role {prev_role}")
 
+    def create_password_secret(
+        self, name: str, username: str, password: str
+    ) -> SnowflakeCursor:
+        return self._execute_query(
+            dedent(
```
sfc-gh-turbaszek, line 89: dedent is unnecessary, we have it in _execute_queries

src/snowflake/cli/plugins/git/commands.py
sfc-gh-turbaszek: Ok, then let's at least add this validation to _repo_path_argument_callback. Today we check only for `snow://` or `@` prefix in `is_stage_path`

src/snowflake/cli/plugins/git/commands.py
```diff
@@ -0,0 +1,236 @@
...
+    * url - address of repository to be used for git clone operation
+
+    * secret - Snowflake secret containing authentication credentials. Not needed if origin repository does not require
+    authentication for RO operations (clone, fetch)
+
+    * API integration - object allowing Snowflake to interact with git repository.
+    """
+    _assure_repository_does_not_exist(repository_name)
+    manager = GitManager()
+    om = ObjectManager()
```
sfc-gh-turbaszek, line 81: We create and use this manger later on, why not reuse it also in `_assure_repository_does_not_exist`?

sfc-gh-pczajka: done

src/snowflake/cli/plugins/git/commands.py
sfc-gh-pczajka: I've added heuristic check for the number of '/' characters. Detailed check happens on snowgit side.

sfc-gh-turbaszek: I don't understand. The message says `Path to the repository root must end with '/'` where do we do `endswith` validation?

sfc-gh-pczajka: We don't - this is only required for a root path (`@repo/tag/tagname/some_dir` is fine), so all correct paths have at least three "/" characters. Checking whether the path is a root path or not would make the logic unnecessarily complicated.
I've added a comment in the code

sfc-gh-turbaszek (approved):

PR #900, opened 2024-03-13
Added additional check for new imports

sfc-gh-jsikorski:
### Pre-review checklist

### Changes description
Adding imports to function description in `snowflake.yml` didn't trigger the definition update - added additional check to fix that.
Also, fixed test for snowpark function with separately created package


src/snowflake/cli/plugins/snowpark/common.py
```diff
@@ -50,6 +52,9 @@ def check_if_replace_is_required(
         )
         return True
 
+    if [imp.lower() for imp in imports] != _get_import_from_resource(resource_json):
```
sfc-gh-turbaszek, line 55: Is the order always the same? Should we perform set operation?

src/snowflake/cli/plugins/snowpark/common.py
```diff
@@ -197,3 +202,15 @@ def format_arg(arg: Argument):
...
     return f"{name}({arguments})"
+
+
+def _get_import_from_resource(resource_json: dict) -> List[str] | None:
+    pattern = re.compile(r"@\w+\.\w+\.([^,]+)")
+    if "imports" in resource_json.keys():
+        return [
+            imp.lower()
+            for imp in pattern.findall(resource_json["imports"])
+            if "app.zip" not in imp
```
sfc-gh-turbaszek, line 213: Why app.zip is hardcoded? Add tests for different source name

sfc-gh-jsikorski: Changed

src/snowflake/cli/plugins/snowpark/common.py
sfc-gh-jsikorski: Good point. Changed

src/snowflake/cli/plugins/snowpark/common.py
```diff
@@ -197,3 +205,17 @@ def format_arg(arg: Argument):
         schema=udf_sproc.schema_name,
     )
     return f"{name}({arguments})"
+
+
+def _get_import_from_resource(
+    resource_json: dict, source_name: str = "app.zip"
```
sfc-gh-turbaszek, line 211: Please do not use hardcoded values like this one. It's often a source of errors

sfc-gh-jsikorski: Changed

src/snowflake/cli/plugins/snowpark/common.py
```diff
@@ -197,3 +205,17 @@ def format_arg(arg: Argument):
...
+) -> Set[str] | None:
+    pattern = re.compile(r"@\w+\.\w+\.([^,]+)")
+    if "imports" in resource_json.keys():
+        return {
+            imp.lower()
+            for imp in pattern.findall(resource_json["imports"])
+            if source_name not in imp
+        }
+    else:
+        return set()
```
sfc-gh-turbaszek, line 219: Unnecessary else

sfc-gh-turbaszek: WDYT about reversing the order?
```py
if "imports" not in resource_json.keys():
    return set()
return {
            imp.lower()
            for imp in pattern.findall(resource_json["imports"])
            if source_name not in imp
        }
```

sfc-gh-turbaszek: What I would recommend is building two sets:
1. Expected imports - including the sources
2. Current imports
then you can do
```py
if expected_imports.difference(current_imports):
   return True
```

If in future, this may help in optimizing uploads

sfc-gh-jsikorski: Done, however i decides to stick with `!=` comparison as it also works, when we try to remove some imports from function definition

src/snowflake/cli/plugins/snowpark/common.py
```diff
@@ -197,3 +203,25 @@ def format_arg(arg: Argument):
         schema=udf_sproc.schema_name,
     )
     return f"{name}({arguments})"
+
+
+def _compare_imports(
+    resource_json: dict, imports: List[str], artifact_file: str
+) -> bool:
+    pattern = re.compile(r"(?:\[@?\w+_\w+\.)?(\w+(?:/\w+)+\.\w+)(?:\])?")
```
sfc-gh-mraba, line 211: I suggest to move this to the top of the file.
In this case it does not matter, but it is a good practice to compile regex only once instead on every function call.

src/snowflake/cli/plugins/snowpark/common.py
```diff
@@ -197,3 +203,25 @@ def format_arg(arg: Argument):
...
+
+    project_imports = {
+        imp
+        for import_string in [*imports, artifact_file]
+        for imp in pattern.findall(import_string.lower())
+    }
+
+    if "imports" not in resource_json.keys():
```
sfc-gh-mraba, line 219: No need to invoke keys when checking for item existence in dict.

sfc-gh-mraba (approved):

PR #904, opened 2024-03-15
Adding app deploy command

sfc-gh-gbloom:
### Pre-review checklist

### Changes description

Adding the `snow app deoply` command, which syncs the local changes to the stage without creating/updating the application (More details in [SNOW-1215781](https://snowflakecomputing.atlassian.net/browse/SNOW-1215781)).

Since the deploy command is basically the same as the run command without the last step of creating/updating the application, the most minimal change would be to add a flag to skip this step in the existing `NativeAppRunProcessor` class.

An alternative approach would be to add a dedicated `DeployProcessor` and refactor the common code up to the base class.

The next step will be to extend the command to accept optional file paths arguments to sync only specific files: `snow app deploy [file]*`.

### Test

Created a basic app:
```
snow app run -c QA3 --warehouse TEST1
```

Running `deploy` with no changes:
```
snow app deploy -c QA3

There are no new files that exist only in your local directory.
There are no new files that exist only on the stage.
There are no existing files that have been modified, or their status is unknown.
Existing files identical to the stage:
README.md
manifest.yml
setup_script.sql
```

Making a change to the readme and running `deploy` again:
```
snow app deploy -c QA3

Existing files modified or status unknown:
README.md
Existing files identical to the stage:
manifest.yml
setup_script.sql
Uploading diff-ed files from your local /Users/gbloom/gbloom-training-app/output/deploy directory to the Snowflake stage.
```

[SNOW-1215781]: https://snowflakecomputing.atlassian.net/browse/SNOW-1215781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -224,3 +224,26 @@ def app_teardown(
...
+    Syncs the local changes to the stage without creating or updating the application.
+    """
+
+    processor = NativeAppRunProcessor(
+        project_definition=cli_context.project_definition,
+        project_root=cli_context.project_root,
+    )
+
+    processor.build_bundle()
+    processor.process(
```
sfc-gh-melnacouzi, line 244: which part deploys to the stage? is it this one?

sfc-gh-gbloom: Yep, I'm piggybacking on [RunProcessor.process()](https://github.com/snowflakedb/snowflake-cli/blob/4198548f9b89cc5758763b7e215e8d377bbf9c13/src/snowflake/cli/plugins/nativeapp/run_processor.py#L352-L401) - see its step 3

sfc-gh-melnacouzi: Doesn't NativeAppRunProcessor also create an application instance? I thought we should stop at copying files to the stage. I suggest you confirm with Cam on this.

sfc-gh-bgoel (commented):
IMO, we should do the alternative suggested in the description, i.e. to have a dedicated `DeployProcessor` class which will handle bundle, package creation and upload to stage without any application creation. This allows for single responsibility to be followed, in this case the ecosystem involved in app package creation.
I understand that the way this is done right now is to introduce minimal changes, perhaps as a poc. But we should refactor this properly if we are to make changes to the main branch.

src/snowflake/cli/plugins/nativeapp/commands.py
sfc-gh-bgoel: @sfc-gh-melnacouzi  The app creation is skipped because of the new flag `skip_app_update=True` in the run processor's `process` signature. It is also passed in as True in line 246 below.

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -224,3 +224,22 @@ def app_teardown(
...
+    """
+    Syncs the local changes to the stage without creating or updating the application.
+    """
+    manager = NativeAppManager(
+        project_definition=cli_context.project_definition,
+        project_root=cli_context.project_root,
+    )
+
+    manager.build_bundle()
+    manager.process()
```
sfc-gh-bdufour, line 243: Shouldn't this just be `manager.deploy()` ? How did the manager pick up the `process()` contract?

sfc-gh-bdufour (commented):
Approach LGTM, other than the one comment I have on this draft.

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -42,173 +31,6 @@ class NativeAppRunProcessor(NativeAppManager, NativeAppCommandProcessor):
...
-
-            for i, queries in enumerate(queued_queries):
-                cc.step(f"Applying package script: {self.package_scripts[i]}")
-                self._execute_queries(queries)
-        except ProgrammingError as err:
-            generic_sql_error_handler(
-                err, role=self.package_role, warehouse=self.package_warehouse
-            )
-
-    def _create_dev_app(self, diff: DiffResult) -> None:
```
sfc-gh-bgoel, line 128: Sorry, where is this `_create_dev_app` being moved to? And do we need to move it at all?

src/snowflake/cli/plugins/nativeapp/commands.py
sfc-gh-gbloom: Yep, you're right - I missed it while refactoring

sfc-gh-gbloom:
Closing this in favor of #921

PR #909, opened 2024-03-18
[SNOW-1246922] Fix printing of env variables in connection list

sfc-gh-pjob:
### Pre-review checklist

### Changes description
  * `snow connection list` - Stopped printing of details of connections which do not exist in config.toml but their name starts with other connection's name
     * For example: connection `int` in config.toml and `SNOWFLAKE_CONNECTION_INT2_PASSWORD` env variable - we were printing the env variable value as part of `int` connection description, we won't do it anymore
   * `snow connection list` - Stopped printing of unknown connection details (other than available via `connection add` command)
     * For example: connection `int` in config.toml  has `foo=bar` setting and there is `SNOWFLAKE_CONNECTION_INT_ABC` env variable - we were printing both values, we won't do it anymore

### DIFF
--- a/src/snowflake/cli/api/config.py
+++ b/src/snowflake/cli/api/config.py
@@ -52,7 +52,7 @@ CONFIG_MANAGER.add_option(
 class ConnectionConfig:
     account: Optional[str] = None
     user: Optional[str] = None
-    password: Optional[str] = None
+    password: Optional[str] = field(default=None, repr=False)
     host: Optional[str] = None
     region: Optional[str] = None
     port: Optional[int] = None
--- a/src/snowflake/cli/api/config.py
+++ b/src/snowflake/cli/api/config.py
@@ -67,13 +67,14 @@ class ConnectionConfig:
 
     @classmethod
     def from_dict(cls, config_dict: dict) -> ConnectionConfig:
-        self = cls()
+        known_settings = {}
+        other_settings = {}
         for key, value in config_dict.items():
-            if hasattr(self, key):
-                setattr(self, key, value)
+            if key in cls.__dict__:
+                known_settings[key] = value
             else:
-                self._other_settings[key] = value
-        return self
+                other_settings[key] = value
+        return cls(**known_settings, _other_settings=other_settings)
 
     def to_dict_of_known_non_empty_values(self) -> dict:
         return {
--- a/src/snowflake/cli/api/config.py
+++ b/src/snowflake/cli/api/config.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import logging
 import os
 from contextlib import contextmanager
+from dataclasses import asdict, dataclass, field
 from pathlib import Path
 from typing import Any, Dict, Optional, Union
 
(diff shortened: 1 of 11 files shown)


tests/test_connection.py
```diff
@@ -172,7 +172,49 @@ def test_lists_connection_information(runner):
                 "database": "dev_database",
                 "host": "dev_host",
                 "port": 8000,
-                "protocol": "dev_protocol",
```
sfc-gh-pjob, line 175: I had to remove `protocol` from expected output because it's not supported via `connection add`, so we won't print it in `connection list`

src/snowflake/cli/api/config.py
```diff
@@ -126,16 +175,23 @@ def config_section_exists(*path) -> bool:
...
 
-def get_connection(connection_name: str) -> dict:
+def get_all_connections() -> dict[str, ConnectionConfig]:
+    return {
+        k: ConnectionConfig.from_dict(connection_dict)
+        for k, connection_dict in get_config_section("connections").items()
+    }
+
+
+def get_connection_dict(connection_name: str) -> dict:
```
sfc-gh-pjob, line 185: I don't want to do too many changes in this PR (to keep it relatively safe), so I decided to keep other usages of "get connection" still as dict. This is the place for further refactoring -> changing it to ConnectionConfig.

src/snowflake/cli/api/config.py
```diff
@@ -47,6 +48,50 @@ class Empty:
...
+    password: Optional[str] = None
+    host: Optional[str] = None
+    region: Optional[str] = None
+    port: Optional[int] = None
+    database: Optional[str] = None
+    schema: Optional[str] = None
+    warehouse: Optional[str] = None
+    role: Optional[str] = None
+    authenticator: Optional[str] = None
+    private_key_path: Optional[str] = None
```
sfc-gh-mraba, line 64: WDY about using pydantic model?

src/snowflake/cli/api/config.py
```diff
@@ -47,6 +48,50 @@ class Empty:
...
+
+    @classmethod
+    def from_dict(cls, config_dict: dict) -> ConnectionConfig:
+        self = cls()
+        for key, value in config_dict.items():
+            if hasattr(self, key):
+                setattr(self, key, value)
+            else:
+                self._other_settings[key] = value
+        return self
```
sfc-gh-mraba, line 76: It's convention, but `self` is meant to represent instance, `config` or `connection_config` should be ok

src/snowflake/cli/api/config.py
```diff
@@ -47,6 +48,50 @@ class Empty:
 )
 
 
+@dataclass
+class ConnectionConfig:
+    account: Optional[str] = None
+    user: Optional[str] = None
+    password: Optional[str] = None
```
sfc-gh-mraba, line 55: This field could be removed from representation `field(default=None, repr=False)`

src/snowflake/cli/api/config.py
sfc-gh-pjob: I thought about it but I wanted to make this change as small as possible for purpose of the patch version and then continue refactoring on the main branch (the same as changes mentioned in https://github.com/snowflakedb/snowflake-cli/pull/909#discussion_r1528699025).

src/snowflake/cli/api/config.py
sfc-gh-pjob: Changing this name causes our precommit hook to fail - probably only `self` instance is allowed to call private members and IMO we should be allowed to call them because even in classmethod, we're still in the context of class, especially in this factory method. I kept `self` as a var name.

```
ruff............................................................................Failed
- hook id: ruff
- exit code: 1

src/snowflake/cli/api/config.py:75:17: SLF001 Private member accessed: `_other_settings`
Found 1 error.
```

src/snowflake/cli/api/config.py
sfc-gh-pjob: Done

sfc-gh-turbaszek (dismissed):

src/snowflake/cli/api/config.py
sfc-gh-pjob: I added one more commit changing `from_dict` function to use 2 dicts as we agreed.

sfc-gh-mraba (approved):

PR #924, opened 2024-03-21
Snow 1232599 refactor package create

sfc-gh-pczajka:
### Pre-review checklist

### Changes description
Flags changes:
  * `--pypi-download` or `--yes` flags have no effect and will cause a warning.
    `create` command always checks against PyPi.
  * `--allow-native-libraries` is renamed to `--allow-shared-libraries`. Old name of the flag is still supported.
  * new `--ignore-anaconda` flag disables package lookup in Snowflake Anaconda channel.
    All dependencies will be downloaded from PyPi.
  * new `--skip-version-check` skips comparing versions of dependencies between requirements and Anaconda.
  * new `--index-url` flag sets up Base URL of the Python Package Index to use for package lookup.
 
Major changes in code:
* `snowflake.cli.plugins.snowpark.package_utils.install_packages` got renamed to `download_packages` + refactored to increase code readability
* `snowflake.cli.plugins.snowpark.package.commands.package_create` is refactored to simplify logic and improve code readability.
* `snowflake.cli.plugins.snowpark.venv.Venv.get_package_dependencies` collects metadata information from downloaded `.whl` files instead of installing packages


src/snowflake/cli/plugins/snowpark/models.py
```diff
@@ -65,6 +67,73 @@ def to_requirement_with_files(self):
...
+            metadata = (root / metadata_path[0]).read_text(encoding="utf-8")
+
+            dep_keyword = "Requires-Dist:"
+            dependencies = [
+                line[len(dep_keyword) :].strip()
+                for line in metadata.splitlines()
+                if line.startswith(dep_keyword)
+            ]
+            name = cls._get_name_from_wheel_filename(wheel_path.name)
+            return cls(name=name, wheel_path=wheel_path, dependencies=dependencies)
```
sfc-gh-turbaszek, line 123: Why not use one of the existing libraries?

src/snowflake/cli/plugins/snowpark/models.py
```diff
@@ -65,6 +67,73 @@ def to_requirement_with_files(self):
...
+    [name] is the name of the package standarized accroding to
+    https://peps.python.org/pep-0491/#escaping-and-unicode
+    """
+
+    name: str
+    wheel_path: Path
+    dependencies: List[str]
+
+    @classmethod
+    def from_wheel(cls, wheel_path: Path):
```
sfc-gh-turbaszek, line 102: What if there's no wheel distribution?

src/snowflake/cli/plugins/snowpark/package/anaconda.py
```diff
@@ -57,11 +64,22 @@ def parse_anaconda_packages(self, packages: List[Requirement]) -> SplitRequireme
...
         return result
+
+
+def get_anaconda_from_snowflake() -> AnacondaChannel:
+    try:
+        return AnacondaChannel.from_snowflake()
+    except HTTPError as err:
+        raise ClickException(
+            f"Accessing Snowflake Anaconda channel failed. Reason {err}"
+        )
```
sfc-gh-turbaszek, line 85: Consider moving this to `from_snowflake`, I'm not sure if this need a factory method

src/snowflake/cli/plugins/snowpark/models.py
sfc-gh-pczajka: I wanted to avoid adding more external dependencies - this logic was simple enough to re-implement

src/snowflake/cli/plugins/snowpark/models.py
sfc-gh-pczajka: `pip wheel` builds and packs a wheel from what is available - this was the easiest approach to have consistent artifacts and avoid "install" at the same time

src/snowflake/cli/plugins/snowpark/package/anaconda.py
sfc-gh-pczajka: fixed

src/snowflake/cli/plugins/snowpark/package/commands.py
```diff
@@ -131,46 +133,127 @@ def package_upload(
...
 ) -> CommandResult:
     """
     Creates a Python package as a zip file that can be uploaded to a stage and imported for a Snowpark Python app.
     """
-    if _deprecated_install_option:
-        install_packages = _deprecated_install_option
+    allow_shared_libraries_pypi_option = {
+        True: PypiOption.YES,
+        False: PypiOption.NO,
+    }[allow_shared_libraries]
```
sfc-gh-turbaszek, line 201: what about ASK?

src/snowflake/cli/plugins/snowpark/package/commands.py
```diff
@@ -131,46 +133,127 @@ def package_upload(
...
+    _deprecated_install_packages: bool = deprecated_pypi_download_option,
     **options,
 ) -> CommandResult:
     """
     Creates a Python package as a zip file that can be uploaded to a stage and imported for a Snowpark Python app.
     """
-    if _deprecated_install_option:
-        install_packages = _deprecated_install_option
+    allow_shared_libraries_pypi_option = {
+        True: PypiOption.YES,
```
sfc-gh-turbaszek, line 199:
```suggestion
        True: YesNoAsk.YES,
```
Consider renaming this option, even if we are deprecating it

src/snowflake/cli/plugins/snowpark/package/commands.py
```diff
@@ -131,46 +133,127 @@ def package_upload(
...
+            )
 
-    lookup_result = lookup(
-        name=name,
-        install_packages=install_packages,
-        allow_native_libraries=allow_native_libraries,
+    packages_dir = SecurePath(".packages")
+    packages_are_downloaded, dependencies = download_packages(
+        anaconda=anaconda,
+        perform_anaconda_check_for_dependencies=not ignore_anaconda,
```
sfc-gh-turbaszek, line 220:
```suggestion
        ignore_anaconda=ignore_anaconda,
```
to be consistent WDYT? instead of using `not`

src/snowflake/cli/plugins/snowpark/package/commands.py
```diff
@@ -131,46 +133,127 @@ def package_upload(
...
+        package_name=name,
+        requirements_file=None,
+        packages_dir=packages_dir,
+        index_url=index_url,
+        allow_shared_libraries=allow_shared_libraries_pypi_option,
```
sfc-gh-turbaszek, line 225: I'm wondering if we have to push this deep into logic, how about structuing this command as
```
download_packages()
native = check_for_native_deps()
if native deps:
    do_what_neccessary()
  
  create_zip_package()
  ```
  
  WDYT? I think pushing all those parameters deep down and mixing flow and action control cause a lot of confusion

File: src/snowflake/cli/plugins/snowpark/package/commands.py [hunk 8]
```diff
@@ -131,46 +133,127 @@ def package_upload(
...
     )
 
-    if not isinstance(lookup_result, (NotInAnaconda, RequiresPackages)):
-        return MessageResult(lookup_result.message)
+    if not packages_are_downloaded:
+        return MessageResult(
+            dedent(
+                f"""
+                Cannot create package for {name}. Please check the package name
+                or try again with --allow-shared-libraries option.
```
sfc-gh-turbaszek, line 234: If we separate download from checking for shared libraries we avoid `or` messages, we can give users clear information what to do next

src/snowflake/cli/plugins/snowpark/package/commands.py
```diff
@@ -131,46 +133,127 @@ def package_upload(
...
     )
-    if isinstance(lookup_result, RequiresPackages):
-        message += "\n" + lookup_result.message
+    if dependencies.snowflake:
+        message += dedent(
+            f"""
+            The package {name} is supported, but does depend on the
+            following Snowflake supported libraries. You should include the
+            following dependencies in you function or procedure requirements:
+            """
```
sfc-gh-turbaszek, line 255: That's not true, `snowflake!=None` means there are dependencies available in Anaconda that needs to be added to packages list because we do not include them in the package.

Basically we are operating on three information:
1. What packages are available in anaconda (dependencies.snowflake)
2. What packages needs to be packed (dependencies.other)
3. Does packages to be packed include shared code (we don't store this info)

src/snowflake/cli/plugins/snowpark/package/manager.py
```diff
@@ -83,7 +39,7 @@ def create_packages_zip(zip_name: str) -> str:
     return file_name
 
 
-def cleanup_after_install(func):
+def cleanup_packages_dir(func):
```
sfc-gh-turbaszek, line 42: In my opinion we should get rid of this wrapper. But we can do this in follow-up PR

src/snowflake/cli/plugins/snowpark/package_utils.py
```diff
@@ -68,165 +66,183 @@ def deduplicate_and_sort_reqs(
...
         raise ClickException(
             "Could not use package name and requirements file simultaneously"
         )
-
     if requirements_file and not requirements_file.exists():
         raise ClickException(f"File {requirements_file.path} does not exists.")
+    if perform_anaconda_check_for_dependencies and not anaconda:
+        raise ClickException(
+            "Cannot perform anaconda checks if anaconda channel is not specified."
+        )
```
sfc-gh-turbaszek, line 133: How users can specify Anaconda channel in the command? This message is implementation detail.

One solution to that is to make AnacondaChannel a singleton so we can cache dependencies and easily access them without having to push instance between commands. WDYt?

src/snowflake/cli/plugins/snowpark/package_utils.py
```diff
@@ -68,165 +66,183 @@ def deduplicate_and_sort_reqs(
...
-
-        assert type(requirements_file) is SecurePath
-        pip_install_result = v.pip_install(str(requirements_file.path))
-        dependencies = v.get_package_dependencies(str(requirements_file.path))
+            requirements_file = SecurePath(v.directory.name) / "requirements.txt"
+            requirements_file.write_text(str(package_name))
 
-        if pip_install_result != 0:
-            log.info(pip_failed_msg.format(pip_install_result))
+        pip_wheel_result = v.pip_wheel(
```
sfc-gh-turbaszek, line 141: Why we decided to use `pip wheel`?

src/snowflake/cli/plugins/snowpark/package_utils.py
```diff
@@ -68,165 +66,183 @@ def deduplicate_and_sort_reqs(
...
-            )
-
-            if len(second_chance_results.snowflake) > 0:
-                log.info(second_chance_msg.format(second_chance_results.snowflake))
-            else:
-                log.info("None of the package dependencies were found on Anaconda")
+        dependencies = v.get_package_dependencies(
+            requirements_file, downloads_dir=downloads_dir.path
+        )
+        dependency_requirements = [d.requirement for d in dependencies]
```
sfc-gh-turbaszek, line 151: Can we merge those two operations?

src/snowflake/cli/plugins/snowpark/package_utils.py
```diff
@@ -68,165 +66,183 @@ def deduplicate_and_sort_reqs(
...
 
-        dependencies_to_be_packed = _get_dependencies_not_avaiable_in_conda(
-            dependencies,
-            second_chance_results.snowflake if second_chance_results else None,
+        log.info(
+            "Downloaded packages: %s",
+            ",".join([d.name for d in dependency_requirements]),
```
sfc-gh-turbaszek, line 155:
```suggestion
        log.info(
            "Downloaded packages: %s",
            ", ".join([d.name for d in dependency_requirements]),
```
To make it more human-readable

src/snowflake/cli/plugins/snowpark/package_utils.py
```diff
@@ -68,165 +66,183 @@ def deduplicate_and_sort_reqs(
...
         )
 
-        log.info("Checking to see if packages have native libraries...")
+        if not perform_anaconda_check_for_dependencies:
```
sfc-gh-turbaszek, line 158:
```suggestion
        if ignore_anaconda:
```
It seems that we are using `not` more than once with this argument 😄

src/snowflake/cli/plugins/snowpark/package_utils.py
```diff
@@ -68,165 +66,183 @@ def deduplicate_and_sort_reqs(
...
+            dependencies_to_be_packed = dependencies
+            split_dependencies = SplitRequirements([], other=dependency_requirements)
+        else:
+            log.info("Checking for dependencies available in Anaconda...")
+            assert anaconda is not None
```
sfc-gh-turbaszek, line 163: Do not use asserts in production code (we should add pre-commit check for this). See:
https://snyk.io/blog/the-dangers-of-assert-in-python/

src/snowflake/cli/plugins/snowpark/package_utils.py
```diff
@@ -68,165 +66,183 @@ def deduplicate_and_sort_reqs(
...
-def _confirm_native_libraries(allow_native_libraries: PypiOption) -> bool:
-    if allow_native_libraries == PypiOption.ASK:
-        return click.confirm("Continue with package installation?", default=False)
+def _log_dependencies_found_in_conda(available_dependencies: List[Requirement]) -> None:
+    if len(available_dependencies) > 0:
+        log.info(
+            "Good news! The following package dependencies can be"
+            " imported directly from Anaconda, and will be excluded from"
+            " the zip: %s",
+            format(", ".join(dep.name for dep in available_dependencies)),
```
sfc-gh-turbaszek, line 237: Let's make it shorter, `Packages available in Anaconda and excluded from the zip archive: `

src/snowflake/cli/plugins/snowpark/package_utils.py
```diff
@@ -68,165 +66,183 @@ def deduplicate_and_sort_reqs(
...
+        )
     else:
-        return allow_native_libraries == PypiOption.YES
+        log.info("None of the package dependencies were found on Anaconda")
```
sfc-gh-turbaszek, line 240: I think the `package create` is a good example of command where we can utilse `phase` and `step` for showing intermediate output instead of using logging.

src/snowflake/cli/plugins/snowpark/package/commands.py
sfc-gh-pczajka: Here I'm translating boolean format into yes/no/ask.
I decided to preserve old code as changing login to boolean, thus removing "ask" option, would be a breaking change.

src/snowflake/cli/plugins/snowpark/package_utils.py
sfc-gh-pczajka: This was the easiest approach to have consistent artifacts and avoid "install" at the same time - pip wheel builds and packs a wheel from whatever distribution is available.
In future we may want to avoid building - for example use pip download for specific architecture - but I figured that we will think more about it then.

src/snowflake/cli/plugins/snowpark/package_utils.py
sfc-gh-pczajka: No, both `dependencies` and `dependency_requirements` are used later. I also hoped to use some neat pythoning here :D

src/snowflake/cli/plugins/snowpark/package_utils.py
sfc-gh-pczajka: done

src/snowflake/cli/plugins/snowpark/package_utils.py
sfc-gh-pczajka: I've recently discovered `# type: ignore`, which was the only reason this assertion was here ;)

src/snowflake/cli/plugins/snowpark/package_utils.py
sfc-gh-pczajka: fixed

src/snowflake/cli/plugins/snowpark/package_utils.py
sfc-gh-pczajka: fixed

src/snowflake/cli/plugins/snowpark/package/commands.py
sfc-gh-pczajka: I wouldn't move the logic outside `download_packages` as is is reused in `snowpark build` command. I changed `if` statements a little:
```
if detected_native_libraries():
    if not _confirm_action()
        FAIL
extract_necessary_packages()
return OK

File: src/snowflake/cli/plugins/snowpark/package_utils.py (hunk 18 above)
Comment (sfc-gh-pczajka) [reply src/snowflake/cli/plugins/snowpark/package_utils.py:236]: I'll address that in PR refactoring `snow snowpark build`

File: src/snowflake/cli/plugins/snowpark/package/commands.py (hunk 6 above)
Comment (sfc-gh-pczajka) [reply src/snowflake/cli/plugins/snowpark/package/commands.py]: done

File: src/snowflake/cli/plugins/snowpark/package_utils.py (hunk 18 above)
Comment (sfc-gh-pczajka) [reply src/snowflake/cli/plugins/snowpark/package_utils.py:236]: To quickly answer here: The problem with converting to "phase"/"step" is that we'll start printing on output without `--verbose` flag, or stop logging to avoid double messages on the output. I think it should be a part of separate refactor (replace `log.info` with `console.step/phase`)

Comment (sfc-gh-pczajka) [reply src/snowflake/cli/plugins/snowpark/package_utils.py:236]: https://snowflakecomputing.atlassian.net/browse/SNOW-1271652

File: src/snowflake/cli/plugins/snowpark/package_utils.py (hunk 11 above)
Comment (sfc-gh-pjob) [reply src/snowflake/cli/plugins/snowpark/package_utils.py]: If it is some kind of "illegal state exception", then IMO it shouldn't be ClickException which is designed to be used to inform users about errors.

Comment (sfc-gh-turbaszek) [reply src/snowflake/cli/plugins/snowpark/package_utils.py]: +1 to what @sfc-gh-pjob says, to me it sounds like `RuntimeError`

File: src/snowflake/cli/plugins/snowpark/package/commands.py [hunk 19]
```diff
@@ -131,46 +133,128 @@ def package_upload(
...
+        ):
+            return MessageResult(
+                f"Package {name} is already available in Snowflake Anaconda Channel."
+            )
 
-    lookup_result = lookup(
-        name=name,
-        install_packages=install_packages,
-        allow_native_libraries=allow_native_libraries,
+    packages_dir = SecurePath(".packages")
```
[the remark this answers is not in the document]
sfc-gh-turbaszek: +1 one to the idea, I would be in favor of doing this change in this PR, it sounds like a part of solution not an addition.

src/snowflake/cli/plugins/snowpark/package/commands.py
sfc-gh-pjob: A separate step is ok too. I just wanted to say that we can have:
```
result:
  - success
  - error1
  - error2
  - errorX
```
And I'm not saying that it's better than separating the logic in this case but I just wanted to show that boolean values are very limited.

sfc-gh-pczajka: I've already changed the output to readable class in https://github.com/snowflakedb/snowflake-cli/pull/937

src/snowflake/cli/plugins/snowpark/package_utils.py
sfc-gh-pczajka: This case is removed in
https://github.com/snowflakedb/snowflake-cli/pull/937, but I get the point :+1:

src/snowflake/cli/plugins/snowpark/package/commands.py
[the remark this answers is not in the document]
sfc-gh-pczajka: I've got rid of pinning to ".packages" in https://github.com/snowflakedb/snowflake-cli/pull/937
I've used a temporary directory instead

RELEASE-NOTES.md
```diff
@@ -6,6 +6,15 @@
...
   available in Snowflake Anaconda channel.
+* `snow snowpark package create`:
+  * `--pypi-download` or `--yes` flags have no effect and will cause a warning.
+    `create` command always checks against PyPi.
+  * `--allow-native-libraries` is replaced with boolean flag `--allow-shared-libraries`.
+    Using old flag will cause a waring.
+  * new `--ignore-anaconda` flag disables package lookup in Snowflake Anaconda channel.
+    All dependencies will be downloaded from PyPi.
+  * new `--skip-version-check` skips comparing versions of dependencies between requirements and Anaconda.
+  * new `--index-url` flag sets up Base URL of the Python Package Index to use for package lookup.
```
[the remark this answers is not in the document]
sfc-gh-pczajka: Moved to `New additions` :)

sfc-gh-pjob (approved):

PR #931, opened 2024-03-22
chore: run codespell

sfc-gh-twhite:
Various typo corrections. Most of these were in docstrings, but some were in message responses. 

It may be worthwhile to add this as a pre-commit hook excluding cli output test assertions where new lines can break spelling and for any non-english, but Snowflake, related terms.


sfc-gh-turbaszek (approved):

sfc-gh-melnacouzi (approved):
lgtm

PR #921, opened 2024-03-20
Add app deploy command

sfc-gh-gbloom:
### Pre-review checklist
   * [x] I've confirmed that instructions included in README.md are still correct after my changes in the codebase.
   * [x] I've added or updated automated unit tests to verify correctness of my new code.
   * [x] I've added or updated integration tests to verify correctness of my new code.
   * [x] I've confirmed that my changes are working by executing CLI's commands manually.
   * [x] I've confirmed that my changes are up-to-date with the target branch.
   * [x] I've described my changes in the release notes.
   * [x] I've described my changes in the section below.

### Changes description

Adding the `snow app deoply` command, which syncs the local changes to the stage without creating/updating the application (More details in [SNOW-1215781](https://snowflakecomputing.atlassian.net/browse/SNOW-1215781)).

Since the deploy command is basically the same as the run command without the last step of creating/updating the application, I moved the shared logic to the base `NativeAppManager` class.

The next step will be to extend the command to accept optional file paths arguments to sync only specific files: `snow app deploy [file]*`.

### Test

Created a basic app:
```
snow app run -c QA3 --warehouse TEST1
```

Running `deploy` with no changes:
```
snow app deploy -c QA3

There are no new files that exist only in your local directory.
There are no new files that exist only on the stage.
There are no existing files that have been modified, or their status is unknown.
Existing files identical to the stage:
README.md
manifest.yml
setup_script.sql
```

Making a change to the readme and running `deploy` again:
```
snow app deploy -c QA3

Existing files modified or status unknown:
README.md
Existing files identical to the stage:
manifest.yml
setup_script.sql
Uploading diff-ed files from your local /Users/gbloom/gbloom-training-app/output/deploy directory to the Snowflake stage.
```

[SNOW-1215781]: https://snowflakecomputing.atlassian.net/browse/SNOW-1215781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #933, opened 2024-03-25
Add optional deploy files argument

sfc-gh-gbloom:
### Pre-review checklist

### Changes description

Following up on #921, this PR introduces the following changes:

1. The option to provide specific files and directories when running `snow app deploy`
2. The `--prune`/`--no-prune` option
3. The `-r`/`--recursive` option
4. Keep track of files copied/symlinked in the build step
5. Fix a bug in `build_md5_map()`

When determining the deploy path from source path, we use the mapping of the files created in the build step. Since we symlink entire directories, nested files will not exist in the mapping. In such cases, we traverse the parents chain until a mapped directory is found.

#### Usage

```
snow app deploy file1 dir1 dir2/file2 -r
```

If no files are specified, all files will be synced to stage (existing behavior).

The file paths are assumed to be relative to the project root.

Absolute paths, directories and patterns are currently not supported.

#### Pruning

When pruning is enabled, files that exist only on the stage will be removed.

When deploying all changes, `--prune` is the default value (existing behavior).
When deploying specific files, `--no-prune` is the default value and the user has to explicitly specify `--prune` to avoid accidentally removing files.

#### Recursive

Same as `cp` command, when specifying specific directories to deploy, the `-r` flag must be provided. When a directory is deployed, all of its content will be synced recursively.

When deploying all changes, `-r` is implied (existing behavior).


src/snowflake/cli/api/exceptions.py
```diff
@@ -128,3 +128,13 @@ def __init__(self):
         super().__init__(
             "Schema not specified. Please update connection to add `schema` parameter, or re-run command using `--schema` option. Use `snow connection list` to list existing connections."
         )
+
```
sfc-gh-bgoel, line 131: Our team has recently tried to avoid introducing too many special exceptions and instead use the generic `ClickException` or its applicable subtype, with an informative message, to raise those exceptions instead. 

That, or maybe reuse/rename [SourceNotFoundError](https://github.com/snowflakedb/snowflake-cli/blob/main/src/snowflake/cli/plugins/nativeapp/artifacts.py#L40) which already throws the error for both file and directory, so you won't need two different exceptions.

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -229,6 +230,11 @@ def app_teardown(
 @app.command("deploy", requires_connection=True)
 @with_project_definition("native_app")
 def app_deploy(
+    files: Optional[List[Path]] = typer.Argument(
+        default=None,
+        show_default=False,
+        help="Files to stage, relative to deploy root. [default: all local changes]",
```
sfc-gh-bgoel, line 236: This help text is used directly to generate our documentation. We should make this a bit more clear, something like:
Path of the file, relative to the deploy root, to be uploaded to a stage. Default: sync all local changes with the stage.

Or something similar.

src/snowflake/cli/plugins/object/stage/diff.py
```diff
@@ -154,12 +159,31 @@ def build_md5_map(list_stage_cursor: SnowflakeCursor) -> Dict[str, str]:
     }
 
 
-def stage_diff(local_path: Path, stage_fqn: str) -> DiffResult:
+def get_absolute_files_to_stage(relative_files_to_stage, local_path) -> List[Path]:
```
sfc-gh-bgoel, line 162: Could we add type hints here to the function signature?

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -432,6 +439,6 @@ def deploy(self) -> DiffResult:
             self._apply_package_scripts()
 
             # 3. Upload files from deploy root local folder to the above stage
-            diff = self.sync_deploy_root_with_stage(self.package_role)
+            diff = self.sync_deploy_root_with_stage(self.package_role, files_to_stage)
```
sfc-gh-bdufour, line 442: nit: files_to_deploy? stage is a noun in snowflake, not a verb (here and elsewhere).

src/snowflake/cli/plugins/object/stage/diff.py
```diff
@@ -154,12 +159,31 @@ def build_md5_map(list_stage_cursor: SnowflakeCursor) -> Dict[str, str]:
...
+    paths = []
+    for file in relative_files_to_stage:
+        absolute_path = Path(os.path.join(local_path, file))
+        if not absolute_path.exists():
+            raise FileDoesNotExistError(file)
+        elif absolute_path.is_dir():
+            raise DirectoryNotSupportedError(file)
```
sfc-gh-bdufour, line 169: why not? It makes a lot of sense to support them for our target use case

src/snowflake/cli/plugins/object/stage/diff.py
sfc-gh-bdufour, line 162: nit: _ prefix? It seems rather specific to local_path == deploy root, so I'd consider it an private API within this module.

src/snowflake/cli/plugins/object/stage/diff.py
```diff
@@ -154,12 +159,31 @@ def build_md5_map(list_stage_cursor: SnowflakeCursor) -> Dict[str, str]:
...
+    stage_fqn: str,
+    files_to_stage: Optional[List[Path]] = None,  # relative to local_path
+) -> DiffResult:
     """
     Diffs the files in a stage with a local folder.
     """
     stage_manager = StageManager()
     local_files = enumerate_files(local_path)
+    if files_to_stage is not None and len(files_to_stage) > 0:
+        local_files = get_absolute_files_to_stage(files_to_stage, local_path)
```
sfc-gh-bdufour, line 186: Why are we filtering here, as opposed to post-processing the diff? We're still paying to compute the remote md5 for all remote files, since it's an upfront cost. The side effect for me is that I don't know how to interpret the diff when files are passed in: for provided files, the diff contains them if they're new or modified, but deleted files are included regardless. That seems odd to me because it looks like we're trying to build some kind of "actionable" diff, but I'm not sure we should delete anything that's remote only in filtered mode. Thoughts?

src/snowflake/cli/api/exceptions.py
sfc-gh-cgorrie: I agree, renaming `SourceNotFoundError` to `LocalFileNotFoundError` or something could be the cleanest solution. Name up to you.

sfc-gh-bdufour:
An additional comment: your description states that the deploy root is CWD, but that's inaccurate, it's the project root that is current. Deploy root is typically a subdirectory of the project root. I still believe that falling back to look for files relative to the project root is a valuable addition to this command.

src/snowflake/cli/plugins/object/stage/diff.py
```diff
@@ -154,12 +161,33 @@ def build_md5_map(list_stage_cursor: SnowflakeCursor) -> Dict[str, str]:
     }
 
 
-def stage_diff(local_path: Path, stage_fqn: str) -> DiffResult:
+def get_absolute_files_to_stage(
+    relative_files_to_stage: List[Path], local_path: Path
+) -> List[Path]:
+    paths = []
+    for file in relative_files_to_stage:
+        absolute_path = Path(os.path.join(local_path, file))
```
sfc-gh-cgorrie, line 169: FYI, this is only an absolute path if `local_path` is an absolute path. I don't think they need to be.

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-cgorrie: Agreed with the name change, it might also make the "filtering" property clearer if it's a lambda that decides whether or not to allow this file to be synced rather than the list of files to be synced (which makes it less clear what happens when a file appears in this list but not in the diff).

src/snowflake/cli/plugins/object/stage/diff.py
sfc-gh-cgorrie: I agree, I think I'd prefer `stage_diff` to always return the full diff (performance-wise, I'm assuming the roundtrip to Snowflake dominates the amount of time taken to list all files vs just one) and then we filter the diff afterwards in `NativeAppManager.sync_deploy_root_with_stafge`.

src/snowflake/cli/plugins/object/stage/diff.py
sfc-gh-gbloom: Just to keep the scope of this PR small. I can follow up with supporting directories as a separate PR.

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -229,6 +230,11 @@ def app_teardown(
 @app.command("deploy", requires_connection=True)
 @with_project_definition("native_app")
 def app_deploy(
+    files: Optional[List[Path]] = typer.Argument(
+        default=None,
+        show_default=False,
+        help=f"""Paths of the files, relative to the deploy root, to be uploaded to a stage. If the path exists remotely but not locally, it will be deleted from the stage [default: sync all local changes to the stage]""",
```
sfc-gh-bgoel, line 236: To be clear, are we specifying files relative to the deploy root or relative or CWD? Because deploy root != CWD. My vote is for the CWD since a user is less likely to remember where the file will be on the stage, but have more familiarity with their actual repo structure.

src/snowflake/cli/plugins/object/stage/diff.py
```diff
@@ -154,7 +161,36 @@ def build_md5_map(list_stage_cursor: SnowflakeCursor) -> Dict[str, str]:
...
+                str(file), "This file does not exist either locally or remotely"
+            )
+        elif path.is_dir():
+            raise ClickException(f"Specifying directories is not supported: '{file}'")
+        else:
+            paths.append(str(relpath))
+    return paths
+
+
+def _filter_from_diff(result: DiffResult, paths_to_keep: Set[str]) -> DiffResult:
```
sfc-gh-bgoel, line 182: From https://github.com/snowflakedb/snowflake-cli/pull/933#discussion_r1539911756, it sounded like we want to keep `stage_diff` untouched, and filter only once the entire result has been returned from the function. This filtering would happen in `sync_deploy_root_with_stage`. Are we deviating from that understanding?

sfc-gh-turbaszek (approved):

RELEASE-NOTES.md
```diff
@@ -22,6 +22,7 @@
 * Added `--if-not-exists` option to `create` commands for `service`, and `compute-pool`. Added `--replace` and `--if-not-exists` options for `image-repository create`.
 * Added support for python connector diagnostic report.
 * Added `snow app deploy` command that creates an application package and syncs the local changes to the stage without creating or updating the application.
+* Added `snow app deploy [files]` option to sync only specific files to the stage. Can also be used to remove files that exists only on the stage.
```
sfc-gh-bdufour, line 25: nit: it's probably just one entry for the entire command

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -229,6 +230,15 @@ def app_teardown(
 @app.command("deploy", requires_connection=True)
 @with_project_definition("native_app")
 def app_deploy(
+    prune: Optional[bool] = typer.Option(
+        False,
```
sfc-gh-bdufour, line 234: Shouldn't default be None so we can tell whether the user specified it explicitly? I'd expect this would be important when no files are provided (`deploy --no-prune` and `deploy` have different semantics)

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -240,6 +250,6 @@ def app_deploy(
     )
 
     manager.build_bundle()
-    manager.deploy()
+    manager.deploy(prune, files)
```
sfc-gh-bdufour, line 253: this is what I meant, isn't this then making deploy without args not prune existing files? That's not what is intended.

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -297,5 +297,5 @@ def process(
             )
             return
 
-        diff = self.deploy()
+        diff = self.deploy(True)
```
sfc-gh-bdufour, line 300: this should be `self.deploy()`, and it should do the right thing by default. prune should default to None, and be resolved to True/False automatically depending on whether files were passed in. Much more intuitive IMHO because the knowledge of special-casing deploy with no args is in a single location.

tests_integration/test_nativeapp.py
```diff
@@ -564,3 +564,172 @@ def test_nativeapp_init_deploy(
                 env=TEST_ENV,
             )
             assert result.exit_code == 0
+
+
+# Tests a simple flow of executing "snow app deploy [files]", verifying that only the specified files are synced to the stage
+@pytest.mark.integration
+def test_nativeapp_init_deploy_files(
```
sfc-gh-bdufour, line 571: The no-arg pruning behaviour for deploy isn't coverage, and needs to be. Let's make sure tests cover the full contract:

- snow app deploy (will prune)
- snow app deploy --prune (will also obviously prune)
- snow app deploy --no-prune (will obviously not prune remote)
- snow app deploy local_file (will not prune)
- snow app deploy local_file --prune/--no-prune

tests_integration/test_nativeapp.py
```diff
@@ -564,3 +564,172 @@ def test_nativeapp_init_deploy(
...
+        finally:
+            # teardown is idempotent, so we can execute it again with no ill effects
+            result = runner.invoke_with_connection_json(
+                ["app", "teardown", "--force"],
+                env=TEST_ENV,
+            )
+            assert result.exit_code == 0
+
+
+# Tests a simple flow of removing a remote-only file from the stage with "snow app deploy remote-file"
```
sfc-gh-bdufour, line 620: you'll remove this in a future update right? This isn't meant to work.

src/snowflake/cli/plugins/object/stage/diff.py
```diff
@@ -154,7 +162,47 @@ def build_md5_map(list_stage_cursor: SnowflakeCursor) -> Dict[str, str]:
...
+            )
+        result.only_on_stage = []
+    return result
+
+
+def stage_diff(
+    local_path: Path,
+    stage_fqn: str,
+    prune: bool = True,
+    files_to_sync: Optional[List[Path]] = None,  # relative to local_path
```
sfc-gh-bdufour, line 204: Didn't we agree to make prune and file filters subsequent passes on the original diff? I don't think this belongs here, let's keep the stage diff logic simple and clean unless there's a really compelling reason to do this here that I'm not seeing.

src/snowflake/cli/plugins/object/stage/diff.py
```diff
@@ -154,7 +162,47 @@ def build_md5_map(list_stage_cursor: SnowflakeCursor) -> Dict[str, str]:
     }
 
 
-def stage_diff(local_path: Path, stage_fqn: str) -> DiffResult:
+def _get_relative_paths_to_sync(
```
sfc-gh-bdufour, line 165: add a doc string to document the contract here

src/snowflake/cli/plugins/object/stage/diff.py
```diff
@@ -154,7 +162,47 @@ def build_md5_map(list_stage_cursor: SnowflakeCursor) -> Dict[str, str]:
...
+                str(file), "This file does not exist either locally or remotely"
+            )
+        elif path.is_dir():
+            raise ClickException(f"Specifying directories is not supported: '{file}'")
+        else:
+            paths.append(str(relpath))
+    return paths
+
+
+def _filter_from_diff(
```
sfc-gh-bdufour, line 183: ditto, doc string for contract

src/snowflake/cli/plugins/object/stage/diff.py
```diff
@@ -154,7 +162,47 @@ def build_md5_map(list_stage_cursor: SnowflakeCursor) -> Dict[str, str]:
     }
 
 
-def stage_diff(local_path: Path, stage_fqn: str) -> DiffResult:
+def _get_relative_paths_to_sync(
+    relative_files_to_sync: List[Path], local_path: Path, remote_paths: Set[str]
+) -> List[str]:
+    paths = []
+    for file in relative_files_to_sync:
+        path = Path(os.path.join(local_path, file))
```
sfc-gh-bdufour, line 170: nit: mixing os.path and pathlib this way is weird. I'd create a Path object for the root dir outside of the loop, and just join the file to it for each iteration. Or just pass both args directly to the ctor, it will join them anyway.

src/snowflake/cli/plugins/object/stage/diff.py
```diff
@@ -154,7 +162,47 @@ def build_md5_map(list_stage_cursor: SnowflakeCursor) -> Dict[str, str]:
     }
 
 
-def stage_diff(local_path: Path, stage_fqn: str) -> DiffResult:
+def _get_relative_paths_to_sync(
+    relative_files_to_sync: List[Path], local_path: Path, remote_paths: Set[str]
```
sfc-gh-bdufour, line 166: local_path is really a root folder here, right? All paths are expected to be relative to it. I'd name it something like local_root to be clear.

tests/__snapshots__/test_help_messages.ambr
```diff
@@ -58,15 +58,26 @@
...
+  │                          [default: sync all local changes to the stage]      │
+  ╰──────────────────────────────────────────────────────────────────────────────╯
   ╭─ Options ────────────────────────────────────────────────────────────────────╮
-  │ --project  -p      TEXT  Path where the Snowflake Native App project         │
-  │                          resides. Defaults to current working directory.     │
-  │ --help     -h            Show this message and exit.                         │
+  │ --prune        --no-prune          If set to true and the specified files    │
+  │                                    exist only remotely, they will be deleted │
+  │                                    from the stage.                           │
+  │                                    [default: no-prune]                       │
```
sfc-gh-bdufour, line 76: That's not really true

tests/__snapshots__/test_help_messages.ambr
```diff
@@ -58,15 +58,26 @@
...
+   Usage: default app deploy [OPTIONS] [FILES]...                                 
                                                                                   
    Creates an application package in your Snowflake account and syncs the local   
    changes to the stage without creating or updating the application.             
                                                                                   
+  ╭─ Arguments ──────────────────────────────────────────────────────────────────╮
+  │   files      [FILES]...  Paths of the files, relative to the deploy root, to │
+  │                          be uploaded to a stage. If the path exists remotely │
+  │                          but not locally, it will be deleted from the stage  │
+  │                          [default: sync all local changes to the stage]      │
```
sfc-gh-bdufour, line 70: I think we need to document that without arguments, deploy is equivalent to deploy --prune. It's really not clear from this text.

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-gbloom: This is the run processor, we don't have a `--prune` flag for `snow app run`

src/snowflake/cli/plugins/nativeapp/commands.py
sfc-gh-gbloom: Yep, you're right! Fixed it

tests_integration/test_nativeapp.py
sfc-gh-gbloom: The test comment missed the "--prune" flag

src/snowflake/cli/plugins/nativeapp/commands.py
sfc-gh-gbloom: I am rewriting this to be relative to project root (which is currently CWD)

tests_integration/test_nativeapp.py
sfc-gh-bdufour: (NB: the team's SOP is to not resolve comments unless they're trivially addressed so that the conversation can happen, please don't resolve immediately in the future)

IMHO these tests are important. Keep in mind that SnowflakeCLI is developed by multiple teams. Having full contract coverage in ITs serves both as a way to prevent regressions, but also as a form of executable documentation. We do leverage parts of the core platform in our logic, so unrelated changes could end up breaking our logic if it's left uncovered. We are generally thorough when it comes to testing new logic. Thoughts?

sfc-gh-bgoel: Bumping this thread, @sfc-gh-gbloom it would be great to have the above test cases all covered, especially because we now have an external team contributing code to our NA plugin and since we do not have the same level of communication with them, these tests will help catch issues early for future development.

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -99,6 +105,39 @@ def ensure_correct_owner(row: dict, role: str, obj_name: str) -> None:
...
+    paths: List[Path], artifacts: List[ArtifactMapping]
+) -> List[Path]:
+    """Maps a list of paths to being relative to deploy root."""
+    new_paths = []
+    for path in paths:
+        new_paths.append(determine_artifacts_file_path(path, artifacts))
+    return new_paths
+
+
+def _get_default_deploy_prune_value(files_argument: Optional[List[Path]]) -> bool:
```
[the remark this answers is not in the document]
sfc-gh-gbloom: That's where I started with, but we have `version_processor` calling `sync_deploy_root_with_stage()` directly.

I preferred adding optional arguments and determine the default value here instead of resolving the default value in 2 different places. WDYT?

tests_integration/test_nativeapp.py
sfc-gh-gbloom: Sure, I added the remaining tests.

I'm all in favor of having full test coverage for our interface, but I think that having an integration test for each permutation is excessive, especially when currently there is a lot of setup/teardown boilerplate.
We could invest in some infra to make these tests slimmer, and I think it's ok for some of the flows to be covered in unit tests.

BTW not that it's not a good goal to have, but I believe we don't have full interface coverage at the moment.

RELEASE-NOTES.md
```diff
@@ -23,7 +23,7 @@
 * Added support for fully qualified image repository names in `spcs image-repository` commands.
 * Added `--if-not-exists` option to `create` commands for `service`, and `compute-pool`. Added `--replace` and `--if-not-exists` options for `image-repository create`.
 * Added support for python connector diagnostic report.
-* Added `snow app deploy` command that creates an application package and syncs the local changes to the stage without creating or updating the application.
+* Added `snow app deploy [files]` command that creates an application package and syncs the local changes to the stage without creating or updating the application. If specific files are not specified, all changes will be synced to the stage.
```
sfc-gh-bgoel, line 26: Nit: we don't need to have `[files]` here, as optional arguments can be excluded.

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -283,3 +283,27 @@ def find_version_info_in_manifest_file(
...
 
     return version_name, patch_name
+
+
+def _determine_artifacts_file_path(
+    source: PurePath, artifacts: List[ArtifactMapping]
+) -> Optional[str]:
+    """Given a source file path (relative to project root) that doesn't necessarily exist on the file system, returns the destination path string for the first matching artifact (relative to deploy root), or None if none matched."""
+    for artifact in artifacts:
+        source_path = PurePath(source)
```
sfc-gh-bgoel, line 293: Nit: you should be able to use `source` directly as is, why do we need source_path?

src/snowflake/cli/plugins/nativeapp/manager.py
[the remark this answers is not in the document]
sfc-gh-bgoel: I see okay, that makes sense.

sfc-gh-bgoel (commented):
lgtm for the code, but will wait for the failing integ tests to pass before approval.

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -283,3 +283,17 @@ def find_version_info_in_manifest_file(
...
     return version_name, patch_name
+
+
+def determine_artifacts_file_path(
+    source: PurePath, artifacts: List[ArtifactMapping]
+) -> Optional[str]:
+    """Given a source file path (relative to project root) that doesn't necessarily exist on the file system, returns the destination path string for the first matching artifact (relative to deploy root), or None if none matched."""
+    for artifact in artifacts:
+        source_path = PurePath(source)
+        if source_path.match(artifact.src):
```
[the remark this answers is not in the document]
sfc-gh-gbloom: I believe that bug doesn't apply here. In this PR I'm not trying to find all matching files (which can cause duplicates), but given a single source path determine the destination path. Hope I'm not missing anything else.

In the example you gave, isn't it the intended behavior? `a/*` shouldn't match `a/b/c.py` -- you have to use `a/**/*` to match directories recursively.

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -283,3 +299,37 @@ def find_version_info_in_manifest_file(
...
+        path_to_target = Path(project_path).relative_to(root)
+        result = Path(path_to_symlink, path_to_target)
+
+        if not result.exists():
+            raise FileNotFoundError(result)
+        return result
+
+    deploy_paths: List[Path] = []
+    for project_path in map(str, project_paths):
+        if project_path in files_mapping:
```
sfc-gh-bdufour, line 331: generally I find it confusing and error-prone that we use strings for project paths. Depending on how these paths are obtained, they may not be in canonical form, but the logic seems to assume they are. Can we make our logic more resilient?

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -229,6 +230,21 @@ def app_teardown(
...
+        None,
+        "--recursive",
+        "-r",
+        help=f"""Controls whether the specified directories should be deployed including all of their contents. When no files are specified (sync all changes), the default is --recursive.""",
+    ),
+    files: Optional[List[Path]] = typer.Argument(
+        default=None,
+        show_default=False,
+        help=f"""Paths of the files, relative to the project root, to be uploaded to a stage. [default: sync all local changes to the stage]""",
+    ),
```
sfc-gh-bdufour, line 247: Wouldn't it be easier to understand if we say that --prune always default to false, --recursive always defaults to false, but deploy with no args is a shorthand for --prune --recursive? This way, there's only one special case: absolutely no arguments (so app deploy --prune implies --no-recursive, is what I'm saying, because an argument was specified and --no-recursive is always the default). Otherwise, you have to worry about files/no-files, which I think is a lot more confusing.

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -283,3 +299,37 @@ def find_version_info_in_manifest_file(
             patch_name = version_info[patch_field]
 
     return version_name, patch_name
+
+
+def project_path_to_deploy_path(
+    project_paths: List[Path], files_mapping: DeployMapping
```
sfc-gh-bdufour, line 305: nit: files_mapping is rather vague, consider a more descriptive name. You call that created_files in a different context too, and that's confusing to me because it's the same thing by a different name.

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -100,6 +104,36 @@ def ensure_correct_owner(row: dict, role: str, obj_name: str) -> None:
...
+            for current_dir, _dirs, files in os.walk(path):
+                for file in files:
+                    deploy_path = Path(current_dir, file).relative_to(deploy_root)
+                    paths.append(str(deploy_path))
+        else:
+            paths.append(str(path.relative_to(deploy_root)))
+    return paths
+
+
+def _assert_no_directories(paths_to_sync):
```
sfc-gh-bdufour, line 121: "assert" is reserved for checking invariants that are never expected to fail, and tends to be disabled in prod (typically for performance). use "verify" here or something similar, as you're checking user input and expect failure as part of normal prod operations (i.e., disabling this check would cause some tests to fail).

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -283,3 +299,37 @@ def find_version_info_in_manifest_file(
             patch_name = version_info[patch_field]
 
     return version_name, patch_name
+
+
+def project_path_to_deploy_path(
```
sfc-gh-bdufour, line 304: nit: this is singular, but acts on a list.

That being said, I do think the list input isn't particularly needed. You could use list comprehension on the caller side and keep this utility function singular for simplicity, something like:

`deploy_paths = [project_path_to_deploy_path(p) for p in project_paths]`

tests/nativeapp/test_manager.py
```diff
@@ -104,6 +111,10 @@ def test_sync_deploy_root_with_stage(
     )
 
 
+def test_sync_deploy_root_with_stage_specific_files():
+    pass
+
+
```
sfc-gh-bdufour, line 117: ?

tests_integration/test_nativeapp.py
```diff
@@ -564,3 +565,269 @@ def test_nativeapp_init_deploy(
...
+            result = runner.invoke_with_connection_json(
+                ["app", "teardown", "--force"],
+                env=TEST_ENV,
+            )
+            assert result.exit_code == 0
+
+
+# Tests that files inside of a symlinked directory are deployed
+@pytest.mark.integration
+def test_nativeapp_deploy_nested_directores(
```
sfc-gh-bdufour, line 696: typo: directories

src/snowflake/cli/plugins/nativeapp/commands.py
sfc-gh-gbloom: Changed it in [this commit](https://github.com/snowflakedb/snowflake-cli/pull/933/commits/8d89da9ef210cf22cc5fe50c48d9af474792912f). WDYT? I'm not sure it's that much simpler, but I'm ok with both ways.

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -283,3 +306,37 @@ def find_version_info_in_manifest_file(
...
+    project_paths: List[Path], files_mapping: DeployMapping
+) -> List[Path]:
+    """Given a list of source paths, returns the deploy destination paths. This function assumes that a build was performed before calling it."""
```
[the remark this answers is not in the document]
sfc-gh-gbloom: I'm verifying that the destination result exists on the filesystem, so the bundle existence is necessary. I could verify it outside of this function and have the coupling somewhere else. Not sure if it will be clearer.

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -310,6 +344,23 @@ def sync_deploy_root_with_stage(self, role: str) -> DiffResult:
...
+        if len(paths_to_sync) > 0:
+            paths_to_sync = [resolve_without_follow(p) for p in paths_to_sync]
+            if not recursive:
+                _verify_no_directories(paths_to_sync)
+            deploy_paths = [
+                project_path_to_deploy_path(p, created_files) for p in paths_to_sync
+            ]
+            paths_to_keep = set(
+                _get_files_to_sync(deploy_paths, self.deploy_root.resolve())
+            )
```
sfc-gh-bdufour, line 358: ok, subjective, but the naming confuses me here... We have deploy_paths, paths_to_sync, paths_to_keep and files_to_sync. Can you replace the names with more descriptive options? I'm losing track of what's what.

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -310,6 +344,23 @@ def sync_deploy_root_with_stage(self, role: str) -> DiffResult:
             % self.deploy_root
         )
         diff: DiffResult = stage_diff(self.deploy_root, self.stage_fqn)
+
+        # If we are syncing specific files/directories, remove everything else from the diff
+        if len(paths_to_sync) > 0:
+            paths_to_sync = [resolve_without_follow(p) for p in paths_to_sync]
```
sfc-gh-bdufour, line 350: nit: Generally not a fan of re-assigning incoming parameters

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -310,6 +344,23 @@ def sync_deploy_root_with_stage(self, role: str) -> DiffResult:
...
+                _verify_no_directories(paths_to_sync)
+            deploy_paths = [
+                project_path_to_deploy_path(p, created_files) for p in paths_to_sync
+            ]
+            paths_to_keep = set(
+                _get_files_to_sync(deploy_paths, self.deploy_root.resolve())
+            )
+            filter_from_diff(diff, paths_to_keep, prune)
+
+        if prune is False:
```
sfc-gh-bdufour, line 361: nit: if not prune? Feels odd to check for False explicitly

tests/stage/test_diff.py
```diff
@@ -215,3 +243,62 @@ def test_sync_local_diff_with_stage(mock_remove, other_directory):
...
+        "only-stage.txt",
+        "only-stage-2.txt",
+    ]
+
+    paths_to_keep = set(["only-stage.txt"])
+    diff = filter_from_diff(diff, paths_to_keep, False)
+
+    assert len(diff.only_on_stage) == 0
+    mock_warning.assert_called_once_with(
+        "The following files exist only on the stage:\n['only-stage.txt']\nUse the --prune flag to delete them from the stage."
```
sfc-gh-bdufour, line 303: The file list could be displayed in a more user-friendly way, please change it.

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -1,14 +1,17 @@
...
+from typing import Dict, List, Optional, Tuple, Union
 
-from click import ClickException
+from click.exceptions import ClickException
 from snowflake.cli.api.constants import DEFAULT_SIZE_LIMIT_MB
 from snowflake.cli.api.project.schemas.native_app.path_mapping import PathMapping
 from snowflake.cli.api.secure_path import SecurePath
 from yaml import safe_load
 
+# Map from source directories and files in the project directory to their path in the deploy directory. Both paths are absolute.
```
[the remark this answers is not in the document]
sfc-gh-gbloom: I started with relative paths, but switched to absolute mostly to canonicalize windows paths (`C:\Users\RUNNER~1\` -> `C:\Users\runneradmin\`).
It also simplifies the code, and doesn't assume that the deploy root is a subdirectory of the project root.

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -100,6 +105,28 @@ def ensure_correct_owner(row: dict, role: str, obj_name: str) -> None:
...
 
+def _get_files_to_sync(paths_to_sync: List[Path], deploy_root: Path) -> List[str]:
+    """Takes a list of paths (files and directories), returning a list of all files recursively, stripping the path to deploy root."""
+    paths = []
+    for path in paths_to_sync:
+        if path.is_dir():
+            for current_dir, _dirs, files in os.walk(path):
+                for file in files:
+                    deploy_path = Path(current_dir, file).relative_to(deploy_root)
+                    paths.append(str(deploy_path))
```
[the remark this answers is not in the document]
sfc-gh-gbloom: Good point. I convert these to a set downstream, didn't realize `Set[Path]` works

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-gbloom: Changed everything to `paths_to_sync`

src/snowflake/cli/plugins/nativeapp/artifacts.py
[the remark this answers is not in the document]
sfc-gh-bdufour: Well, the deploy root could be anything, we should never assume that it's within the project directory.

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -310,6 +350,25 @@ def sync_deploy_root_with_stage(self, role: str) -> DiffResult:
             % self.deploy_root
         )
         diff: DiffResult = stage_diff(self.deploy_root, self.stage_fqn)
+
+        # If we are syncing specific files/directories, remove everything else from the diff
+        if len(paths_to_sync) > 0:
+            resolved_paths_to_sync = [resolve_without_follow(p) for p in paths_to_sync]
+            if not recursive:
+                _verify_no_directories(resolved_paths_to_sync)
```
sfc-gh-bdufour, line 358: shouldn't this apply even when paths to sync isn't empty? At this point, the CLI options are irrelevant, we should honour the arguments that were passed to this call without any assumptions.

src/snowflake/cli/plugins/stage/diff.py
```diff
@@ -149,12 +152,34 @@ def build_md5_map(list_stage_cursor: SnowflakeCursor) -> Dict[str, str]:
...
 
-def stage_diff(local_path: Path, stage_fqn: str) -> DiffResult:
+def filter_from_diff(
+    diff: DiffResult, paths_to_sync: Set[str], prune: bool
+) -> DiffResult:
+    """Modifies the given diff, keeping only the provided paths. If prune is false, remote-only paths will be empty."""
+    diff.different = [i for i in diff.different if i in paths_to_sync]
+    diff.only_local = [i for i in diff.only_local if i in paths_to_sync]
+    if prune:
+        diff.only_on_stage = [i for i in diff.only_on_stage if i in paths_to_sync]
```
sfc-gh-bdufour, line 167: you handle prune in a different location too, sounds odd to have to different places where we try to give that flag some semantics. I don't think the one here is needed, right?

src/snowflake/cli/plugins/stage/diff.py
```diff
@@ -149,12 +152,34 @@ def build_md5_map(list_stage_cursor: SnowflakeCursor) -> Dict[str, str]:
...
+    else:
+        files_not_removed = [i for i in diff.only_on_stage if i in paths_to_sync]
+        if len(files_not_removed) > 0:
+            files_not_removed_str = "\n".join(files_not_removed)
+            cc.warning(
+                f"The following files exist only on the stage:\n{files_not_removed_str}\n\nUse the --prune flag to delete them from the stage."
+            )
```
sfc-gh-bdufour, line 174: sorry, didn't occur to me before, but this doesn't belong in a utility method. We're inferring an action from a filter operation, but we shouldn't. This method should be reusable in the future.

tests/nativeapp/test_artifacts.py
```diff
@@ -163,3 +168,51 @@ def test_too_many_files(project_definition_files):
...
+    # Source files
+    touch("file")
+    touch("dir/nested_file1")
+    touch("dir/nested_dir/nested_file2")
+    # Build
+    os.mkdir("deploy")
+    os.symlink("file", "deploy/file")
+    os.symlink(Path("dir").resolve(), Path("deploy/dir"))
+
+    files_mapping = {
```
sfc-gh-bdufour, line 212: These tests don't cover the full contract. I'd add test cases where the directory was renamed too, and a case where no mapping is found.

tests_integration/test_nativeapp.py
```diff
@@ -564,3 +565,264 @@ def test_nativeapp_init_deploy(
...
+    with pushd(Path(os.getcwd(), project_name)):
+        result = runner.invoke_with_connection_json(
+            ["app", "deploy"],
+            env=TEST_ENV,
+        )
+        assert result.exit_code == 0
+
+        try:
+            # delete a file locally
+            os.remove(os.path.join("app", "manifest.yml"))
```
sfc-gh-bdufour, line 606: nit: consider deleting a different file, the fact that we don't check for a manifest now is somewhat of a bug.

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -237,9 +244,11 @@ def build_bundle(
             if len(source_paths) == 1:
                 # copy a single file as the given destination path
                 symlink_or_copy(source_paths[0], dest_path)
+                created_files[source_paths[0].resolve()] = dest_path
             else:
                 # refuse to map multiple source files to one destination (undefined behaviour)
                 raise TooManyFilesError(dest_path)
+    return created_files
```
sfc-gh-bdufour, line 251: nit: I keep getting back to this name, and something feels off. I think it's because in my mind, bundle is something that is essentially acting on a diff. So it becomes unclear here if "created" means newly added files, or all files that are supposed to be mapped. It's the latter of course, reading the code, but the name should reflect that. How about something a lot more generic like "deploy_root_file_mapping" / "deploy_root_files" / "mapped_files" etc? Might want the team's input on that one, naming is hard.

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-gbloom: You mean apply when the paths _are_ empty? (full deploy)
Yes, added this verification in `build_bundle()` too

src/snowflake/cli/plugins/stage/diff.py
sfc-gh-gbloom: This is the deploying specific files flow. The full deploy flow doesn't run this function. I can consider moving the other check to run only on a full deploy.

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-bdufour: Oops, yeah, I said the opposite of what I meant.

tests/nativeapp/test_artifacts.py
sfc-gh-gbloom: When the user specifies a file that exists in project root but isn't mapped by any artifact pattern, `FileNotFoundError` will be raised. Added a test for it.

Not sure I understand the renamed directory test -- can you explain?

sfc-gh-bdufour: I mean foo in src goes to bar in dest. To make sure we compose paths correctly. Currently your tests assume src & dest have the same names.

sfc-gh-gbloom: Got it. Changed the file/dir names in these tests to demonstrate they don't have to match.

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-gbloom: Renamed to `mapped_files`

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -228,7 +235,9 @@ def build_bundle(
 
             # copy all files as children of the given destination path
             for source_path in source_paths:
-                symlink_or_copy(source_path, dest_path / source_path.name)
+                dest_child_path = dest_path / source_path.name
+                symlink_or_copy(source_path, dest_child_path)
+                mapped_files[source_path.resolve()] = dest_child_path
```
sfc-gh-bgoel, line 240: Might be related to the discussion above, but why:
1. Is `source_path` being resolved?
2. Is only `source_path` being resolved but not `dest_child_path`?

Also, can you clarify the following:
If the user specified a glob pattern, then `source_path` will be the directories or files that qualify the source pattern. In case of directories, even though their contents are symlinked as well, `mapped_files` will only contain a kv pair of the directory to the dest, and not its contents.

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -283,3 +294,31 @@ def find_version_info_in_manifest_file(
             patch_name = version_info[patch_field]
 
     return version_name, patch_name
+
+
+def project_path_to_deploy_path(
+    project_path: Path, mapped_files: ArtifactDeploymentMap
```
sfc-gh-bgoel, line 300: Nit: is this `project_path` or `source_path`? `project_path` can be misleading to only mean the path to the project, i.e. root of the project.

src/snowflake/cli/plugins/stage/diff.py
```diff
@@ -149,12 +151,30 @@ def build_md5_map(list_stage_cursor: SnowflakeCursor) -> Dict[str, str]:
...
+    diff: DiffResult, paths_to_sync: Set[str], prune: bool
+) -> List[str]:
+    """Modifies the given diff, keeping only the provided paths. If prune is false, remote-only paths will be empty and the non-removed paths will be returned."""
+    diff.different = [i for i in diff.different if i in paths_to_sync]
+    diff.only_local = [i for i in diff.only_local if i in paths_to_sync]
+    files_not_removed = []
+    if prune:
+        diff.only_on_stage = [i for i in diff.only_on_stage if i in paths_to_sync]
+    else:
+        files_not_removed = [i for i in diff.only_on_stage if i in paths_to_sync]
```
sfc-gh-bgoel, line 169: Lines 167 and 169 rely on the same list comprehension. Consider assigning it to a placeholder/variable before checking for prune, to reduce duplicating code.

src/snowflake/cli/plugins/nativeapp/artifacts.py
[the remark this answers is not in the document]
sfc-gh-cgorrie: In an ideal world, these would be relative to their respective roots, but it's documented well and if we need them we can always use `relative_to`.

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-cgorrie: Agreed, I think `source_path` is a better name.

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-gbloom: 1. Both src/dest parts in `ArtifactDeploymentMap` are resolved, mostly to canonicalize windows paths `(C:\Users\RUNNER~1\` -> `C:\Users\runneradmin\`)
2. `dest_child_path` is also resolved (`dest_path` is resolved a few lines above)

Yes, `mapped_files` won't contain the contents of symlinked directories. It's only the files that we actually copied to the deploy directory.
You can see the logic in `project_path_to_deploy_path()` for cases where we look for a file under a symlinked directory.

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -229,17 +230,44 @@ def app_teardown(
 @app.command("deploy", requires_connection=True)
 @with_project_definition("native_app")
 def app_deploy(
+    prune: Optional[bool] = typer.Option(
+        default=None,
+        help=f"""Controls whether files that exist only remotely will be deleted from the stage.""",
```
sfc-gh-cgorrie, line 235:
```suggestion
        help=f"""If set, matched files that do not exist locally will also be deleted from the stage.""",
```

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -100,6 +106,20 @@ def ensure_correct_owner(row: dict, role: str, obj_name: str) -> None:
         raise UnexpectedOwnerError(obj_name, role, actual_owner)
 
 
+def _get_paths_to_sync(paths_to_sync: List[Path], deploy_root: Path) -> List[str]:
+    """Takes a list of paths (files and directories), returning a list of all files recursively, stripping the path to deploy root."""
```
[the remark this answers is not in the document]
sfc-gh-gbloom: This function won't convert Windows paths to Unix paths. Noted that the returned paths are relative to deploy root.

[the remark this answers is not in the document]
sfc-gh-cgorrie: We are deferring the windows fix to another PR, yes?

sfc-gh-mheavin (commented):
Left some thoughts for your consideration

src/snowflake/cli/plugins/nativeapp/manager.py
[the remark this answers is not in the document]
sfc-gh-gbloom: Yep

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -100,6 +106,20 @@ def ensure_correct_owner(row: dict, role: str, obj_name: str) -> None:
         raise UnexpectedOwnerError(obj_name, role, actual_owner)
 
 
+def _get_paths_to_sync(paths_to_sync: List[Path], deploy_root: Path) -> List[str]:
```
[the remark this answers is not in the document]
sfc-gh-bgoel: Yes, not visible to the customer using the CLI. But visible to the public in general since this is open source. 
Regardless, good idea to still have some information on what this function does.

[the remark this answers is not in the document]
sfc-gh-gbloom: yep, internal

sfc-gh-cgorrie (approved):

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -229,17 +230,44 @@ def app_teardown(
 @app.command("deploy", requires_connection=True)
 @with_project_definition("native_app")
 def app_deploy(
+    prune: Optional[bool] = typer.Option(
+        default=None,
+        help=f"""Whether to delete specified files from the stage if they don't exist locally. If set, the command deletes files that exist in the stage, but not in the local filesystem.""",
+    ),
+    recursive: Optional[bool] = typer.Option(
+        None,
```
sfc-gh-turbaszek, line 238: Personally I would use `False` as default for boolean value

sfc-gh-turbaszek (approved):

src/snowflake/cli/plugins/nativeapp/commands.py
sfc-gh-gbloom: I needed it to be `None` here because there is the special flow of running `snow app deploy` with no arguments, in which case the default value is `True`

PR #944, opened 2024-03-27
Fixed version parsing in AnacondaChannel

sfc-gh-pczajka:
### Pre-review checklist

### Changes description
- Change AnacondaChannel to support multiple versions of a package
- Use `packaging.requirements` to check whether package version fulfills specs
- Moved AnacondaChannel tests to `test_anaconda.py` instead of `test_utils`
- Fix package comparison in `snowpark.common. _get_snowflake_packages_delta`


src/snowflake/cli/plugins/snowpark/package/anaconda.py
```diff
@@ -1,45 +1,79 @@
...
     )
 
-    def __init__(self, packages):
-        self._packages = packages
+    def __init__(self, packages: Dict[str, Set[str]]):
+        """[packages] should be a dictionary mapping package name to set of its available versions"""
+        self._packages = {
+            _standarize_name(package_name): versions
+            for package_name, versions in packages.items()
+        }
```
sfc-gh-turbaszek, line 35: We are performing `for` loop here and `from_snowflake` I think it would be much better to encapsulate this logic in a single place (probably in `from_snowflake`). Using `for` loop instead of comprehensions may be better in such case. It will also make tests much cleaner (although you need to adjust them).

src/snowflake/cli/plugins/snowpark/package/anaconda.py
```diff
@@ -1,45 +1,79 @@
...
+            return any(
+                parse(version) in package_specifiers
+                for version in self._packages[package_name]
+            )
+        except (InvalidVersion, InvalidRequirement):
+            # fail-safe for non-pep508 formats
+            return False
+
+    def package_latest_version(self, package: Requirement) -> str | None:
+        package_name = _standarize_name(package.name)
```
sfc-gh-turbaszek, line 57: We are using this every time we work with package. Should we consider creating our own custom `Requirement` that would implement this as property?

src/snowflake/cli/plugins/snowpark/package/anaconda.py
```diff
@@ -1,45 +1,79 @@
...
+        if package_name not in self._packages:
+            return None
+        try:
+            versions = {parse(v) for v in self._packages[package_name]}
+        except InvalidVersion:
+            versions = self._packages[package_name]
+        return str(max(versions))
```
sfc-gh-turbaszek, line 64: Are we sure the max work properly here?

src/snowflake/cli/plugins/snowpark/package_utils.py
```diff
@@ -38,11 +37,13 @@ def parse_requirements(
...
     reqs: List[Requirement] = []
     if requirements_file.exists():
-        with requirements_file.open(
-            "r", read_file_limit_mb=DEFAULT_SIZE_LIMIT_MB, encoding="utf-8"
-        ) as f:
-            for req in requirements.parse(f):
-                reqs.append(req)
+        for line in requirements_file.read_text(
+            file_size_limit_mb=DEFAULT_SIZE_LIMIT_MB
+        ).splitlines():
```
sfc-gh-turbaszek, line 42: Isn't there `readlines` method?

src/snowflake/cli/plugins/snowpark/package_utils.py
```diff
@@ -38,11 +37,13 @@ def parse_requirements(
...
+            # remove comments
+            line = line.split("#")[0].strip()
```
sfc-gh-turbaszek, line 44: Wouldn't `re.sub` be a bit cleaner?

src/snowflake/cli/plugins/snowpark/package_utils.py
```diff
@@ -72,7 +73,11 @@ def get_snowflake_packages() -> List[str]:
         with requirements_file.open(
             "r", read_file_limit_mb=DEFAULT_SIZE_LIMIT_MB, encoding="utf-8"
         ) as f:
-            return [req for line in f if (req := line.split("#")[0].strip())]
+            return [
+                Requirement.parse(req).to_name_and_version()
+                for line in f
+                if (req := line.split("#")[0].strip())
```
sfc-gh-turbaszek, line 79: Duplicated code fragment, consider introducing a method for cleaning comments.

tests/snowpark/test_anaconda.py
```diff
@@ -0,0 +1,113 @@
...
+    assert split_requirements.snowflake[0].specs == [("==", "1.4.4")]
+    assert split_requirements.other[0].name == "FuelSDK"
+    assert split_requirements.other[0].specifier is True
+    assert split_requirements.other[0].specs == [(">=", "0.9.3")]
+    assert split_requirements.other[1].name == "Pamela"
+    assert split_requirements.other[1].specs == [("==", "1.0.1")]
+
+
+@patch("snowflake.cli.plugins.snowpark.package.anaconda.requests")
+def test_anaconda_packages_streamlit(mock_requests):
```
sfc-gh-turbaszek, line 84: What's the purpose of this test?

src/snowflake/cli/plugins/snowpark/package/anaconda.py
sfc-gh-pczajka: For parsed versions - yes
For non-pep508 - might work not so good. The Anaconda example was `jpeg` package, which was available in versions `9c`, `9d` and `9e`.

As it is only used in in `snowpark package lookup`, I think good alternative might be simply list all possible versions for every package, WDYT?

tests/snowpark/test_anaconda.py
sfc-gh-pczajka: ¯\_(ツ)_/¯
I got it moved from `test_common`, it looks like a duplicate of `test_anaconda_packages`. Removed

src/snowflake/cli/plugins/snowpark/package/anaconda.py
sfc-gh-pczajka: Fixed.

src/snowflake/cli/plugins/snowpark/package/anaconda.py
sfc-gh-pczajka: Done. I've added pre-commit hook checking whether `requirements` library is not used instead

sfc-gh-turbaszek: Have you pushed the changes?

sfc-gh-pczajka: done

src/snowflake/cli/plugins/snowpark/package_utils.py
sfc-gh-pczajka: fixed

src/snowflake/cli/plugins/snowpark/package_utils.py
sfc-gh-pczajka: After refactor it is used only in `snowflake.cli.plugins.snowpark.package_utils.parse_requirements`

src/snowflake/cli/plugins/snowpark/package_utils.py
sfc-gh-pczajka: Not in pathlib - I'd have to convert that back to "with file.open(...)"

src/snowflake/cli/plugins/snowpark/package/anaconda.py
sfc-gh-pczajka: Done :)

src/snowflake/cli/plugins/snowpark/commands.py
```diff
@@ -363,11 +380,15 @@ def build(
...
             )
+            anaconda: Optional[AnacondaChannel] = (
+                None if ignore_anaconda else AnacondaChannel.from_snowflake()
+            )
 
             download_result = package_utils.download_unavailable_packages(
                 requirements=requirements,
                 target_dir=packages_dir,
                 ignore_anaconda=ignore_anaconda,
+                anaconda=anaconda,
```
sfc-gh-turbaszek, line 391: Does this method requires anaconda or packages available in anaconda? If the latter is true the we can do something like
```
anaconda_packages = [] if ignore_anaconda else AnacondaChannel.from_snowflake().packages

download_result = package_utils.download_unavailable_packages(anaconda_packages=anaconda_packages, ...)
```

In this way we have clear interface. WDYT?

src/snowflake/cli/plugins/snowpark/commands.py
```diff
@@ -390,9 +411,10 @@ def build(
                         "Try again with --allow-shared-libraries."
                     )
             if download_result.packages_available_in_anaconda:
-                _write_requirements_file(
-                    snowpark_paths.snowflake_requirements_file,
-                    download_result.packages_available_in_anaconda,
+                _write_snowflake_requirements_file(
+                    file_path=snowpark_paths.snowflake_requirements_file,
+                    anaconda=anaconda,
```
sfc-gh-turbaszek, line 416: Same here, do we really need anaconda?

src/snowflake/cli/plugins/snowpark/common.py
```diff
@@ -18,28 +20,25 @@ def check_if_replace_is_required(
...
     stage_artifact_file: str,
 ) -> bool:
     import logging
 
     log = logging.getLogger(__name__)
     resource_json = _convert_resource_details_to_dict(current_state)
-    anaconda_packages = resource_json["packages"]
+    deployed_packages = resource_json["packages"]
     log.info(
         "Found %s defined Anaconda packages in deployed %s...",
```
sfc-gh-turbaszek, line 33:
```suggestion
        "Found %d defined Anaconda packages in deployed %s...",
```
Nit which we can fix using this occasion

src/snowflake/cli/plugins/snowpark/models.py
```diff
@@ -46,8 +55,8 @@ class SplitRequirements:
     snowflake-supported vs other packages.
     """
 
-    snowflake: List[Requirement]
-    other: List[Requirement]
+    in_snowflake: List[Requirement]
+    unavailable: List[Requirement]
```
sfc-gh-turbaszek, line 59: One of the most important changes, love it! 🚀

src/snowflake/cli/plugins/snowpark/package/anaconda.py
sfc-gh-pczajka: @sfc-gh-turbaszek changed in https://github.com/snowflakedb/snowflake-cli/pull/944/commits/45bd493fa94437d09da0ad48ef32e232c4758091

src/snowflake/cli/plugins/snowpark/common.py
```diff
@@ -18,28 +20,25 @@ def check_if_replace_is_required(
...
         object_type,
     )
     log.info("Checking if app configuration has changed...")
-    updated_package_list = _get_snowflake_packages_delta(
-        anaconda_packages,
-    )
 
-    if updated_package_list:
-        diff = len(updated_package_list) - len(anaconda_packages)
+    if _snowflake_requirements_differ(deployed_packages, packages):
```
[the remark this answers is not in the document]
sfc-gh-pczajka: It is contents of `requirements.snowflake.txt`, which is `packages` argument for `create or replace`. It is named `packages` through the whole `deploy` logic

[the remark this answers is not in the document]
sfc-gh-turbaszek: Should we name this more appropriately? I think ambiguous names were one of the main problems of this part of code

src/snowflake/cli/plugins/snowpark/common.py
sfc-gh-pczajka: done

src/snowflake/cli/plugins/snowpark/commands.py
sfc-gh-pczajka: Refactored - I moved this to AnacondaChannel

src/snowflake/cli/plugins/snowpark/common.py
sfc-gh-pczajka: done

src/snowflake/cli/plugins/snowpark/commands.py
sfc-gh-pczajka: Refactored - I've added `AnacondaChannel.empty()`, which ignores all packages. This way `download_unavailable_packages` can forget about `ignore_anacoda=true` case

sfc-gh-turbaszek: Nice and clean 🚀

sfc-gh-turbaszek (approved):

PR #942, opened 2024-03-27
Stage copy with directory structure

sfc-gh-astus:
### Pre-review checklist
   * [x] I've confirmed that instructions included in README.md are still correct after my changes in the codebase.
   * [x] I've added or updated automated unit tests to verify correctness of my new code.
   * [x] I've added or updated integration tests to verify correctness of my new code.
   * [x] I've confirmed that my changes are working by executing CLI's commands manually.
   * [x] I've confirmed that my changes are up-to-date with the target branch.
   * [x] I've described my changes in the release notes.
   * [x] I've described my changes in the section below.

### Changes description
`stage copy` from stage will reproduce the directory structure locally.

PR #949, opened 2024-03-29
Added execute immediate command

sfc-gh-astus:
### Pre-review checklist

### Changes description

Added command for execute immediate `snow object stage execute`


tests/__snapshots__/test_snow_connector.ambr
```diff
@@ -1,53 +0,0 @@
-# serializer version: 1
```
sfc-gh-astus, line 1: Old, not used snapshots 🤷

tests/__snapshots__/test_help_messages.ambr
```diff
@@ -916,405 +916,22 @@
   
   '''
 # ---
-# name: test_help_messages[git.copy]
```
sfc-gh-astus, line 919: `--snapshot-update` 😱

tests/object/stage/__snapshots__/test_stage.ambr
```diff
@@ -23,6 +23,45 @@
...
+  exe/s1.sql - SUCCESS
+  exe/s2.sql - FAILURE
+  exe/s3.sql - SUCCESS
+  +------------------------------+
+  | File       | Status  | Error |
+  |------------+---------+-------|
+  | exe/s1.sql | SUCCESS | None  |
+  | exe/s2.sql | FAILURE | Error |
+  | exe/s3.sql | SUCCESS | None  |
+  +------------------------------+
```
sfc-gh-turbaszek, line 37: I'm wondering if we really want to show duplicated info, won't table be enough?

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -137,3 +147,97 @@ def create(self, stage_name: str, comment: Optional[str] = None) -> SnowflakeCur
...
+        sql_parameters = self._parse_execute_parameters(parameters)
+        results = []
+        for file in sorted_list:
+            results.append(self._call_execute_immediate(file, sql_parameters, on_error))
+
+        return results
+
+    def _get_files_list_from_stage(self, stage_path: str) -> List[str]:
+        stage_name = stage_path.split("/")[0]
+        files_list_result = self.list_files(stage_name).fetchall()
```
sfc-gh-turbaszek, line 172: Weren't we supposed to use `iter_stage` from #942?

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -137,3 +147,97 @@ def create(self, stage_name: str, comment: Optional[str] = None) -> SnowflakeCur
...
+
+        sql_parameters = self._parse_execute_parameters(parameters)
+        results = []
+        for file in sorted_list:
+            results.append(self._call_execute_immediate(file, sql_parameters, on_error))
+
+        return results
+
+    def _get_files_list_from_stage(self, stage_path: str) -> List[str]:
+        stage_name = stage_path.split("/")[0]
```
sfc-gh-turbaszek, line 171: Use `self.get_stage_name_from_path`

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -137,3 +147,97 @@ def create(self, stage_name: str, comment: Optional[str] = None) -> SnowflakeCur
...
+        if len(files_list_result) == 0:
+            raise ClickException(f"No files found on stage '{stage_name}'")
+
+        return [f["name"] for f in files_list_result]
+
+    def _filter_files_list(
+        self, stage_path: str, files_on_stage: List[str]
+    ) -> List[str]:
+        if stage_path.startswith("@"):
+            stage_path = stage_path[1:]
```
sfc-gh-turbaszek, line 183: Encapsulate in a method

src/snowflake/cli/plugins/object/stage/manager.py
sfc-gh-pczajka, line 172: If I'm correct, `list_files` should already work well with path's subdirectories, just the argument name is unfortunate:
```
▶ snow object stage list XD/fromdisc
ls @XD/fromdisc
+---------------------------------------------------------------------------------------------------------+
| name                          | size | md5                              | last_modified                 |
|-------------------------------+------+----------------------------------+-------------------------------|
| xd/fromdisc/file1_on_main.txt | 16   | eaa66e3eff4a7c1f310314fc0d7e44ed | Tue, 27 Feb 2024 16:44:10 GMT |
| xd/fromdisc/file2_on_main.txt | 32   | 2ee58a50ba86be95e57dcaff04a038b3 | Tue, 27 Feb 2024 16:44:09 GMT |
+---------------------------------------------------------------------------------------------------------+
```
We can avoid filtering result and simplify logic

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -137,3 +147,97 @@ def create(self, stage_name: str, comment: Optional[str] = None) -> SnowflakeCur
...
+
+        if not stage_path.__contains__("/"):
+            filtered_list = files_on_stage
+        elif has_magic(stage_path):
```
sfc-gh-turbaszek, line 187:
```suggestion
        elif glob.has_magic(stage_path):
```
Consider doing this way to make it easier to understand

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -137,3 +147,97 @@ def create(self, stage_name: str, comment: Optional[str] = None) -> SnowflakeCur
...
+
+    def execute(
+        self,
+        stage_path: str,
+        on_error: OnErrorType,
+        parameters: Optional[List[str]] = None,
+    ):
+        files_list = self._get_files_list_from_stage(stage_path)
+        filtered_list = self._filter_files_list(stage_path, files_list)
+        sorted_list = sorted(
```
sfc-gh-pczajka, line 159:
```suggestion
        sorted_file_list = sorted(
```

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -137,3 +147,97 @@ def create(self, stage_name: str, comment: Optional[str] = None) -> SnowflakeCur
...
+
+        if not stage_path.__contains__("/"):
+            filtered_list = files_on_stage
+        elif has_magic(stage_path):
+            filtered_list = fnmatch.filter(files_on_stage, stage_path)
+        else:
+            if files_on_stage.__contains__(stage_path):
+                filtered_list = [stage_path]
+            else:
+                filtered_list = fnmatch.filter(files_on_stage, f"{stage_path}*")
```
sfc-gh-turbaszek, line 193: Can't it be `elif / else`?

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -137,3 +147,97 @@ def create(self, stage_name: str, comment: Optional[str] = None) -> SnowflakeCur
...
+
+        if len(filtered_list) == 0:
```
sfc-gh-turbaszek, line 195:
```suggestion
        if not filtered_list:
```

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -137,3 +147,97 @@ def create(self, stage_name: str, comment: Optional[str] = None) -> SnowflakeCur
...
+            raise ClickException(f"No files matched pattern '{stage_path}'")
+
+        return filtered_list
+
+    def _stage_files_comparator(self, o1, o2):
```
sfc-gh-turbaszek, line 200: What `o1` and `o2` stands for?

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -137,3 +147,97 @@ def create(self, stage_name: str, comment: Optional[str] = None) -> SnowflakeCur
...
+        o2_directories = o2.count("/")
+
+        if o1_directories > o2_directories:
+            return 1
+        elif o1_directories < o2_directories:
+            return -1
+        else:
+            if o1 >= o2:
+                return 1
+            else:
```
sfc-gh-turbaszek, line 211: Same, `elif / else`?

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -137,3 +147,97 @@ def create(self, stage_name: str, comment: Optional[str] = None) -> SnowflakeCur
...
+        self,
+        stage_path: str,
+        on_error: OnErrorType,
+        parameters: Optional[List[str]] = None,
+    ):
+        files_list = self._get_files_list_from_stage(stage_path)
+        filtered_list = self._filter_files_list(stage_path, files_list)
+        sorted_list = sorted(
+            filtered_list, key=cmp_to_key(self._stage_files_comparator)
+        )
```
sfc-gh-turbaszek, line 161: Consider adding comments to explain what we are filtering and sorting

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -137,3 +147,97 @@ def create(self, stage_name: str, comment: Optional[str] = None) -> SnowflakeCur
...
             query += f" comment='{comment}'"
         return self._execute_query(query)
+
+    def execute(
+        self,
+        stage_path: str,
+        on_error: OnErrorType,
+        parameters: Optional[List[str]] = None,
+    ):
+        files_list = self._get_files_list_from_stage(stage_path)
```
sfc-gh-pczajka, line 157:
```suggestion
        all_files_list = self._get_files_list_from_stage(stage_path)
```

src/snowflake/cli/plugins/object/stage/commands.py
```diff
@@ -121,3 +123,26 @@ def stage_diff(
     """
     diff: DiffResult = stage_diff(Path(folder_name), stage_name)
     return ObjectResult(str(diff))
+
+
+@app.command("execute", requires_connection=True)
+def execute(
+    stage_path: str = typer.Argument(
+        None, help="Execute immediate all files from the path."
```
sfc-gh-pczajka, line 131: Do we want to hint that patterns like `@stage/*.sql` work? After reading only help messages I thought it was not a case

tests/object/stage/__snapshots__/test_stage.ambr
sfc-gh-astus: Messages are printed immediately after receiving the result, so user can see the progress. `--silent` flag hides messages. WDYT?

src/snowflake/cli/plugins/object/stage/manager.py
sfc-gh-astus: I hoped so, but it was easier to filter files as list and you wanted to execute them in order, so I need whole list to sort it.

I can use `iter_stage` filter files one by one, collect and sort, but filtering will not be that efficient like in current way.

WDYT?

src/snowflake/cli/plugins/object/stage/manager.py
sfc-gh-astus: Done

src/snowflake/cli/plugins/object/stage/manager.py
sfc-gh-astus: Agree

src/snowflake/cli/plugins/object/stage/manager.py
sfc-gh-astus: It can 😄

src/snowflake/cli/plugins/object/stage/manager.py
sfc-gh-astus: `object1`, `object2`

Renamed them

src/snowflake/cli/plugins/object/stage/commands.py
sfc-gh-astus: Good idea

sfc-gh-turbaszek: We should also have integrations tests for such cases, in fact it would be good to have a template of project where we can test this

sfc-gh-astus: I think unit test is enough, filter logic is on our side and base on it we execute files on stage. Unit test should cover all or almost all cases https://github.com/snowflakedb/snowflake-cli/pull/949/files#diff-cee995a234660b328c01de5358fc05ea47cd49d77aee3df528787cac5b292e99R484 . If you see any more cases let me know.

tests/object/stage/__snapshots__/test_stage.ambr
sfc-gh-astus: Only if it is the same query (`select * from table`), in this case is many queries, so can't stream table. This is how sql command works. If we really want to stream table we should create separate ticket for it.

sfc-gh-turbaszek: Let's create the ticket, having option to stream to a table from generic generator not only cursor will be useful

src/snowflake/cli/plugins/object/stage/commands.py
```diff
@@ -121,3 +123,27 @@ def stage_diff(
...
+        help="What to do when an error occurs. Defaults to break.",
+    ),
+    parameters: Optional[List[str]] = typer.Option(
+        None,
+        help='Parameters for the template. For example: `--parameters "<key>=<value>"`, string values should be in single quotes.',
+    ),
+    **options,
+):
+    """
+    Execute immediate all files from the stage path.
```
sfc-gh-turbaszek, line 146: Let's add info about glob pattern support

src/snowflake/cli/plugins/object/stage/commands.py
```diff
@@ -121,3 +123,27 @@ def stage_diff(
...
+    """
+    results = StageManager().execute(stage_path, on_error, parameters)
```
sfc-gh-turbaszek, line 148: Pass argument using keywords, it's easier to refactor later on

src/snowflake/cli/plugins/object/stage/commands.py
```diff
@@ -121,3 +123,27 @@ def stage_diff(
     """
     diff: DiffResult = stage_diff(Path(folder_name), stage_name)
     return ObjectResult(str(diff))
+
+
+@app.command("execute", requires_connection=True)
+def execute(
+    stage_path: str = typer.Argument(
+        None,
+        help="Execute immediate all files from the path. Files can be filtered with glob like pattern, e.g. `@stage/**/*.sql`.",
```
sfc-gh-turbaszek, line 132:
```suggestion
        help="Stage path with files to be execute. For example @stage/dev.",
```

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -54,20 +66,24 @@ def quote_stage_name(name: str) -> str:
 
         return standard_name
 
+    @staticmethod
+    def remove_stage_prefix(stage_path: str) -> str:
+        if stage_path.startswith("@"):
```
sfc-gh-turbaszek, line 71: Should we transform the stage path with `get_standard_stage_prefix` first or are we sure we won't get `snow://` here?

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -137,3 +153,96 @@ def create(self, stage_name: str, comment: Optional[str] = None) -> SnowflakeCur
...
+    def _filter_files_list(
+        self, stage_path: str, files_on_stage: List[str]
+    ) -> List[str]:
+        stage_path = self.remove_stage_prefix(stage_path)
+
+        if not stage_path.__contains__("/"):
+            filtered_list = files_on_stage
+        elif glob.has_magic(stage_path):
+            filtered_list = fnmatch.filter(files_on_stage, stage_path)
+        elif files_on_stage.__contains__(stage_path):
```
sfc-gh-turbaszek, line 196:
```suggestion
        elif stage_path in files_on_stage:
```
Same here

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -137,3 +153,96 @@ def create(self, stage_name: str, comment: Optional[str] = None) -> SnowflakeCur
...
+            filtered_list = [stage_path]
+        else:
+            filtered_list = fnmatch.filter(files_on_stage, f"{stage_path}*")
+
+        if not filtered_list:
+            raise ClickException(f"No files matched pattern '{stage_path}'")
+
+        return filtered_list
```
sfc-gh-turbaszek, line 204:
```suggestion
        if "/" not in stage_path:
            return files_on_stage
        
        if glob.has_magic(stage_path):
            return fnmatch.filter(files_on_stage, stage_path)
        
        if stage_path in files_on_stage:
            return[stage_path]
        
        return fnmatch.filter(files_on_stage, f"{stage_path}*")
```

and move
```py
if not filtered_list:
    raise ClickException(f"No files matched pattern '{stage_path}'")
```
To the top level logic. WDYT?

Also consider adding comments about what we do in all of the if cases

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -137,3 +153,96 @@ def create(self, stage_name: str, comment: Optional[str] = None) -> SnowflakeCur
...
+
+    def _stage_files_comparator(self, file_path_1, file_path_2):
+        file_path_1_directories = file_path_1.count("/")
+        file_path_2_directories = file_path_2.count("/")
```
sfc-gh-turbaszek, line 208:
```suggestion
        file_1_depth = file_path_1.count("/")
        file_2_depth = file_path_2.count("/")
```
WDYT?

src/snowflake/cli/plugins/object/stage/commands.py
```diff
@@ -121,3 +123,27 @@ def stage_diff(
...
+    on_error: OnErrorType = typer.Option(
+        OnErrorType.BREAK.value,
+        "--on-error",
+        help="What to do when an error occurs. Defaults to break.",
+    ),
+    parameters: Optional[List[str]] = typer.Option(
+        None,
+        help='Parameters for the template. For example: `--parameters "<key>=<value>"`, string values should be in single quotes.',
+    ),
+    **options,
```
sfc-gh-turbaszek, line 143: It should be hidden for now

src/snowflake/cli/plugins/object/stage/commands.py
sfc-gh-pczajka: I wouldn't change the example, at first I assumed that only prefix-filtering is allowed - IMO it's good to have a hint

src/snowflake/cli/plugins/object/stage/manager.py
sfc-gh-pczajka: 👍

src/snowflake/cli/plugins/object/stage/commands.py
```diff
@@ -121,3 +123,27 @@ def stage_diff(
     """
     diff: DiffResult = stage_diff(Path(folder_name), stage_name)
     return ObjectResult(str(diff))
+
+
+@app.command("execute", requires_connection=True)
+def execute(
+    stage_path: str = typer.Argument(
+        None,
```
sfc-gh-pczajka, line 131:
```suggestion
        ...,
        show_default=False,
```
This way you'll get "missing argument" error, not "expected str (...), got NoneType" :)

tests/object/stage/__snapshots__/test_stage.ambr
sfc-gh-astus: Created

src/snowflake/cli/plugins/object/stage/commands.py
sfc-gh-astus: Added

src/snowflake/cli/plugins/object/stage/commands.py
```diff
@@ -121,3 +123,27 @@ def stage_diff(
...
+        help="Execute immediate all files from the path. Files can be filtered with glob like pattern, e.g. `@stage/**/*.sql`.",
+    ),
+    on_error: OnErrorType = typer.Option(
+        OnErrorType.BREAK.value,
+        "--on-error",
+        help="What to do when an error occurs. Defaults to break.",
+    ),
+    parameters: Optional[List[str]] = typer.Option(
+        None,
+        help='Parameters for the template. For example: `--parameters "<key>=<value>"`, string values should be in single quotes.',
```
sfc-gh-pczajka, line 141:
```suggestion
        help='Parameters for the template. For example: `--parameters "<key>=<value>"`, string values should be in single quotes.',
        show_default=False,
```

src/snowflake/cli/plugins/object/stage/commands.py
```diff
@@ -121,3 +123,27 @@ def stage_diff(
...
+    stage_path: str = typer.Argument(
+        None,
+        help="Execute immediate all files from the path. Files can be filtered with glob like pattern, e.g. `@stage/**/*.sql`.",
+    ),
+    on_error: OnErrorType = typer.Option(
+        OnErrorType.BREAK.value,
+        "--on-error",
+        help="What to do when an error occurs. Defaults to break.",
+    ),
+    parameters: Optional[List[str]] = typer.Option(
```
sfc-gh-pczajka, line 139: How about making this `List[str]` which defaults to `[]`? Less cases to keep in mind

src/snowflake/cli/plugins/object/stage/commands.py
sfc-gh-pczajka, line 141: Currently usage of this is `--parameters "key1=value1 --parameters "key2=value2"...`. I'd change the name to `parameter` (singular seems more intuitive for me)
Also - do we want to add some abbreviation (`--param` / `-p`)? Just not to write `--parameter` for absolute every value

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -137,3 +153,96 @@ def create(self, stage_name: str, comment: Optional[str] = None) -> SnowflakeCur
...
+        if file_path_1_directories > file_path_2_directories:
+            return 1
+        elif file_path_1_directories < file_path_2_directories:
+            return -1
+        elif file_path_1 >= file_path_2:
+            return 1
+        else:
+            return -1
+
+    def _parse_execute_parameters(
```
sfc-gh-pczajka, line 219: `_parse_execute_parameters`, `_stage_files_comparator` and `_filter_files_list` don't use `self` properties - I'd define them as `@staticmethod` or even define outside `StageManager`

src/snowflake/cli/plugins/object/stage/manager.py
sfc-gh-astus: It looks much better now

src/snowflake/cli/plugins/object/stage/commands.py
sfc-gh-astus: Added

src/snowflake/cli/plugins/object/stage/commands.py
sfc-gh-astus: We always use Optional with None

src/snowflake/cli/plugins/object/stage/manager.py
sfc-gh-astus: I'm not sure about `snow://` so I added it

src/snowflake/cli/plugins/object/stage/commands.py
sfc-gh-astus: I added `*` to example `For example @stage/dev/*`

src/snowflake/cli/plugins/object/stage/commands.py
sfc-gh-astus: I unified variables with notebooks and snowsql, now is `-D`

src/snowflake/cli/plugins/object/stage/manager.py
sfc-gh-astus: Now only `_parse_execute_parameters` doesn't use `self`, I added `@staticmethod`

src/snowflake/cli/plugins/object/stage/manager.py
sfc-gh-turbaszek: We have this `startswith` in two places. Should we adjust the `get_standard_stage_prefix` to return always the `@...` version? In this way we fill have one place where we handle the `snow://` scenario

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -164,3 +186,90 @@ def create(self, stage_name: str, comment: Optional[str] = None) -> SnowflakeCur
...
+
+        # sort filtered files in alphabetical order with directories at the end
+        sorted_file_list = sorted(
+            filtered_file_list, key=lambda f: (path.dirname(f), path.basename(f))
+        )
+
+        sql_variables = self._parse_execute_variables(variables)
+        results = []
+        for file in sorted_file_list:
+            results.append(self._call_execute_immediate(file, sql_variables, on_error))
```
sfc-gh-turbaszek, line 211: use keywords arguments to make refactoring easier later on

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -164,3 +186,90 @@ def create(self, stage_name: str, comment: Optional[str] = None) -> SnowflakeCur
...
+
+        return results
+
+    def _get_files_list_from_stage(self, stage_path: str) -> List[str]:
+        stage_name = self.get_stage_name_from_path(stage_path)
+        files_list_result = self.list_files(stage_name).fetchall()
+
+        if len(files_list_result) == 0:
```
sfc-gh-turbaszek, line 219:
```suggestion
        if not files_list_result:
```

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -164,3 +186,90 @@ def create(self, stage_name: str, comment: Optional[str] = None) -> SnowflakeCur
...
+        sql_variables = self._parse_execute_variables(variables)
+        results = []
+        for file in sorted_file_list:
+            results.append(self._call_execute_immediate(file, sql_variables, on_error))
+
+        return results
+
+    def _get_files_list_from_stage(self, stage_path: str) -> List[str]:
+        stage_name = self.get_stage_name_from_path(stage_path)
+        files_list_result = self.list_files(stage_name).fetchall()
```
sfc-gh-turbaszek, line 217: How big this last can be? Do we want to fetchall or do fetch one by one? Something to think about for future

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -164,3 +186,90 @@ def create(self, stage_name: str, comment: Optional[str] = None) -> SnowflakeCur
...
+        # Execute everything if only stage name was provided
+        if "/" not in stage_path:
+            return files_on_stage
+
+        # Filter with fnmatch if contains `*` or `?`
+        if glob.has_magic(stage_path):
+            return fnmatch.filter(files_on_stage, stage_path)
+
+        # Exact file path was provided if stage_path in file list
+        if stage_path in files_on_stage:
```
sfc-gh-turbaszek, line 238: How about not existing file?  If the file does not exist then the later comment is not true (directory was not privded).

How about `if Path(stage_path).suffix in SUPPORTED_FILES_FORMATS` where `SUPPORTED_FILES_FORMATS = {".sql"}`. WDYT?

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -164,3 +186,90 @@ def create(self, stage_name: str, comment: Optional[str] = None) -> SnowflakeCur
...
+
+        return [f["name"] for f in files_list_result]
+
+    def _filter_files_list(
+        self, stage_path: str, files_on_stage: List[str]
+    ) -> List[str]:
+        stage_path = self.remove_stage_prefix(stage_path)
+
+        # Execute everything if only stage name was provided
+        if "/" not in stage_path:
```
sfc-gh-turbaszek, line 230: What about `@my_stage/`? Would `len(Path(stage_path).part) == 1`  be more accurate?

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -164,3 +186,90 @@ def create(self, stage_name: str, comment: Optional[str] = None) -> SnowflakeCur
...
+    def _parse_execute_variables(variables: Optional[List[str]]) -> Optional[str]:
+        if not variables:
+            return None
+
+        query_parameters = []
+        for p in variables:
+            if "=" not in p:
+                raise ClickException(f"Invalid variable: '{p}'")
+
+            key, value = p.split("=")
```
sfc-gh-turbaszek, line 254: What about `"key=val=ue""` ?

sfc-gh-astus: Good point, check and test added

src/snowflake/cli/plugins/object/stage/commands.py
```diff
@@ -129,6 +130,36 @@ def stage_diff(
     return ObjectResult(str(diff))
 
 
+@app.command("execute", requires_connection=True)
+def execute(
+    stage_path: str = typer.Argument(
+        ...,
+        help="Stage path with files to be execute. For example `@stage/dev/*`.",
```
sfc-gh-pczajka, line 137:
```suggestion
        help="Stage path with files to be execute. For example `@stage/dev/*`.",
        show+default=False,
```
to get rid of "[default: None]" from
```
╭─ Arguments ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ *    stage_path      TEXT  Stage path with files to be execute. For example `@stage/dev/*`. [default: None] [required]                  │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

tests_integration/test_stage.py
```diff
@@ -128,3 +128,54 @@ def test_stage_get_recursive(
...
+        assert result.exit_code == 0, result.output
+        assert contains_row_with(result.json, {"status": "UPLOADED"})
+
+    result = runner.invoke_with_connection_json(
+        ["object", "stage", "execute", stage_name]
+    )
+    assert result.exit_code == 0
+    assert result.json == snapshot
+
+    #
```
sfc-gh-pczajka, line 170:
```suggestion
    # TODO: uncomment when templating will be supported
    #
```

src/snowflake/cli/plugins/object/stage/commands.py
sfc-gh-astus: Added

tests_integration/test_stage.py
sfc-gh-astus: It is more about availability than support, so adjusted message

src/snowflake/cli/plugins/object/stage/manager.py
sfc-gh-astus: `ls @stage` returns files in alphabetical order ignoring directories, so we need to have whole filtered list to execute files in right order.

src/snowflake/cli/plugins/object/stage/manager.py
sfc-gh-astus: It depends if you provide stage that does not exist, command raises exception from Snowflake. If you provide file does not exist on existing stage message is `No files matched pattern 'exe/*.txt'`. 

I forgot to add extension check.

sfc-gh-pczajka (approved):

src/snowflake/cli/plugins/object/stage/manager.py
sfc-gh-turbaszek: Why is it invalid? I think it's valid, we just have to make the parsing right. `p.split("=", 1)` should solve the issue.

I would consider extracting this parsing logic to separate method that would return a list of {key: ..., value: ...}  dictionaries. There will be other places where we will want to use similar parsing.

src/snowflake/cli/plugins/object/stage/manager.py
```diff
@@ -164,3 +183,102 @@ def create(self, stage_name: str, comment: Optional[str] = None) -> SnowflakeCur
...
+    ) -> Dict:
+        try:
+            stage_path_prefixed = self.get_standard_stage_prefix(file)
+            query = (
+                f"execute immediate from {self.quote_stage_name(stage_path_prefixed)}"
+            )
+            if variables:
+                query += variables
+            self._execute_query(query)
+            cli_console.step(f"{file} - SUCCESS")
```
sfc-gh-turbaszek, line 278: I would reverse the order
```suggestion
            cli_console.step(f"SUCCESS - {file}")
```
to achieve more consistent output:
```
SUCCESS - foo.sql
SUCCESS - bar/foo.sql
SUCCESS - foo/bar/arr.sql
```
instead of 
```
foo.sql - SUCCESS
bar/foo.sql - SUCCESS
foo/bar/arr.sql - SUCCESS
```

WDYT?

sfc-gh-astus: I agree, changed

src/snowflake/cli/plugins/object/stage/manager.py
sfc-gh-astus: Extracted

sfc-gh-turbaszek (approved):

sfc-gh-turbaszek (approved):

PR #965, opened 2024-04-04
Add support for hidden session parameters --session-token and --master--token

sfc-gh-lmonteromarin:
### Pre-review checklist

### Changes description
...


src/snowflake/cli/api/cli_global_context.py
```diff
@@ -28,6 +28,8 @@ def __init__(self):
         self._diag_log_path: Optional[Path] = None
         self._diag_allowlist_path: Optional[Path] = None
         self._temporary_connection: bool = False
+        self._session_token: Optional[str] = None
+        self._master_token: Optional[str] = None
```
sfc-gh-turbaszek, line 32: This looks pretty good! Is there any public documentation for those parameters?

sfc-gh-lmonteromarin: I haven't found any, and I think someone mentioned that this is not something that is usually exposed to customers (although it can be created and seen by users of the Node SDK). That is the reason why I believe that maybe this should be kept as a hidden parameter.
I will ask more about this and write here what I find.

PR #967, opened 2024-04-05
Add support for hidden session parameters --session-token and --master--token

sfc-gh-lmonteromarin:
### Pre-review checklist

### Changes description
Add support for the session paramters `--session-token` and `--master-token`, which can be used to share a connection created with the Python connector or the Node connector.


sfc-gh-turbaszek:
@sfc-gh-lmonteromarin should we add an integration tests that verifies this behaviour? I think it would be enough to:
1. Extract session tokens from snowpark session that we use in integration tests
2. Execute `snow sql -q "set a = 42" -x ...` with passing session parameters, the variable should be persisted in the session
4. Execute `snow sql -q "select $a" -x ...` to validate that we use the same session.

WDYT?

sfc-gh-lmonteromarin:
> @sfc-gh-lmonteromarin should we add an integration tests that verifies this behaviour? I think it would be enough to:
> 
> 1. Extract session tokens from snowpark session that we use in integration tests
> 2. Execute `snow sql -q "set a = 42" -x ...` with passing session parameters, the variable should be persisted in the session
> 3. Execute `snow sql -q "select $a" -x ...` to validate that we use the same session.
> 
> WDYT?

This sounds good to me, I will add it.

tests_integration/test_session_token.py
```diff
@@ -0,0 +1,41 @@
...
+            "sql",
+            "-q",
+            "select $a as dummy",
+            "-x",
+            "--account",
+            snowflake_session.account,
+            "--session-token",
+            session_token,
+            "--master-token",
+            master_token,
```
sfc-gh-turbaszek, line 35: Seeing the errors from integration tests I'm wondering if those parameters changes the session life time.  I'm guessing that existing tests start to fail after this one. To check it you can run `pytest -p no:randomly` to avoid random order of tests.

sfc-gh-lmonteromarin: Thanks, I'll review that.

sfc-gh-lmonteromarin: I have checked this and it seems like it was closing the session after invoking the CLI with the session token and master token. I have changed the code to set connection parameter `server_session_keep_alive` to `True` when the session token and master token are used, since in those cases, the process that creates the connection (and shares it) should be the one responsible for deleting the session when it is no longer used.

sfc-gh-turbaszek (dismissed):
Looks good to me :rocket:

src/snowflake/cli/api/cli_global_context.py
```diff
@@ -166,6 +182,8 @@ def _collect_not_empty_connection_attributes(self):
             "schema": self.schema,
             "role": self.role,
             "warehouse": self.warehouse,
+            "session_token": self.session_token,
```
sfc-gh-melnacouzi, line 185: do we need both to be present for this to work? If so, do we check anywhere if they are both present or both absent? (i.e. should we error out if user provides master-token without providing session-token?)

sfc-gh-lmonteromarin: You are right about this: these tokens must be used together. I will add appropriate error messages, thank you!

sfc-gh-lmonteromarin: I have added three error messages (with their corresponding unit tests). Please let me know if you think the messages can be improved.

1. For the case in which a session token or master token has been provided but the user has not specified that the connection is temporary
2. For the case in which only the session token was provided
3. For the case in which only the master token was provided
image
image
image

sfc-gh-melnacouzi (approved):

sfc-gh-astus (approved):

PR #1029, opened 2024-04-29
Add teardown cascade option

sfc-gh-gbloom:
### Pre-review checklist

### Changes description

Adding the `--cascade` option when running `snow app teardown` which will be passed to the `DROP APPLICATION` command. A list of the affected objects will be printed to the user.

Without it, when the application owns account level objects outside of the app such as compute pools/warehouses/databases, `snow app teardown` will fail with a general error: `Could not successfully execute the Snowflake SQL statements`. The specific SQL error is shown with `--debug`.

### Interactive mode

In interactive mode, if `--cascade`/`--no-cascade` was not explicitly provided, the user will be prompted for an input with the list of objects that will be dropped if approved. The following demo shows the `abort` (default), `no` and `yes` flows:

https://github.com/snowflakedb/snowflake-cli/assets/162226501/91774b98-a069-439e-9cd7-5d7151dd6546

### No cascade

Teardown with no explicit `--cascade`/`--no-cascade`:
```
$ snow app teardown
╭─ Error ───────────────────────────────────────────────────────────────────────╮
│ The following application objects are owned application spcs_na_5_gbloom:     │
│ (WAREHOUSE) SPCS_NA_5_GBLOOM_WH                                               │
│                                                                               │
│ Aborting, re-run teardown again with --cascade or --no-cascade to specify     │
│ whether these objects should be dropped along with the application.           │
╰───────────────────────────────────────────────────────────────────────────────╯
```

Teardown with explicit `--no-cascade`:
```
snow app teardown --no-cascade
The following objects are owned by application spcs_na_5_gbloom:
(WAREHOUSE) SPCS_NA_5_GBLOOM_WH
Dropping application spcs_na_5_gbloom now.
╭─ Error ───────────────────────────────────────────────────────────────────────╮
│                                                                               │
│                                                                               │
│     Could not successfully execute the Snowflake SQL statements.              │
│                                                                               │
│                 drop application spcs_na_5_gbloom                             │
│                                                                               │
╰───────────────────────────────────────────────────────────────────────────────╯
```

### Integration & manual tests

The added integration tests fails on CI due to insufficient privileges for the CI role. I verified the tests pass locally with a test user. I committed the test and marked it to be skipped.

I also manually created an app that owns a warehouse and ran the same flows:
* `app teardown`
* `app teardown --force --no-cascade`
* `app teardown --cascade`


src/snowflake/cli/plugins/nativeapp/teardown_processor.py
```diff
@@ -89,9 +97,28 @@ def drop_application(self, auto_yes: bool):
...
                     return  # The user desires to keep the app, therefore exit gracefully
 
-        # 4. All validations have passed, drop object
+        # 4. If cascade was not explicitly set, check for objects owned by the app
+        if cascade is None:
+            has_compute_pools = False  # SHOW OBJECTS OWNED BY APPLICATION
+            if has_compute_pools:
+                if interactive:
+                    cascade = typer.confirm(
+                        "There are objects owned by this app. Should they be deleted?"
```
sfc-gh-bdufour, line 106: We should show the user what these are, otherwise the question is really difficult to answer.

Also, any kind of "should I do <something destructive>" should probably have 3 possible answers: yes, no, abort.

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
```diff
@@ -89,9 +97,28 @@ def drop_application(self, auto_yes: bool):
...
+                    )
+                else:
+                    cc.warning(
+                        "There are objects owned by this app. Please explicitly set --cascade if they should be deleted, or --no-cascade otherwise."
```
sfc-gh-bdufour, line 110: workding: set -> specify

Also, @sfc-gh-bgoel, what's the official guidance on using 'app' vs 'application' in command output?

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -213,6 +213,11 @@ def app_open(
 @with_project_definition("native_app")
 def app_teardown(
     force: Optional[bool] = ForceOption,
+    cascade: Optional[bool] = typer.Option(
+        None,
+        help=f"""When enabled, will also drop all objects owned by the application within the account. Defaults to false.""",
```
sfc-gh-bgoel, line 218: "application object" instead of "application".

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
sfc-gh-bgoel: 1. Yes, thank you for bumping this! @sfc-gh-gbloom you should use "application object" here instead of "app/application".
2. I'l like to display object names and types that are owned by the application object so the user knows the side effects of cascade (both here and line 106 like Bruno commented). Having to execute a `snow sql "show objects ..."` is an additional command we can save them the effort of typing if we just display the objects here. WDYT?

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
```diff
@@ -89,9 +97,28 @@ def drop_application(self, auto_yes: bool):
                     cc.message(f"Did not drop application object {self.app_name}.")
                     return  # The user desires to keep the app, therefore exit gracefully
 
-        # 4. All validations have passed, drop object
+        # 4. If cascade was not explicitly set, check for objects owned by the app
+        if cascade is None:
```
sfc-gh-cgorrie, line 101: How do `cascade` and `auto_yes` (populated by `--force`) interact? What would `--force --no-cascade` do?

sfc-gh-cgorrie: To give background `--force` was added so that we could cleanup the app / app pkg even if they weren't created by Snowflake CLI or they had a different owner than expected.

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
sfc-gh-bdufour: +1 on displaying the names, I had the same comment elsewhere

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
sfc-gh-gbloom: The behavior will be:
`--force`: cascade
`--force --no-cascade`: no cascade

sfc-gh-cgorrie: This was my suggestion, but now I'm remembering that you can teardown without cascading to the objects (I thought it blocked it). Yeah we should not have `--force` imply `--cascade`

sfc-gh-cgorrie: My bad @sfc-gh-gbloom sorry about that

sfc-gh-gbloom: 👍

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
```diff
@@ -16,33 +16,55 @@
...
                 raise SnowflakeSQLExecutionError(drop_query)
 
             cc.message(f"Dropped {object_type} {object_name} successfully.")
 
-    def drop_application(self, auto_yes: bool):
+    def _application_objects_to_str(
+        self, application_objects: ApplicationObject
+    ) -> str:
+        """
+        Returns a list in a "(Object Type) Object Name" format:
```
sfc-gh-cgorrie, line 56: Worth documenting here that database- and schema-level object names appear to be fully qualified.

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
```diff
@@ -89,9 +111,53 @@ def drop_application(self, auto_yes: bool):
...
+                if interactive:
+                    cascade = typer.confirm(
+                        f"""\
+                            The following application objects are owned by this application:
+                            {application_objects_str}
+                            Should they be deleted?
+                        """
+                    )
+                else:
+                    cc.warning(
```
sfc-gh-cgorrie, line 144: Is this a warning, or an error?

tests_integration/test_nativeapp.py
```diff
@@ -564,3 +564,35 @@ def test_nativeapp_init_deploy(
                 env=TEST_ENV,
             )
             assert result.exit_code == 0
+
+
+# Executing "snow app teardown --cascade", verifying that the application and application objects owned by the application are dropped
+@pytest.mark.integration
+def test_nativeapp_teardown_cascade(
```
sfc-gh-cgorrie, line 571: Considering the length of the file, consider splitting the native apps integration tests up into multiple files.

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
sfc-gh-gbloom: Added the list of the files in question -- see PR description for a demo.

Do you think it's reasonable to assume users know to abort the script? (ie Ctrl+C)
Typer has nice out of the box support for boolean values, but by switching to yes/no/abort I will need to parse the input myself and it might get inconsistent with other boolean values. WDYT?

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
sfc-gh-gbloom: Listing the affected files now in all flows

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
sfc-gh-gbloom: Raising an exception now

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -57,6 +57,8 @@
 from snowflake.connector import ProgrammingError
 from snowflake.connector.cursor import DictCursor
 
+ApplicationObject = TypedDict("ApplicationObject", {"name": str, "type": str})
```
sfc-gh-bdufour, line 60: OwnedObject? ApplicationObject makes me think these are application instances (since instances are objects).

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
sfc-gh-bdufour: This comment seems fair and not yet resolved.

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
```diff
@@ -89,9 +112,60 @@ def drop_application(self, auto_yes: bool):
...
+        application_objects = self.get_application_objects()
+        if len(application_objects) > 0:
+            application_objects_str = self._application_objects_to_str(
+                application_objects
+            )
+            if cascade is True:
+                cc.message(
+                    dedent(
+                        f"""\
+                            The following application objects will be deleted:
```
sfc-gh-bdufour: How about "The following objects are owned by application <app name> and will be dropped"

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
```diff
@@ -89,9 +112,60 @@ def drop_application(self, auto_yes: bool):
...
+                            {application_objects_str}
+                        """
+                    )
+                )
+            elif cascade is False:
+                cc.message(
+                    dedent(
+                        f"""\
+                            The following application objects will not be deleted:
```
sfc-gh-bdufour: ditto, but consider adding a line stating what happens to them. How are their ownership affected?

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
```diff
@@ -89,9 +112,60 @@ def drop_application(self, auto_yes: bool):
...
+                    )
+                )
+            elif interactive:
+                if interactive:
+                    cascade = typer.confirm(
+                        dedent(
+                            f"""\
+                        The following application objects are owned by this application:
+                        {application_objects_str}
+                        Do you confirm to delete these?
```
sfc-gh-bdufour: How about 

```
The following objects are owned by application <app name>:
<list goes here>
Delete these objects along with the application? [y/n/ABORT]
```

If the user just hits enter, it's not a strong enough signal, we should abort the teardown step IMHO.

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
```diff
@@ -89,9 +112,60 @@ def drop_application(self, auto_yes: bool):
...
+                    """
+                        )
+                    )
+            else:
+                raise ClickException(
+                    dedent(
+                        f"""\
+                    The following application objects are owned by this application:
```
sfc-gh-bdufour, line 138: Similar:
```
The following objects are owned by application <app name>:
<list goes here>

Aborting, re-run teardown again with --cascade or --no-cascade to specify whether these objects should be dropped along with the application.
```

Please check with Mike about the wording of course.

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
```diff
@@ -89,9 +112,60 @@ def drop_application(self, auto_yes: bool):
...
+                    {application_objects_str}
+                    Please explicitly set --cascade if they should be deleted, or transfer ownership and run teardown again.
+                """
+                    )
+                )
+        elif cascade is None:
+            cascade = False
```
sfc-gh-bdufour, line 145: Is this reachable? If so, how?

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -405,6 +407,16 @@ def get_existing_app_pkg_info(self) -> Optional[dict]:
                 "application packages", self.package_name, name_col=NAME_COL
             )
 
+    def get_application_objects(self) -> List[ApplicationObject]:
```
sfc-gh-bdufour, line 410: `get_objects_owned_by_application()` ? see my other comment.

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
sfc-gh-gbloom: We run the drop command even though we expect it to fail.
Hypothetically, if the drop is somehow successful, I would assume the ownership is transferred to the account or current role?

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
sfc-gh-gbloom: Yes, if cascade is unset and there are no objects owned by the application. In this case I prefer the value to be explicitly `False` instead of `None`

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
sfc-gh-gbloom: If the user didn't respond with yes/y/no/n they will be asked again ([source](https://github.com/pallets/click/blob/main/src/click/termui.py#L230-L252))

sfc-gh-bdufour: I've mentioned a few times now that I think abort should be an option here, given that yes and no both have consequences for the user. You seem to disagree though, so I suggest you drive the discussion with the team as a whole. It's quite possible I'm a minority.

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
sfc-gh-bdufour: You should verify and state it for the user, not make assumptions IMHO.

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
```diff
@@ -89,9 +113,38 @@ def drop_application(self, auto_yes: bool):
...
+                application_objects
+            )
+            if cascade is True:
+                cc.message(
+                    f"The following objects are owned by application {self.app_name} and will be dropped:\n{application_objects_str}"
+                )
+            elif cascade is False:
+                cc.message(
+                    f"The following objects are owned by application {self.app_name} but will not be dropped:\n{application_objects_str}"
+                )
```
sfc-gh-cgorrie, line 129: As discussed, is this ever true? Will `drop application` allow orphaned objects?

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
sfc-gh-gbloom: WDYT?
```
$ snow app teardown --no-cascade
The following objects are owned by application spcs_na_5_gbloom:
(WAREHOUSE) SPCS_NA_5_GBLOOM_WH
Dropping application spcs_na_5_gbloom now.
╭─ Error ───────────────────────────────────────────────────────────────────────╮
│                                                                               │
│                                                                               │
│     Could not successfully execute the Snowflake SQL statements.              │
│                                                                               │
│                 drop application spcs_na_5_gbloom                             │
│                                                                               │
╰───────────────────────────────────────────────────────────────────────────────╯
```

`drop_generic_object()` should probably display the SQL error message instead of a generic failure text. We can consider doing it as part of different task.

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
sfc-gh-gbloom: Added the `[y/n/ABORT]` prompt

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
sfc-gh-gbloom: `drop application` will not allow orphaned objects, we expect this command to fail.
Changed the wording to just reflect the state in case the behavior will change in the future (automatic ownership transfer, for example)

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
```diff
@@ -89,9 +112,45 @@ def drop_application(self, auto_yes: bool):
...
+                    f"The following objects are owned by application {self.app_name} and will be dropped:\n{application_objects_str}"
+                )
+            elif cascade is False:
+                cc.message(
+                    f"The following objects are owned by application {self.app_name}:\n{application_objects_str}"
+                )
+            elif interactive:
+                if interactive:
+                    api_integration = typer.prompt(
+                        f"The following objects are owned by application {self.app_name}:\n{application_objects_str}\n\nWould you like to drop these objects as well as the application? [y/n/ABORT]"
```
sfc-gh-cgorrie, line 132:
```suggestion
                        f"The following objects are owned by application {self.app_name}:\n{application_objects_str}\n\nWould you like to drop these objects as well, or just the application? [y/n/ABORT]"
```

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
```diff
@@ -89,9 +112,45 @@ def drop_application(self, auto_yes: bool):
...
+                    )
+                    if api_integration in ["y", "yes", "Y", "Yes", "YES"]:
+                        cascade = True
+                    elif api_integration in ["n", "no", "N", "No", "NO"]:
+                        cascade = False
+                    else:
+                        raise typer.Abort()
+            else:
+                cc.message(
+                    f"The following application objects are owned application {self.app_name}:\n{application_objects_str}\n\nRe-run teardown again with --cascade or --no-cascade to specify whether these objects should be dropped along with the application."
```
sfc-gh-cgorrie, line 142:
```suggestion
                    f"The following application objects are owned by application {self.app_name}:\n{application_objects_str}\n\nRe-run teardown again with --cascade or --no-cascade to specify whether these objects should be dropped along with the application."
```

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
sfc-gh-gbloom: Not sure about this wording, the answer is no longer yes/no this way

sfc-gh-cgorrie: My point was that "no" could either mean "drop neither" or "just drop the application". You're not being clear about which option maps to which outcome.

tests/__snapshots__/test_help_messages.ambr
```diff
@@ -416,14 +416,26 @@
...
+  │                                          enable this option if interactive   │
+  │                                          mode is not specified and if you    │
+  │                                          want perform potentially            │
+  │                                          destructive actions. Defaults to    │
+  │                                          unset.                              │
+  │ --cascade          --no-cascade          When enabled, will also drop all    │
+  │                                          application objects owned by the    │
+  │                                          application within the account.     │
+  │                                          Defaults to false.                  │
+  │ --interactive  -i                        When enabled, this option displays  │
```
sfc-gh-mheavin, line 431: When enabled, the command displays...

(out of curiosity, if not terminals, where are the prompts displayed?)

sfc-gh-mheavin (commented):
Some suggestions for your consideration

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
sfc-gh-mheavin: What do you think of:

Would you like to drop these objects in addition to the application?

tests/__snapshots__/test_help_messages.ambr
```diff
@@ -416,14 +416,26 @@
...
-  │ --help     -h            Show this message and exit.                         │
+  │ --force                                  When enabled, this option causes    │
+  │                                          the command to implicitly approve   │
+  │                                          any prompts that arise. You should  │
+  │                                          enable this option if interactive   │
+  │                                          mode is not specified and if you    │
+  │                                          want perform potentially            │
+  │                                          destructive actions. Defaults to    │
+  │                                          unset.                              │
+  │ --cascade          --no-cascade          When enabled, will also drop all    │
```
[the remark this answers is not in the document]
sfc-gh-gbloom: The `teardown` command drops the application. `teardown --cascade` will also drop all application objects this application owns. Taking the suggestion above (this is the auto generated file).

tests/__snapshots__/test_help_messages.ambr
sfc-gh-gbloom: When a human runs the CLI we can ask questions interactively as the script executes. When a script runs it, the "answers" should be provided in advance via options.

tests/__snapshots__/test_help_messages.ambr
```diff
@@ -416,14 +416,26 @@
...
-  │                          should enable this option if interactive mode is    │
-  │                          not specified and if you want perform potentially   │
-  │                          destructive actions. Defaults to unset.             │
-  │ --project  -p      TEXT  Path where the Snowflake Native App project         │
-  │                          resides. Defaults to current working directory.     │
-  │ --help     -h            Show this message and exit.                         │
+  │ --force                                  When enabled, this option causes    │
+  │                                          the command to implicitly approve   │
+  │                                          any prompts that arise. You should  │
+  │                                          enable this option if interactive   │
```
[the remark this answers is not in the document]
sfc-gh-gbloom: As an example, we use it ourselves to clean up the temp applications after our tests

sfc-gh-turbaszek (dismissed):

sfc-gh-cgorrie (approved):
LGTM, some minor nitpicks

sfc-gh-cgorrie (approved):

tests/__snapshots__/test_help_messages.ambr
sfc-gh-bdufour: nit: I think Mike's suggestion for "this option displays" -> "this command displays" is a good one

sfc-gh-bdufour (approved):

tests/__snapshots__/test_help_messages.ambr
sfc-gh-gbloom: Sure, let's align the wording across all commands/options? It's not really tied to this PR

sfc-gh-pjob (approved):

PR #1032, opened 2024-05-01
[NADE] Collect external Snowpark functions for NA Codegen

sfc-gh-bgoel:
### Pre-review checklist
   * [ ] I've added or updated integration tests to verify correctness of my new code. - N/A
   * [ ] I've confirmed that my changes are working by executing CLI's commands manually. - N/A
   * [x] I've confirmed that my changes are up-to-date with the target branch. - as of May 7, 9AM EST
   * [ ] I've described my changes in the release notes. - N/A

### Changes description
Part of setup script codegen for native apps. Executes the collection of snowpark objects from a given py file and output it as a json/dict to be further used in subsequent PRs which will spin up SQL DDL from the output.


src/snowflake/cli/plugins/nativeapp/setup_script_compiler/annotation_processors/snowpark_extension_function.py
```diff
@@ -0,0 +1,316 @@
...
+    QUERY_TAG = "QUERY_TAG"
+    COLUMN = "COLUMN"
+    PROCEDURE = "PROCEDURE"
+    TABLE_FUNCTION = "TABLE_FUNCTION"
+    DYNAMIC_TABLE = "DYNAMIC_TABLE"
+    AGGREGATE_FUNCTION = "AGGREGATE_FUNCTION"
+    CTE = "CTE"
+
+
+def get_handler_path_without_suffix(file_path: Path) -> str:
```
sfc-gh-bgoel, line 67: Use path split here instead of replace.

src/snowflake/cli/api/project/schemas/native_app/path_mapping.py
```diff
@@ -6,7 +6,7 @@
 from snowflake.cli.api.project.schemas.updatable_model import UpdatableModel
 
 
-class AnnotationProcessor(UpdatableModel):
+class Processor(UpdatableModel):
```
sfc-gh-bgoel, line 9: Updated name as `AnnotationProcessor` is too specific, a processor does not have to act only on the annotations.

.pre-commit-config.yaml
```diff
@@ -48,6 +48,7 @@ repos:
           ^src/snowflake/cli/api/utils/rendering.py$|
           ^src/snowflake/cli/plugins/spcs/common.py$|
           ^src/snowflake/cli/plugins/snowpark/venv.py$
+          ^src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py$|
```
sfc-gh-bgoel, line 51: To escape the check on `print` statements in the code.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -0,0 +1,352 @@
...
+with open({py_file}, mode='r', encoding='utf-8') as udf_code:
+    code = udf_code.read()
+
+
+snowflake.snowpark.context._is_execution_environment_sandboxed_for_client = True
+snowflake.snowpark.context._should_continue_registration = __SNOWFLAKE_CLI_NATIVE_APP_INTERNAL_callback_replacement()
+snowflake.snowpark.session._is_execution_environment_sandboxed_for_client = True
+
+# Remove reference to the callback from globals() since the function is already assigned to snowflake.snowpark.context._should_continue_registration
+del __SNOWFLAKE_CLI_NATIVE_APP_INTERNAL_callback_replacement
```
sfc-gh-bdufour, line 132: nit: that's a bit odd, I would have done something like:

```
g = globals()
del g['<internal name>']
exec(code, g)
```

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -0,0 +1,352 @@
...
+    exec(code, globals())
+except Exception:
+    sys.exit(10)
+
+print(__SNOWFLAKE_CLI_NATIVE_APP_INTERNAL_callback_return_list)
+    """
+
+    completed_process = execute_script_in_sandbox(script_source=script_source, **kwargs)
+    if completed_process.returncode != 0:
+        raise ClickException(f"Unable to fetch Snowpark objects from {py_file}")
```
sfc-gh-bdufour, line 144: You're using this for control flow by exception. That's generally not a great pattern (especially for something as generic as ClickException). Your code here should catch the sandbox exceptions and return an obviously incorrect value (None or something like that) that the calling code can inspect and act on.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -0,0 +1,352 @@
...
+
+try:
+    exec(code, globals())
+except Exception:
+    sys.exit(10)
+
+print(__SNOWFLAKE_CLI_NATIVE_APP_INTERNAL_callback_return_list)
+    """
+
+    completed_process = execute_script_in_sandbox(script_source=script_source, **kwargs)
```
sfc-gh-bdufour, line 142: NB: you'll need to pass in the deploy root as CWD here (my code supports it), and to be safe, we should probably put in a timeout too. Something short should do, but just in case, let's go with 30s or more

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -0,0 +1,352 @@
...
+        # src: john/doe/folder1/**/* (in this case, all files and directories under src need to be considered)
+        # dest: stagepath/
+        elif "**" in artifact_src:
+            all_source_paths = list(self.project_root.glob(artifact_src))
+            src_py_files.extend(get_all_python_files(all_source_paths))
+
+        # Case 3: When artifact has the following src/dest pairing
+        # src: john/doe/folder1/*
+        # dest: stagepath/ (in this case, the directories under folder1 will be symlinked, which means files inside those directories also need to be considered due to implicit availability from directory symlink)
+        elif is_glob(artifact_src):
```
sfc-gh-bdufour, line 250: Guy added a way to keep track of the mapping during bundle, we should just use that. There is no reason to do resolve globbing here again. It's bad for separation of concerns and maintainability.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function.py
```diff
@@ -0,0 +1,343 @@
...
+            dest_path=file_path,
+            deploy_root=deploy_root,
+        )
+
+
+def get_object_type_as_text(object_type: TempObjectType) -> str:
+    return object_type.value.replace("_", " ")
+
+
+class ExtensionFunctionProperties:
```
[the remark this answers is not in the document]
sfc-gh-bgoel: The data from snowpark is not enough to create the SQL statements. For example:
1. We need access to the destination file (in deploy root) to create the handler. 
2. We need to validate the raw imports with access to the user's filesystem.
3. `native_app_params` is used to set the value to `self.schema` and `self.application_roles`. When not found, it needs to warn the user accordingly.

and so on.

All this enrichment and validation needs to happen before the SQL can be created. And because they are all related to the snowpark data itself, and have nothing to do with SQL just yet, I wanted to centralize them in one place, i.e. ExtensionFunctionProperties.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/processor_base.py
```diff
@@ -0,0 +1,21 @@
+from abc import ABC, abstractmethod
+from pathlib import Path
+
+from snowflake.cli.api.project.schemas.native_app.native_app import NativeApp
+from snowflake.cli.api.project.schemas.native_app.path_mapping import PathMapping
+
+
+class BuiltInAnnotationProcessor(ABC):
```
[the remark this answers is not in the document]
sfc-gh-bgoel: I forgot to change this after I changed `path_mapping.py`. Will do it now.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -0,0 +1,352 @@
...
+    # Create the code snippet to be executed in the sandbox
+
+    script_source = f"""
+import functools
+import sys
+
+try:
+    import snowflake.snowpark
+except ModuleNotFoundError:
+    sys.exit(1)
```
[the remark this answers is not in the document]
sfc-gh-bgoel: Actually, do we even need to surround this in try except? Because if the import fails due to `ModuleNotFoundError`, that will abort the rest of the execution, and the exception should bubble up to the subprocess.stderr. Is that an incorrect understanding?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -0,0 +1,352 @@
...
+        extension_function_dict["secrets"] = extension_function_properties.secrets
+        extension_function_dict["inline_python_code"] = extension_function_properties.inline_python_code
+        extension_function_dict["native_app_params"] = extension_function_properties.native_app_params
+        extension_function_dict["raw_imports"] = extension_function_properties.raw_imports
+        extension_function_dict["replace"] = extension_function_properties.replace
+        extension_function_dict["if_not_exists"] = extension_function_properties.if_not_exists
+        extension_function_dict["execute_as"] = extension_function_properties.execute_as
+        extension_function_dict["anonymous"] = extension_function_properties.anonymous
+        raw_func = extension_function_properties.func
+        extension_function_dict["func"] = raw_func.__name__ if isinstance(raw_func, Callable) else raw_func
```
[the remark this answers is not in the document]
sfc-gh-bgoel: From snowpark, func can either be a `Callable` or a `Tuple[str, str]`. I have not seen an example of the latter, but `__name__` can only be used on the `Callable`. When type is `Tuple[str, str]`, the entry at index 1 is used. I can directly assign that here too "... else raw_func[1]`, but I instead used it to throw a warning to the user that we don't do tuples in our context, and hence codegen will be skipped.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -0,0 +1,352 @@
...
+        return extension_function_dict
+
+
+    def __SNOWFLAKE_CLI_NATIVE_APP_INTERNAL_callback_append_to_list(callback_return_list, extension_function_properties):
+        extension_function_dict = __SNOWFLAKE_CLI_NATIVE_APP_INTERNAL_transform_snowpark_object_to_json(extension_function_properties)
+        #TODO: print(extension_function_dict) here in case of downstream UDF creation failures
```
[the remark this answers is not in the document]
sfc-gh-bgoel: This is my DM from this morning, we can discuss.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bgoel: True, if we want to delete just they key from the globals() dict. But in this case, since the snowpark callback has already been assigned the callable, I deleted the reference to the callable. Not strongly opinionated, so can go either way.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bgoel: Ack, can rebase onto your changes once merged.

src/snowflake/cli/plugins/nativeapp/codegen/compiler.py
```diff
@@ -0,0 +1,38 @@
...
+            if (isinstance(processor, str) and processor == SNOWPARK_PROCESSOR) or (
+                isinstance(processor, Processor)
+                and processor.name == SNOWPARK_PROCESSOR
+            ):
+                SnowparkAnnotationProcessor(
+                    project_definition=project_definition,
+                    project_root=project_root,
+                    deploy_root=deploy_root,
+                    artifact_to_process=artifact,
+                    processor=processor,
```
sfc-gh-turbaszek, line 37: Why processor takes processor as an input?

sfc-gh-bgoel: There are two different types of processor here, maybe I should make that clear in the code too:
1. One is a `SnowparkAnnotationProcessor`, the class that will do the heavy lifting for us.
2. The class needs access to the `processor` dictionary that was part of project definition, and this is what is passed in here. 
Would you prefer making that distinction in code? I can add in the change.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bgoel: Synced with @sfc-gh-gbloom . Seems like neither the mapping nor `source_path_to_deploy_path` will address the use here. The mapping can be "incomplete" in that it doesn't contain symlinked kv pairs. 

Also, for `source_path_to_deploy_path`, I'd need access to
1. each source (file) path for a given dir, and pass it to `source_path_to_deploy_path`, i.e. I need enumeration anyway for all source paths
2. the exact file location in the deploy root, not just a parent path that could be returned from `source_path_to_deploy_path`.

tests/testing_utils/files_and_dirs.py
```diff
@@ -40,10 +42,11 @@ def temp_local_dir(files: Dict[str, Union[str, bytes]]) -> Generator[Path, None,
         for relpath, contents in files.items():
             path = Path(tmpdir, relpath)
             path.parent.mkdir(parents=True, exist_ok=True)
-            mode = "wb" if isinstance(contents, bytes) else "w"
-            encoding = None if isinstance(contents, bytes) else "UTF-8"
-            with open(path, mode=mode, encoding=encoding) as fh:
-                fh.write(contents)
+            if path.suffix != "":
```
sfc-gh-cgorrie, line 45: This looks like it's attempting to avoid making files that have `None` as their values (as they represent directories), maybe we just explicitly look for that? Directories can have dots in them.

If I understood the use case correctly, my suggestion would be something like:

```python
is_directory = (contents is None)
if not is_directory:
```

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/callback_source.py.jinja
```diff
@@ -0,0 +1,137 @@
...
+    global __snowflake_cli_native_app_internal_callback_return_list
+
+    def __snowflake_cli_native_app_internal_transform_snowpark_object_to_json(
+        extension_function_properties,
+    ):
+
+        {% raw %}extension_function_dict = {}
+        extension_function_dict[
+            "object_type"
+        ] = extension_function_properties.object_type.name
```
sfc-gh-bdufour, line 40: nit: would it be easier to read as:

```
fn = external_function_properties
extension_function_dict = {
    "object_type": fn.object_type.name,
    "input_sql_types": fn.input_sql_types,
    # [ ... ]
}
```

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function_utils.py
```diff
@@ -0,0 +1,104 @@
...
+    # Input raw_imports can look like [("tests/resources/test_udf_dir/test_udf_file.py", "resources.test_udf_dir.test_udf_file")]
+    for valid_import in raw_imports:
+        # TODO: should add leading "/" for paths on stage?
+        if isinstance(valid_import, str):
+            # Similar to session.add_import("tests/resources/test_udf_dir/test_udf_file.py"), no checks on path being absolute
+            # Or session.add_import("/tmp/temp.txt")
+            if valid_import.startswith("/"):
+                all_urls.append(valid_import)
+            else:
+                all_urls.append(f"/{valid_import}")
```
sfc-gh-bdufour, line 79: sorry, can you explain what's going on here?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -0,0 +1,276 @@
...
+    env_props = processor.properties["env"]
+    env_type = env_props.get("type", None)
+
+    if env_type is None:
+        return {}
+
+    if env_type.upper() == ExecutionEnvironmentType.CONDA.name:
+        env_name = env_props.get("name", None)
+        if env_name is None:
+            raise MissingProjectDefinitionPropertyError(
```
sfc-gh-bdufour, line 53: This seems odd. Why raise an exception that reads like a warning? Not specifying the name should be fine, if we can't find an env we'll just bail out anyway.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -0,0 +1,276 @@
...
+    )
+
+    try:
+        completed_process = execute_script_in_sandbox(
+            script_source=script_source,
+            cwd=deploy_root,
+            timeout=DEFAULT_TIMEOUT,
+            **kwargs,
+        )
+    except SandboxExecutionError as sdbx_err:
```
sfc-gh-bdufour, line 89: If we hit one of these, it's likely to affect all python files the same way. The user will then see the same line for every single script file. Fine for now, but maybe we should change the exceptions a bit so that we can distinguish between "the sandboxing is broken" and "this specific script didn't work".

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function_utils.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def _get_all_imports(raw_imports: List):
+    """
+    Sets validated imports in string form to be used in the creation of the SQL statement.
+    """
+    all_urls: List[str] = []
+    # Input raw_imports can look like [("tests/resources/test_udf_dir/test_udf_file.py", "resources.test_udf_dir.test_udf_file")]
+    for valid_import in raw_imports:
+        # TODO: should add leading "/" for paths on stage?
```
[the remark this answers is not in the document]
sfc-gh-bgoel: Using an example here:
```
create or replace function ext_code_schema.py_echo_fn(STR string)
  RETURNS STRING
  LANGUAGE PYTHON
  RUNTIME_VERSION=3.8
  PACKAGES=('snowflake-snowpark-python')
  HANDLER='echo.echo_fn'
  IMPORTS=('/echo.py');
```
Here `imports` has a leading / for its path. And the Todo is discuss if we need to do that, or if we expect our customer to provide it and we don't validate/add.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function_utils.py
sfc-gh-bgoel: Same as above

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -0,0 +1,276 @@
...
+def _execute_in_sandbox(
+    py_file: str, deploy_root: Path, kwargs: Dict[str, Any]
+) -> Optional[List[Dict[str, Any]]]:
+    # Create the code snippet to be executed in the sandbox
+    script_source = jinja_render_from_file(
+        template_path=Path("./callback_source.py.jinja"), data={"py_file": py_file}
+    )
+
+    try:
+        completed_process = execute_script_in_sandbox(
```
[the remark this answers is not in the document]
sfc-gh-bgoel: Is there a scenario where `SandboxExecutionError` is not thrown but the return code is no-zero?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bgoel: I can turn this into a warning instead.

tests/testing_utils/files_and_dirs.py
sfc-gh-bgoel: I like that suggestion, maybe we can even simplify it further. 
```
path = Path(tmpdir, relpath)
...
if path.is_file():
    ...write to file...
```

But, there can be a case where we pass in `contents = None` even if `path` is a file. In that case, I assume we wanted to create a file, but not write anything to it. In that case, I can add a check:
```
if path.is_file:
    if contents is None:
          f = open(path, "x") # creates new file
    else:
          ... write to file as before ...
```

WDYT?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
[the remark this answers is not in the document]
sfc-gh-bdufour: Very much so, yes

src/snowflake/cli/api/project/schemas/native_app/path_mapping.py
```diff
@@ -19,4 +19,22 @@ class AnnotationProcessor(UpdatableModel):
 class PathMapping(UpdatableModel):
     src: str
     dest: Optional[str] = None
-    processors: Optional[List[Union[str, AnnotationProcessor]]] = None
+    processors: Optional[List[Union[str, ProcessorMapping]]] = None
+
+    @field_validator("processors")
+    @classmethod
+    def transform_processors(
```
sfc-gh-bdufour, line 26: Neat!

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/callback_source.py.jinja
```diff
@@ -0,0 +1,117 @@
...
+) and hasattr(snowflake.snowpark.context, "_should_continue_registration")
+
+if not found_correct_version:
+    print(
+        "Did not find the minimum required version for snowflake-snowpark-python package. Please upgrade to v1.15.0 or higher.",
+        file=sys.stderr,
+    )
+    sys.exit(1)
+
+orig_globals = globals().copy()
```
sfc-gh-bdufour, line 26: We need to snapshot before we pollute the global space with `found_correct_version`

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -0,0 +1,289 @@
...
+        """
+
+        kwargs = (
+            _determine_virtual_env(processor_mapping)
+            if processor_mapping is not None
+            else {}
+        )
+
+        # 1. Get all src.py -> dest.py mapping
+        src_py_file_to_dest_py_file_map = self.get_src_py_file_to_dest_py_file_map(
```
sfc-gh-bdufour, line 167: note: I'm working on an alternative to this, but let's go with this implementation for now so we can keep making progress. This code will be replaced soon.

src/snowflake/cli/plugins/nativeapp/utils.py
```diff
@@ -81,3 +81,33 @@ def verify_exists(paths_to_sync: List[Path]):
...
+def filter_files(generator: Generator, predicate_func: Callable):
+    """
+    Filter file paths based on a given predicate function.
+    """
+    for file_path in generator:
+        if predicate_func(file_path):
+            yield file_path
+
+
+def is_single_quoted(name: str) -> bool:
```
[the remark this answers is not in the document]
sfc-gh-bdufour: the purpose of single quoting is ambiguous then. Snowflake has a specific syntax for single quoted strings, including how to escape nesting quotation marks. I'm not super comfortable with this helper, it's too generic/ambiguous and could be misused. I'd move it to your own module, at the very least, instead of making it a general utility.

src/snowflake/cli/plugins/nativeapp/utils.py
```diff
@@ -81,3 +81,33 @@ def verify_exists(paths_to_sync: List[Path]):
...
+    for root, _, files in os.walk(directory):
+        for file in files:
+            yield Path(os.path.join(root, file))
+
+
+def is_python_file(file_path: Path):
+    """
+    Checks if the given file is a python file.
+    """
+    return file_path.suffix == ".py"
```
sfc-gh-bdufour, line 100: nit: this doesn't need to be a general utility (or not yet), I'd keep it localized to the logic that uses it.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bgoel: I've added a TODO here noting the impending change.

sfc-gh-bdufour (approved):

PR #1033, opened 2024-05-02
Add the ability to execute a python script in virtual environments

sfc-gh-bdufour:
### Pre-review checklist
   * [N/A] I've added or updated integration tests to verify correctness of my new code.

### Changes description

Added a utility method to execute a python script in a virtual environment. This a building block for an upcoming native applications feature.

### Testing

I tested the logic on both Windows and Mac using a manually written driver and a combination of different python versions installed globally, conda environment and venv environments. Also added unit tests to cover 100% of the logic.

### DIFF
--- a/src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
+++ b/src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
@@ -2,11 +2,9 @@ import os
 import shutil
 import subprocess
 import sys
-import tempfile
-from contextlib import contextmanager
 from enum import Enum
 from pathlib import Path
-from typing import Optional
+from typing import Optional, Sequence, Union
 
 from click.exceptions import ClickException
 
@@ -38,26 +36,30 @@ def _is_ms_windows() -> bool:
     return sys.platform == "win32"
 
 
-@contextmanager
-def _temp_script_file(script_source: str):
-    script_file = tempfile.NamedTemporaryFile(mode="w", suffix=".py", delete=True)
-    try:
-        script_file.write(script_source)
-        script_file.flush()
+def _execute_python_interpreter(
+    python_executable: Optional[Union[str, Path, Sequence[str]]], script_source: str
+) -> subprocess.CompletedProcess:
+    if not python_executable:
+        raise SandboxExecutionError("No python executable found")
 
-        yield script_file.name
-    finally:
-        script_file.close()
+    if isinstance(python_executable, str) or isinstance(python_executable, Path):
+        args = [python_executable]
+    else:
+        args = [arg for arg in python_executable]
+    args.append("-")
+    return subprocess.run(args, capture_output=True, text=True, input=script_source)
 
 
 def _execute_in_venv(
-    script_source: str, venv_path: Optional[Path] = None
+    script_source: str, venv_path: Optional[Union[str, Path]] = None
 ) -> subprocess.CompletedProcess:
     resolved_venv_path = None
     if venv_path is None:
         active_venv_dir = _get_active_venv_dir()
         if active_venv_dir is not None:
             resolved_venv_path = Path(active_venv_dir).resolve()
+    elif isinstance(venv_path, str):
+        resolved_venv_path = Path(venv_path).resolve()
     else:
         resolved_venv_path = venv_path.resolve()
 
@@ -82,10 +84,7 @@ def _execute_in_venv(
             f"No venv python executable found: {resolved_venv_path}"
         )
 
-    with _temp_script_file(script_source) as script_file:
-        return subprocess.run(
-            [python_executable, script_file], capture_output=True, text=True
-        )
+    return _execute_python_interpreter(python_executable, script_source)
 
 
 def _execute_in_conda_env(
@@ -97,46 +96,33 @@ def _execute_in_conda_env(
     if conda_env is None:
         raise SandboxExecutionError("No conda environment found")
 
-    if shutil.which("conda") is None:
+    conda_exec = shutil.which("conda")
+    if conda_exec is None:
         raise SandboxExecutionError(
             "conda command not found, make sure it is installed on your system and in your PATH"
         )
 
-    with _temp_script_file(script_source) as script_file:
-        # conda run removes the need to activate the environment, as would typically be done in an interactive shell
-        return subprocess.run(
-            [
-                "conda",
-                "run",
-                "-n",
-                conda_env,
-                "--no-capture-output",
-                "python3",
-                script_file,
-            ],
-            capture_output=True,
-            text=True,
-        )
+    # conda run removes the need to activate the environment, as would typically be done in an interactive shell
+    return _execute_python_interpreter(
+        [conda_exec, "run", "-n", conda_env, "--no-capture-output", "python"],
+        script_source,
+    )
 
 
-def _execute_with_system_python(script_source: str) -> subprocess.CompletedProcess:
+def _execute_with_system_path_python(script_source: str) -> subprocess.CompletedProcess:
     python_executable = (
         shutil.which("python3") or shutil.which("python") or sys.executable
     )
-    if not python_executable:
-        raise SandboxExecutionError("No python executable found")
 
-    with _temp_script_file(script_source) as script_file:
-        return subprocess.run(
-            [python_executable, script_file], capture_output=True, text=True
-        )
+    return _execute_python_interpreter(python_executable, script_source)
 
 
 class ExecutionEnvironmentType(Enum):
     AUTO_DETECT = 0  # auto-detect the current python interpreter by looking for an active virtual environment
     VENV = 1  # use the python interpreter specified by a venv environment
     CONDA = 2  # use the python interpreter specified by a conda environment
-    SYSTEM_DEFAULT = 3  # use the system's default python interpreter
+    SYSTEM_PATH = 3  # search for a python interpreter in the system path
+    CURRENT = 4  # Use the python interpreter that is currently executing (i.e. `sys.executable`)
 
 
 def execute_script_in_sandbox(
@@ -166,10 +152,12 @@ def execute_script_in_sandbox(
         elif _is_conda_active():
             return _execute_in_conda_env(script_source)
         else:
-            return _execute_with_system_python(script_source)
+            return _execute_with_system_path_python(script_source)
     elif env_type == ExecutionEnvironmentType.VENV:
         return _execute_in_venv(script_source, kwargs.get("path"))
     elif env_type == ExecutionEnvironmentType.CONDA:
         return _execute_in_conda_env(script_source, kwargs.get("name"))
-    else:
-        return _execute_with_system_python(script_source)
+    elif env_type == ExecutionEnvironmentType.SYSTEM_PATH:
+        return _execute_with_system_path_python(script_source)
+    else:  # ExecutionEnvironmentType.CURRENT
+        return _execute_python_interpreter(sys.executable, script_source)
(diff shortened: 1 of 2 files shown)


src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
```diff
@@ -0,0 +1,201 @@
...
+    timeout: Optional[int],
+) -> subprocess.CompletedProcess:
+    if not python_executable:
+        raise SandboxExecutionError("No python executable found")
+
+    if isinstance(python_executable, str) or isinstance(python_executable, Path):
+        args = [python_executable]
+    else:
+        args = [arg for arg in python_executable]
+    args.append("-")
```
sfc-gh-bgoel, line 52: What is the significance of `-` ?

src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
```diff
@@ -0,0 +1,201 @@
...
+    Executes a python script in a sandboxed environment, and returns its output. The script is executed in a different
+    process. The execution environment is determined by the `env_type` argument. By default, the logic will attempt
+    to auto-detect the correct environment by looking for an active venv or conda environment. If none can be found, it
+    will use the system's default python executable, as determined by the user's path. As a last resort, the current
+    python execution environment will be used (still in a subprocess).
+
+    Parameters:
+        script_source (str): The python script to be executed, as a string.
+        env_type: The type of execution environment to use (default: ExecutionEnvironmentType.AUTO_DETECT).
+        cwd (Optional[Union[str, Path]]): An optional path to use as the current directory when exeecuting the script.
```
sfc-gh-bgoel, line 171: 1. typo in `executing`.
2. Why would we need a `cwd`? The python file must be in the project root.

src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
```diff
@@ -0,0 +1,201 @@
...
+        timeout (Optional[int]): An optional timeout in seconds when executing the script. Defaults to no timeout.
```
sfc-gh-bgoel, line 172: For both `timeout` and `cwd`, are these values ever going to differ from their defaults? If yes, where can a user provide them, the PDF file?

src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
```diff
@@ -0,0 +1,201 @@
...
+    elif env_type == ExecutionEnvironmentType.CONDA:
+        return _execute_in_conda_env(
+            script_source, kwargs.get("name"), cwd=cwd, timeout=timeout
+        )
+    elif env_type == ExecutionEnvironmentType.SYSTEM_PATH:
+        return _execute_with_system_path_python(script_source, cwd=cwd, timeout=timeout)
+    else:  # ExecutionEnvironmentType.CURRENT
+        return _execute_python_interpreter(
+            sys.executable, script_source, cwd=cwd, timeout=timeout
+        )
```
sfc-gh-bgoel, line 201: Right now, the caller of `execute_script_in_sandbox` only passes in `VENV` or `CONDA`, and in the absence of both, `AUTO_DETECT` is the default.
How would `SYSTEM_PATH` (line 196) and `CURRENT` (line 198) become true? Are these supposed to come from the user as part of the `properties` of the processor?

(I do see that `_execute_with_system_path_python` is used in `AUTO_DETECT` too, in addition to line 197)

src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
```diff
@@ -0,0 +1,201 @@
...
+        text=True,
+        input=script_source,
+        timeout=timeout,
+        cwd=cwd,
+    )
+
+
+def _execute_in_venv(
+    script_source: str,
+    venv_path: Optional[Union[str, Path]] = None,
```
sfc-gh-bgoel, line 65: `venv_path` is determined from the PDF, and passed on to this function as a string. Is using `Path` as a type needed here?

src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
```diff
@@ -0,0 +1,201 @@
...
+    if not resolved_venv_path.is_dir():
+        raise SandboxExecutionError(
+            f"venv path should be an existing directory: {resolved_venv_path}"
+        )
+
+    # find the python interpreter for this environment. There is no need to activate environment prior to invoking the
+    # interpreter, as venv maintains the invariant that invoking any of the scripts will set up the virtual environment
+    # correctly. activation scripts are only used for convenience in interactive shells.
+    if _is_ms_windows():
+        python_executable = resolved_venv_path / "Scripts" / "python.exe"
```
sfc-gh-bgoel, line 91: Can this `python.exe` for a venv be referring to an incompatible major version of python, such as `python 2.x`? Do we need to check for that? Specifying venv in my question as running the CLI itself requires 3.8 or higher.

src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
```diff
@@ -0,0 +1,201 @@
...
+def _execute_in_conda_env(
+    script_source: str,
+    env_name: Optional[str] = None,
+    cwd: Optional[Union[str, Path]] = None,
+    timeout: Optional[int] = None,
+) -> subprocess.CompletedProcess:
+    conda_env = env_name
+    if conda_env is None:
+        conda_env = _get_active_conda_env()
+    if conda_env is None:
```
sfc-gh-bgoel, line 114: Nit: Nest under 112? Would prevent an additional check when `conda_env` is not None from the user.

src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+def _execute_with_system_path_python(
+    script_source: str,
+    cwd: Optional[Union[str, Path]] = None,
+    timeout: Optional[int] = None,
+) -> subprocess.CompletedProcess:
+    python_executable = (
+        shutil.which("python3") or shutil.which("python") or sys.executable
+    )
```
sfc-gh-bgoel, line 139: Ditto here about check for python versions.

sfc-gh-cgorrie (commented):
Can we add two tests:

- blocking versions < 3.8 (and making this minimum python version a constant somewhere)
- testing when the underlying script fails somehow (e.g. InterpreterError)

src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
sfc-gh-bdufour: '-' is a standard argument that means 'read from stdin'. Python supports it and interprets it to mean that the script to execute will be piped into the process via stdin. Works on both unix and windows, I tested them. Actually passing a file (I tried that first) doesn't work well on Windows, I ran into permission issues because the subprocess couldn't read the temp file created in the parent.

src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
sfc-gh-bdufour: I wrote the utility function to be generally useful. Felt like a minor decision that can increase reusability.

src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
```diff
@@ -0,0 +1,201 @@
...
+    cwd: Optional[Union[str, Path]] = None,
+    timeout: Optional[int] = None,
+) -> subprocess.CompletedProcess:
+    resolved_venv_path = None
+    if venv_path is None:
+        active_venv_dir = _get_active_venv_dir()
+        if active_venv_dir is not None:
+            resolved_venv_path = Path(active_venv_dir).resolve()
+    elif isinstance(venv_path, str):
+        resolved_venv_path = Path(venv_path).resolve()
```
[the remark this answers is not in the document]
sfc-gh-bdufour: That's true, but my brain doesn't like creating a new object when none is needed :)

src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
sfc-gh-bdufour: This is a utility function, it shouldn't be opinionated about python versions. If we want to check for supported versions, we should do it in the annotation processor preamble.

src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
sfc-gh-bdufour: Answered above

src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
sfc-gh-bdufour: Bhumika, CWD for your use case should be the deploy root, and the file should be relative to that. We need bundle to execute before we start loading stuff, otherwise we can't be sure we're seeing what Snowpark is meant to see.

src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
sfc-gh-bdufour: Cam got what I was trying to do. This is a general function, so it's not too opinionated about these params.

src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
sfc-gh-cgorrie: That's fair, wherever the context screams "snowpark!" that's where we need it.

src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
sfc-gh-bdufour: Again, it's a general function. You use what you need, but you won't need everything right now.

src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
sfc-gh-bgoel: Thank you for explaining, this makes sense.

src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
sfc-gh-bgoel: Ack

src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
sfc-gh-bgoel: Ack

src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
sfc-gh-bgoel: Ack

src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py
sfc-gh-bgoel: Yes,  I mis-typed, I did mean in the project directory not the root. But regardless, this makes sense.

sfc-gh-bdufour:
> Can we add two tests:
> 
> * blocking versions < 3.8 (and making this minimum python version a constant somewhere)
> * testing when the underlying script fails somehow (e.g. InterpreterError)

As discussed, we're not blocking python versions here. As for errors, I added a test that simulates a non-zero exit code. However, if the script fails to execute, that happens in a subprocess so there is nothing to catch. Same with bad command line args or anything like that. The exceptions are raised for timeouts, but I meant for those to propagate (they might be retryable on the client side), or when check_returncode is called, but that's in client code if at all. The sandboxing logic does not need to (or want to) call that.

sfc-gh-cgorrie (approved):
OK with things going in as-is!

sfc-gh-bgoel (approved):
+1, 🚀

PR #1036, opened 2024-05-06
Adding name parameter to job creation and using execute job service sql command

zaramzamzam:
### Changes description

In the hidden `snow spcs job` command I added a new parameter (`name`) this can be used in order to specify the name of the job and not rely on a randomly generated id.
Additionally this changes executing the job with the `EXECUTE SERVICE ...` SQL command to using `EXECUTING JOB SERVICE ...` which is used in the tutorials.

### DIFF
--- a/src/snowflake/cli/plugins/spcs/jobs/manager.py
+++ b/src/snowflake/cli/plugins/spcs/jobs/manager.py
@@ -7,7 +7,7 @@ from snowflake.connector.cursor import SnowflakeCursor
 
 
 class JobManager(SqlExecutionMixin):
-    def create(self, compute_pool: str, spec_path: Path, name: str | None) -> SnowflakeCursor:
+    def create(self, compute_pool: str, spec_path: Path, name: str) -> SnowflakeCursor:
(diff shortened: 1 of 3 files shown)


src/snowflake/cli/plugins/spcs/jobs/manager.py
```diff
@@ -7,11 +7,18 @@
...
         return self._execute_schema_query(
             f"""\
-            EXECUTE SERVICE
+            EXECUTE JOB SERVICE
+            IN COMPUTE POOL {compute_pool}
+            NAME={name}
+            FROM SPECIFICATION $$
+            {spec}
+            $$
+            """ if name else f"""\
```
sfc-gh-yiwang, line 20: NAME is a mandatory property. It must be provided.

sfc-gh-yiwang:
@zaramzamzam Thanks for the contribution. The current job api represents deprecated syntax thus it is hidden. We are going to reintroduce syntax support for spcs job in one of the next versions. Please stay tuned.

PR #1042, opened 2024-05-07
Add feature flags usage to info and telemetry

sfc-gh-turbaszek:
### Pre-review checklist
   * [] I've described my changes in the release notes.

### Changes description
Include feature flags information in `--info` callback and telemetry data

### DIFF
--- a/src/snowflake/cli/app/telemetry.py
+++ b/src/snowflake/cli/app/telemetry.py
@@ -127,5 +127,8 @@ def check_feature_flags_usage():
 
     flags = {}
     for flag_cls in FeatureFlagMixin.__subclasses__():
+        # To skip test classes
+        if flag_cls.__name__.startswith("_"):
+            continue
         flags.update({f.name: f.is_enabled() for f in flag_cls})
     return flags
--- a/src/snowflake/cli/app/telemetry.py
+++ b/src/snowflake/cli/app/telemetry.py
@@ -8,6 +8,7 @@ from typing import Any, Dict, Union
 import click
 from snowflake.cli.__about__ import VERSION
 from snowflake.cli.api.cli_global_context import cli_context
+from snowflake.cli.api.config import get_feature_flags_section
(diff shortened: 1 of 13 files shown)


sfc-gh-pjob (changes requested):
Release notes are not updated. IMO it is an improvement.

src/snowflake/cli/app/telemetry.py
```diff
@@ -117,3 +120,15 @@ def log_command_usage():
...
+
+def check_feature_flags_usage():
+    from snowflake.cli.api.feature_flags import FeatureFlagMixin
+
+    flags = {}
+    for flag_cls in FeatureFlagMixin.__subclasses__():
+        # To skip test classes
+        if flag_cls.__name__.startswith("_"):
+            continue
+        flags.update({f.name: f.is_enabled() for f in flag_cls})
```
sfc-gh-pjob, line 133: Can't we print it similarly in `snow --info`? Now the reader has to know what is the default value of the feature flag to be sure how our CLI is configured.

sfc-gh-turbaszek: Very good point, the purpose is different but it's worth knowing if the value was set or not. Adjusted the code.

sfc-gh-pjob: Hmmm, the last changes in telemetry are good but my comment was more about info presented in `snow --info` that we can print there the same info as we send in telemetry and I think you didn't change anything there.

sfc-gh-turbaszek: Correct. The purpose is different. In telemetry we want to know what was really happening during execution time. In `--info` we want to se what users have in configuration. Telemetry approach scans all flags. Info approach works only on flags known to users. This is deliberate decision to avoid exposing all feature flags to users in case of using --info.

sfc-gh-pjob (approved):

sfc-gh-pjob (approved):

PR #1043, opened 2024-05-07
Fix invalid default value for secrets filed

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
Fixes default value for secrets fields. This was mostly causing warning in tests. There should be no impact on users code.


sfc-gh-cgorrie (approved):
Looks good to me from NADE side. One nitpick: [this line](https://github.com/snowflakedb/snowflake-cli/blob/main/src/snowflake/cli/plugins/snowpark/commands.py#L237) should probably use `secrets.keys()` instead of just `secrets`, and better type hinting in this function is recommended so `object_definition` is of type `Callable`.

sfc-gh-pczajka (approved):

PR #1045, opened 2024-05-08
[SNOW-1345064] Fix stage path handling on Windows

sfc-gh-bdufour:
### Pre-review checklist

### Changes description

The current stage diff code assumes that Unix path separators are used on the local execution environment. This is not the case on Windows, leading to incorrect stage path separators being used. This change introduces an explicit `StagePath` type to make sure that stage paths are handled correctly throughout the diff & deploy logic. I also ended up fixing a hidden stage diff command that has always been broken.

### Testing

Updated all existing tests to pass with the new logic, and removed the workaround for the Windows bug from previous commits. Also manually verified the behaviour of `snow app deploy` with various arguments as well as  `snow app run` on both Mac and Windows. I manually checked the contents of the remote app package stages to make sure the files were published correctly.

### DIFF
--- a/src/snowflake/cli/plugins/nativeapp/manager.py
+++ b/src/snowflake/cli/plugins/nativeapp/manager.py
@@ -321,7 +321,19 @@ class NativeAppManager(SqlExecutionMixin):
     ) -> DiffResult:
         """
         Ensures that the files on our remote stage match the artifacts we have in
-        the local filesystem. Returns the DiffResult used to make changes.
+        the local filesystem.
+
+        Args:
+            role (str): The name of the role to use for queries and commands.
+            prune (bool): Whether to prune artifacts from the stage that don't exist locally.
+            recursive (bool): Whether to traverse directories recursively.
+            local_paths_to_sync (List[Path], optional): List of local paths to sync. Defaults to None to sync all
+             local paths. Note that providing an empty list here is equivalent to None.
+            mapped_files: the file mapping computed during the `bundle` step. Required when local_paths_to_sync is
+             provided.
+
+        Returns:
+            A `DiffResult` instance describing the changes that were performed.
         """
 
         # Does a stage already exist within the application package, or we need to create one?
@@ -346,7 +358,9 @@ class NativeAppManager(SqlExecutionMixin):
         diff: DiffResult = compute_stage_diff(self.deploy_root, self.stage_fqn)
 
         files_not_removed = []
-        if local_paths_to_sync is not None:
+        if local_paths_to_sync:
+            assert mapped_files is not None
+
             # Deploying specific files/directories
             resolved_paths_to_sync = [
                 resolve_without_follow(p) for p in local_paths_to_sync
--- a/src/snowflake/cli/plugins/nativeapp/manager.py
+++ b/src/snowflake/cli/plugins/nativeapp/manager.py
@@ -51,8 +51,8 @@ from snowflake.cli.plugins.nativeapp.utils import verify_exists, verify_no_direc
 from snowflake.cli.plugins.stage.diff import (
     DiffResult,
     StagePath,
+    compute_stage_diff,
     preserve_from_diff,
-    stage_diff,
     sync_local_diff_with_stage,
     to_stage_path,
 )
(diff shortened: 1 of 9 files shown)


src/snowflake/cli/plugins/stage/diff.py
```diff
@@ -18,24 +18,25 @@
 
 log = logging.getLogger(__name__)
 
+StagePath = PurePosixPath  # alias PurePosixPath as StagePath for clarity
```
sfc-gh-cgorrie, line 21: Great idea

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -336,37 +343,40 @@ def sync_deploy_root_with_stage(
             "Performing a diff between the Snowflake stage and your local deploy_root ('%s') directory."
             % self.deploy_root
         )
-        diff: DiffResult = stage_diff(self.deploy_root, self.stage_fqn)
+        diff: DiffResult = compute_stage_diff(self.deploy_root, self.stage_fqn)
 
         files_not_removed = []
-        if len(paths_to_sync) > 0:
+        if local_paths_to_sync is not None:
```
sfc-gh-cgorrie, line 349: I agree that we shouldn't be sending in an empty array for `local_paths_to_sync` (None is better), but what happens if we do? Sensible error message? Does everything work the same as `None`?

sfc-gh-turbaszek (approved):

src/snowflake/cli/plugins/stage/diff.py
```diff
@@ -156,109 +167,123 @@ def build_md5_map(list_stage_cursor: DictCursor) -> Dict[str, str]:
...
+        result.only_on_stage.append(stage_filename)
 
     return result
 
 
-def get_stage_path_from_file(filepath: str):
-    parent = str(Path(filepath).parent)
-    stage_path = "" if parent == "." else parent
-    return stage_path
+def get_stage_subpath(stage_path: StagePath) -> str:
```
sfc-gh-cgorrie, line 229: Can we call this `get_stage_parent_path` instead?

sfc-gh-cgorrie (commented):
A couple thoughts here, moving onto manual testing

src/snowflake/cli/plugins/stage/diff.py
sfc-gh-bdufour: Well it was yours a long time ago :D

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-bdufour: This change was really about removing the mutable default object, because Python is weird that way and it's bad practice. I ended up going with "if you pass a list, we use it" for the semantics. I can change things so that empty is the same as None, but right now you'd likely end up with an empty diff instead.

src/snowflake/cli/plugins/stage/diff.py
sfc-gh-bdufour: Context: I went with that this is called when used, because it's not quite the parent, it's empty if the parent is ".". Weird contract, I didn't want to go overboard with the refactorings so I kept this method but renamed it to avoid the even more ambiguous name it had before.

However, "." -> "" isn't a particularly big part of the contract, so I can rename as suggested.

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-cgorrie: OK, as long as it's clear in the docstrings I'm fine with that!

src/snowflake/cli/plugins/stage/diff.py
sfc-gh-cgorrie: I'll leave it up to you. I can see both arguments.

sfc-gh-bdufour: Ah, now I remember. I was trying to preserve the invariant that anywhere "stage_path" or "stage path" is used, it means a `StagePath` instance in the code. This is not a proper stage path, so I avoided the term on purpose. Will leave as is then, if that's alright with you. I don't want to introduce confusion by making `stage_parent_path` return a string.

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-bdufour: Added full doc string, but since passing an empty list makes no sense, I flipped to the logic to make `None` equivalent to `[]`, it's a fair callout.

sfc-gh-cgorrie (approved):
Manual testing passed, everything LGTM

sfc-gh-turbaszek (approved):

sfc-gh-turbaszek (approved):

sfc-gh-bgoel (approved):
lgtm

PR #1048, opened 2024-05-09
[NADE] Add feature check to allow codegen

sfc-gh-bgoel:
### Pre-review checklist
   * [ ] I've added or updated automated unit tests to verify correctness of my new code - N/A
   * [ ] I've confirmed that my changes are working by executing CLI's commands manually - N/A

### Changes description
Add a feature check in `build_bundle` to allow for codegen.

### DIFF
--- a/src/snowflake/cli/plugins/nativeapp/manager.py
+++ b/src/snowflake/cli/plugins/nativeapp/manager.py
@@ -31,6 +31,9 @@ from snowflake.cli.plugins.nativeapp.artifacts import (
     source_path_to_deploy_path,
     translate_artifact,
 )
+from snowflake.cli.plugins.nativeapp.codegen.compiler import (
+    _find_and_execute_processors,
+)
 from snowflake.cli.plugins.nativeapp.constants import (
     ALLOWED_SPECIAL_COMMENTS,
     COMMENT_COL,
@@ -47,6 +50,7 @@ from snowflake.cli.plugins.nativeapp.exceptions import (
     MissingPackageScriptError,
     UnexpectedOwnerError,
 )
+from snowflake.cli.plugins.nativeapp.feature_flags import FeatureFlag
 from snowflake.cli.plugins.nativeapp.utils import verify_exists, verify_no_directories
 from snowflake.cli.plugins.stage.diff import (
     DiffResult,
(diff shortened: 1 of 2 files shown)


src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -47,6 +50,7 @@
     MissingPackageScriptError,
     UnexpectedOwnerError,
 )
+from snowflake.cli.plugins.nativeapp.feature_flags import FeatureFlag
```
sfc-gh-turbaszek, line 53: I know this is not the gist of the change but should we consider renaming `FeatureFlag` to `NativeAppFeatureFlags`? In this way we may avoid any ambiguity with existing `snowflake.cli.api.feature_flags.FeatureFlag.` WDYT?

sfc-gh-bdufour: Isn't it common already to have the same files in multiple plugin directories if they define the same concepts? I'm thinking commands and manager, for example. Native apps code shouldn't rely on feature flags outside of its own IMHO, so at least now there shouldn't be ambiguity.

sfc-gh-bdufour:
Bhumika, not following the target branch here. Shouldn't we target feature-snowpark-annotations instead? Let's not create complex branching/merging structures, it will be a pain to deal with later. This should stack on top of your other work, on the common feature branch.

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -309,7 +313,14 @@ def build_bundle(self) -> ArtifactDeploymentMap:
         """
         Populates the local deploy root from artifact sources.
         """
-        return build_bundle(self.project_root, self.deploy_root, self.artifacts)
+        mapped_files = build_bundle(self.project_root, self.deploy_root, self.artifacts)
+        if FeatureFlag.ENABLE_SETUP_SCRIPT_GENERATION.is_enabled():
+            _find_and_execute_processors(
```
sfc-gh-bdufour, line 318: remind me: where does the output of this step go?

sfc-gh-bgoel: We do not have the generate DDL piece in yet, which I can direct to stdout. 
If you just want to see the JSON output for now, I can output that instead. Will add the change in the next commit.

sfc-gh-bgoel (commented):
Bruno, yes you are right that the target branch needs to be the feature branch I created. But this should either be 1. Automatically changed by github once I merge bgoel-collect-external-functions or 2. I retarget it to feature-snowflake-snowpark.

The current way of targeting was to isolate changes from just this enablement.

sfc-gh-bdufour (approved):

PR #1056, opened 2024-05-10
[NADE] DDL Generation for Extension Functions

sfc-gh-bgoel:
### Pre-review checklist
   * [ ] I've confirmed that my changes are working by executing CLI's commands manually - TBD

### Changes description
Part of Setup Script Codegen. The changes introduced in this PR collect the information from snowpark callback, validate/sanitize/update that information to ultimately generate SQL statements.


src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function_utils.py
```diff
@@ -13,6 +13,107 @@
...
+    has_eai = (
+        ex_fn.get("external_access_integrations", None)
+        and len(ex_fn["external_access_integrations"]) > 0
+    )
+    ex_fn["external_access_integrations"] = (
+        ex_fn["external_access_integrations"] if has_eai else None
+    )
+
+    has_secrets = ex_fn.get("secrets", None) and len(ex_fn["secrets"]) > 0
+    ex_fn["secrets"] = ex_fn["secrets"] if has_secrets else None
```
sfc-gh-bdufour, line 74: that's a lot of code to sanitize fields that are optional. Not sure if we really need it, something like:
```
secrets = ex_fn.get("secrets")
if secrets:
     pass  # use it
```

doesn't seem so bad. If we do want this much sanitization, we could just introduce a helper to do it for multiple fields:

```
sanitize_list_attribute(ex_fn, "secrets")
sanitize_list_attribute(ex_fn, "inline_python_code")
...
```

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -299,3 +307,140 @@ def get_src_py_file_to_dest_py_file_map(
...
+{inline_python_code}
+$$
+"""
+    else:
+        inline_python_code_in_sql = ""
+
+    try:
+        assert ex_fn["runtime_version"] is not None
+        assert ex_fn["handler"] is not None
+    except AssertionError as err:
```
sfc-gh-bdufour, line 408: ditto, that's not a pattern we should use

tests/testing_utils/fixtures.py
```diff
@@ -338,3 +338,54 @@ def native_app_project_instance():
             },
         }
     )
+
+
+@pytest.fixture()
+def native_app_codegen_instance():
+    return {
+        "object_type": "TABLE_FUNCTION",
+        "object_name": "SNOWPARK_TEMP_FUNCTION_WZUNHMZJKA",
```
sfc-gh-bdufour, line 347: for now I would prefer to focus on the case where the name was explicitly provided, and even make that an strong requirement for the users. The other stuff we do is really just a heuristic. We could deliver without, if need be.

tests/testing_utils/fixtures.py
```diff
@@ -338,3 +338,54 @@ def native_app_project_instance():
...
+        "execute_as": "owner",
+        "anonymous": False,
+        "func": "<lambda>",
+        "schema": "data",
+        "application_roles": ["app_admin", "app_viewer"],
+    }
+
+    # {
+    #     "object_type": "FUNCTION",
+    #     "object_name": "SNOWPARK_TEMP_FUNCTION_S782HXBR30",
```
sfc-gh-bdufour, line 372: nit: clean up?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -299,3 +307,140 @@ def get_src_py_file_to_dest_py_file_map(
...
+def generate_create_sql_ddl_statements(ex_fn: Dict[str, Any]) -> str:
+    """
+    Generates a "CREATE FUNCTION/PROCEDURE ... " SQL DDL statement based on a dictionary of extension function properties.
+    Logic for this create statement has been lifted from snowflake-snowpark-python v1.15.0 package.
+    """
+
+    object_type = ex_fn["object_type"]
+    object_name = ex_fn["object_name"]
+
+    if object_type == "PROCEDURE" and ex_fn.get("anonymous", False):
```
sfc-gh-bdufour, line 349: nit: why would anonymous need a default vs making your sanitization pass handle it?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -299,3 +307,140 @@ def get_src_py_file_to_dest_py_file_map(
...
+        cc.warning(
+            dedent(
+                f"""{object_type.replace(' ', '-')} {object_name} cannot be an anonymous procedure in a Snowflake Native App.
+                    Skipping generation of 'CREATE FUNCTION/PROCEDURE ...' SQL statement for this object."""
+            )
+        )
+        return ""
```
sfc-gh-bdufour, line 356: nit: returning None would be clearer on failure

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -299,3 +307,140 @@ def get_src_py_file_to_dest_py_file_map(
...
+        cc.warning(f"Could not generate SQL DDL due to incorrect information:\n{err}")
+        return ""
+
+    create_query = dedent(
+        f"""\
+CREATE{replace_in_sql}
+{_get_object_type_as_text(object_type)} {'IF NOT EXISTS' if ex_fn["if_not_exists"] else ''}{object_name}({sql_func_args})
+{ex_fn["return_sql"]}
+LANGUAGE PYTHON
+RUNTIME_VERSION={ex_fn["runtime_version"]} {imports_in_sql}{packages_in_sql}{external_access_integrations_in_sql}{secrets_in_sql}
```
sfc-gh-bdufour, line 418: just thought of this, but do we default the runtime version to the one from the sandbox executing the user's code? If not, we should add a task for that.

tests/testing_utils/fixtures.py
sfc-gh-bgoel: I don't believe we can distinguish between Snowpark's temp name and a user-provided name that happens to begin with this "SNOWPARK_TEMP_" prefix. We do provide precedence to objects that do not begin with "SNOWPARK_TEMP_" prefix. Worst case where a user provided this prefix, we will add the schema name to the rest of the object_name.

sfc-gh-bdufour: We should probably make another change to the Snowpark lib to optionally avoid generating names. We need to be able to tell when that's happening, and a prefix isn't the right way. It's too brittle. That being said, we can have the logic to handle the prefix specially for now, but we should really tackle the real solution before that logic is forced to live forever.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bdufour: Sounds like Snowpark is already doing what I thought it should, so that's OK for now. Currently, the runtime will always be populated by Snowpark then? If so, that's good enough.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function_utils.py
```diff
@@ -5,14 +5,190 @@
...
+
+    optional_expected_type: Optional[List] = []
+    default_raw_imports: List[Union[str, Tuple[str, str]]] = []
+    _sanitize_ex_fn_attribute(
+        attr="raw_imports",
+        ex_fn=ex_fn,
+        expected_type=type(optional_expected_type),
+        default_value=default_raw_imports,
+    )
+    if len(ex_fn["raw_imports"]) == 0:
```
sfc-gh-bdufour, line 109: why is this an error?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function_utils.py
```diff
@@ -5,14 +5,190 @@
...
+        raise MalformedExtensionFunctionError(
+            _create_missing_attr_str(attribute="raw_imports", py_file=py_file)
+        )
+
+    _sanitize_ex_fn_attribute(
+        attr="schema", ex_fn=ex_fn, make_uppercase=True, expected_type=str
+    )
+    if ex_fn["schema"] is None:
+        raise MalformedExtensionFunctionError(
```
sfc-gh-bdufour, line 118: NB: Eventually we'll want to switch to accumulating warnings/errors instead of relying on exceptions like this. This way, we can control the output and the flow of the codegen process.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -220,15 +220,30 @@ def process(
...
+                    can_proceed = sanitize_extension_function_data(
+                        ex_fn=entity, py_file=dest_file
+                    )
+                    if not can_proceed:
+                        cc.warning(
+                            f"Skipping generation of 'CREATE FUNCTION/PROCEDURE ...' SQL statement for this object."
+                        )
+                        continue
+                    enrich_entity(
                         entity=entity,
```
sfc-gh-bdufour, line 232: nit: I fit it confusing that ex_fn and entity are both used for the same concept. Can we just keep ex_fn for clarity?

tests/nativeapp/codegen/snowpark/test_extension_function_utils.py
```diff
@@ -7,6 +8,68 @@
...
+    )
+    assert ex_fn["team"] == ["nade"]
+
+    with pytest.raises(ef_utils.MalformedExtensionFunctionError):
+        ef_utils._sanitize_ex_fn_attribute(  # noqa: SLF001
+            attr="team", ex_fn=ex_fn, make_uppercase=True, expected_type=Dict
+        )
+
+
+def test_create_missing_attr_str():
```
sfc-gh-bdufour, line 63: that's just an internal helper, IMHO this kind of stuff should be tested through other behaviour. It makes it so much harder to refactor code when we test the implementation instead of the contract.

tests/nativeapp/codegen/snowpark/test_extension_function_utils.py
```diff
@@ -7,6 +8,68 @@
 from tests.testing_utils.files_and_dirs import temp_local_dir
 
 
+def test_sanitize_ex_fn_attribute():
```
sfc-gh-bdufour, line 11: See my comment below about testing implementation vs contract. For example, say we want to split up this method per parameter type, we have to rewrite all tests.

tests/nativeapp/codegen/snowpark/test_extension_function_utils.py
```diff
@@ -92,3 +181,127 @@ def test_get_all_imports():
...
+    ex_fn["schema"] = "core"
+    with pytest.raises(ef_utils.MalformedExtensionFunctionError) as err:
+        ef_utils.sanitize_extension_function_data(ex_fn=ex_fn, py_file=some_path)
+    assert err.value.message.__contains__("runtime_version")
+
+    ex_fn["runtime_version"] = "3.8"
+    assert ef_utils.sanitize_extension_function_data(ex_fn=ex_fn, py_file=some_path)
+    # This check is because we cannot stop the function execution beyond required keys, but there are other tests below that go in depth
+
+    assert ex_fn["object_type"] == "FUNCTION"
```
sfc-gh-bdufour, line 249: nit: dictionaries define equality, so you can just assert the entire dictionary at once. As a bonus, this makes sure there are no extra keys.

tests/nativeapp/codegen/snowpark/test_python_processor.py
```diff
@@ -286,3 +288,147 @@ def test_execute_in_sandbox_none_entity_case2(mock_jinja, mock_sandbox):
...
+    actual = generate_grant_sql_ddl_statements(ex_fn=ex_fn)
+    expected = "GRANT USAGE ON TABLE FUNCTION CORE.MYFUNC\nTO APPLICATION ROLE APP_ADMIN;\n\nGRANT USAGE ON TABLE FUNCTION CORE.MYFUNC\nTO APPLICATION ROLE APP_VIEWER;\n"
+    assert actual == expected
+
+
+def test_generate_sql_ddl_statements_empty(native_app_project_instance):
+    dir_structure = {
+        "a/b/c/main.py": "# this is a file\n",
+        "a/b/c/data.py": "# this is a file\n",
+        "output/deploy": None,
```
sfc-gh-bdufour, line 343: The utility function takes files, not directories. This looks out of place, I _think_ it will try to create an empty file named `deploy`

tests/nativeapp/codegen/snowpark/test_python_processor.py
```diff
@@ -286,3 +288,147 @@ def test_execute_in_sandbox_none_entity_case2(mock_jinja, mock_sandbox):
...
+
+
+def test_generate_grant_sql_ddl_statements():
+    ex_fn = {
+        "object_type": "TABLE_FUNCTION",
+        "object_name": "CORE.MYFUNC",
+        "application_roles": ["APP_ADMIN", "APP_VIEWER"],
+    }
+    actual = generate_grant_sql_ddl_statements(ex_fn=ex_fn)
+    expected = "GRANT USAGE ON TABLE FUNCTION CORE.MYFUNC\nTO APPLICATION ROLE APP_ADMIN;\n\nGRANT USAGE ON TABLE FUNCTION CORE.MYFUNC\nTO APPLICATION ROLE APP_VIEWER;\n"
```
sfc-gh-bdufour, line 335: hard to read, why not use snapshots here too?

tests/nativeapp/codegen/snowpark/test_python_processor.py
```diff
@@ -286,3 +288,147 @@ def test_execute_in_sandbox_none_entity_case2(mock_jinja, mock_sandbox):
...
+    native_app_codegen_instance, snapshot
+):
+    native_app_codegen_instance["replace"] = False
+    native_app_codegen_instance["all_imports"] = ""
+    native_app_codegen_instance["all_packages"] = ""
+    native_app_codegen_instance["external_access_integrations"] = None
+    native_app_codegen_instance["secrets"] = None
+    native_app_codegen_instance["execute_as"] = None
+    native_app_codegen_instance["inline_python_code"] = None
+    assert generate_create_sql_ddl_statements(native_app_codegen_instance) == snapshot
```
sfc-gh-bdufour, line 318: I like these snapshot tests for DDL quite a bit, but IMHO we need quite a few more to achieve the desired coverage of the codegen contract.

tests/nativeapp/codegen/snowpark/test_python_processor.py
```diff
@@ -286,3 +288,147 @@ def test_execute_in_sandbox_none_entity_case2(mock_jinja, mock_sandbox):
...
+        }
+        result = snowpark_processor.generate_sql_ddl_statements(
+            dest_file_py_file_to_collected_entities
+        )
+        assert len(result) == 1
+
+
+@mock.patch(
+    "snowflake.cli.plugins.nativeapp.codegen.snowpark.python_processor.generate_grant_sql_ddl_statements"
+)
```
sfc-gh-bdufour, line 392: This is odd, why do we need to mock our own logic?

tests/nativeapp/codegen/snowpark/test_python_processor.py
```diff
@@ -286,3 +288,147 @@ def test_execute_in_sandbox_none_entity_case2(mock_jinja, mock_sandbox):
...
+            project_root=local_path,
+            deploy_root=Path(local_path, "output/deploy"),
+        )
+        dest_file_py_file_to_collected_entities: Dict[Path, Optional[Any]] = {
+            Path("stagepath", "main.py"): [native_app_codegen_instance]
+        }
+        result = snowpark_processor.generate_sql_ddl_statements(
+            dest_file_py_file_to_collected_entities
+        )
+        assert len(result) == 1
```
sfc-gh-bdufour, line 387: isn't this a really weak test?

tests/nativeapp/codegen/snowpark/test_python_processor.py
```diff
@@ -286,3 +288,147 @@ def test_execute_in_sandbox_none_entity_case2(mock_jinja, mock_sandbox):
...
+
+        dest_file_py_file_to_collected_entities: Dict[Path, Optional[Any]] = {
+            Path("stagepath", "main.py"): [native_app_codegen_instance]
+        }
+        mock_generate.return_value = "SOME_STRING"
+        result = snowpark_processor.generate_sql_ddl_statements(
+            dest_file_py_file_to_collected_entities
+        )
+        assert len(result) == 1
+        mock_generate.assert_called_once()
```
sfc-gh-bdufour, line 434: I think I need your help to understand the general strategy you're following for these tests. What's your condition to determine when sufficient testing has been added? I ask because I'm struggling to understand, with what's there now, how much we trust the implementation.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function_utils.py
sfc-gh-bgoel: Added TODO.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function_utils.py
```diff
@@ -5,14 +5,190 @@
...
+
+
+class MalformedExtensionFunctionError(ClickException):
+    """Required extension function attribute is missing."""
+
+    def __init__(self, message: str):
+        super().__init__(message=message)
+
 
 TEMP_OBJECT_NAME_PREFIX = "SNOWPARK_TEMP_"
```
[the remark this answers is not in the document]
sfc-gh-bgoel: Added TODO.

tests/nativeapp/codegen/snowpark/test_extension_function_utils.py
sfc-gh-bgoel: Made some changes to the two testing files. Closing comments that relate to this as an acknowledgement as they would be on outdated code state, please open new comments for feedback.

As of the commit after this reply, `test_extension_function_utils.py` cov stands at 99%, and `test_python_processor.py` is at 98%.

tests/nativeapp/codegen/snowpark/test_extension_function_utils.py
```diff
@@ -1,94 +1,195 @@
...
-        dest_path = Path(local_path, "output", "deploy", input_param)
-        actual = ef_utils._get_handler_path_without_suffix(  # noqa: SLF001
-            file_path=dest_path,
-            suffix_str_to_rm=".py",
-            deploy_root=Path(local_path, "output", "deploy"),
-        )
-        assert actual == expected
+# --------------------------------------------------------
+# ------------- get_object_type_as_text ------------------
+# --------------------------------------------------------
```
sfc-gh-bgoel, line 8: This is a large test file, added these comments to create sections.

tests/nativeapp/codegen/snowpark/test_extension_function_utils.py
```diff
@@ -1,94 +1,195 @@
...
-        object_name, schema, handler
+def test_sanitize_extension_function_data_required_keys(snapshot):
+
+    ex_fn = {}
+    some_path = Path("some/path")
+
+    # Test for absence or malformed object_type, object_name and return_sql
+    with pytest.raises(ef_utils.MalformedExtensionFunctionError) as err:
+        ef_utils.sanitize_extension_function_data(ex_fn=ex_fn, py_file=some_path)
+    assert err.value.message.__contains__("object_type")
```
sfc-gh-bdufour, line 37: why not `assert "object_type" in err.value.message` ? calling `__contains__` directly is very odd.

tests/nativeapp/codegen/snowpark/test_extension_function_utils.py
```diff
@@ -1,94 +1,195 @@
...
+
+    ex_fn["object_type"] = ["function"]
+    with pytest.raises(ef_utils.MalformedExtensionFunctionError) as err:
+        ef_utils.sanitize_extension_function_data(ex_fn=ex_fn, py_file=some_path)
+    assert err.value.message.__contains__("object_type")
+
+    ex_fn["object_type"] = "function"
+    with pytest.raises(ef_utils.MalformedExtensionFunctionError) as err:
+        ef_utils.sanitize_extension_function_data(ex_fn=ex_fn, py_file=some_path)
+    assert err.value.message.__contains__("object_name")
```
sfc-gh-bdufour, line 47: took me a minute to figure out the pattern here. You're basically building up to a a correct dictionary, so you keep asserting that the next thing to fix is flagged. Maybe add that as a comment to help the next person reading this?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function_utils.py
```diff
@@ -5,14 +5,253 @@
...
+
+    Parameters:
+        name (str): Any arbitrary string
+    Returns:
+        A string that has replaced underscores with spaces.
+    """
+    return name.replace("_", " ")
+
+
+def _sanitize_ex_fn_str_attribute(
```
sfc-gh-bdufour, line 39: nit: I would either remove `ex_fn` here (since the module name makes it clear that extension functions are the scope), or expand it. I don't think this name will be clear in the long run, after we're done with the initial feature.

sfc-gh-bdufour (approved):

PR #1058, opened 2024-05-10
Snow 1369491 snow object create command

sfc-gh-pczajka:
### Pre-review checklist

### Changes description
Add hidden "create" command, which is a thin layer over Snowflake REST API CREATE commands. Supported objects:
https://docs.snowflake.com/LIMITEDACCESS/rest-api/reference/index.html


src/snowflake/cli/plugins/object/commands.py
```diff
@@ -21,6 +21,15 @@
 ObjectArgument = typer.Argument(
     help="Type of object. For example table, procedure, streamlit.",
     case_sensitive=False,
+    show_default=False,
+)
+ObjectDefinitionArgument = typer.Argument(
+    help="""Object definition in JSON format, for example \'{"name": "my_database", "owner": "owner_role"}\',
```
sfc-gh-turbaszek, line 27: Is this a valid example?

sfc-gh-pczajka: I've checked - "owner" is not settable field, it is ignored by request (that's why my previous check passed). I've changed it to "comment"

tests_integration/test_object.py
```diff
@@ -118,3 +121,55 @@ def test_show_drop_image_repository(runner, test_database, snowflake_session):
     )
     assert result_drop.exit_code == 0, result_drop.output
     assert f"{repo_name} successfully dropped" in result_drop.output
+
+
+@pytest.mark.parametrize(
```
sfc-gh-pczajka, line 126: `table`, `task`, `warehouse` work when called manually, integration test role *probably* don't have enough permissions

src/snowflake/cli/plugins/object/commands.py
```diff
@@ -21,6 +21,17 @@
...
     help="Type of object. For example table, procedure, streamlit.",
     case_sensitive=False,
+    show_default=False,
+)
+ObjectDefinitionArgument = typer.Argument(
+    help="""Object definition in JSON format, for example
+\'{"name": "my_db", "comment": "created with Snowflake CLI"}\', or provided as a list of key=value pairs,
+for example: name=my_db 'comment=created with Snowflake CLI'.
+
+Check https://docs.snowflake.com/LIMITEDACCESS/rest-api/reference/ for the full list of available parameters
```
sfc-gh-turbaszek, line 31: Do not link to private docs, we should not merge the changes as long as APIs are not in PuPr.

sfc-gh-turbaszek (changes requested):
Private feature atm.

src/snowflake/cli/plugins/object/commands.py
sfc-gh-pczajka: I left TODO to be filled, and changed base branch from main to `object-create`, which will collect PRs and merge them together into main when APIs will gave gone PuPr.

src/snowflake/cli/plugins/object/manager.py
```diff
@@ -52,3 +55,83 @@ def object_exists(self, *, object_type: str, name: str):
...
+        try:
+            result = self._send_rest_request(url, method="get")
+            return bool(result) or result == []
+        except InterfaceError as err:
+            if "404 Not Found" in str(err):
+                return False
+            raise err
+
+    def _get_rest_api_create_url(self, object_type: str):
+        plural_object_type = _get_object_names(object_type).sf_plural_name.replace(
```
sfc-gh-turbaszek, line 99: This is based on `OBJECT_TO_NAMES` which is a hardcoded list of supported objects. It does not sounds like something we agreed on (having hardcoded list).

sfc-gh-pczajka: I used that as it is a "sure way" of correctly pluralize the type, but I see what you mean.

Looking at a list of currently supported objects, I replaced that with a simple heuristic

src/snowflake/cli/plugins/object/manager.py
```diff
@@ -52,3 +55,85 @@ def object_exists(self, *, object_type: str, name: str):
...
+            full_url=full_url,
+            headers=headers,
+            token=rest.token,
+            data=json.dumps(data if data else {}),
+            no_retry=True,
+        )
+
+    def _url_exists(self, url):
+        try:
+            result = self._send_rest_request(url, method="get")
```
sfc-gh-astus, line 91: Instead of `get` you could use `head`.

src/snowflake/cli/plugins/object/manager.py
```diff
@@ -52,3 +55,85 @@ def object_exists(self, *, object_type: str, name: str):
...
+            return bool(result) or result == []
```
sfc-gh-astus, line 92: Why don't check http code?

src/snowflake/cli/plugins/object/manager.py
```diff
@@ -52,3 +55,85 @@ def object_exists(self, *, object_type: str, name: str):
...
+            if not response:
+                raise ClickException(
+                    dedent(
+                        """                An unexpected error occurred while creating the object. Try again with --debug for more info.
+                Most probable reasons:
+                  * object you are trying to create already exists
+                  * role you are using do not have permissions to create this object"""
+                    )
+                )
+            return response["status"]
```
sfc-gh-astus, line 137: Not checking http code?

tests_integration/test_object.py
sfc-gh-astus: I think we should grant create for them and test it, but this will requires extending of our cleanup function. Can be done in separate task.

src/snowflake/cli/plugins/object/commands.py
```diff
@@ -92,3 +104,45 @@ def describe(
...
+        try:
+            return _parse_list_to_dict(object_definition)
+        except ClickException:
+            raise json_err
+
+
+@app.command(name="create", requires_connection=True, hidden=True)
+def create(
+    object_type: str = ObjectArgument,
+    object_definition: List[str] = ObjectDefinitionArgument,
```
sfc-gh-turbaszek, line 141: Have you considered separating the inputs into two: 
* `object_definition: dict` - where we use callback to parse the data
* `object_attirutes: List[str]` - where we support key=value specification

Won't that help with logic complexity?

src/snowflake/cli/plugins/object/commands.py
```diff
@@ -92,3 +104,45 @@ def describe(
...
+
+
+def _parse_object_definition(object_definition: List[str]) -> Dict[str, Any]:
+    import json
+
+    def _parse_list_to_dict(object_definition: List[str]) -> Dict[str, Any]:
+        payload = {}
+        for item in object_definition:
+            try:
+                key, value = item.split("=", 1)
```
sfc-gh-turbaszek, line 116: Should we be able to use `parse_key_value_variables` as we do in `snow sql` and `snow stage execute`?

src/snowflake/cli/plugins/object/commands.py
sfc-gh-pczajka: Yes, but the usage would be more painful. My thoughts:
If we wanted to have two arguments, both need to be optional, which means we have to use `typer.Option`, which require flags. So usage would be `-D key=value -D key=value` or `--json '{"name": "my_db"}'`

My personal preference is simpler usage (without multiple `-D`), but I'm open for discussion

src/snowflake/cli/plugins/object/commands.py
sfc-gh-pczajka: Nice catch, didn't know about that one!

sfc-gh-pczajka: fixed

src/snowflake/cli/plugins/object/commands.py
sfc-gh-pczajka: Fixed to `snow init [ATTRIBUTES] [--json DEFINITION]` format

src/snowflake/cli/plugins/object/manager.py
sfc-gh-pczajka: I'd love to - we don't get it back from python connector :(

src/snowflake/cli/plugins/object/manager.py
sfc-gh-pczajka: I tried - our REST API seem to not support HEAD requests (it always return 404)

src/snowflake/cli/plugins/object/manager.py
sfc-gh-pczajka: I'd love to - we don't get that back from python connector:
```
➜ snow object create database name=mordo1
{'status': 'Database MORDO1 successfully created.'}
Database MORDO1 successfully created.

```

sfc-gh-astus (approved):

src/snowflake/cli/plugins/object/commands.py
```diff
@@ -35,6 +35,29 @@
 ObjectArgument = typer.Argument(
     help="Type of object. For example table, procedure, streamlit.",
     case_sensitive=False,
+    show_default=False,
+)
+# TODO: add documentation link
+ObjectAttributesArgument = typer.Argument(
+    None,
+    help="""Object attributes provided as a list of key=value pairs,
+for example name=my_db 'comment=created with Snowflake CLI'.
```
sfc-gh-turbaszek, line 44: why one has quotes and other does not?

sfc-gh-turbaszek, line 44: Is there any possibility for nested attirbutes?

src/snowflake/cli/plugins/object/commands.py
```diff
@@ -106,3 +129,36 @@ def describe(
...
+
+    if object_json:
+        import json
+
+        object_data = json.loads(object_json)
+
+    elif object_attributes:
+        object_data = {
+            v.key: v.value for v in parse_key_value_variables(object_attributes)
+        }
```
sfc-gh-turbaszek, line 156: Any support for nested attirbutes?

src/snowflake/cli/plugins/object/manager.py
```diff
@@ -66,3 +69,85 @@ def object_exists(self, *, object_type: str, name: str):
...
+        return rest.fetch(
+            method=method,
+            full_url=full_url,
+            headers=headers,
+            token=rest.token,
+            data=json.dumps(data if data else {}),
+            no_retry=True,
+        )
+
+    def _url_exists(self, url):
```
sfc-gh-turbaszek, line 103: Add type hints

src/snowflake/cli/plugins/object/manager.py
```diff
@@ -66,3 +69,85 @@ def object_exists(self, *, object_type: str, name: str):
...
+        except InterfaceError as err:
+            if "404 Not Found" in str(err):
+                return False
+            raise err
+
+    def _get_rest_api_create_url(self, object_type: str):
+        if object_type.endswith("y"):
+            plural_object_type = object_type[:-1].lower() + "ies"
+        else:
+            plural_object_type = object_type.lower() + "s"
```
sfc-gh-turbaszek, line 116: Let's separate this to separate util method

src/snowflake/cli/plugins/object/manager.py
```diff
@@ -66,3 +69,85 @@ def object_exists(self, *, object_type: str, name: str):
...
+
+        url_prefix = "/api/v2"
```
sfc-gh-turbaszek, line 118: Make this a global constant

src/snowflake/cli/plugins/object/manager.py
```diff
@@ -66,3 +69,85 @@ def object_exists(self, *, object_type: str, name: str):
...
+
+        url = f"{url_prefix}/{plural_object_type}/"
+        if self._url_exists(url):
+            return url
```
sfc-gh-turbaszek, line 122: Add comments explaining why this pre-check

src/snowflake/cli/plugins/object/manager.py
```diff
@@ -66,3 +69,85 @@ def object_exists(self, *, object_type: str, name: str):
...
+    def create(self, object_type: str, object_data: Dict[str, Any]) -> str:
+        url = self._get_rest_api_create_url(object_type)
+        if not url:
+            return f"Create operation for type {object_type} is not supported. Try using `sql -q 'CREATE ...'` command"
+        try:
+            response = self._send_rest_request(url=url, method="post", data=object_data)
+            if not response:
+                raise ClickException(
+                    dedent(
+                        """                An unexpected error occurred while creating the object. Try again with --debug for more info.
```
sfc-gh-turbaszek, line 145: Won't this be duplicated in the output due to global handling?

src/snowflake/cli/plugins/object/manager.py
```diff
@@ -66,3 +69,85 @@ def object_exists(self, *, object_type: str, name: str):
...
+                Most probable reasons:
+                  * object you are trying to create already exists
+                  * role you are using do not have permissions to create this object"""
+                    )
+                )
+            return response["status"]
+        except BadRequest:
+            raise ClickException("Incorrect object definition.")
```
sfc-gh-turbaszek, line 153: Is there any information from the error that we can expose to the users? What exactly was wrong?

src/snowflake/cli/plugins/object/manager.py
sfc-gh-pczajka: No, python connector surpresses the exception. I've added a "workaround", but I think it's worth to discuss with connector team to change that behaviour.

src/snowflake/cli/plugins/object/manager.py
sfc-gh-pczajka: No, the only information we get is "409 bad request", which, according to documentation, means malformatted payload (for example misspelled argument)

src/snowflake/cli/plugins/object/commands.py
sfc-gh-pczajka: If the value contains spaces, it should be quoted. I just checked, `comment='created with Snowflake CLI'` works as well, changed.

src/snowflake/cli/plugins/object/commands.py
sfc-gh-pczajka: Nice catch, fixed

src/snowflake/cli/plugins/object/manager.py
sfc-gh-pczajka: Added a docstring and refactored so it is more clear

src/snowflake/cli/plugins/object/manager.py
sfc-gh-pczajka: done

src/snowflake/cli/plugins/object/manager.py
sfc-gh-pczajka: Fixed, this is the first use

src/snowflake/cli/plugins/object/commands.py
```diff
@@ -106,3 +129,36 @@ def describe(
...
+def create(
+    object_type: str = ObjectArgument,
+    object_attributes: Optional[List[str]] = ObjectAttributesArgument,
+    object_json: str = ObjectDefinitionJsonOption,
+    **options,
+):
+    """Create an object of a given type. List of supported objects
+    and parameters: https://docs.snowflake.com/LIMITEDACCESS/rest-api/reference/"""
+    if object_attributes and object_json:
+        raise ClickException(
```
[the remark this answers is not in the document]
sfc-gh-pczajka: https://snowflakecomputing.atlassian.net/browse/SNOW-1482252

src/snowflake/cli/plugins/object/manager.py
```diff
@@ -66,3 +69,85 @@ def object_exists(self, *, object_type: str, name: str):
...
+            return url
+
+        db = self._conn.database
+        url = f"{url_prefix}/databases/{db}/{plural_object_type}/"
+        if self._url_exists(url):
+            return url
+
+        schema = self._conn.schema
+        url = f"{url_prefix}/databases/{db}/schemas/{schema}/{plural_object_type}/"
+        if self._url_exists(url):
```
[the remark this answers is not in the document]
sfc-gh-pczajka: Done

tests_integration/test_object.py
sfc-gh-pczajka: https://snowflakecomputing.atlassian.net/browse/SNOW-1482578

tests_integration/test_object.py
```diff
@@ -132,3 +135,55 @@ def test_show_drop_image_repository(runner, test_database, snowflake_session):
...
+            "schema",
+            {
+                "name": "test_schema",
+                "comment": "test_schema = schema for testing stuff",
+            },
+        ),
+        ("image-repository", {"name": "test_image_repo"}),
+    ],
+)
+@pytest.mark.integration
```
sfc-gh-turbaszek, line 154: Add integration tests for expected errors so we have a check that nothing changes on server side

sfc-gh-turbaszek (approved):

PR #1060, opened 2024-05-11
Fix setup script generation errors

sfc-gh-bdufour:
### Pre-review checklist
   * [N/A] I've added or updated automated unit tests to verify correctness of my new code.
   * [N/A] I've added or updated integration tests to verify correctness of my new code.

### Changes description

Fixed misc errors in the script generation logic to enable manual end-to-end testing via the `snow app bundle` command.


tests/nativeapp/codegen/snowpark/test_callback_source.py
```diff
@@ -1,48 +0,0 @@
...
-
-with open(
-    "./tests/nativeapp/codegen/snowpark/expected_callback_source.py",
-    mode="r",
-    encoding="utf-8",
-) as udf_code:
-    rendered_src = udf_code.read()
-
-
-def test_rendering(temp_dir):
```
sfc-gh-bdufour, line 24: this test couldn't be salvaged. It is very brittle and really doesn't test anything interesting. I decided to just remove it instead. We should test the template through the normal entry point instead.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function_utils.py
```diff
@@ -131,5 +131,5 @@ def _enrich_entity(
         handler=entity["handler"],
     )
     entity["all_imports"] = _get_all_imports(
-        raw_imports=entity["raw_imports"], suffix_str=suffix_str
+        raw_imports=entity["raw_imports"] or [], suffix_str=suffix_str
```
sfc-gh-bgoel, line 134: Acking preventing errors such as "NoneType is not iterable". Does that mean that in my sanitize/add_defaults function, it would be better to have the default values be an empty value (list, dict, str) instead of None ?
That way, you won't need to keep track of it when calling functions.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -202,7 +216,7 @@ def process(
 
                 cc.message(f"This is the file path in deploy root: {dest_file}\n")
                 cc.message("This is the list of collected entities:")
-                cc.message(collected_entities)
+                cc.message(pprint.pformat(collected_entities))
```
sfc-gh-bgoel, line 219: + `cc.message` comment above.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function_utils.py
sfc-gh-bdufour: I prefer to deal with a model that's been sanitize, personally. This was just a quick way to achieve this for processors in particular.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -89,6 +100,8 @@ def _execute_in_sandbox(
             timeout=DEFAULT_TIMEOUT,
             **kwargs,
         )
+        cc.message(f"stdout: {completed_process.stdout}")
+        cc.message(f"stderr: {completed_process.stderr}")
```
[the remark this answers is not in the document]
sfc-gh-bdufour: No, we wouldn't. This is just for now as we're actively developing. It was useful to me to see what the sandbox was outputting.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bdufour: Ditto, very temporary.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function_utils.py
sfc-gh-bgoel: Okay, I'll edit the defaults in my PR to not always be None.

PR #1065, opened 2024-05-14
Add variables support in sql command

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
...

### DIFF
--- a/src/snowflake/cli/api/commands/flags.py
+++ b/src/snowflake/cli/api/commands/flags.py
@@ -489,7 +489,7 @@ def project_definition_option(project_name: str):
 
     def _callback(project_path: Optional[str]):
         dm = DefinitionManager(project_path)
-        project_definition = getattr(dm.project_definition, project_name, None)
+        project_definition = dm.project_definition
         project_root = dm.project_root
 
         if not project_definition:
@@ -501,6 +501,7 @@ def project_definition_option(project_name: str):
         cli_context_manager.set_project_root(project_root)
         return project_definition
 
+    # TODO: inline this with decorator
     if project_name == "native_app":
         project_name_help = "Snowflake Native App"
     elif project_name == "streamlit":
--- a/src/snowflake/cli/api/commands/flags.py
+++ b/src/snowflake/cli/api/commands/flags.py
@@ -12,6 +12,7 @@ import typer
 from click import ClickException
 from snowflake.cli.api.cli_global_context import cli_context_manager
 from snowflake.cli.api.console import cli_console
+from snowflake.cli.api.exceptions import MissingConfiguration
 from snowflake.cli.api.output.formats import OutputFormat
 
 DEFAULT_CONTEXT_SETTINGS = {"help_option_names": ["--help", "-h"]}
@@ -483,13 +484,13 @@ def execution_identifier_argument(sf_object: str, example: str) -> typer.Argumen
     )
 
 
-def project_definition_option(project_name: str):
+def project_type_option(project_name: str):
     from snowflake.cli.api.exceptions import NoProjectDefinitionError
     from snowflake.cli.api.project.definition_manager import DefinitionManager
 
     def _callback(project_path: Optional[str]):
         dm = DefinitionManager(project_path)
-        project_definition = dm.project_definition
+        project_definition = getattr(dm.project_definition, project_name, None)
         project_root = dm.project_root
 
         if not project_definition:
@@ -501,7 +502,6 @@ def project_definition_option(project_name: str):
         cli_context_manager.set_project_root(project_root)
         return project_definition
 
-    # TODO: inline this with decorator
     if project_name == "native_app":
         project_name_help = "Snowflake Native App"
     elif project_name == "streamlit":
@@ -515,6 +515,34 @@ def project_definition_option(project_name: str):
         "--project",
         help=f"Path where the {project_name_help} project resides. "
         f"Defaults to current working directory.",
+        callback=_callback,
+        show_default=False,
+    )
+
+
+def project_definition_option(optional: bool = False):
+    from snowflake.cli.api.project.definition_manager import DefinitionManager
+
+    def _callback(project_path: Optional[str]):
+        try:
+            dm = DefinitionManager(project_path)
+            project_definition = dm.project_definition
+            project_root = dm.project_root
+        except MissingConfiguration:
+            if optional:
+                project_definition = None
+                project_root = None
+            else:
+                raise
+        cli_context_manager.set_project_definition(project_definition)
+        cli_context_manager.set_project_root(project_root)
+        return project_definition
+
+    return typer.Option(
+        None,
+        "-p",
+        "--project",
+        help=f"Path where Snowflake project resides. Defaults to current working directory.",
         callback=_callback,
         show_default=False,
     )
--- a/src/snowflake/cli/api/project/schemas/project_definition.py
+++ b/src/snowflake/cli/api/project/schemas/project_definition.py
@@ -1,8 +1,7 @@
 from __future__ import annotations
 
-from typing import Dict, Optional
+from typing import Dict, Optional, Union
 
-from click import ClickException
 from packaging.version import Version
 from pydantic import Field, field_validator
 from snowflake.cli.api.project.schemas.native_app.native_app import NativeApp
@@ -17,9 +16,8 @@ _latest_version = "1.1"
 
 
 class ProjectDefinition(UpdatableModel):
-    definition_version: str = Field(
+    definition_version: Union[str, int] = Field(
         title="Version of the project definition schema, which is currently 1",
-        default=_latest_version,
     )
     native_app: Optional[NativeApp] = Field(
         title="Native app definitions for the project", default=None
@@ -45,8 +43,9 @@ class ProjectDefinition(UpdatableModel):
     @field_validator("definition_version")
     @classmethod
     def _is_supported_version(cls, version: str) -> str:
+        version = str(version)
         if version not in _supported_version:
-            raise ClickException(
+            raise ValueError(
                 f'Version {version} is not supported. Supported versions: {", ".join(_supported_version)}'
             )
         return version
--- a/src/snowflake/cli/api/project/schemas/project_definition.py
+++ b/src/snowflake/cli/api/project/schemas/project_definition.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
-from typing import Optional
+import os
+from typing import Optional, List, Union
 
 from pydantic import Field
 from snowflake.cli.api.project.schemas.native_app.native_app import NativeApp
@@ -9,11 +10,15 @@ from snowflake.cli.api.project.schemas.streamlit.streamlit import Streamlit
 from snowflake.cli.api.project.schemas.updatable_model import UpdatableModel
 
 
+# todo: update examples
+_supported_version = ("1", "1.1")
+_latest_version = "1.1"
+
+
 class ProjectDefinition(UpdatableModel):
-    definition_version: int = Field(
+    definition_version: str = Field(
         title="Version of the project definition schema, which is currently 1",
-        ge=1,
-        le=1,
+        default=_latest_version
     )
     native_app: Optional[NativeApp] = Field(
         title="Native app definitions for the project", default=None
@@ -25,3 +30,13 @@ class ProjectDefinition(UpdatableModel):
     streamlit: Optional[Streamlit] = Field(
         title="Streamlit definitions for the project", default=None
     )
+    env: Optional[List[Variable]] = Field(title="Environment specification for this project.", default=None)
+
+
+VariableType = Union[str, bool, int, float]
+
+
+class Variable(UpdatableModel):
+    name: str = Field(title="Name of variable.")
+    value: VariableType = Field(title="Value of variable.")
+
--- a/src/snowflake/cli/api/project/schemas/project_definition.py
+++ b/src/snowflake/cli/api/project/schemas/project_definition.py
@@ -8,11 +8,10 @@ from snowflake.cli.api.project.schemas.native_app.native_app import NativeApp
 from snowflake.cli.api.project.schemas.snowpark.snowpark import Snowpark
 from snowflake.cli.api.project.schemas.streamlit.streamlit import Streamlit
 from snowflake.cli.api.project.schemas.updatable_model import UpdatableModel
-from snowflake.cli.api.utils.models import DictWithEnvironFallback
+from snowflake.cli.api.utils.models import EnvironWithDefinedDictFallback
 
 # todo: update examples
 _supported_version = ("1", "1.1")
-_latest_version = "1.1"
 
 
 class ProjectDefinition(UpdatableModel):
@@ -37,8 +36,8 @@ class ProjectDefinition(UpdatableModel):
 
     @field_validator("env")
     @classmethod
-    def _convert_env(cls, env: Optional[Dict]) -> DictWithEnvironFallback:
-        variables = DictWithEnvironFallback(env if env else {})
+    def _convert_env(cls, env: Optional[Dict]) -> EnvironWithDefinedDictFallback:
+        variables = EnvironWithDefinedDictFallback(env if env else {})
         return variables
 
     @field_validator("definition_version")
--- a/src/snowflake/cli/api/project/schemas/project_definition.py
+++ b/src/snowflake/cli/api/project/schemas/project_definition.py
@@ -1,14 +1,14 @@
 from __future__ import annotations
 
-from typing import Optional, List, Union
+from typing import List, Optional, Union
 
 from click import ClickException
+from packaging.version import Version
 from pydantic import Field, field_validator
 from snowflake.cli.api.project.schemas.native_app.native_app import NativeApp
 from snowflake.cli.api.project.schemas.snowpark.snowpark import Snowpark
 from snowflake.cli.api.project.schemas.streamlit.streamlit import Streamlit
 from snowflake.cli.api.project.schemas.updatable_model import UpdatableModel
-
 
 # todo: update examples
 _supported_version = ("1", "1.1")
@@ -18,7 +18,7 @@ _latest_version = "1.1"
 class ProjectDefinition(UpdatableModel):
     definition_version: str = Field(
         title="Version of the project definition schema, which is currently 1",
-        default=_latest_version
+        default=_latest_version,
     )
     native_app: Optional[NativeApp] = Field(
         title="Native app definitions for the project", default=None
@@ -30,9 +30,11 @@ class ProjectDefinition(UpdatableModel):
     streamlit: Optional[Streamlit] = Field(
         title="Streamlit definitions for the project", default=None
     )
-    env: Optional[List[Variable]] = Field(title="Environment specification for this project.", default=None)
+    env: Optional[List[Variable]] = Field(
+        title="Environment specification for this project.", default=None
+    )
 
-    @field_validator('definition_version')
+    @field_validator("definition_version")
     @classmethod
     def _is_supported_version(cls, version: str) -> str:
         if version not in _supported_version:
@@ -40,6 +42,9 @@ class ProjectDefinition(UpdatableModel):
                 f'Version {version} is not supported. Supported versions: {", ".join(_supported_version)}'
             )
         return version
+
+    def meets_version_requirement(self, required_version: str) -> bool:
+        return Version(self.definition_version) >= Version(required_version)
 
 
 VariableType = Union[str, bool, int, float]
--- a/src/snowflake/cli/api/project/schemas/project_definition.py
+++ b/src/snowflake/cli/api/project/schemas/project_definition.py
@@ -39,9 +39,6 @@ class ProjectDefinition(UpdatableModel):
     @classmethod
     def _convert_env(cls, env: Optional[Dict]) -> DictWithEnvironFallback:
         variables = DictWithEnvironFallback(env if env else {})
-        for key in variables:
-            # Accessing value first checks for env var, so in this way we update the in-memory state
-            variables[key] = variables[key]
         return variables
 
     @field_validator("definition_version")
--- a/src/snowflake/cli/api/project/schemas/project_definition.py
+++ b/src/snowflake/cli/api/project/schemas/project_definition.py
@@ -9,6 +9,7 @@ from snowflake.cli.api.project.schemas.native_app.native_app import NativeApp
 from snowflake.cli.api.project.schemas.snowpark.snowpark import Snowpark
 from snowflake.cli.api.project.schemas.streamlit.streamlit import Streamlit
 from snowflake.cli.api.project.schemas.updatable_model import UpdatableModel
+from snowflake.cli.api.utils.models import DictWithEnvironFallback
 
 # todo: update examples
 _supported_version = ("1", "1.1")
@@ -31,8 +32,17 @@ class ProjectDefinition(UpdatableModel):
         title="Streamlit definitions for the project", default=None
     )
     env: Optional[List[Variable]] = Field(
-        title="Environment specification for this project.", default=None
+        title="Environment specification for this project.",
+        default=None,
+        validation_alias="env",
     )
+
+    @field_validator("env")
+    @classmethod
+    def _convert_env(cls, env: Optional[List[Variable]]) -> DictWithEnvironFallback:
+        if env is None:
+            return DictWithEnvironFallback({})
+        return DictWithEnvironFallback({v.name: v.value for v in env})
 
     @field_validator("definition_version")
     @classmethod
@@ -53,3 +63,11 @@ VariableType = Union[str, bool, int, float]
 class Variable(UpdatableModel):
     name: str = Field(title="Name of variable.")
     value: VariableType = Field(title="Value of variable.")
+
+
+class ProjectDefinitionGetter:
+    def __init__(self, project: ProjectDefinition):
+        self.__project = project
+
+    def __getitem__(self, item):
+        return s
--- a/src/snowflake/cli/api/project/schemas/project_definition.py
+++ b/src/snowflake/cli/api/project/schemas/project_definition.py
@@ -38,7 +38,11 @@ class ProjectDefinition(UpdatableModel):
     @field_validator("env")
     @classmethod
     def _convert_env(cls, env: Optional[Dict]) -> DictWithEnvironFallback:
-        return DictWithEnvironFallback(env if env else {})
+        variables = DictWithEnvironFallback(env if env else {})
+        for key in variables:
+            # Accessing value first checks for env var, so in this way we update the in-memory state
+            variables[key] = variables[key]
+        return variables
 
     @field_validator("definition_version")
     @classmethod
--- a/src/snowflake/cli/api/project/schemas/project_definition.py
+++ b/src/snowflake/cli/api/project/schemas/project_definition.py
@@ -1,9 +1,9 @@
 from __future__ import annotations
 
-import os
 from typing import Optional, List, Union
 
-from pydantic import Field
+from click import ClickException
+from pydantic import Field, field_validator
 from snowflake.cli.api.project.schemas.native_app.native_app import NativeApp
 from snowflake.cli.api.project.schemas.snowpark.snowpark import Snowpark
 from snowflake.cli.api.project.schemas.streamlit.streamlit import Streamlit
@@ -32,6 +32,15 @@ class ProjectDefinition(UpdatableModel):
     )
     env: Optional[List[Variable]] = Field(title="Environment specification for this project.", default=None)
 
+    @field_validator('definition_version')
+    @classmethod
+    def _is_supported_version(cls, version: str) -> str:
+        if version not in _supported_version:
+            raise ClickException(
+                f'Version {version} is not supported. Supported versions: {", ".join(_supported_version)}'
+            )
+        return version
+
 
 VariableType = Union[str, bool, int, float]
 
@@ -39,4 +48,3 @@ VariableType = Union[str, bool, int, float]
 class Variable(UpdatableModel):
     name: str = Field(title="Name of variable.")
     value: VariableType = Field(title="Value of variable.")
-
--- a/src/snowflake/cli/api/project/schemas/project_definition.py
+++ b/src/snowflake/cli/api/project/schemas/project_definition.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import List, Optional, Union
+from typing import Dict, Optional
 
 from click import ClickException
 from packaging.version import Version
@@ -31,7 +31,7 @@ class ProjectDefinition(UpdatableModel):
     streamlit: Optional[Streamlit] = Field(
         title="Streamlit definitions for the project", default=None
     )
-    env: Optional[List[Variable]] = Field(
+    env: Optional[Dict] = Field(
         title="Environment specification for this project.",
         default=None,
         validation_alias="env",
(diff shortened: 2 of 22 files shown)


src/snowflake/cli/api/commands/flags.py
```diff
@@ -485,7 +485,7 @@ def project_definition_option(project_name: str):
 
     def _callback(project_path: Optional[str]):
         dm = DefinitionManager(project_path)
-        project_definition = getattr(dm.project_definition, project_name, None)
+        project_definition = dm.project_definition
```
sfc-gh-turbaszek, line 488: This change alters what is provided by `@with_project_definition` decorator. Instead of part of project definition (snowpark, streamlit, native_app) it provides full definition.

The only purpose of keeping `@with_project_definition(NAME)` is to add `-p/--project` with "nice" description that relates to project type.

However, we already have some name mapping:
https://github.com/snowflakedb/snowflake-cli/blob/532d1f10a8be36105ec689ce597d20ac6bbd8e50/src/snowflake/cli/api/commands/flags.py#L500-L505

So my suggestion would be to get rid of this "nice description" thing in favor of simplicity. In this way we can move the functionality of this decorator to `app.command(uses_project=True)`.

@sfc-gh-cgorrie @sfc-gh-pjob wdyt?

src/snowflake/cli/api/project/schemas/project_definition.py
```diff
@@ -1,19 +1,24 @@
...
 
 
+# todo: update examples
+_supported_version = ("1", "1.1")
+_latest_version = "1.1"
+
+
 class ProjectDefinition(UpdatableModel):
-    definition_version: int = Field(
+    definition_version: str = Field(
```
sfc-gh-turbaszek, line 19: Had to change it from integer to string to enable minimal semantic versioning.

src/snowflake/cli/api/utils/rendering.py
```diff
@@ -88,22 +94,151 @@ def jinja_render_from_file(
...
+            f"{context_key} in user defined data. The `{context_key}` variable is reserved for CLI usage."
+        )
+
+    project_definition = cli_context.project_definition
+    context_data = {context_key: {"env": _EnvGetter(os.environ)}}
+    # If there's project definition file then resolve variables from it
+    if project_definition and project_definition.meets_version_requirement("1.1"):
+        context_data = _resolve_variables_in_project(
+            project_definition=project_definition, context_key=context_key
+        )
```
sfc-gh-turbaszek, line 160: This feature will work only for snowflake.yml with 1.1 or later version

src/snowflake/cli/api/commands/flags.py
sfc-gh-melnacouzi: should uses_project=True be the default?

src/snowflake/cli/api/utils/rendering.py
```diff
@@ -88,22 +94,151 @@ def jinja_render_from_file(
         print(rendered_result)
 
 
-class _AttrGetter:
+def _is_key_in_environ(key: str):
+    return key.upper() in os.environ or key in os.environ
+
+
+def _get_value_from_environ(key: str):
+    if key.upper() in os.environ:
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: I can see we return key as a backup. I think it's safer to check key first, and then upper as a backup.

src/snowflake/cli/api/utils/rendering.py
```diff
@@ -88,22 +94,151 @@ def jinja_render_from_file(
...
+        return os.environ[key.upper()]
+    return os.environ[key]
```
sfc-gh-melnacouzi, line 104: I think we should do key first, and if not found, try upper.

sfc-gh-cgorrie: I'm not sure we should be trying to retrofit case-insensitivity onto a case-sensitive platform.

See https://docs.python.org/3/library/os.html#os.environ -- it already takes care of this on relevant platforms. So we can use `key` as-is.

src/snowflake/cli/api/utils/rendering.py
```diff
@@ -88,22 +94,151 @@ def jinja_render_from_file(
         print(rendered_result)
 
 
-class _AttrGetter:
+def _is_key_in_environ(key: str):
+    return key.upper() in os.environ or key in os.environ
```
sfc-gh-cgorrie, line 98:
```suggestion
    return key in os.environ
```
See below for rationale.

src/snowflake/cli/api/utils/rendering.py
sfc-gh-melnacouzi: even better.

src/snowflake/cli/api/utils/rendering.py
```diff
@@ -88,22 +94,151 @@ def jinja_render_from_file(
...
+
+class _ContextGetter:
+    """
+    Getter for project definition data. It replaces `env` attribute with _EnvGetter
+    that enables fallback to environment variables.
+    """
+
+    def __init__(self, project_definition: ProjectDefinition):
+        self._project = project_definition
+        self.variables = {v.name: v.value for v in project_definition.env}
```
sfc-gh-melnacouzi, line 134: Should we not use variables, and bundle everything under env like we discussed earlier? Could the _EnvGetter handle this instead?

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -172,7 +172,7 @@ def app_run(
 
     processor = NativeAppRunProcessor(
         project_definition=cli_context.project_definition,
-        project_root=cli_context.project_root,
+        project_root=cli_context.project_root.native_app,
```
sfc-gh-melnacouzi, line 175: did you mean to update project_definition instead?

src/snowflake/cli/api/commands/flags.py
sfc-gh-turbaszek: I don't think so, [explicite is better than implicite](https://peps.python.org/pep-0020/#the-zen-of-python). Not all command require access to project file. Imho the decorators should inform developer what is used by the command  instead of expecting them to know default values and behaviors. WDYT?

src/snowflake/cli/api/utils/rendering.py
sfc-gh-turbaszek: Should we consider using best practices for naming environment variables? For example I'm referring to
https://google.github.io/styleguide/shellguide.html#constants-and-environment-variable-names

Honestly, lower-case env vars looks strange and uncommon. I'm fine with doing a fallback to key but let's not force users to use CAPS names in snowflake.yml only to be able to write `export CAPS` instead of `export caps`. WDYT?

src/snowflake/cli/api/utils/rendering.py
```diff
@@ -88,22 +94,151 @@ def jinja_render_from_file(
...
-        raise ValueError(f"{context_key} in user defined data")
-    context_data = {context_key: {"env": _AttrGetter(os.environ)}}
-    return {**data, **context_data}
+    if context_key in external_data:
+        raise ClickException(
+            f"{context_key} in user defined data. The `{context_key}` variable is reserved for CLI usage."
+        )
+
+    project_definition = cli_context.project_definition
+    context_data = {context_key: {"env": _EnvGetter(os.environ)}}
```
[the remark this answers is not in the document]
sfc-gh-turbaszek: Why introduce this inconsistency?

project definition present = ctx.env.foo
no pdf = env.foo

src/snowflake/cli/plugins/nativeapp/commands.py
sfc-gh-turbaszek: Yes, thanks for catching this!

src/snowflake/cli/api/utils/rendering.py
sfc-gh-cgorrie: I'm sure CLI supports the latter on Windows (if not, I'm not sure how that ended up being the case!).

I guess the point is, why should we be opinionated about env var case sensitivity? It's already "solved" at the OS level. It's worth trying out comparables and seeing how they react to case sensitivity, as that will likely be the users' expectation. If other tooling makes things case insensitive, I'm OK with the approach here, but that isn't the case with what I'm seeing, e.g.

- Github actions uses case sensitive env vars: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable

sfc-gh-turbaszek: > I  would guess that CLI supports the latter on Windows (if not, I'm curious how that ended up being the case!).

It's not CLI, it's the platform that supports it due to case-insensitivity. For case-sensitive platforms (which are the only concern) we expect variables to be upper case. 

Ok, let's use the keys. If someone want to follow best practices they still are free to do so using upper-case names. We just have to document this properly.

src/snowflake/cli/api/utils/rendering.py
```diff
@@ -88,24 +96,122 @@ def jinja_render_from_file(
...
+        if var.startswith(ctx_env_prefix):
+            required_variables.append(var[len(ctx_env_prefix) :])
+    return len(found_variables), required_variables
 
 
 def snowflake_cli_jinja_render(content: str, data: Dict | None = None) -> str:
-    data = _add_project_context(data or dict())
+    data = _add_project_context(
+        external_data=data or dict(), project_definition=cli_context.project_definition
+    )
```
sfc-gh-turbaszek, line 216: I think this should be part of `ProjectDefinition` validation. When we read the yaml file we should also perform the templates expansions so we don't need to worry about it in too many place. WDYT @sfc-gh-melnacouzi ?

tests/api/utils/test_rendering.py
```diff
@@ -63,3 +70,68 @@ def test_that_undefined_variables_raise_error():
...
+    result = _add_project_context({}, project_definition=pdf)
+    assert result == {
+        "ctx": ProjectDefinition(
+            definition_version="1.1",
+            native_app=None,
+            snowpark=None,
+            streamlit=None,
+            env={"A": 42, "B": "b=42", "C": "b=42 and 42"},
+        )
+    }
```
sfc-gh-turbaszek, line 112: @sfc-gh-cgorrie @sfc-gh-melnacouzi if we can agree to use `key:value` format for `env` schema then this looks pretty promising. Also

tests/project/test_project_schemas.py
```diff
@@ -63,3 +64,15 @@ def test_project_schema_is_updated_correctly_from_dict(
     native_app_project_instance.update_from_dict(update_dict)
     assert native_app_project_instance.native_app.name == "napp_test"
     assert native_app_project_instance.native_app.package.distribution == "external"
+
+
+def test_project_definition_work_for_int_version():
+    p = ProjectDefinition(definition_version=1)
+    assert p.definition_version == "1"
```
sfc-gh-turbaszek, line 71: To assure backward compatibility

src/snowflake/cli/api/utils/rendering.py
```diff
@@ -88,24 +96,123 @@ def jinja_render_from_file(
...
+        # Templates are reserved only to string variables
+        if not isinstance(value, str):
+            continue
+
+        # This value requires
+        include_template, required_variables = _search_for_required_variables(value)
+
+        if key in required_variables:
+            raise ClickException(
+                f"Cycle detected in environment definition for `{key}` variable."
```
sfc-gh-turbaszek, line 179: We should check for `A -> B, B -> A` cycle

src/snowflake/cli/api/utils/rendering.py
```diff
@@ -88,24 +96,123 @@ def jinja_render_from_file(
...
+            unresolved_keys.append(key)
+
+    return context_data
+
 
-    def __getattr__(self, item):
-        if item not in self._data_dict:
-            raise AttributeError(f"No attribute {item}")
-        return self._data_dict[item]
+def _check_variables_consistency(variables_data: DictWithEnvironFallback):
```
sfc-gh-melnacouzi, line 158: nit: rename to _get_variables_with_depedencies? The rest is just validation, so doesn't have to affect the name.

src/snowflake/cli/api/utils/rendering.py
```diff
@@ -88,24 +96,123 @@ def jinja_render_from_file(
...
+    variables_with_dependencies = _check_variables_consistency(variables_data)
+    # Sort keys so we start from the shortest having lower probability of including more than one variable
+    unresolved_keys = sorted(list(variables_with_dependencies), reverse=True)
+
+    env = get_snowflake_cli_jinja_env()
+    context_data = {_CONTEXT_KEY: project_definition}
+    while unresolved_keys:
+        key = unresolved_keys.pop()
+        value = variables_data[key]
+        if not isinstance(value, str):
```
sfc-gh-melnacouzi, line 144: didn't we check this earlier?

src/snowflake/cli/api/utils/rendering.py
```diff
@@ -88,24 +96,123 @@ def jinja_render_from_file(
...
+        DictWithEnvironFallback, project_definition.env
+    )
+
+    variables_with_dependencies = _check_variables_consistency(variables_data)
+    # Sort keys so we start from the shortest having lower probability of including more than one variable
+    unresolved_keys = sorted(list(variables_with_dependencies), reverse=True)
+
+    env = get_snowflake_cli_jinja_env()
+    context_data = {_CONTEXT_KEY: project_definition}
+    while unresolved_keys:
```
sfc-gh-melnacouzi, line 141: how do we guarantee no infinite loop?

src/snowflake/cli/api/utils/rendering.py
```diff
@@ -88,24 +96,123 @@ def jinja_render_from_file(
...
+        key = unresolved_keys.pop()
+        value = variables_data[key]
+        if not isinstance(value, str):
+            continue
+
+        try:  # try to evaluate the template given current state of know variables
+            new_value = env.from_string(value).render(context_data)
```
sfc-gh-melnacouzi, line 148: where do environment variables play a role here? I don't see them in context_data

src/snowflake/cli/api/utils/rendering.py
```diff
@@ -88,24 +96,123 @@ def jinja_render_from_file(
...
+            if string_includes_template(new_value):
+                unresolved_keys.append(key)
+            variables_data[key] = env.from_string(value).render(context_data)
```
sfc-gh-melnacouzi, line 151: why not set:
```
variables_data[key] = new_value
```
?

src/snowflake/cli/api/commands/flags.py
```diff
@@ -515,6 +516,34 @@ def _callback(project_path: Optional[str]):
...
+                raise
+        cli_context_manager.set_project_definition(project_definition)
+        cli_context_manager.set_project_root(project_root)
+        return project_definition
+
+    return typer.Option(
+        None,
+        "-p",
+        "--project",
+        help=f"Path where Snowflake project resides. Defaults to current working directory.",
```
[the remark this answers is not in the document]
sfc-gh-turbaszek: I agree, but to limit the scope of the change and was planning to do this clean up as a followup. Will that work for you?

src/snowflake/cli/api/project/schemas/project_definition.py
```diff
@@ -25,3 +29,26 @@ class ProjectDefinition(UpdatableModel):
...
+    env: Optional[Dict] = Field(
+        title="Environment specification for this project.",
+        default=None,
+        validation_alias="env",
+    )
+
+    @field_validator("env")
+    @classmethod
+    def _convert_env(cls, env: Optional[Dict]) -> DictWithEnvironFallback:
+        return DictWithEnvironFallback(env if env else {})
```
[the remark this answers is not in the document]
sfc-gh-turbaszek: Yes. as suggested in other comment it would be good to move the expansion/rendering step from sql to project definition. So any use of `cli_context.project_definition` would be accessing expanded env/definition. But I decided to clean this up as a follow-up

src/snowflake/cli/api/project/util.py
```diff
@@ -180,3 +180,7 @@ def identifier_to_show_like_pattern(identifier: str) -> str:
     matching this identifier
     """
     return f"'{escape_like_pattern(unquote_identifier(identifier))}'"
+
+
+def supported_since(version: str):
```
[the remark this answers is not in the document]
sfc-gh-turbaszek: Good catch, will remove!

src/snowflake/cli/api/utils/rendering.py
```diff
@@ -88,24 +96,123 @@ def jinja_render_from_file(
...
+    """
+    if _CONTEXT_KEY in external_data:
+        raise ClickException(
+            f"{_CONTEXT_KEY} in user defined data. The `{_CONTEXT_KEY}` variable is reserved for CLI usage."
+        )
+
+    context_data = {_CONTEXT_KEY: {"env": DictWithEnvironFallback({})}}
+    # If there's project definition file then resolve variables from it
+    if project_definition and project_definition.meets_version_requirement("1.1"):
+        context_data = _resolve_variables_in_project(project_definition)
```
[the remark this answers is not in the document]
sfc-gh-turbaszek: Good idea!

src/snowflake/cli/api/utils/rendering.py
sfc-gh-turbaszek: Accessing `ctx.env` checks environment because `env` is `DictWithEnvironFallback`. So accessing any attribute will first look for environment variable.

src/snowflake/cli/api/utils/rendering.py
sfc-gh-turbaszek: True, removed and added typing

src/snowflake/cli/api/utils/rendering.py
sfc-gh-melnacouzi: I agree. I will work on this as part of the project definition global expansion, since I have to do this for variables in all of the pdf.

sfc-gh-melnacouzi: for the env part though, I think we should restrict the code from using it, or if we let code use it, we should have the DictWithEnvironFallback instance so code doesn't access what's written in env section directly.

src/snowflake/cli/api/commands/flags.py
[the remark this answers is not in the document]
sfc-gh-melnacouzi: sounds good.

src/snowflake/cli/api/project/schemas/project_definition.py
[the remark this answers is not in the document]
sfc-gh-melnacouzi: makes sense. thx

src/snowflake/cli/api/utils/rendering.py
sfc-gh-turbaszek: Renamed

src/snowflake/cli/api/utils/rendering.py
sfc-gh-turbaszek: Changed 👌

src/snowflake/cli/api/project/schemas/project_definition.py
```diff
@@ -25,3 +29,30 @@ class ProjectDefinition(UpdatableModel):
...
+        default=None,
+        validation_alias="env",
+    )
+
+    @field_validator("env")
+    @classmethod
+    def _convert_env(cls, env: Optional[Dict]) -> DictWithEnvironFallback:
+        variables = DictWithEnvironFallback(env if env else {})
+        for key in variables:
+            # Accessing value first checks for env var, so in this way we update the in-memory state
```
sfc-gh-melnacouzi, line 43: why is that necessary?

src/snowflake/cli/api/utils/rendering.py
sfc-gh-turbaszek: `unresolved_keys` are returned from `_get_variables_with_dependencies` that makes sure
1. None of the variables depends on unknown variable (ctx.env.something) (possible infinite loop because of unknown error)
2. There's no cycles between between variables

So `unresolved_keys` is a list of variables defined in `env` that have some cross-dependencies between `env` variables. And thanks to above resolving all of them is possible.

src/snowflake/cli/api/utils/rendering.py
```diff
@@ -88,24 +96,122 @@ def jinja_render_from_file(
...
+    Look for &{...} pattern in  variable value. Returns a list of env variables required
+    to expand this template.`
+    """
+    ctx_env_prefix = f"{_CONTEXT_KEY}.env."
+    found_variables = re.findall(r"(&{([\.\w ]+)})+", variable_value)
+    required_variables = []
+    for _, variable in found_variables:
+        var: str = variable.strip()
+        if var.startswith(ctx_env_prefix):
+            required_variables.append(var[len(ctx_env_prefix) :])
```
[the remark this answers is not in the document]
sfc-gh-turbaszek: I can error. This will cause that `env` variables cannot reference other parts of definition file. But this can be fine for now. Do we agree to raise error and remove support for cross-project references in `env` for now?

src/snowflake/cli/api/utils/rendering.py
```diff
@@ -88,24 +97,138 @@ def jinja_render_from_file(
...
+    for key in list(nodes):
+        q = deque([key])
+        visited: List[str] = []
+        while q:
+            curr = q.popleft()
+            if visited and curr == key:
+                raise ClickException(
+                    "Cycle detected between variables: {}".format(" -> ".join(visited))
+                )
+            visited.append(curr)
```
sfc-gh-melnacouzi, line 180: what's the point of visited, since we never check the content?

src/snowflake/cli/api/utils/rendering.py
```diff
@@ -88,24 +97,138 @@ def jinja_render_from_file(
...
+
+
+def _check_for_cycles(nodes: defaultdict):
+    nodes = nodes.copy()
+    for key in list(nodes):
+        q = deque([key])
+        visited: List[str] = []
+        while q:
+            curr = q.popleft()
+            if visited and curr == key:
```
sfc-gh-melnacouzi, line 176: you probably need to check if curr in visited instead. What if you have a -> b -> c -> b -> c -> b ... this might cause infinite loop (I think)

src/snowflake/cli/api/utils/rendering.py
sfc-gh-turbaszek: Persisting cycle so we can construct message, I considered this useful

src/snowflake/cli/api/project/schemas/project_definition.py
sfc-gh-turbaszek: A leftover, no longer neccessary

src/snowflake/cli/api/utils/rendering.py
sfc-gh-turbaszek: good point, fixed

sfc-gh-melnacouzi (dismissed):

src/snowflake/cli/api/utils/rendering.py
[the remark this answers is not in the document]
sfc-gh-melnacouzi: yes, I agree to raise error for now. I will remove the error / tighten the check as part of my changes later.

sfc-gh-melnacouzi (dismissed):

src/snowflake/cli/api/utils/rendering.py
```diff
@@ -88,24 +97,152 @@ def jinja_render_from_file(
...
+
+
+def _remove_ctx_env_prefix(text: str) -> str:
+    prefix = "ctx.env."
+    if text.startswith(prefix):
+        return text[len(prefix) :]
+    return text
+
+
+def string_includes_template(text: str) -> bool:
```
sfc-gh-melnacouzi, line 124: There is an issue loading yaml using the &{} template, so we will hit the chicken and egg issue here. We need to parse the yaml before we expand the templates.
Should we change the syntax - choose something parsable by yaml for project file and eventually deprecate the sql-only syntax?
Cam suggested ${}. Let me know what you think

sfc-gh-bdufour: @sfc-gh-turbaszek This seems to be a critical problem with the chosen syntax. Not argument against remaining compatible with the legacy tool in SQL text, but we need to ensure that we don't end up breaking yaml parsing/generating tools because it's the _only_ syntax we support. We already know that customers like to generate a number of their project files, so that seems like an important requirement. Thoughts?

src/snowflake/cli/api/project/schemas/project_definition.py
```diff
@@ -1,19 +1,23 @@
...
+from pydantic import Field, field_validator
 from snowflake.cli.api.project.schemas.native_app.native_app import NativeApp
 from snowflake.cli.api.project.schemas.snowpark.snowpark import Snowpark
 from snowflake.cli.api.project.schemas.streamlit.streamlit import Streamlit
 from snowflake.cli.api.project.schemas.updatable_model import UpdatableModel
+from snowflake.cli.api.utils.models import DictWithEnvironFallback
+
+# todo: update examples
+_supported_version = ("1", "1.1")
+_latest_version = "1.1"
```
sfc-gh-pjob, line 15: I don't see usage of this variable.

src/snowflake/cli/api/utils/models.py
```diff
@@ -0,0 +1,14 @@
+import os
+
+
+class DictWithEnvironFallback(dict):
```
sfc-gh-pjob, line 4:
```suggestion
class EnvironWithDefinedDictFallback(dict):
```

src/snowflake/cli/api/utils/rendering.py
```diff
@@ -88,24 +97,152 @@ def jinja_render_from_file(
...
+    Updates the external data with variables from snowflake.yml definition file.
+    """
+    if _CONTEXT_KEY in external_data:
+        raise ClickException(
+            f"{_CONTEXT_KEY} in user defined data. The `{_CONTEXT_KEY}` variable is reserved for CLI usage."
+        )
+
+    context_data = _resolve_variables_in_project(project_definition)
+
+    context_data.update(external_data)
```
sfc-gh-pjob, line 113: Is this correct direction of updating? Should it really be external -> context, not context -> external? IMO method comment suggests the second direction.

src/snowflake/cli/api/commands/flags.py
sfc-gh-pjob: IMO it looks good :)

src/snowflake/cli/api/utils/rendering.py
sfc-gh-pjob: I think I agree with @sfc-gh-turbaszek that we can support using variables only as part of strings specified in quotes and cast strings properly in models if needed.

But we have to confirm that it works correctly with multiline strings which do not require quotes.

sfc-gh-bdufour: We have to think long term here IMHO. We should be allowing templating in files other than the project definition file (it's on NADE's roadmap for the near future anyway). What if someone is using a numerical port number in a container spec? We can't assume strings are the only use case.

Also, we should not assume a specific parser and/or generator is used. What we make our users write should be valid YAML so that they are free to load their files in any editor, IDE or tool they choose, and it won't fail on them.

sfc-gh-turbaszek: So the solution you propose is to use different templating for snowflake.yml. I'm fine with that.  But let's be mindful we will end up with 4 templates:
- {{ ... }} for server side
- &{ ... }  for SQL
- ${ ... }  for snowflake.yaml
- ??? for creating templates of yamls that can be rendered, this syntaxt has to be different than all previous

sfc-gh-melnacouzi: To add on that, the idea of templating is that you replace the template with the value before reading it. In that case, if we don't use the quote, the template marker should be replaced by a non-quoted string, which is a valid YAML - which is why it doesn't seem right to force users with quoting it.

Would we consider the idea of using {{ }} instead, and have snow CLI only parse stuff that starts with ctx. ?

sfc-gh-turbaszek: Looking for ctx requires parsing before expanding. Plus one can have SQL template in snsowlfake.yml that uses `{{ ctx }}`. Templates within templates within templates 🤯

src/snowflake/cli/api/utils/rendering.py
sfc-gh-melnacouzi: I think direction doesn't matter because external_data is at a separate level in this case (top level), while the rest is under ctx.env. My understanding, external_data is only there for backward compatibility with snow sql, and will not be used for non sql stuff. @sfc-gh-turbaszek correct me if I'm wrong.

src/snowflake/cli/api/utils/rendering.py
sfc-gh-melnacouzi: I think we have to parse before expanding, and do it iteratively - at least for snowflake.yaml, because vars are defined there. Could you please elaborate on the second part about sql templates in snowflake.yml?

sfc-gh-turbaszek: Imagine we will add support for embedding SQL in snowflake.yml. Then users may be tempted to 
```
sql:
   run: "EXECUTE IMMEDIATE  SELECT * FROM {{ ctx }}"
env:
   ...
```
In this way usage of `{{ }}` will be ambiguous. Thus, we need different template pattern.

sfc-gh-melnacouzi: It's the same context, so it shouldn't matter where it's replaced. Or am I missing something?

sfc-gh-bdufour: No no, I'm not proposing to use 4 different template syntax options. `{{ }}` is the server side one, so that ship has sailed. I'm proposing that we pick something for the client side that works for both SQL and YAML, and make that the client-side template. Then we also support `&{ }` as a legacy feature (maybe it's even a feature flag?). We monitor its use for deprecation purposes. If you use &{}, you use it exclusively, and only in a SQL file. Everybody else should be using the new and improved syntax that doesn't break YAML parsers. So really, it's only 2 options, other than legacy, which is something we must support regardless.

src/snowflake/cli/api/utils/rendering.py
sfc-gh-turbaszek: Correct, and in current implementation there's no option for external data to override context data (ctx key is reserved), so direction does not matter.

src/snowflake/cli/api/utils/rendering.py
sfc-gh-turbaszek: Having legacy for sql, still leaves us with 3 templates (legacy, jinja, something for yaml) + one more if someone wants to generate snowflake.yml, or do I miss something?

sfc-gh-bdufour: Why would snowflake.yml be any different? But yes, 3 total, 2 actively supported and 1 legacy. I do think that's the minimal number. Happy to involve the product folks here if you think it would be worthwhile.

src/snowflake/cli/api/project/schemas/project_definition.py
sfc-gh-turbaszek: Removed

src/snowflake/cli/api/utils/models.py
sfc-gh-turbaszek: Renamed

sfc-gh-pjob (dismissed):

sfc-gh-melnacouzi (approved):

PR #1067, opened 2024-05-14
Snow 1272319 alias snowcli core object commands

sfc-gh-pczajka:
### Pre-review checklist

### Changes description
Alias `snow object` commands for Snowflake-CLI core objects: git-repository, stage, streamlit, procedures and functions.


src/snowflake/cli/plugins/git/commands.py
```diff
@@ -260,3 +276,31 @@ def execute(
...
+@app.command("list", requires_connection=True)
+def list_(
+    like: str = GitRepositoryLikeOption,
+    scope: Tuple[str, str] = ScopeOption,
+    **options,
+):
+    """Lists all available git repository stages."""
+    object_list(
+        object_type=GIT_REPOSITORY_OBJECT_TYPE, like=like, scope=scope, **options
+    )
```
sfc-gh-turbaszek, line 290: This looks like a bit of repetition. Maybe we can use approach similar to what we do for init commands?
https://github.com/snowflakedb/snowflake-cli/blob/7ef04cab05f2819db55f91c9c9f2ce2acc62c071/src/snowflake/cli/plugins/snowpark/commands.py#L78

sfc-gh-pczajka: Nice idea!
I created `object.command_aliases.add_object_command_aliases` for that. I've only left Snowpark unchanged, as it differs from the rest - asks for object type (procedure/function)

PR #1072, opened 2024-05-15
[NADE] Add generated DDL to Setup Script

sfc-gh-bgoel:
### Pre-review checklist

### Changes description
Part of the setup script codegen. This PR takes the generated DDL from the previous few steps and adds the content to new SQL files. These SQL files are further added to the setup script for Native Apps.


src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -313,3 +315,63 @@ def get_src_py_file_to_dest_py_file_map(
...
+    """
+    Adds an 'execute immediate' to setup script for every SQL file in the map
+    """
+    # For every SQL file, add SQL statement 'execute immediate' to setup script.
+    setup_file_path = find_setup_script_file(deploy_root=deploy_root)
+    for _, sql_file in dest_file_to_sql_file.items():
+        sql_file_relative_path = sql_file.relative_to(deploy_root)  # Path on stage
+        with open(setup_file_path, "a") as file:
+            file.write("\n")
+            file.write(f"EXECUTE IMMEDIATE FROM {sql_file_relative_path}")
```
sfc-gh-bgoel, line 377: Are we still doing the template variable replacement `<% functions %>` given it is just the extension functions being added to the file?

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -268,6 +268,40 @@ def find_manifest_file(deploy_root: Path) -> Path:
     )
 
 
+def find_and_read_manifest_file(deploy_root: Path) -> Any:
```
sfc-gh-bdufour, line 271: why `-> Any`? This utility function should be opinionated about what it returns, or take arguments to control str vs bytes for example.

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -268,6 +268,40 @@ def find_manifest_file(deploy_root: Path) -> Path:
...
+
+    manifest_content = find_and_read_manifest_file(deploy_root=deploy_root)
+
+    if (artifacts in manifest_content) and (
+        setup_script in manifest_content[artifacts]
+    ):
+        setup_script_rel_path = manifest_content[artifacts][setup_script]
+        # src_path = os.path.realpath(setup_script_rel_path)
+        file_name = Path(deploy_root / setup_script_rel_path)
+        if file_name.exists():
```
sfc-gh-bdufour, line 295: `is_file` is better than `exists` here, it could be a directory.

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -268,6 +268,40 @@ def find_manifest_file(deploy_root: Path) -> Path:
...
+            return Path(file_name)
```
sfc-gh-bdufour, line 296: nit: it's already a `Path` object

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -268,6 +268,40 @@ def find_manifest_file(deploy_root: Path) -> Path:
...
+        else:
+            raise SourceNotFoundError(file_name)
```
sfc-gh-bdufour, line 298: nit: it's not a source, it's the deployed manifest we couldn't find.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -313,3 +315,63 @@ def get_src_py_file_to_dest_py_file_map(
         # TODO: Does this case 5 need to be considered?
 
         return src_py_file_to_dest_py_file_map
+
+
+def _create_new_sql_file_at_path(parent_dir: Path, file_name: str) -> Optional[Path]:
+    """
+    Creates an empty SQL file at the same location specified as the parent directory, with file_name as part of the new file's name.
+    """
+    sql_file_name = f"__{file_name}_{uuid.uuid4().hex}.sql"
```
sfc-gh-bdufour, line 324: Not a fan of using uuid4 here for a few reasons. The most important one is that it won't be stable across bundle operations, so it will make any kind of diff unusable. The other is that UUID are terrible for humans. If I need a human to ever read a random ID, I much prefer to use base 36 random numbers. But really I think we should do something stable here, and avoid collisions in a more deterministic way.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -313,3 +315,63 @@ def get_src_py_file_to_dest_py_file_map(
...
+    Creates SQL files at predetermined location as specified in the input dictionary,
+    and writes the contents from the map to the new file.
+    """
+    dest_file_to_sql_file: Dict[Path, Path] = {}
+    for dest_file, all_ddl in dest_file_py_file_to_ddl_map.items():
+        if all_ddl is None or all_ddl.strip() == "":
+            continue
+
+        sql_file_path = _create_new_sql_file_at_path(
+            parent_dir=dest_file.parent, file_name=dest_file.name.split(".")[0]
```
sfc-gh-bdufour, line 350: the `split` piece is just `Path.stem` I believe since we're dealing with python files.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -313,3 +315,63 @@ def get_src_py_file_to_dest_py_file_map(
...
+
+def _create_new_sql_file_at_path(parent_dir: Path, file_name: str) -> Optional[Path]:
+    """
+    Creates an empty SQL file at the same location specified as the parent directory, with file_name as part of the new file's name.
+    """
+    sql_file_name = f"__{file_name}_{uuid.uuid4().hex}.sql"
+    sql_file_path = Path(parent_dir, sql_file_name)
+    try:
+        f = open(sql_file_path, "x")
+        f.close()
```
sfc-gh-bdufour, line 328: nit: not a huge fan of doing the open .. close thing twice (one here, one below). Feels like we can just rehash this logic a little to make it more intuitive.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -313,3 +315,63 @@ def get_src_py_file_to_dest_py_file_map(
...
+
+
+def create_and_write_to_sql_files(dest_file_py_file_to_ddl_map: Dict[Path, str]):
+    """
+    Creates SQL files at predetermined location as specified in the input dictionary,
+    and writes the contents from the map to the new file.
+    """
+    dest_file_to_sql_file: Dict[Path, Path] = {}
+    for dest_file, all_ddl in dest_file_py_file_to_ddl_map.items():
+        if all_ddl is None or all_ddl.strip() == "":
```
sfc-gh-bdufour, line 346: nit: weird, by this point I'd expect things to have been filtered out so that the generation logic doesn't have to deal with unexpected state. Also, your dict says Path -> str, so None is really unexpected here.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -313,3 +315,63 @@ def get_src_py_file_to_dest_py_file_map(
...
+
+        sql_file_path = _create_new_sql_file_at_path(
+            parent_dir=dest_file.parent, file_name=dest_file.name.split(".")[0]
+        )
+        if sql_file_path is None:
+            # We lose information in this step, since the file already exists and we do not overwrite or append to that
+            # as we do not know what already exists in that file.
+            continue
+
+        with open(sql_file_path, "a") as file:
```
sfc-gh-bdufour, line 357: "a" looks odd here, if we're generating we'll want to overwrite the file each time when we use a stable naming scheme.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bdufour: Yeah, that's a great question. My opinion here is likely to be unpopular. PL tomorrow?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bgoel: The reason I avoided Path.stem is for this reason:
```
PurePosixPath('my/library.tar.gz').stem
'library.tar'
```
Not sure about a user's play here, but if they name a file like `foobar.<ext>.py` for some reason, then `Path.stem` will output `foobar.<ext>`, but I just want the file name. 

This'll depend on how the sandbox/snowpark responds to being passed in a file like that.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bgoel: Yes true, this is just an additional check, because this function can be generalized for any other create and write and we may not have control over what are the contents in the map outside the snowpark processor. I can remove it for now if you strongly feel one way.

I double checked about the type hint here, but at least in py3.11:
```
>>> my_dict: Dict[str, str] = {}
>>> my_dict["hello"] = None
>>> my_dict
{'hello': None}
```

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bgoel: Updated with the regular `FileNotFoundError`, or do we want ClickException? I like the built in python exc `FileNotFoundError` since a user is much more likely to be familiar with this without reading too much into the error msgs/logs.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bgoel: Made an update to the naming scheme, open for discussion. Thought about `GENERATED_BY_SNOWFLAKECLI` prefix or suffix, just like we use for comments in SQL DDL of application package and application. Open for discussion.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -317,17 +316,13 @@ def get_src_py_file_to_dest_py_file_map(
...
 
 
-def _create_new_sql_file_at_path(parent_dir: Path, file_name: str) -> Optional[Path]:
+def _generate_new_sql_file_name(parent_dir: Path, file_name: str) -> Optional[Path]:
     """
-    Creates an empty SQL file at the same location specified as the parent directory, with file_name as part of the new file's name.
+    Generates a SQL filename using the provided parent directory, with file_name as part of the new file's name.
     """
-    sql_file_name = f"__{file_name}_{uuid.uuid4().hex}.sql"
+    sql_file_name = f"__cli_codegen__{file_name}.sql"
```
sfc-gh-bdufour, line 323: better for sure, but my favorite way to solve this is to reduce the problem to a single source of collisions: we introduce some kind of "generated" directory (we call it something like _generated) and add an option to the PDF to redefine it if need be, the same way you can change the deploy root. This way, most users don't need to worry about it, it's clean and easy to inspect. Diffs work. Filenames make sense. But if by any chance there's a collision, we tell the user something akin to "you've already used our directory, delete/rename it, or change it in your config", then we bail out of code generation altogether.

sfc-gh-bgoel: Would `_generated` live at `output/deploy` ? With the current implementation, generated SQL files live next to their python files in the deploy root. If there is a `_generated` folder, then:
1. Do we essentially duplicate the path within `__generated` for the SQL file? e.g. if there is a `output/deploy/a/b/c.py`, will `c.sql` live at `output/deploy/__generated/a/b/c.sql`, or even ? 
2.  The user will need to inspect the `__generated` folder keeping the directory structure of `output/deploy` in mind. Might be too much to keep track of at once?
3. How will the upload work? Deploy root is a mirror of the stage, does that mean that the py file and the sql file will live at different paths in the stage?
4. I think depending on this, I'll implement the logic of the handler property, which is currently on hold.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bdufour: Python types are hints, they don't impact the behaviour at runtime because Python is very much dynamically typed. But using variables in a way that's inconsistent with type hints is just bad for understanding. Not to mention it's annoying to see the squigglies in an IDE :)

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bdufour: Let's discuss tomorrow then.

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bdufour: Doesn't FNFE cause the CLI to crash horribly? I thought ClickException subclasses are handled gracefully and not much else. I was just pointing out that SourceNotFound is used for a particular set of errors, and doesn't seem to match here. The manifest might very well be present, but if the user doesn't have a rule to deploy it, we still won't find it. Maybe we need a new kind of error?

src/snowflake/cli/api/project/schemas/native_app/native_app.py
```diff
@@ -24,6 +24,10 @@ class NativeApp(UpdatableModel):
         title="Folder at the root of your project where the build step copies the artifacts.",
         default="output/deploy/",
     )
+    generated_root: Optional[str] = Field(
+        title="Folder at the root of your deploy_root where the build step checks in the generated artifacts. This is also the name of the SQL file that will be created.",
```
sfc-gh-bdufour, line 28: How about simply:
```
Subdirectory of the deploy root where files generated by the Snowflake CLI will be written.
```

We shouldn't make this overly specific to the main file. I think we can play ostrich with that one for now.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/models.py
```diff
@@ -28,7 +28,11 @@ class NativeAppExtensionFunction(_CallableBase):
     )
     packages: Optional[List[str]] = Field(
         title="List of packages (with optional version constraints) to be loaded for the function",
-        default={},
+        default=[],
+    )
+    raw_imports: Optional[List[Union[str, Tuple[str, str]]]] = Field(
```
sfc-gh-bdufour, line 33: IMHO this should go to imports, not expose yet another field to represent the same data. This way we'll be consistent with Snowpark entities defined in the PDF. We can make the driver generate the correct list directly, as with the other fields.

Also, this change is unrelated to the DDL file generation change, I would have likely made 2 different PRs for these.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -149,22 +152,27 @@ def __init__(
...
         self.deploy_root = deploy_root
+        self.generated_root = generated_root
 
     def process(
         self,
         artifact_to_process: PathMapping,
         processor_mapping: Optional[ProcessorMapping],
         **kwargs,
-    ) -> str:  # String output is temporary until we have better e2e testing mechanism
+    ) -> Dict[
```
sfc-gh-bdufour, line 173: I was just waiting for the file output to remove the hack. I don't think we should do this. First, it breaks all existing snapshots, so it's hard to check for regressions. Second, it doesn't let us test the actual output anyway, so it's of very limited value.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -199,20 +209,88 @@ def process(
                 if grant_statements is not None:
                     cc.message(grant_statements)
                     collected_output.append(grant_statements)
+            extension_fns_ddl = "\n".join(collected_output)
+            dest_file_py_file_to_ddl_map[py_file] = extension_fns_ddl
+
+        write_to_sql = kwargs.get(
```
sfc-gh-bdufour, line 215: why do we need this? We always want to write the files IMHO.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -179,8 +187,10 @@ def process(
             bundle_map, processor_mapping
         )
 
-        collected_output = []
+        dest_file_py_file_to_ddl_map: Dict[Path, str] = {}
```
sfc-gh-bdufour, line 190: nit: I'm not sure we need to buffer. We can just write the SQL files for each python file for which generation succeeded. If we do this, we can populate the main file at the very end, we just need to remember the SQL files to `execute immediate` from the main file. That's a lot less information in memory for users that have thousands of UDFs/SPROCs.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -199,20 +209,88 @@ def process(
...
+
+        write_to_sql = kwargs.get(
+            "write_to_sql", True
+        )  # temporary until we have better e2e testing mechanism
+
+        if write_to_sql:
+            dest_file_to_sql_file = create_and_write_to_sql_files(
+                dest_file_py_file_to_ddl_map=dest_file_py_file_to_ddl_map,
+                deploy_root=self.deploy_root,
+                generated_root=self.generated_root,
```
sfc-gh-bdufour, line 223: nit: combine into just a destination directory? No need to have other parts of the logic worry about what the relationship of generated root is to the deploy root. This way it would be really easy to introduce subdirs of the generated root in the future, for example.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -199,20 +209,88 @@ def process(
...
+                    f"{extension_fn.handler} in {py_file} in the deploy root contains an import specifying a stage path. Skipping validation of this import."
+                )
+                extension_fn.imports.append(import_str)
+            elif import_str == "/":
+                # Do not validate, but add to imports
+                # Specifying only "/" will result in a Snowflake error, but do not block a user if it exists.
+                cc.warning(
+                    f"{import_str} may not be a valid input. Skipping validation of this import."
+                )
+                extension_fn.imports.append(import_str)
```
sfc-gh-bdufour, line 259: Why is this a special case we need to handle?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -199,20 +209,88 @@ def process(
...
+                    generated_root=self.generated_root,
+                )
+
+        return dest_file_py_file_to_ddl_map
+
+    def _normalize_raw_imports(
+        self,
+        extension_fn: NativeAppExtensionFunction,
+        py_file: Path,
+        resolved_deploy_root: Path,
```
sfc-gh-bdufour, line 239: the deploy root fed to the ctor is already resolved and absolute (IIRC, please check, but that was the intention)

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -199,20 +209,88 @@ def process(
...
                     collected_output.append(grant_statements)
+            extension_fns_ddl = "\n".join(collected_output)
+            dest_file_py_file_to_ddl_map[py_file] = extension_fns_ddl
+
+        write_to_sql = kwargs.get(
+            "write_to_sql", True
+        )  # temporary until we have better e2e testing mechanism
+
+        if write_to_sql:
+            dest_file_to_sql_file = create_and_write_to_sql_files(
```
sfc-gh-bdufour, line 220: Subjective, but that's a strange primitive to me. It does a bit too much? Why is it returning a map? Conceptually, I'd probably do something like:

```py
generated_sql_dir = self.deploy_root / self.generated_dir  # I'm assume generated dir is relative here, but adjust as needed
all_sql_files = []
for py_file, extension_fns in collected_extension_functions_by_path.items():
    sql_file_rel_path = generate_ddl(py_file, extensions_fns, output_dir = self.deploy_root / self.generated_root
    if sql_file_path:  # if DDL generation fails, None is returned
        all_sql_files.append(sql_file_rel_path)
```

if all_sql_files:
    # generate main file here will one execute immediate per file in the list

# weave setup script here

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -343,3 +422,95 @@ def generate_grant_sql_ddl_statements(
...
+    Creates SQL files at predetermined location as specified in the input dictionary,
+    and writes the contents from the map to the new file.
+    """
+    dest_file_to_sql_file: Dict[Path, Path] = {}
+    absolute_deploy_root = resolve_without_follow(deploy_root)
+    absolute_generated_root = resolve_without_follow(generated_root)
+
+    for dest_file, all_ddl in dest_file_py_file_to_ddl_map.items():
+        if all_ddl.strip() == "":
+            continue
```
sfc-gh-bdufour, line 457: nit: possible? It shouldn't be. We should even be able to `assert` that it's the case.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -343,3 +422,95 @@ def generate_grant_sql_ddl_statements(
         grant_sql_statements.append(grant_sql_statement)
 
     return "\n".join(grant_sql_statements)
+
+
+def generate_new_sql_file_name(parent_dir: Path, file_name: str) -> Optional[Path]:
```
sfc-gh-bdufour, line 427: subjective, but this logic seems to be more complicated than it needs to be, something like:

```py
relative_py_file = py_file.relative_to(self.deploy_root)
sql_file = py_file.with_suffix(".sql")
if sql_file.exists():
    # oops

relative_py_file.parent.mkdir(exist_ok=True, parents=True)  # if needed, adjust as you see fit
```

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -343,3 +422,95 @@ def generate_grant_sql_ddl_statements(
...
+            file_name=f"{dest_file.stem}",
+        )
+        if sql_file_path is None:
+            # We lose information in this step, since the file already exists and we skip the overwrite.
+            continue
+
+        sql_file_path.parent.mkdir(parents=True, exist_ok=True)
+
+        with open(sql_file_path, "w+") as file:
+            file.write(all_ddl)
```
sfc-gh-bdufour, line 473: nit: see above, DDL is generated in pieces, it would be preferable to write it in pieces, otherwise we do a lot of work in memory for no particular reason (these files could be big)

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -343,3 +422,95 @@ def generate_grant_sql_ddl_statements(
...
+
+        dest_file_to_sql_file[dest_file] = sql_file_path
+
+    return dest_file_to_sql_file
+
+
+def edit_setup_script_with_exec_imm_sql(
```
sfc-gh-bdufour, line 480: note: this is going to be replaced in the next PR? Appending is obviously broken.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/models.py
```diff
@@ -29,8 +29,12 @@ class NativeAppExtensionFunction(_CallableBase):
     )
     packages: Optional[List[str]] = Field(
         title="List of packages (with optional version constraints) to be loaded for the function",
-        default={},
+        default=[],
     )
+    # raw_imports: Optional[List[Union[str, Tuple[str, str]]]] = Field(
```
sfc-gh-bdufour, line 34: forgot this one

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -201,21 +215,74 @@ def process(
                     cc.message(grant_statements)
                     collected_output.append(grant_statements)
 
-                self.deannotate(py_file, extension_fns)
+                with open(sql_file, "a") as file:
```
sfc-gh-bdufour, line 218: why append? Isn't it a new file?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/models.py
sfc-gh-bdufour, line 34: missed one :)

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -181,7 +187,15 @@ def process(
         )
 
         collected_output = []
+        collected_sql_files: List[Path] = []
+        resolved_generated_root = resolve_without_follow(self.generated_root)
```
sfc-gh-bdufour, line 191: nit: I still think we should make it an invariant that Paths passed into this class are resolved upfront, making this unnecessary. If you want, we can assert that all provided paths are absolute at least in the ctor.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -361,3 +429,60 @@ def generate_grant_sql_ddl_statements(
         grant_sql_statements.append(grant_sql_statement)
 
     return "\n".join(grant_sql_statements)
+
+
+def generate_new_sql_file_name(
```
sfc-gh-bdufour, line 434: nit: move this to the class, and access deploy and generated roots directly from `self`?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -361,3 +429,60 @@ def generate_grant_sql_ddl_statements(
...
+    Generates a SQL filename for the generated root from the python file, and creates its parent directories.
+    """
+    relative_py_file = py_file.relative_to(deploy_root)
+    sql_file = Path(generated_root, relative_py_file.with_suffix(".sql"))
+    if sql_file.exists():
+        cc.warning(
+            f"""\
+            File {sql_file} already exists, will append SQL statements to this file.
+        """
+        )
```
sfc-gh-bdufour, line 447: Impossible path?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -361,3 +429,60 @@ def generate_grant_sql_ddl_statements(
...
+):
+    """
+    Adds an 'execute immediate' to setup script for every SQL file in the map
+    """
+    # Create a __generated.sql in the __generated folder
+    generated_file_path = Path(generated_root, f"{generated_root.stem}.sql")
+    generated_file_path.parent.mkdir(exist_ok=True, parents=True)
+
+    if generated_file_path.exists():
+        cc.warning(
```
sfc-gh-bdufour, line 463: nit: mention why? As a user I'd be pretty puzzled to see this message.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -361,3 +429,60 @@ def generate_grant_sql_ddl_statements(
...
+        )
+        return
+
+    # For every SQL file, add SQL statement 'execute immediate' to __generated.sql script.
+    for sql_file in collected_sql_files:
+        sql_file_relative_path = sql_file.relative_to(
+            deploy_root
+        )  # Path on stage, without the leading slash
+        with open(generated_file_path, "a") as file:
+            file.write(f"\nEXECUTE IMMEDIATE FROM '/{sql_file_relative_path}';")
```
sfc-gh-bdufour, line 474: flip this around, you should not keep opening and closing the same file to append, it's expensive. You should open the file, write in multiple batches, then close.

tests/nativeapp/codegen/snowpark/test_python_processor.py
```diff
@@ -168,27 +173,253 @@ def test_generate_grant_sql_ddl_statements(native_app_extension_function, snapsh
...
+        )
+        assert sql_file_path == Path(generated_root, Path(py_file).with_suffix(".sql"))
+
+
+# --------------------------------------------------------
+# --------- create_and_write_to_sql_files ----------------
+# --------------------------------------------------------
+
+
+# def test_create_and_write_to_sql_files_no_existing_generate(snapshot):
```
sfc-gh-bdufour, line 206: ?

tests/nativeapp/codegen/snowpark/test_python_processor.py
```diff
@@ -237,14 +484,39 @@ def test_process_with_collected_functions(
...
-            deploy_root=Path(local_path, "output/deploy"),
+            deploy_root=deploy_root,
+            generated_root=generated_root,
         ).process(
             artifact_to_process=native_app_project_instance.native_app.artifacts[0],
             processor_mapping=processor_mapping,
         )
-        assert output == snapshot
+        # Cannot compare keys as these are temporary paths that change per test run
+        # assert output == snapshot
```
sfc-gh-bdufour, line 501: nit: sure, but why have this assertion here and commented out?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -201,21 +215,74 @@ def process(
...
+        py_file: Path,
+        deploy_root: Path,
+    ):
         if extension_fn.name is None:
             # The extension function was not named explicitly, use the name of the Python function object as its name
             extension_fn.name = extension_fn.handler
 
         # Compute the fully qualified handler
+        # If user defined their udf as @udf(lambda: x, ...) then extension_fn.handler is <lambda>.
+        # Further, de-annotating @udf will likely remove the presence of the function in the py file at all.
```
sfc-gh-bdufour, line 276: It will not

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -201,21 +215,74 @@ def process(
...
+        extension_fn: NativeAppExtensionFunction,
+        py_file: Path,
+        deploy_root: Path,
+    ):
         if extension_fn.name is None:
             # The extension function was not named explicitly, use the name of the Python function object as its name
             extension_fn.name = extension_fn.handler
 
         # Compute the fully qualified handler
+        # If user defined their udf as @udf(lambda: x, ...) then extension_fn.handler is <lambda>.
```
sfc-gh-bdufour, line 275: These shouldn't even be present in the output from the collection subprocess, are you saying they're currently not filtered out?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bgoel: Same as one of my comments above. This is technically not possible unless a user themselves creates a `__generated` folder in deploy root, for whatever reason. And then checked in a `py_file_name.sql` in that folder. 
Highly unlikely they do that, but that's why the warning and the appending to write to the file instead of overwriting it entirely.

sfc-gh-bgoel: One mismatch currently in this code:
If the main __generated.sql exists, I flag that error and return, not appending.
But if the generated subfiles exists, I append to those. 
We can make this behavior consistent. I dont have a strong opinion either way, so WDYT?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bdufour: If you feel it's absolutely necessary, but really, relying on CWD is broken anyway, so passing the deploy / project roots around as resolved, absolute directories is really the only contract that makes sense to me. I don't like the idea that a class could silently just infer something wrong from the CWD if we ever mess up the logic. I'd much rather have the code fail during development, hence the suggestion to use assertions. Strong contracts make code easy to maintain.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bdufour: It shouldn't be possible. If the user already has a generated dir, we should refuse to generate _anything_ (in the compiler) unless they change the generated dir.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bdufour: Yeah, they'll be there alright, and they're tricky to filter out because Python doesn't expose the fact that they're lambdas in a reliable way. `inspect` might give us clues, or we can filter out anything that has the same name as `lambda None` (just to make sure we're a bit resilient to changes in names, although that's pretty unlikely).

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bdufour: Speculative complexity. We probably wouldn't reuse the helper from there, we might want to introduce a base class for generation-based logic, etc. IMHO write the simplest code, most readable code there is, and refactor aggressively.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bdufour: See above, I think we stop generation if the user has the own __generated dir (or whatever the config value is set to). We preserve the invariant that we generate to an empty directory. If there's a clash, it means that something's off with symlinks or something like that, I'd probably want an error message to figure out what's going on. It's very much unexpected.

src/snowflake/cli/api/project/schemas/native_app/native_app.py
```diff
@@ -24,6 +24,10 @@ class NativeApp(UpdatableModel):
         title="Folder at the root of your project where the build step copies the artifacts.",
         default="output/deploy/",
     )
+    generated_root: Optional[str] = Field(
+        title="Subdirectory of the deploy root where files generated by the Snowflake CLI will be written.",
+        default="__generated/",
```
sfc-gh-cgorrie, line 29: Any reason we don't want this nested under the `output/` directory? To me, this is output.

sfc-gh-bdufour: it's under `output/deploy/__generated` now because it needs to end up on the stage.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bgoel: Ack, the changes are coming in the next commit. I have limited path resolution requirement to processors, not to the compiler. If we discover further use of resolved paths elsewhere in the compiler, we can then refactor the paths to the compiler itself be resolved.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bgoel: I am not sure if this refusal should be at the compiler level just yet, given we don't know how other processors may be written in the future or what their behavior might be.
So I have added the refusal just to the Snowpark Processor for now, before we even start doing any of the collection work. Idea being, if you're not writing the collected functions into a file, then why are you collecting the functions at all? Printing to console is one reason, but not one I'm convinced needs all this work.

src/snowflake/cli/api/project/schemas/native_app/native_app.py
sfc-gh-cgorrie: Oh. I swear "Subdirectory of deploy root" wasn't there when I reviewed 😂 ignore this

sfc-gh-bgoel: All files under this generated directory also get uploaded to the stage, with the `__generated` parent directory preserved and visible on the stage. Which means if `output/deploy` mirrors the stage, then `__generated` should belong directly under `deploy`.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bdufour: Fair enough

sfc-gh-bdufour (approved):

PR #1073, opened 2024-05-15
Introduce BundleMap to encapsulate artifact path mappings

sfc-gh-bdufour:
### Pre-review checklist

### Changes description

Introduced BundleMap as an abstraction that encapsulates src -> dest mapping in native app projects. This will enable other parts of the logic (e.g. artifact processors) to query the mapping in a much cleaner way. There are some changes in behaviour introduced as part of this change, most notably enforcing that all src & dest mappings are relative paths.

### Testing

* New logic is fully unit tested, including as many corner cases in the contract as I could think of
* Verified the previous logic manually before capturing the corner cases
* Verified the implementation by running `snow app bundle` manually

### DIFF
--- a/src/snowflake/cli/plugins/nativeapp/artifacts.py
+++ b/src/snowflake/cli/plugins/nativeapp/artifacts.py
@@ -50,7 +50,7 @@ class TooManyFilesError(ClickException):
 
     def __init__(self, dest_path: Path):
         super().__init__(
-            f"{dedent(str(self.__doc__))})\ndest_path = {dest_path}".strip()
+            f"{dedent(str(self.__doc__))}\ndestination = {dest_path}".strip()
         )
         self.dest_path = dest_path
 
@@ -402,7 +402,9 @@ def delete(path: Path) -> None:
         spath.rmdir(recursive=True)  # remove dir and all contains
 
 
-def symlink_or_copy(src: Path, dst: Path, makedirs=True, overwrite=True) -> None:
+def symlink_or_copy(
+    src: Path, dst: Path, deploy_root: Path, makedirs=True, overwrite=True
+) -> None:
     """
     Tries to create a symlink to src at dst; failing that (i.e. in Windows
     without Administrator / Developer Mode) copies the file from src to dst instead.
@@ -413,6 +415,14 @@ def symlink_or_copy(src: Path, dst: Path, makedirs=True, overwrite=True) -> None
     sdst = SecurePath(dst)
     if makedirs:
         sdst.parent.mkdir(parents=True, exist_ok=True)
+
+    # Verify that the mapping isn't accidentally trying to create a file in the project source through symlinks.
+    # We need to ensure we're resolving symlinks for this check to be effective.
+    resolved_dst = dst.resolve()
+    resolved_deploy_root = deploy_root.resolve()
+    if resolved_deploy_root not in resolved_dst.parents:
+        raise NotInDeployRootError(dest_path=dst, deploy_root=deploy_root, src_path=src)
+
     if overwrite:
         delete(dst)
     try:
@@ -478,7 +488,7 @@ def build_bundle(
     for (absolute_src, absolute_dest) in bundle_map.all_mappings(
         absolute=True, expand_directories=False
     ):
-        symlink_or_copy(absolute_src, absolute_dest)
+        symlink_or_copy(absolute_src, absolute_dest, deploy_root=deploy_root)
 
     return bundle_map
 
--- a/src/snowflake/cli/plugins/nativeapp/artifacts.py
+++ b/src/snowflake/cli/plugins/nativeapp/artifacts.py
@@ -3,16 +3,13 @@ from __future__ import annotations
 import os
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Callable, Dict, Iterator, List, Optional, Tuple, Union
 
 from click.exceptions import ClickException
 from snowflake.cli.api.constants import DEFAULT_SIZE_LIMIT_MB
 from snowflake.cli.api.project.schemas.native_app.path_mapping import PathMapping
 from snowflake.cli.api.secure_path import SecurePath
 from yaml import safe_load
-
-# Map from source directories and files in the project directory to their path in the deploy directory. Both paths are absolute.
-ArtifactDeploymentMap = Dict[Path, Path]
 
 
 class DeployRootError(ClickException):
@@ -33,26 +30,13 @@ class ArtifactError(ClickException):
         super().__init__(msg)
 
 
-class GlobMatchedNothingError(ClickException):
+class SourceNotFoundError(ClickException):
     """
-    No files were found that matched the provided glob pattern.
+    No match was not found for the specified source in the project directory
     """
 
     def __init__(self, src: str):
         super().__init__(f"{self.__doc__}: {src}")
-
-
-class SourceNotFoundError(ClickException):
-    """
-    The specifically-referenced source file or directory was not found
-    in the project directory.
-    """
-
-    path: Path
-
-    def __init__(self, path: Path):
-        super().__init__(f"{self.__doc__}\npath = {path}")
-        self.path = path
 
 
 class TooManyFilesError(ClickException):
@@ -75,20 +59,17 @@ class NotInDeployRootError(ClickException):
     (use "./" instead to copy into the deploy root).
     """
 
-    artifact_src: str
-    dest_path: Path
+    dest_path: Union[str, Path]
     deploy_root: Path
 
-    def __init__(self, artifact_src: str, dest_path: Path, deploy_root: Path):
+    def __init__(self, dest_path: Union[Path, str], deploy_root: Path):
         super().__init__(
             f"""
             {self.__doc__}
-            artifact_src = {artifact_src}
             dest_path = {dest_path}
             deploy_root = {deploy_root}
             """
         )
-        self.artifact_src = artifact_src
         self.dest_path = dest_path
         self.deploy_root = deploy_root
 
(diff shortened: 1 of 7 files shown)


src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -361,21 +361,25 @@ def sync_deploy_root_with_stage(
...
             if not recursive:
                 verify_no_directories(resolved_paths_to_sync)
-            deploy_paths_to_sync = [
-                source_path_to_deploy_path(p, mapped_files)
-                for p in resolved_paths_to_sync
-            ]
-            verify_exists(deploy_paths_to_sync)
+
+            deploy_paths_to_sync = []
+            for resolved_path in resolved_paths_to_sync:
```
sfc-gh-bdufour, line 374: I went for minimal disturbance here because the focus was on `BundleMap` as a new abstraction. Once it's in, I'll rehash this logic to integrate better, but I'll likely need to tweak the APIs in the `BundleMap` class to do so. I didn't want the extra noise here.

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -75,37 +60,300 @@ class NotInDeployRootError(ClickException):
...
+        """
+        Yields a (src, dest) pair for each deployed artifact in the project. Each pair corresponds to a single file
+        in the project. Source directories are resolved as needed to resolve their contents.
+
+        Arguments:
+            self: this instance
+            absolute (bool): Specifies whether the yielded paths should be joined with the project or deploy roots,
+             as appropriate.
+            walk_directories (bool): Specifies whether directory to directory mappings should be expanded to
+             resolve their contained files.
```
sfc-gh-cgorrie, line 252: For what it's worth, when I looked at the usage of `all_mappings` I thought `walk_directories` would do the opposite (i.e. would directories show up in the generated pairs, or just the files?) but now I realize the question is whether or not we go through the directory boundary when it's one of the members.

This is a major nitpick, but `expand_directories` I think would be more precise.

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -361,21 +361,25 @@ def sync_deploy_root_with_stage(
...
+                verify_exists(resolved_path)
+                deploy_path = bundle_map.to_deploy_path(resolved_path)
+                if deploy_path is None:
+                    raise ClickException(f"No artifact found for {resolved_path}")
+                deploy_paths_to_sync.append(deploy_path)
+
             stage_paths_to_sync = _get_stage_paths_to_sync(
-                deploy_paths_to_sync, self.deploy_root.resolve()
+                deploy_paths_to_sync, resolve_without_follow(self.deploy_root)
```
sfc-gh-cgorrie, line 382: Is this a bugfix?

tests/nativeapp/test_artifacts.py
```diff
@@ -44,6 +46,512 @@ def dir_structure(path: Path, prefix="") -> List[str]:
...
+        bundle_map,
+        {
+            "app/setup.sql": None,
+            ".": None,
+            "/not/in/project": None,
+        },
+    )
+
+
+@pytest.mark.parametrize("mapping_factory", [ArtifactMapping, path_mapping_factory])
```
sfc-gh-cgorrie, line 115: These parameterizations are just giving us a bit more test coverage because we're using `isinstance` instead of duck typing (are `src` and `dest` both strings?) in `BundleMap`. I'm not suggesting a change, just thinking aloud

tests/nativeapp/test_artifacts.py
```diff
@@ -44,6 +46,512 @@ def dir_structure(path: Path, prefix="") -> List[str]:
     return parts
 
 
+@pytest.fixture
+def bundle_map():
```
sfc-gh-cgorrie, line 50: How about a test like the following, I'm curious what happens with these changes:

```
- src: src/snowpark
  dest: ./snowpark
- src: README.md
  dest: ./snowpark/
```

sfc-gh-cgorrie (commented):
Some food for thought, though it looks pretty much ready

tests/nativeapp/test_artifacts.py
sfc-gh-bdufour: Added, and it works (hopefully) exactly as you'd expect: it's allowed because in both cases the destination is a directory and there are no file collisions.

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -75,37 +60,300 @@ class NotInDeployRootError(ClickException):
...
+        canonical_dest = self._canonical_dest(dest)
+
+        src_is_dir = absolute_src.is_dir()
+        if map_as_child:
+            # Make sure the destination is a child of the original, since this was requested
+            canonical_dest = canonical_dest / canonical_src.name
+            dest_is_dir = src.is_dir()
+
+        if canonical_src in self._src_to_dest:
+            raise TooManyFilesError(canonical_dest)
```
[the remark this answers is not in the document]
sfc-gh-bdufour: Sure does, silly me. Will fix.

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-bdufour: Kind of, it ends up making things inconsistent on Windows because the specific resolution logic you use matters when doing prefix comparison. We use resolve_without_follow everywhere except for here, and without this change some integ tests fail due to the weird ~1 expansion windows does sometimes.

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bdufour: I used `walk` because `os.walk` was the analogy in my head, by expand is fine too.

tests/nativeapp/test_artifacts.py
sfc-gh-bdufour: There are two entry points for add, at least conceptually. One uses a PathMapping as input, the other uses and ArtifactMapping. They're really pretty much equivalent, except that PathMapping carries around the optional `dest` (and soon processors). TBH I want to remove ArtifactMapping now that we moved over to pydantic. That class made sense when we were getting the original contents of the yaml file as a dictionary, but it's very redundant now. I wrote tests this way because removing ArtifactMapping would then be trivial :)

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -33,28 +31,15 @@ def __init__(self, msg: str):
         super().__init__(msg)
 
 
-class GlobMatchedNothingError(ClickException):
+class SourceNotFoundError(ClickException):
     """
-    No files were found that matched the provided glob pattern.
+    No match was not found for the specified source in the project directory
```
sfc-gh-bgoel, line 36: typo: No match was found for the ...

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -75,37 +60,300 @@ class NotInDeployRootError(ClickException):
     (use "./" instead to copy into the deploy root).
     """
 
-    artifact_src: str
```
sfc-gh-bgoel, line 78: My guess is that the `artifact_src` was there just for an FYI, so a user will know the source to the file that should have been there in the deploy root. Do we not want to provide that information anyway? Maybe having the artifact src in there helps the user narrow the error faster in snowflake.yml, as opposed to having multiple artifact src's end up in the same dest path and combing through each of them to figure out why it threw an error?

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -75,37 +60,300 @@ class NotInDeployRootError(ClickException):
...
         self.dest_path = dest_path
         self.deploy_root = deploy_root
 
 
 @dataclass
 class ArtifactMapping:
     """
-    Used to keep track of equivalent paths / globs so we can copy
-    artifacts from the project folder to the deploy root.
+    Used to keep track of equivalent paths / globs so we can copy artifacts from the project folder to the deploy root.
```
sfc-gh-bgoel, line 81: Aren't we getting rid of this in favor of `PathMapping` pydantic model?

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -75,37 +60,300 @@ class NotInDeployRootError(ClickException):
...
+            )
+        return resolved_src
+
+    def _absolute_dest(self, dest: Path) -> Path:
+        if dest.is_absolute():
+            resolved_dest = resolve_without_follow(dest)
+        else:
+            resolved_dest = resolve_without_follow(self._deploy_root / dest)
+        if (
+            self._deploy_root != resolved_dest
```
sfc-gh-bgoel, line 318: This is to eliminate condition when src is checked directly under deploy root? In which case dest = deploy_root?

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -75,37 +60,300 @@ class NotInDeployRootError(ClickException):
...
+            raise SourceNotFoundError(src)
+
+    def add(self, mapping: Union[ArtifactMapping, PathMapping]) -> None:
+        """
+        Adds an artifact mapping rule to this instance.
+        """
+        if isinstance(mapping, ArtifactMapping):
+            self._add_mapping(mapping.src, mapping.dest)
+        elif isinstance(mapping, PathMapping):
+            self._add_mapping(mapping.src, mapping.dest)
```
sfc-gh-bgoel, line 199: Not clear on why the distinction here. I thought part of `PathMapping` is used to create `ArtifactMapping` ?

sfc-gh-bgoel (commented):
This work was much needed to avoid all the confusion between absolute, resolves, relatives when dealing with paths. Thank you!

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bdufour: The issue is I don't always have a source given the abstractions in place, and it would make the code really tricky if we had to thread it through in all cases. I added it back as an optional piece, on a best effort basis. Does that work?

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bdufour: In due time, this PR is already big enough. I have a few follow-ups planned.

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bdufour: Think of it as 2 distinct APIs for now. Our existing code uses ArtifactMapping everywhere, but the upcoming use by processors will rely on PathMapping instead. So I handle both, until I can unify them.

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bdufour: This is basically enforcing that any destination path has to be the deploy root or one of its children. The check to make sure we don't clobber the deploy root happens elsewhere, so this logic is good enough here.

tests/nativeapp/test_artifacts.py
sfc-gh-bdufour: I should expand: we don't currently have the ability to detect deep collisions (I added a skipped test that illustrates that). The new logic does not fix this, but I have some thoughts on how to make it possible. It will be a bigger change, but it would also help with some of our outstanding bugs.

tests/nativeapp/test_artifacts.py
sfc-gh-cgorrie: That makes a ton of sense. Optimize for the clarity of the refactor you know is coming.

tests/nativeapp/test_artifacts.py
sfc-gh-cgorrie: Can we add an assert that `README.md` doesn't end up in the `snowpark` source folder?

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bgoel: That works!

sfc-gh-bgoel (dismissed):
LGTM!

tests/nativeapp/test_artifacts.py
sfc-gh-bdufour: Yeah, it was possible in the old logic to pollute the source directory. Added logic to explicitly disallow that, + an integ test to verify.

sfc-gh-cgorrie (approved):
LGTM, thanks for adding the integration test

PR #1076, opened 2024-05-16
[NADE] Add Native Apps Codegen for Snowpark Annotations

sfc-gh-bgoel:
### Pre-review checklist

### Changes description
This PR introduces code to:
1. Collect external Snowpark functions from python files for setup script code generation
2. Generate DDL from the collected functions
3. Check if feature flag is enabled, if yes, then allow code generation

There will be a follow up PR to main to take the generated DDL and write to new SQL files.


src/snowflake/cli/plugins/nativeapp/codegen/compiler.py
```diff
@@ -0,0 +1,73 @@
...
+    UnsupportedArtifactProcessorError,
+)
+from snowflake.cli.plugins.nativeapp.codegen.snowpark.python_processor import (
+    SnowparkAnnotationProcessor,
+)
+
+SNOWPARK_PROCESSOR = "snowpark"
+
+# Singleton dictionary of all processors created and shared between different artifact objects
+cached_processors: Dict[str, ArtifactProcessor] = {}
```
sfc-gh-cgorrie, line 22: This is a code smell to me. Should we introduce some sort of context object or attach it to an existing one?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/callback_source.py.jinja
```diff
@@ -0,0 +1,118 @@
...
+            "inline_python_code": ext_fn.inline_python_code,
+            "raw_imports": ext_fn.raw_imports,
+            "replace": ext_fn.replace,
+            "if_not_exists": ext_fn.if_not_exists,
+            "execute_as": ext_fn.execute_as,
+            "anonymous": ext_fn.anonymous,
+            # Set func based on type
+            "func": ext_fn.func.__name__
+            if isinstance(ext_fn.func, Callable)
+            else ext_fn.func,
```
sfc-gh-cgorrie, line 63: Nit: indent the last two lines

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function_utils.py
```diff
@@ -0,0 +1,368 @@
...
+    attr: str,
+    make_uppercase: bool = False,
+    py_file: Optional[Path] = None,
+    raise_err: bool = False,
+):
+    """
+    Sanitizes a single key-value pair of the specified dictionary. As part of the sanitization,
+    it goes through a few checks. A key must be created if it does not already exist.
+    Then, it checks the type of the value of the key, i.e. if it is of type str, and if it contains any leading or trailing whitespaces.
+    A user is able to specity if they want to re-assign a key to an uppercase instance of its original value.
```
sfc-gh-cgorrie, line 52:
```suggestion
    A user is able to specify if they want to re-assign a key to an uppercase instance of its original value.
```

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function_utils.py
```diff
@@ -0,0 +1,368 @@
...
+    Get a handler for an extension function based on the file path on the stage. If a specific suffix needs to be removed from the path,
+    then that is also taken into account.
+    """
+    # rel_file_path = file_path.absolute().relative_to(deploy_root.resolve())  # No leading slash
+    # stem, suffix = os.path.splitext(rel_file_path)
+    # if (suffix_str_to_rm is not None) and (suffix == suffix_str_to_rm):
+    #     file_parts = Path(stem).parts
+    # else:
+    #     file_parts = rel_file_path.parts
+    # return ".".join(file_parts)
```
sfc-gh-cgorrie, line 268: What's happening with these commented-out lines?

src/snowflake/cli/plugins/nativeapp/utils.py
```diff
@@ -82,3 +82,22 @@ def verify_no_directories(paths_to_sync: List[Path]):
 def verify_exists(path: Path):
     if not path.exists():
         raise ClickException(f"The following path does not exist: {path}")
+
+
+def get_all_file_paths_under_dir(directory: Path):
+    """
+    Recursively traverse the given directory and yield file paths. If the directory is a symlink, then the function follows the behavior of os.walk for symlinks.
+    The directory may be a relative path to the project root.
```
sfc-gh-cgorrie, line 90: Calling out that this relies on cwd == project root. Might want to add to the docstring.

src/snowflake/cli/plugins/nativeapp/utils.py
```diff
@@ -82,3 +82,22 @@ def verify_no_directories(paths_to_sync: List[Path]):
...
+    """
+    for root, _, files in os.walk(directory):
+        for file in files:
+            yield Path(os.path.join(root, file))
+
+
+def filter_files(generator: Generator, predicate_func: Callable):
```
sfc-gh-cgorrie, line 97: Is this utility necessary? Can it not be a generator comprehension? https://peps.python.org/pep-0289/

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function_utils.py
```diff
@@ -0,0 +1,368 @@
...
+    Parameters:
+        name (str): Any arbitrary string
+    Returns:
+        A string that has replaced underscores with spaces.
+    """
+    return name.replace("_", " ")
+
+
+def _sanitize_str_attribute(
+    ex_fn: Dict[str, Any],
```
sfc-gh-cgorrie, line 42: I see `ex_fn` a lot but I'm really not sure what it conceptually maps to. Is this like a property bag for a sproc/udf? Maybe we can use a type alias and add a docstring there?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function_utils.py
```diff
@@ -0,0 +1,368 @@
...
+    """
+    if py_file is None:
+        raise ValueError("Python file path must not be None.")
+    return f"Required attribute '{attribute}' of extension function is missing or incorrectly defined for an extension function defined in python file {py_file.absolute()}."
+
+
+def _is_function_wellformed(ex_fn: Dict[str, Any]) -> bool:
+    """
+    Checks if the specified dictionary contains a key called 'func'.
+    if it does, then the value must be of type str or a list of fixes size 2.
```
sfc-gh-cgorrie, line 124:
```suggestion
    if it does, then the value must be of type str or a list of fixed size 2.
```

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function_utils.py
```diff
@@ -0,0 +1,368 @@
...
+    It further checks the item at 1st index of this list.
+    """
+    if ex_fn.get("func", None):
+        if isinstance(ex_fn["func"], str):
+            return ex_fn["func"].strip() != ""
+        elif isinstance(ex_fn["func"], list) and (len(ex_fn["func"]) == 2):
+            return isinstance(ex_fn["func"][1], str) and ex_fn["func"][1].strip() != ""
```
sfc-gh-cgorrie, line 131: Why do we only have to check the 2nd string?

tests/nativeapp/codegen/snowpark/test_python_processor.py
```diff
@@ -0,0 +1,327 @@
...
+    assert generate_create_sql_ddl_statements(native_app_codegen_full_json) == snapshot
+
+
+def test_generate_create_sql_ddl_statements_none():
+    ex_fn = {
+        "object_type": "PROCEDURE",
+        "object_name": "CORE.MYFUNC",
+        "anonymous": True,
+    }
+    assert generate_create_sql_ddl_statements(ex_fn=ex_fn) is None
```
sfc-gh-cgorrie, line 170: Why does this generate `None`? Shouldn't it throw an error if the input is malformed instead?

sfc-gh-cgorrie (commented):
I'm not ready to sign off yet, still understanding the python processor and what's going on in callback_source.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function_utils.py
sfc-gh-bdufour: yeah, it's an extension function property bag expressed as a dict. Rather temporary I'd say, the validation piece here is very difficult to reason about already.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -0,0 +1,368 @@
...
+            if processor_mapping is not None
+            else {}
+        )
+
+        bundle_map = BundleMap(
+            project_root=self.project_root, deploy_root=self.deploy_root
+        )
+        bundle_map.add(artifact_to_process)
+
+        # 2. Get raw extension functions through Snowpark callback
```
sfc-gh-cgorrie, line 174: There is no #1

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -0,0 +1,368 @@
...
+                    kwargs=kwargs,
+                )
+            except Exception as exc:
+                cc.warning(
+                    f"Error processing extension functions in {src_file}: {exc}"
+                )  # Display the actual file for the user to inspect
+                cc.warning("Skipping generating code of all objects from this file.")
+                collected_raw_ex_fns = None
+
+            if (collected_raw_ex_fns is None) or (len(collected_raw_ex_fns) == 0):
```
sfc-gh-cgorrie, line 196:
```suggestion
            if not collected_raw_ex_fns:
```

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -0,0 +1,368 @@
...
+    if inline_python_code:
+        inline_python_code_in_sql = f"""\
+AS $$
+{inline_python_code}
+$$
+"""
+    else:
+        inline_python_code_in_sql = ""
+
+    create_query = dedent(
```
sfc-gh-cgorrie, line 331: What's the point of dedent here? It's already dedented.

src/snowflake/cli/plugins/nativeapp/codegen/compiler.py
sfc-gh-bgoel: I'd like to understand more on why this is a concerning issue.
Regardless, I could make this a part of the existing global cli context. However, most of the properties on the existing cli context class are universal, i.e. used across various command subgroups. But this processors dict is a very specific part of native apps and within native apps, the artifacts that it processes. So is adding it to a global context object justified?

If we want to leverage this new codegen implementation for wider use beyond NA going forward, e.g. in pure Snowpark, then it might make sense to promote it to be universal.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function_utils.py
sfc-gh-bgoel: There are two types that you could have passed to a func property in Snowpark: Callable and Tuple. 
The tuple construction is this: (file_path, func_name). We don't really do any validation on the file_path yet in our MVP. That leaves the func_name, which we are checking to ensure that is not blank.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function_utils.py
sfc-gh-bgoel: The handler work is a bit more complicated, and is being resolved as part of the follow up PR. Until then, it stays "NotImplementedHandler". I have removed the commented lines to avoid confusion, but now you have function arguments that are essentially unused, but I can leave them as such for now.

src/snowflake/cli/plugins/nativeapp/utils.py
sfc-gh-bgoel: Actually for this and the function above, I am not longer using these two. The original function that was using these two utilities has been replaced by Bruno's BundleMap implementation. So I have deleted these two utilities.

tests/nativeapp/codegen/snowpark/test_python_processor.py
sfc-gh-bgoel: From a user's perspective, defining an anonymous procedure is not a malformed input/ask. It is only in the Native Apps context that anonymous procedures do not exist. As a result, when these two properties are detected in python_processor.generate_create_sql_ddl_statements, we only print a warning to the user that we are skipping generation and moving on to the next entity, no exceptions thrown. Hence the return type in that scenario is a None.

src/snowflake/cli/plugins/nativeapp/codegen/compiler.py
sfc-gh-cgorrie: We can certainly make a team decision that the current approach needs to be re-written if we want wider use. Generally I expect there to be one context-level object for each global-ish lifetime and for us to jam things in there. This isn't a problem right now because we're only ever working with one project.

My suggestion would be to call out the cases in which this cached mapping would cause incorrect behaviour in the comment above, I think we can accept it for now but we need to be mindful if/when we move out into a library.

tests/nativeapp/codegen/snowpark/test_python_processor.py
sfc-gh-cgorrie: Got it, so none == skipped because irrelevant here. I'd like this part of the contract called out in the docstring.

src/snowflake/cli/plugins/nativeapp/codegen/compiler.py
sfc-gh-bgoel: I've added more clarification, lmk.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function_utils.py
sfc-gh-bgoel: We actually never use the func when it is of type `Tuple`. There is an exception thrown in one of the other methods `_get_handler()` if it detects the Tuple. What we can do is bring that exception to this method instead of further down. But as it stands, we don't need to perform file validation because we don't act on this func type.

sfc-gh-cgorrie (approved):
LGTM. As discussed offline, we can get to some incremental refactors later.

PR #1079, opened 2024-05-16
Fix creation of the 3rd+ patch on an app pkg version

sfc-gh-cgorrie:
### Pre-review checklist

### Changes description
`NativeAppRunProcessor.get_existing_version_info` rewritten to return the most-recent patch row when trying to get information about a specific version.

A new integration test was added to ensure that 3 patches can be added to a version through the CLI.

### DIFF
--- a/src/snowflake/cli/plugins/nativeapp/run_processor.py
+++ b/src/snowflake/cli/plugins/nativeapp/run_processor.py
@@ -9,10 +9,16 @@ from click import UsageError
 from snowflake.cli.api.console import cli_console as cc
 from snowflake.cli.api.exceptions import SnowflakeSQLExecutionError
 from snowflake.cli.api.project.schemas.native_app.native_app import NativeApp
+from snowflake.cli.api.project.util import (
+    identifier_to_show_like_pattern,
+    unquote_identifier,
+)
+from snowflake.cli.api.utils.cursor import find_all_rows
 from snowflake.cli.plugins.nativeapp.constants import (
     ALLOWED_SPECIAL_COMMENTS,
     COMMENT_COL,
     LOOSE_FILES_MAGIC_VERSION,
+    PATCH_COL,
     SPECIAL_COMMENT,
     VERSION_COL,
 )
@@ -30,7 +36,7 @@ from snowflake.cli.plugins.nativeapp.policy import PolicyBase
 from snowflake.cli.plugins.stage.diff import DiffResult
 from snowflake.cli.plugins.stage.manager import StageManager
 from snowflake.connector import ProgrammingError
-from snowflake.connector.cursor import SnowflakeCursor
+from snowflake.connector.cursor import DictCursor, SnowflakeCursor
 
 UPGRADE_RESTRICTION_CODES = {93044, 93055, 93045, 93046}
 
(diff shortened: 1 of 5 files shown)


src/snowflake/cli/api/sql_execution.py
```diff
@@ -188,13 +193,21 @@ def show_specific_object(
...
-            )
 
-        show_obj_row = find_first_row(
+        matching_rows = find_all_rows(
             show_obj_cursor,
             lambda row: row[name_col] == unquote_identifier(unqualified_name),
         )
-        return show_obj_row
+
+        if len(matching_rows) == 0:
```
sfc-gh-bgoel, line 202: Is this condition same as line 194, do we need an explicit check?

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -146,6 +146,7 @@ def get_existing_version_info(self, version: str) -> Optional[dict]:
                     version,
                     name_col=VERSION_COL,
                     in_clause=f"in application package {self.package_name}",
+                    pick_if_multiple=lambda rows: rows[-1],
```
sfc-gh-bgoel, line 149: I forget the output rules here, why are we only returning the last row? Is some order guaranteed in the output that `-1` will always be deterministic?

src/snowflake/cli/api/sql_execution.py
sfc-gh-cgorrie: No, `None` means that the query was not successfully executed, whereas this is looking for a successfully-executed query with a zero-length list.

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-cgorrie: I had made an assumption here that the patches were in order, and we'd be interested in the last patch. Though now I'm thinking about it I don't think there's any such guarantees stated. Right now, everything that calls `get_existing_version_info` just throws away the result after it determines if it's `None` or not, so it's moot -- which is why I'm kind of thinking maybe we don't need such a deep change and can just call `show versions like <version name>` directly...

WDYT?

sfc-gh-bgoel: I like the idea here to keep it simple, I think `show_specific_object` is probably an overkill for our kind of query here. We could implement our own python `max` on the patch column for a given version to really determine the max patch seen so far.

[if only this could be treated as regular sql :D ]
```
with t1 as (show versions like <> in application package)
select max(patch) from t1
```

sfc-gh-cgorrie: OK, if you feel the same way then it's settled, I'll go for a simpler implementation and ping you again when this is ready for re-review.

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -136,24 +142,34 @@ def get_all_existing_versions(self) -> SnowflakeCursor:
...
-                    name_col=VERSION_COL,
-                    in_clause=f"in application package {self.package_name}",
+                query = f"show versions like {identifier_to_show_like_pattern(version)} in application package {self.package_name}"
+                cursor = self._execute_query(query, cursor_class=DictCursor)
+
+                if cursor.rowcount is None:
+                    raise SnowflakeSQLExecutionError(query)
+
+                matching_rows = find_all_rows(
+                    cursor, lambda row: row[VERSION_COL] == unquote_identifier(version)
```
sfc-gh-cgorrie, line 159: Extra check needed because we could have different versions that match our like clause, e.g. `"v1"` and `V1`

sfc-gh-turbaszek (approved):

sfc-gh-bdufour (dismissed):

sfc-gh-bdufour (approved):

sfc-gh-bgoel (approved):
LGTM

PR #1085, opened 2024-05-17
Use original argument names rather than Snowpark generated ones

sfc-gh-bdufour:
### Pre-review checklist

### Changes description

When generating DDL from Snowpark annotations, we should use the user's original argument names. Currently, Snowpark generated names are used.

### Testing

Tested manually. This functionality will be tested via integ tests when the feature works end-to-end.


src/snowflake/cli/plugins/nativeapp/codegen/snowpark/callback_source.py.jinja
```diff
@@ -58,9 +78,7 @@ def __snowflake_cli_native_app_internal_callback_replacement():
             "execute_as": ext_fn.execute_as,
             "anonymous": ext_fn.anonymous,
             # Set func based on type
-            "func": ext_fn.func.__name__
-            if isinstance(ext_fn.func, Callable)
-            else ext_fn.func,
+            "func": ext_fn.func.__name__ if isinstance(ext_fn.func, Callable) else ext_fn.func,
```
sfc-gh-bgoel, line 81: How did this pass the lint check!

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/callback_source.py.jinja
```diff
@@ -34,15 +35,34 @@ def __snowflake_cli_native_app_internal_callback_replacement():
...
     ):
-
         {% raw %}ext_fn = extension_function_properties
+        input_args = None
+
+        # Try to fetch the original argument names from the extension function
+        if isinstance(ext_fn.func, Callable):
+            try:
+                original_arg_names = inspect.getfullargspec(ext_fn.func)[0]
+                start_index = len(original_arg_names) - len(ext_fn.input_args)
```
sfc-gh-bgoel, line 45: Why the subtraction? Shouldn't len of `original_arg_names` be the same as len of`ext_fn.input_args` as the latter is created by the former, i.e. start_index is always 0?

sfc-gh-bdufour: No, because some external functions take a session, but that's implicit in SQL. input_args is just the "real" args, but the python args can include more.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/callback_source.py.jinja
sfc-gh-bdufour: it's a jinja file, it's exempt :)

sfc-gh-bgoel (approved):
LGTM! Thank you!

sfc-gh-bdufour:
Dropped in favour of #1096

PR #1089, opened 2024-05-20
Add support for multiple -f flags in snow sql

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
Support executing multiple files in `snow sql` command by passing multiple `-f` flags.


src/snowflake/cli/api/utils/cursor.py
```diff
@@ -18,3 +18,9 @@ def find_first_row(
 ) -> Optional[dict]:
     """Returns the first row that matches the predicate, or None."""
     return next(_rows_generator(cursor, predicate), None)
+
+
+def join_cursors(cursors: List[Iterable[SnowflakeCursor]]) -> Iterable[SnowflakeCursor]:
+    while cursors:
+        cur = cursors.pop(0)
```
sfc-gh-cgorrie, line 25: nit: list needs to continually be modified, `for cur in cursors:` feels simpler and more performant given `cursors` is a `List`

src/snowflake/cli/plugins/sql/manager.py
```diff
@@ -35,9 +35,30 @@ def execute(
 
         if std_in:
             query = sys.stdin.read()
-        elif file:
-            query = SecurePath(file).read_text(file_size_limit_mb=UNLIMITED)
+        elif files and len(files) == 1:
+            # If there's only one file we need to handle case of single query
+            query = SecurePath(files[0]).read_text(file_size_limit_mb=UNLIMITED)
```
sfc-gh-cgorrie, line 40: Just curious, why a special case for one file?

sfc-gh-cgorrie, line 40: Non-blocking: I'd suggest removing the special case for a single file and instead tally all queries you ever found to figure out if there is a "single statement" or not, but maybe you are working around some property of `split_statements` (does it erroneously return non-statements if a file is empty of non-comment SQL?).

sfc-gh-cgorrie (approved):
NADE-owned files LGTM, just a couple thoughts. And it would be nice to see a new test to codify what exactly happens when empty files are included in the file list.

src/snowflake/cli/api/utils/cursor.py
sfc-gh-turbaszek: Good point, adjusted

src/snowflake/cli/plugins/sql/manager.py
sfc-gh-turbaszek: Good point, changed

src/snowflake/cli/plugins/sql/manager.py
```diff
@@ -3,30 +3,30 @@
...
 from jinja2 import UndefinedError
 from snowflake.cli.api.secure_path import UNLIMITED, SecurePath
 from snowflake.cli.api.sql_execution import SqlExecutionMixin
+from snowflake.cli.api.utils.cursor import join_cursors
 from snowflake.cli.api.utils.rendering import snowflake_sql_jinja_render
 from snowflake.cli.plugins.sql.snowsql_templating import transpile_snowsql_templates
 from snowflake.connector.cursor import SnowflakeCursor
 from snowflake.connector.util_text import split_statements
 
+SingleStatement = bool
```
sfc-gh-pczajka, line 18: How about
```suggestion
IsSingleStatement = bool
```

src/snowflake/cli/api/utils/cursor.py
```diff
@@ -18,3 +18,8 @@ def find_first_row(
 ) -> Optional[dict]:
     """Returns the first row that matches the predicate, or None."""
     return next(_rows_generator(cursor, predicate), None)
+
+
+def join_cursors(cursors: List[Iterable[SnowflakeCursor]]) -> Iterable[SnowflakeCursor]:
```
sfc-gh-pczajka, line 23: How about using `itertools.chain.from_iterable` instead?
https://docs.python.org/3/library/itertools.html#itertools.chain.from_iterable

PR #1096, opened 2024-05-22
[NADE] Rewrote extension function collection to use a Pydantic-based schema validator

sfc-gh-bdufour:
### Pre-review checklist

### Changes description

Rewrote the extension function collection logic to:
- Avoid executing `__main__` guards when user code is loaded (sadly, we are forced to leak a few variables in the user code sandbox due to this change)
- Disable stdout and stderr while user code is executing
- Follow the existing Snowpark entity definition whenever possible
- Disallow some of the really complex Snowpark scenarios that are not planned to be supported, at least in the initial deliverable
- Perform validation of the collected entities using a Pydantic model

### Testing

Tested manually via invoking the CLI with the feature flag enabled on a representative sample application. This is partial work towards an upcoming feature and is not yet fully testable end to end.


src/snowflake/cli/plugins/nativeapp/codegen/snowpark/callback_source.py.jinja
```diff
@@ -12,107 +14,144 @@ except ModuleNotFoundError as exc:
...
+__snowflake_global_collected_extension_fn_json = []
+
+def __snowflake_internal_create_extension_fn_registration_callback():
+    def __snowflake_internal_try_extract_lineno(extension_fn):
+        try:
+            return inspect.getsourcelines(extension_fn)[1]
+        except Exception:
+            return None
+
+    def __snowflake_internal_extract_name(extension_fn):
```
sfc-gh-bgoel, line 37: nit: `__snowflake_internal_extract_object_name`. For a second I confused it with the original arg name impl.

Also, should we have a short docstring describing each sub-function here?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/callback_source.py.jinja
```diff
@@ -12,107 +14,144 @@ except ModuleNotFoundError as exc:
...
+                    return None
+
+        except Exception:
+            # ignore any exception and fall back to using the object name reported from Snowpark
+            pass
+
+        return extension_fn.object_name
+
+    def __snowflake_internal_create_package_list(extension_fn):
+        return [pkg_spec.strip() for pkg_spec in extension_fn.all_packages.split(",")]
```
sfc-gh-bgoel, line 53: What's the use case here? Are we ever getting `'pkg_spec'` with leading or trailing spaces?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/callback_source.py.jinja
```diff
@@ -12,107 +14,144 @@ except ModuleNotFoundError as exc:
...
+        try:
+            args_spec = inspect.getfullargspec(extension_fn.func)
+            original_arg_names = args_spec[0]
+            start_index = len(original_arg_names) - len(extension_fn.input_sql_types)
+            signature = []
+            defaults_start_index = len(original_arg_names) - len(args_spec.defaults or [])
+            for i in range(len(extension_fn.input_sql_types)):
+                arg = {
+                    'name': original_arg_names[start_index + i],
+                    'type': extension_fn.input_sql_types[i]
```
sfc-gh-bgoel, line 66: 👍

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/callback_source.py.jinja
```diff
@@ -12,107 +14,144 @@ except ModuleNotFoundError as exc:
...
+        # Failed to extract the original arguments through reflection, fall back to alternative approach
+        return [
+            {"name": input_arg.name, "type": input_type}
+            for (input_arg, input_type) in zip(extension_fn.input_args, extension_fn.input_sql_types)
+        ]
+
+    def __snowflake_internal_extension_fn_to_json(extension_fn):
+        if not isinstance(extension_fn.func, Callable):
+            # Unsupported case: extension function is a tuple
+            return
```
sfc-gh-bgoel, line 86: Does the user find out why their code gen failed in this scenario?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/callback_source.py.jinja
```diff
@@ -12,107 +14,144 @@ except ModuleNotFoundError as exc:
...
     )
     sys.exit(1)
 
-__snowflake_cli_native_app_internal_callback_return_list: List[Any] = []
+__snowflake_global_collected_extension_fn_json = []
+
+def __snowflake_internal_create_extension_fn_registration_callback():
+    def __snowflake_internal_try_extract_lineno(extension_fn):
+        try:
+            return inspect.getsourcelines(extension_fn)[1]
```
sfc-gh-bgoel, line 33: [1] to omit the decorator?
Also, would it make sense to return `inspect.getsourcelines(extension_fn)[1:]` instead, especially for de-annotation work, instead of only `def` of the function?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/callback_source.py.jinja
```diff
@@ -12,107 +14,144 @@ except ModuleNotFoundError as exc:
...
+    def __snowflake_internal_extension_fn_to_json(extension_fn):
+        if not isinstance(extension_fn.func, Callable):
+            # Unsupported case: extension function is a tuple
+            return
+
+        # Collect basic properties of the extension function
+        extension_fn_json = {
+            "function_type": extension_fn.object_type.name,
+            "lineno": __snowflake_internal_try_extract_lineno(extension_fn.func),
+            "name": __snowflake_internal_extract_name(extension_fn),
```
sfc-gh-bgoel, line 92: If this returns `None`, should we even proceed with the rest? Because user must have not provided a `name` attribute on the decorator, but we require it.  In that case, it might be worth shifting this check up after line 86.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/callback_source.py.jinja
```diff
@@ -12,107 +14,144 @@ except ModuleNotFoundError as exc:
...
+            "handler": extension_fn.func.__name__,
+            "imports": extension_fn.all_imports or [],
+            "packages": __snowflake_internal_create_package_list(extension_fn),
+            "runtime": extension_fn.runtime_version,
+            "returns": extension_fn.return_sql.upper().replace("RETURNS ", "").strip(),
+            "signature": __snowflake_internal_make_extension_fn_signature(extension_fn),
+            "external_access_integrations": extension_fn.external_access_integrations or [],
+            "secrets": extension_fn.secrets or {},
+        }
```
sfc-gh-bgoel, line 101: Some missing keys:
1. raw_imports: unvalidated here is fine, the validation is part of the weaving DDL PR. We do need access to the raw imports that a user has passed into the decorator, because it may or may not be the path to this exact file. And in our case, with Session being None always, `all_imports` is guaranteed to be empty.
2. anonymous: we need to check this, either here in this file or later in generation, about anonymous procedures because we skip generating DDL for that.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/models.py
```diff
@@ -0,0 +1,41 @@
...
+from pydantic import Field
+from snowflake.cli.api.project.schemas.snowpark.callable import _CallableBase
+from snowflake.cli.api.project.schemas.updatable_model import IdentifierField
+
+
+class ExtensionFunctionTypeEnum(str, Enum):
+    PROCEDURE = "PROCEDURE"
+    FUNCTION = "FUNCTION"
+    TABLE_FUNCTION = "TABLE_FUNCTION"
+    AGGREGATE_FUNCTION = "AGGREGATE_FUNCTION"
```
sfc-gh-cgorrie, line 15: Any reason why we need underscores here? Is there somewhere we're using them that would parse incorrectly, or is it more of a defense against these perhaps being used in a future context where they'd be treated as two separate literals?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/callback_source.py.jinja
```diff
@@ -12,107 +14,144 @@ except ModuleNotFoundError as exc:
...
+    if global_key.startswith("__snowflake_internal"):
+        del globals()[global_key]
+
+del globals()["global_key"] # make sure to clean up the loop variable as well
+
 try:
-    exec(code, orig_globals)
+    import importlib
+    with contextlib.redirect_stdout(None):
+        with contextlib.redirect_stderr(None):
```
sfc-gh-bgoel, line 147: Nit: can be combined into one line:
```
with contextlib.redirect_stdout(None), contextlib.redirect_stderr(None):
    ...
```

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/models.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class NativeAppExtensionFunction(_CallableBase):
+    function_type: ExtensionFunctionTypeEnum
+    lineno: Optional[int] = Field(
+        title="The line number of the extension function", default=None
+    )
+    name: Optional[str] = Field(title="The name of the external function", default=None)
```
sfc-gh-bgoel, line 23: typo: extension

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/models.py
```diff
@@ -0,0 +1,41 @@
...
+    packages: Optional[List[str]] = Field(
+        title="List of packages (with optional version constraints) to be loaded for the function",
```
sfc-gh-bgoel, line 25: nit: should we specify these packages are already single quoted?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/models.py
```diff
@@ -0,0 +1,41 @@
...
+
+class ExtensionFunctionTypeEnum(str, Enum):
+    PROCEDURE = "PROCEDURE"
+    FUNCTION = "FUNCTION"
+    TABLE_FUNCTION = "TABLE_FUNCTION"
+    AGGREGATE_FUNCTION = "AGGREGATE_FUNCTION"
+
+
+class NativeAppExtensionFunction(_CallableBase):
+    function_type: ExtensionFunctionTypeEnum
```
sfc-gh-cgorrie, line 19: Nit: just `type` is clearer as some of these aren't functions, they're procedures

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/models.py
```diff
@@ -0,0 +1,41 @@
...
+        alias="schema",
+    )
+    application_roles: Optional[List[str]] = Field(
+        title="Application roles granted usage to the function",
+        default=[],
+    )
+    execute_as_caller: Optional[bool] = Field(
+        title="Determine whether the procedure is executed with the privileges of "
+        "the owner or with the privileges of the caller",
+        default=None,
```
sfc-gh-cgorrie, line 40: Default is none but it's not clear whether that means callers' or owners'

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -154,217 +162,180 @@ def process(
...
                 )
-
-            # 4. Enrich the raw extension functions by setting additional properties
-            for raw_ex_fn in filtered_collection:
-                sanitize_extension_function_data(ex_fn=raw_ex_fn, py_file=dest_file)
-                enrich_ex_fn(
-                    ex_fn=raw_ex_fn,
-                    py_file=dest_file,
-                    deploy_root=self.deploy_root,
+                cc.message(create_stmt)
```
sfc-gh-bgoel, line 190: Is this intended to be in the final version, or just for debugging for now?

sfc-gh-cgorrie (commented):
Made a few comments. I still need to dive through the `callback_source.py.jinja` file.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -154,217 +162,180 @@ def process(
...
+        extension_fn.handler = f"{py_file.stem}.{extension_fn.handler}"
 
-            collected_ex_fns = dest_file_py_file_to_collected_raw_ex_fns[
-                py_file
-            ]  # Collected entities is List[Dict[str, Any]]
-            if collected_ex_fns is None:
-                continue
+        if extension_fn.imports is None:
+            extension_fn.imports = []
+        extension_fn.imports.append(f"/{to_stage_path(py_file)}")
```
sfc-gh-bgoel, line 213: 1. `NativeAppExtensionFunction` does not have an `imports` attribute.
2. What if the user has already specified the correct path as part of the decorator? This will duplicate that, unless we make it a set.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -154,217 +162,180 @@ def process(
...
-LANGUAGE PYTHON
-RUNTIME_VERSION={ex_fn["runtime_version"]} {imports_in_sql}{packages_in_sql}{external_access_integrations_in_sql}{secrets_in_sql}
-HANDLER='{ex_fn["handler"]}'{execute_as_sql}
-{inline_python_code_in_sql}"""
+    if extension_fn.function_type == ExtensionFunctionTypeEnum.PROCEDURE:
+        if extension_fn.execute_as_caller:
+            create_query += f"\nEXECUTE AS CALLER"
+        else:
+            create_query += f"\nEXECUTE AS OWNER"
+    create_query += ";\n"
```
sfc-gh-bgoel, line 312: 👍

sfc-gh-bgoel (commented):
The overall simplification is much appreciated!

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/callback_source.py.jinja
sfc-gh-bdufour: AFAIK getsourcelines only returns the first line number of the function declaration. It won't help with deannotation, sadly.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/callback_source.py.jinja
sfc-gh-bdufour: From the documentation they'd be able to see it. We don't have a good way to collect/report warnings to the parent process, so I left it out. We can add that in the future, but it's out of scope now IMHO.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/callback_source.py.jinja
sfc-gh-bdufour: I strip the Snowpark prefix here, which means we use the handler's name in the generation logic. I think that's intuitive enough to keep it.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/callback_source.py.jinja
sfc-gh-bdufour: I left raw imports out on purpose. We can discuss if you want. Anonymous should be filtered out here instead of exposing it through collected functions.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/models.py
sfc-gh-bdufour: I can remove the underscores so that YAML looks pretty.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/models.py
sfc-gh-bdufour: I'm OK being a bit lenient here TBH

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bdufour: Definitely temporary

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bdufour: 1. Yes it does, inherited from the Snowpark model
2. Yup, we'll duplicate it for now. Not the final version. We need to rehash how we handle imports anyway. Just focusing on enabling e2e testing in here.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/callback_source.py.jinja
sfc-gh-bdufour: I moved away from "object name" in the schema, so this was intentional. Didn't get the confusion, what's that impl you refer to?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/models.py
sfc-gh-bdufour: default in SF is Owner, so changed to False

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/callback_source.py.jinja
sfc-gh-bdufour: Oh I see what you're asking. `getsourcelines` is peculiar, it returns a pair of `(source code of the function, first lineno)`. So I discard the source code and keep the line number.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/models.py
sfc-gh-bdufour: type isn't allowed in source code by the linter (it shadows a python built-in), but I made the change in the schema at least.

sfc-gh-cgorrie: Ah, I didn't realize that. OK with whatever in that case.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/callback_source.py.jinja
sfc-gh-bgoel: +1, let's have this discussion offline.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bgoel: 1. ack
2. I see, can be part of the PR I'm working on.

sfc-gh-bgoel (approved):
I'm good with this, except for the raw_imports that we'll catch up on. But that stays a one liner change if we do add it here. Hence approving.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/callback_source.py.jinja
```diff
@@ -12,107 +14,155 @@ except ModuleNotFoundError as exc:
...
+        try:
+            return inspect.getsourcelines(extension_fn)[1]
+        except Exception:
+            return None
+
+    def __snowflake_internal_extract_extension_fn_name(extension_fn):
+        try:
+            import snowflake.snowpark._internal.utils as snowpark_utils
+
+            if hasattr(snowpark_utils, 'TEMP_OBJECT_NAME_PREFIX'):
```
sfc-gh-cgorrie, line 41: Is this a version guard? Is it necessary given the v1.15.0 check above and the exception wrapped around?

sfc-gh-bdufour: No, it's pretty much "undoing" a bad transformation, but pulling the authority from Snowpark's internal APIs. Even with the version check passing, they could still rename / remove this field, and we'd fall back to the generic logic. Not ideal, but we have a channel open with Snowpark to push improvements upon this initial solution. That is one IMHO.

PR #1098, opened 2024-05-22
[NADE] De-annotate Snowpark functions when generation was successful

sfc-gh-bdufour:
### Pre-review checklist

### Changes description

Snowpark in the cloud cannot handle `@sproc`, `@udf`, etc annotations, so they must be removed prior to uploading python contents to the stage. For now the logic is rather strict, and only de-annotates functions once generation was confirmed to be successful. This is so that the user will see an error in their setup script if anything went wrong during generation, instead of masking potential errors that the Snowpark validation would have uncovered.

### Testing

* Tested manually using a representative sample app
* Added full coverage via unit tests (minus some defensive, future-proof paths in AST handling code that can't be covered) for the core deannotation logic

I will focus on adding more tests that focus on file-based output after the SQL generation to file is checked in (currently WIP).


src/snowflake/cli/plugins/nativeapp/codegen/snowpark/models.py
```diff
@@ -21,7 +21,7 @@ class NativeAppExtensionFunction(_CallableBase):
         alias="type",
     )
     lineno: Optional[int] = Field(
-        title="The line number of the extension function", default=None
+        title="The start line number of the extension function (1-based)", default=None
```
sfc-gh-cgorrie, line 24:
```suggestion
        title="The starting line number of the extension function (1-based)", default=None
```

tests/nativeapp/codegen/snowpark/test_extension_function_utils.py
```diff
@@ -86,3 +88,203 @@ def test_ensure_all_string_literals():
...
+        signature=[
+            Argument(name="first", type="int"),
+            Argument(name="second", type="int"),
+        ],
+    )
+
+    non_annotated_code = dedent(
+        """
+    if __name__ == "__main__":
+        print("All your base are belong to us")
```
sfc-gh-cgorrie, line 223: What you say?!

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function_utils.py
```diff
@@ -83,3 +85,102 @@ def ensure_all_string_literals(values: Sequence[str]) -> List[str]:
         A list with all values transformed to be valid string literals (as necessary).
     """
     return [ensure_string_literal(value) for value in values]
+
+
+class _FunctionDefAccumulator(ast.NodeVisitor):
```
sfc-gh-cgorrie, line 90: I could use a docstring to understand how this visitor fits in. To me it looks like:

> Compiles a list of function definitions that match a set of `NativeAppExtensionFunction`s.

tests/nativeapp/codegen/snowpark/__snapshots__/test_extension_function_utils.ambr
```diff
@@ -1,33 +1,97 @@
...
+  from snowflake.snowpark.functions import col, sum, sproc
+  
+  
+  @custom
+  #: @sproc(native_app_params={'schema': 'math', 'application_roles': ['app_public', 'app_admin']})
+  def sproc_sum(session: Session, first: int, second: int) -> int:
+      return first + second
+  
+  #: @udf(native_app_params={'schema': 'math', 'application_roles': ['app_public', 'app_admin']})
+  #: @module.annotation
```
sfc-gh-cgorrie, line 17: Why does this one get commented-out?

sfc-gh-cgorrie (commented):
A few questions

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/callback_source.py.jinja
```diff
@@ -30,7 +30,7 @@ __snowflake_global_collected_extension_fn_json = []
 def __snowflake_internal_create_extension_fn_registration_callback():
     def __snowflake_internal_try_extract_lineno(extension_fn):
         try:
-            return inspect.getsourcelines(extension_fn)[1]
+            return inspect.getsourcelines(extension_fn)[1] + 1
```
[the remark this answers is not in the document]
sfc-gh-bgoel: iirc it is zero-indexing vs one-indexing line number confusion that the +1 is trying to resolve. 
e.g. `getsourcelines` returns something like this (copied from [stack overflow](https://stackoverflow.com/questions/57035364/inspect-getsourcelinesobject-shows-oserror-if-run-from-python-shell-but-gives))

```
(['def f1(p,r):\n', '\t"""Return f1 score from p,r"""\n', '\treturn 2*p*r/(p+r)\n'], 3)
```

So `[1]` gives line number, but it is likely zero indexed and hence we add +1 to ease readability.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/extension_function_utils.py
```diff
@@ -83,3 +85,102 @@ def ensure_all_string_literals(values: Sequence[str]) -> List[str]:
...
+
+        return False
+
+
+def _get_decorator_id(node: ast.AST) -> Optional[str]:
+    if isinstance(node, ast.Name):
+        return node.id
+    elif isinstance(node, ast.Attribute):
+        return f"{_get_decorator_id(node.value)}.{node.attr}"
+    if isinstance(node, ast.Call):
```
sfc-gh-bgoel, line 123: nit: Why is this an additional `if` instead of `elif` ?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -264,6 +267,21 @@ def collect_extension_functions(
...
+        self, py_file: Path, extension_fns: List[NativeAppExtensionFunction]
+    ):
+        with open(py_file, "r", encoding="utf-8") as f:
+            code = f.read()
+
+        if py_file.is_symlink():
+            # if the file is a symlink, make sure we don't overwrite the original
+            py_file.unlink()
+
+        new_code = deannotate(code, extension_fns)
```
sfc-gh-bgoel, line 280: nit: worth differentiating between extension_function_utils.deannotate vs this annotate? I understand that one way is to see the use of `self.deannotate` vs regular `deannotate`, but could help to differentiate for readability.

tests/nativeapp/codegen/snowpark/__snapshots__/test_extension_function_utils.ambr
```diff
@@ -1,33 +1,97 @@
...
+  '''
+  import snowflake.snowpark
+  
+  from typing import Optional
+  from snowflake.snowpark.session import Session
+  from snowflake.snowpark.functions import col, sum, sproc
+  
+  
+  @custom
+  #: @sproc(native_app_params={'schema': 'math', 'application_roles': ['app_public', 'app_admin']})
```
sfc-gh-bgoel, line 12: In case a user has a linter installed like ours where the annotation can be spread out over multiple lines, will this include a multi-line comment?

tests/nativeapp/codegen/snowpark/__snapshots__/test_extension_function_utils.ambr
sfc-gh-bgoel: +1

tests/nativeapp/codegen/snowpark/__snapshots__/test_extension_function_utils.ambr
sfc-gh-bdufour: Right! Added this case, it will work as expected, thanks for the suggestion.

tests/nativeapp/codegen/snowpark/__snapshots__/test_extension_function_utils.ambr
sfc-gh-cgorrie: Ahh ok, I missed that it was an allowlist (thought it was a denylist for some reason). Let me re-review with this context.

sfc-gh-cgorrie (approved):
LGTM. One minor question is why we have the "annotations to preserve" logic if we don't use it anywhere -- I'm guessing this is coming in a subsequent PR soon?

sfc-gh-bdufour:
> LGTM. One minor question is why we have the "annotations to preserve" logic if we don't use it anywhere -- I'm guessing this is coming in a subsequent PR soon?

Pretty much. I figured it would probably be a worthy addition to the processor's config. It should be a small change to thread it through.

PR #1104, opened 2024-05-23
Fix env merging of snowflake.yml and snowflake.local.yml

sfc-gh-melnacouzi:
### Pre-review checklist

### Changes description
Fix env merging of snowflake.yml and snowflake.local.yml
Before the change, if I have env section in both snowflake.yml and snowflake.local.yml, env of snowflake.local.yml will clobber all of the env section in snowflake.yml.
Expected behavior is to clobber only the env variables defined in snowflake.local.yml.

This PR fixes this.

### DIFF
--- a/src/snowflake/cli/api/utils/models.py
+++ b/src/snowflake/cli/api/utils/models.py
@@ -8,8 +8,8 @@ class EnvironWithDefinedDictFallback(Dict):
     def __getattr__(self, item):
         try:
             return self[item]
-        except KeyError:
-            raise AttributeError
+        except KeyError as e:
+            raise AttributeError(e)
 
     def __getitem__(self, item):
         if item in os.environ:
--- a/src/snowflake/cli/api/utils/models.py
+++ b/src/snowflake/cli/api/utils/models.py
@@ -1,9 +1,15 @@
+from __future__ import annotations
+
 import os
+from typing import Any, Dict
 
 
-class EnvironWithDefinedDictFallback(dict):
+class EnvironWithDefinedDictFallback(Dict):
     def __getattr__(self, item):
-        return self[item]
+        try:
+            return self[item]
+        except KeyError:
+            raise AttributeError
 
     def __getitem__(self, item):
         if item in os.environ:
@@ -12,3 +18,6 @@ class EnvironWithDefinedDictFallback(dict):
 
     def __contains__(self, item):
         return item in os.environ or super().__contains__(item)
+
+    def update_from_dict(self, update_values: Dict[str, Any]):
+        return super().update(update_values)
--- a/tests/test_sql.py
+++ b/tests/test_sql.py
@@ -337,6 +337,17 @@ def test_uses_variables_from_snowflake_yml(
 
 
 @mock.patch("snowflake.cli.plugins.sql.commands.SqlManager._execute_string")
+def test_uses_variables_from_snowflake_local_yml(
+    mock_execute_query, project_directory, runner
+):
+    with project_directory("sql_templating"):
+        result = runner.invoke(["sql", "-q", "select &{ ctx.env.sf_var_override }"])
+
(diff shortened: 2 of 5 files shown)


sfc-gh-turbaszek (approved):

src/snowflake/cli/api/utils/models.py
```diff
@@ -1,9 +1,15 @@
...
 
 
-class EnvironWithDefinedDictFallback(dict):
+class EnvironWithDefinedDictFallback(Dict):
     def __getattr__(self, item):
-        return self[item]
+        try:
+            return self[item]
+        except KeyError:
+            raise AttributeError
```
sfc-gh-bdufour, line 12: Shouldn't we say something about which attribute it was? Python itself does something like '<object> has no atttribute <name>'

src/snowflake/cli/api/utils/models.py
```diff
@@ -1,9 +1,15 @@
+from __future__ import annotations
+
 import os
+from typing import Any, Dict
 
 
-class EnvironWithDefinedDictFallback(dict):
+class EnvironWithDefinedDictFallback(Dict):
```
sfc-gh-bdufour, line 7: why? `dict` refers to the built-in dictionary class, it seemed fine to me before.

src/snowflake/cli/api/utils/models.py
```diff
@@ -12,3 +18,6 @@ def __getitem__(self, item):
 
     def __contains__(self, item):
         return item in os.environ or super().__contains__(item)
+
+    def update_from_dict(self, update_values: Dict[str, Any]):
```
sfc-gh-bdufour, line 22: the original intention was to limit this to `UpdatableModel`. Should this inherit from it? Not quite sure how this change fits in the bigger picture.

tests/test_sql.py
```diff
@@ -336,6 +336,17 @@ def test_uses_variables_from_snowflake_yml(
...
 
+@mock.patch("snowflake.cli.plugins.sql.commands.SqlManager._execute_string")
+def test_uses_variables_from_snowflake_local_yml(
+    mock_execute_query, project_directory, runner
+):
+    with project_directory("sql_templating"):
+        result = runner.invoke(["sql", "-q", "select &{ ctx.env.sf_var_override }"])
+
+    assert result.exit_code == 0
+    mock_execute_query.assert_called_once_with("select foo_value_override")
```
sfc-gh-bdufour, line 347: It would be nice to check that sf_var is still defined too? Or is that covered elsewhere?

src/snowflake/cli/api/utils/models.py
sfc-gh-melnacouzi: The signature of update_from_dict was using it, so I thought it makes sense to be consistent. If that's a problem, I can change it back.

src/snowflake/cli/api/utils/models.py
sfc-gh-melnacouzi: AttributeError only takes an error as input. That will at least help identify what attribute value it is. does this address your concern?
for now, it's only used with hasattr() which only returns false if there is attributerror.

src/snowflake/cli/api/utils/models.py
sfc-gh-melnacouzi: No. I tried to inherit from both UpdatableModel and dict, but they conflict. 
The way it's currently done, we use type Dict when parsing yml, which is recognized by Pydantic, and then transform to EnvironWithDefinedDictFallback during validation. From that moment on, Pydantic is not involved anymore, and the resulting types are only used for update_from_dict, which is not Pydantic specific (we added it to all Pydantic model in order not to repeat that).

sfc-gh-bdufour (approved):

sfc-gh-turbaszek (approved):

PR #1124, opened 2024-05-27
Clean up output of the setup script generation step

sfc-gh-bdufour:
### Pre-review checklist

### Changes description

- Remove debugging output for setup script generation, replaced with a proper phase + steps output


tests/nativeapp/codegen/snowpark/test_python_processor.py
```diff
@@ -25,6 +28,56 @@
...
+        return f"d {p}"
+    else:
+        return f"f {p}"
+
+
+def all_paths_under_dir(root: Path) -> List[Path]:
+    assert root.is_dir()
+
+    paths: Set[Path] = set()
+    for subdir, dirs, files in os.walk(root):
```
sfc-gh-cgorrie, line 56: `os.walk` does not have a guaranteed order, so we should sort `dirs` and `files` to ensure the output is stable.

sfc-gh-cgorrie (commented):
Just one note. It's possible the os.walk stuff will have a different order on Windows / other systems so I think it's best to sort the returned dirs / files so we ensure the snapshots match.

tests/nativeapp/codegen/snowpark/test_python_processor.py
sfc-gh-bdufour: Absolutely, that happens below, the entire output list is sorted.

sfc-gh-cgorrie: Lol 🤦

sfc-gh-cgorrie (approved):

src/snowflake/cli/plugins/nativeapp/codegen/compiler.py
```diff
@@ -54,19 +55,28 @@ def compile_artifacts(self):
...
-                    processor_mapping=processor,
-                )
-                if artifact_processor is None:
-                    raise UnsupportedArtifactProcessorError(
-                        processor_name=processor.name
-                    )
-                else:
-                    artifact_processor.process(
-                        artifact_to_process=artifact, processor_mapping=processor
+            if artifact.processors:
```
sfc-gh-bgoel, line 60: Do we need the extra `for` loop? I understand that this is not really a long list of artifacts, but wondering if we can combine it with something like this:

```
for artifact in self.artifacts:
     if not artifact.processors:
          continue
     with cc.phase("Invoking artifact processors"
     for processor in artifact.processors:
          ....
```

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -173,7 +172,7 @@ def process(
         artifact_to_process: PathMapping,
         processor_mapping: Optional[ProcessorMapping],
         **kwargs,
-    ) -> str:  # String output is temporary until we have better e2e testing mechanism
+    ) -> None:
```
sfc-gh-bgoel, line 175: 👍

sfc-gh-bgoel (approved):
Left a suggestion, not a blocker.

src/snowflake/cli/plugins/nativeapp/codegen/compiler.py
sfc-gh-bdufour: If we have a start/end API we could use instead of the context manager API, it would be easy. Sadly, we do not.

PR #1134, opened 2024-05-28
Add templating for project definition file available for all modules

sfc-gh-melnacouzi:
### Pre-review checklist
   * [x] I've added or updated integration tests to verify correctness of my new code. (Will add these separately)

### Changes description
* Added templating support for project definition file.
  * Templates can now be used within the main section of the project definition file.
  * Resolved values of the project definition file are available to all modules.


src/snowflake/cli/api/commands/decorators.py
```diff
@@ -67,7 +68,9 @@ def _decorator(func: Callable):
                     "project_definition",
                     inspect.Parameter.KEYWORD_ONLY,
                     annotation=Optional[str],
-                    default=project_type_option(project_name),
+                    default=project_type_option(project_name)
+                    if project_name
+                    else project_definition_option(),
```
sfc-gh-cgorrie, line 73: Is this a short-term fix? It seems to me like we should just make `project_type_option` take an `Optional[str]` as well, I'm not sure what having two paths gets us with such similar names / purposes.

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,191 @@
...
+from snowflake.cli.api.utils.rendering import get_snowflake_cli_jinja_env
+
+
+@dataclass
+class GraphNode:
+    path: str
+    value: str | None = None
+
+
+def _get_referenced_vars(ast_node, attr_chain: List[str] = []) -> List[List[str]]:
```
sfc-gh-cgorrie, line 20: I'm having trouble visualizing what this type represents (`List[List[str]]`). Can you provide a type alias?

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,191 @@
...
+def _check_for_cycles(original_graph):
+    graph = copy.deepcopy(original_graph)
+    while len(graph) > 0:
+        node = _find_node_with_no_dependencies(graph)
+        if node == None:
+            raise RecursionError("Cycle detected in project definition file template")
+        _remove_node_from_graph(graph, node)
+
+
+def _build_dependency_graph(env, referenced_vars, context_without_env):
```
sfc-gh-cgorrie, line 83: Consider https://docs.python.org/3/library/graphlib.html#graphlib.TopologicalSorter

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -133,7 +133,7 @@ def app_bundle(
 
 
 @app.command("run", requires_connection=True)
-@with_project_definition("native_app")
+@with_project_definition()
```
sfc-gh-cgorrie, line 136: Do we still get the behaviour that the command exits with a fatal error if the `native_app` section doesn't exist?

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,191 @@
...
+
+def _get_value_from_var_path(env: Environment, context: dict, var_path: str):
+    # given a variable path (e.g. ctx.env.test), return evaluated value based on context
+    # fall back to env variables and escape them so we stop the chain of variables
+    ref_str = "<% " + var_path + " %>"
+    template = env.from_string(ref_str)
+    try:
+        return template.render(context)
+    except UndefinedError as e:
+        # check in env variables
```
sfc-gh-cgorrie, line 53: Is it possible / preferable to do this ahead-of-time? I'd expect whatever calls this function to already be populating `context` with the env vars.

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+def _find_node_with_no_dependencies(graph):
+    for node, deps in graph.items():
+        if len(deps) == 0:
+            return node
+    return None
+
+
+def _remove_node_from_graph(graph, node):
```
sfc-gh-cgorrie, line 67: Worth adding a docstring that explains this deletes a leaf and all edges that point to it, or maybe tweak the name a bit.

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,191 @@
...
+        attr_chain = [getattr(ast_node, "attr")] + attr_chain
+    elif isinstance(ast_node, nodes.Name):
+        all_referenced_vars = [[getattr(ast_node, "name")] + attr_chain]
+
+    for child_node in ast_node.iter_child_nodes():
+        all_referenced_vars = all_referenced_vars + _get_referenced_vars(
+            child_node, attr_chain
+        )
+
+    return all_referenced_vars
```
sfc-gh-cgorrie, line 35: Is each var a string (e.g. `ctx.env.abc`)?

src/snowflake/cli/api/commands/decorators.py
sfc-gh-turbaszek: Imho the `project_name` option does not create enough value to have it. My take would be that this decorator should provide full project definition. If command requires a part of it (for example streamlit) then the command should access it.

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+def _get_referenced_vars_from_str(env: Environment, template_str: str) -> List[str]:
```
sfc-gh-turbaszek, line 38: How about either making those methods part of custom `Environment` class or introducing a processor class?

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,191 @@
...
+    # given a variable path (e.g. ctx.env.test), return evaluated value based on context
+    # fall back to env variables and escape them so we stop the chain of variables
+    ref_str = "<% " + var_path + " %>"
+    template = env.from_string(ref_str)
+    try:
+        return template.render(context)
+    except UndefinedError as e:
+        # check in env variables
+        try:
+            return f"{env.block_start_string} raw {env.block_end_string}{template.render({'ctx': {'env': os.environ}})}{env.block_start_string} endraw {env.block_end_string}"
```
sfc-gh-turbaszek, line 55: I don't know what's happening there

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-turbaszek: To populate the context with env variables we need to know for what variables we are looking for. This excludes env-only variables The solution was having a custom dict that joins values with os.environ

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-melnacouzi: yes

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-melnacouzi: when evaluating environment variables coming from os, we do not want to expand them further, which is why i am using {% raw %} {% endraw %} templates. I don't like that though. I will see if I can do better.

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-melnacouzi: I will change this to DFS, it seems simpler and more efficient.

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,191 @@
...
+
+def _remove_node_from_graph(graph, node):
+    del graph[node]
+    for n, deps in graph.items():
+        if node in deps:
+            deps.remove(node)
+
+
+def _check_for_cycles(original_graph):
+    graph = copy.deepcopy(original_graph)
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: will change this to DFS and combine it with evaluation.

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,191 @@
...
+
+    value = _get_value_from_var_path(jinja_env, context_with_env, node.path)
+    template = jinja_env.from_string(value)
+    node.value = template.render(my_context)
+    return node.value
+
+
+def render_project_template(project):
+    jinja_env = get_snowflake_cli_jinja_env()
+    context_without_env = {"ctx": project.model_dump(exclude_unset=True)}
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: This contains the model without the classes, which does not use EnvironWithDefinedDictFallback, and therefore, use the env values from the pdf directly. I will simplify the logic to only use one of the 2 contexts.

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,191 @@
...
+    context_with_env = {"ctx": project}
+    combined_pdf_as_str = yaml.dump(project.model_dump(exclude_unset=True))
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: yes - it's the same - we will use this as the jinja template to be evaluated at the end.

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,191 @@
...
+    # env vars are used as backup as leaf nodes (just for graph purposes)
+    dependencies_graph = _build_dependency_graph(
+        jinja_env, referenced_vars, context_without_env
+    )
+    _check_for_cycles(dependencies_graph)
+
+    # store extra node information (var_path to GraphNode object map)
+    graph_nodes_map = {}
+    for path in dependencies_graph.keys():
+        graph_nodes_map[path] = GraphNode(path=path)
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: dependencies_graph is the graph. graph_nodes_map is only for the metadata of the nodes.  I tried using __hash__ and __eq__ in Graph and combining them, but I found the logic getting more verbose (unless you still have key to node mapping)

src/snowflake/cli/plugins/nativeapp/commands.py
sfc-gh-melnacouzi: good point - I will revert this logic, but keep this no-arg option available for future commands that might need it.

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-fcampbell, line 154:
```suggestion
    combined_pdf_as_str = yaml.safe_dump(project.model_dump(exclude_unset=True))
```
?

src/snowflake/cli/api/utils/graph.py
```diff
@@ -0,0 +1,61 @@
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+
+
+class Graph:
+    @dataclass
+    class Node:
+        key: str
+        status: str | None = None
```
sfc-gh-melnacouzi, line 10: this should be _status

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,163 @@
...
+        variable_appended = True
+        all_referenced_vars.add(Variable(variable_attr_chain))
+
+    for child_node in ast_node.iter_child_nodes():
+        all_referenced_vars.update(
+            _get_referenced_vars(child_node, variable_attr_chain)
+        )
+
+    if variable_appended:
+        variable_attr_chain.popleft()
```
sfc-gh-bdufour, line 79: [continued] ah, so I guess this chain manipulation that confused me is intended for bookkeeping during the recursive call only, and here you undo the change before returning? This code needs, at the very least, some comments to explain what's going on.

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,163 @@
...
+    # Var nodes end in Name node, and starts with Getattr node.
+    # Example: ctx.env.test will look like Getattr(test) -> Getattr(env) -> Name(ctx)
+    all_referenced_vars = set()
+
+    variable_appended = False
+    if isinstance(ast_node, nodes.Getattr):
+        variable_attr_chain.appendleft(getattr(ast_node, "attr"))
+        variable_appended = True
+    elif isinstance(ast_node, nodes.Name):
+        variable_attr_chain.appendleft(getattr(ast_node, "name"))
```
sfc-gh-bdufour, line 69: why not just ast_node.name? The `getattr` call is usually necessary only when you don't know the name a priori, but here we're using it with a constant.

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-bdufour: They're not stored as strings though, they're `Variable` instances (hint hint, we need type annotations throughout)

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,163 @@
...
+        return self.get_key() == other.get_key()
+
+
+@dataclass(eq=False)
+class TemplateGraphNode(Graph.Node):
+    templated_value: str | None = None
+    rendered_value: str | None = None
+    variable: Variable | None = None
+
+    def store_in_context(self, context):
```
sfc-gh-bdufour, line 38: this whole logic feels a bit complex for what it is. If variable `ctx.env.foo` needs to be stored, then we just need loop through all the names and getOrCreate a dictionary for each (python has that primitive in dict, it's called `setdefault`), and insert the value on the last step. Should be a simple loop. Am I missing anything?

src/snowflake/cli/api/utils/graph.py
```diff
@@ -0,0 +1,61 @@
...
+        dependencies: set[Graph.Node] = field(default_factory=set)
```
sfc-gh-bdufour, line 11: nit: edges? I think that's the standard term. Dependencies is your use for this graph.

src/snowflake/cli/api/utils/graph.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def get_all_nodes(self):
+        return self._graph_nodes_map.values()
+
+    def add(self, node: Node):
+        if node.key in self._graph_nodes_map:
+            raise KeyError(f"Node key {node.key} already exists")
+        self._graph_nodes_map[node.key] = node
+
+    def add_dependency(self, key1: str, key2: str):
```
sfc-gh-bdufour, line 35: nit: this is a directed graph, 1 and 2 are unclear, rename to include the direction of the edge (some kind of from/to naming)

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-bdufour: I expected the merge to happen explicitly too. I'm quite confused but the logic here. It looks like it tries to resolve variables with only env vars initially, or something like that? Maybe this works, but I can't quite convince myself it does, and in any case it feels rather counter-intuitive. I'd prefer to find a clearer way to express what's going on. Happy to discuss offline.

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-melnacouzi: mypy was complaining about name not being defined, so I was forced to use that.

sfc-gh-bdufour: Would something like `# type: ignore` help then (or type hints)? In cases where the tooling fails us badly, then we tend to prefer disabling the tooling over crippling the code.

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-fcampbell: can also just make them all defaultdicts: https://stackoverflow.com/a/8702435

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,163 @@
...
+def _get_referenced_vars(
+    ast_node, variable_attr_chain: deque = deque()
+) -> set[Variable]:
+    # Var nodes end in Name node, and starts with Getattr node.
+    # Example: ctx.env.test will look like Getattr(test) -> Getattr(env) -> Name(ctx)
+    all_referenced_vars = set()
+
+    variable_appended = False
+    if isinstance(ast_node, nodes.Getattr):
+        variable_attr_chain.appendleft(getattr(ast_node, "attr"))
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: wow, that's interesting! Default is probably not needed here.
I modified this according to a previous comment to optimize and pass the list and append on the go, but I think I will revert back to the previous version of this function as it was cleaner.

RELEASE-NOTES.md
```diff
@@ -8,6 +8,9 @@
 * Added `snow notebook` commands:
   * `snow notebook execute` enabling head-less execution of a notebook.
   * `snow notebook create` proving an option to create a Snowflake Notebook from a file on stage.
+* Added templating support for project definition file.
+  * Templates can now be used within the main section of the project definition file.
+  * Resolved values of the project definition file are available to all modules.
```
sfc-gh-cgorrie, line 13: What is a module here? The different `native_app:` / `streamlit:` / etc. top-level collections?

RELEASE-NOTES.md
```diff
@@ -8,6 +8,9 @@
 * Added `snow notebook` commands:
   * `snow notebook execute` enabling head-less execution of a notebook.
   * `snow notebook create` proving an option to create a Snowflake Notebook from a file on stage.
+* Added templating support for project definition file.
+  * Templates can now be used within the main section of the project definition file.
```
sfc-gh-cgorrie, line 12: How can a customer know what the main section of `snowflake.yml` is?

src/snowflake/cli/api/project/definition_manager.py
```diff
@@ -100,6 +101,16 @@ def _user_definition_file_if_available(project_path: Path) -> Optional[Path]:
             DefinitionManager.USER_DEFINITION_FILENAME, project_path
         )
 
+    @functools.cached_property
+    def _project_properties(self) -> ProjectProperties:
+        return load_project(self._project_config_paths)
+
     @functools.cached_property
```
sfc-gh-cgorrie, line 108: Does this need to be cached, given the only value it depends on is cached?

src/snowflake/cli/api/project/definition_manager.py
```diff
@@ -100,6 +101,16 @@ def _user_definition_file_if_available(project_path: Path) -> Optional[Path]:
...
     def project_definition(self) -> ProjectDefinition:
-        return load_project_definition(self._project_config_paths)
+        return self._project_properties.project_definition
+
+    @functools.cached_property
+    def template_context(self) -> dict:
+        definition = self._project_properties.raw_project_definition
```
sfc-gh-cgorrie, line 114: Does this mean we can use e.g. `<% ctx.native_app.name %>`? Are we documenting this or is it a "hidden feature"?

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,196 @@
...
+class Variable:
+    def __init__(self, vars_chain):
+        self._vars_chain = list(vars_chain)
+        self.templated_value = None
+        self.rendered_value = None
+
+    def get_key(self):
+        return ".".join(self._vars_chain)
+
+    def is_env_var(self):
```
sfc-gh-cgorrie, line 21: Suggest that this is a property

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,196 @@
...
+        )
+
+    def get_env_var_name(self) -> str:
+        if not self.is_env_var():
+            raise KeyError(
+                f"Referenced variable {self.get_key()} is not an environment variable"
+            )
+        return self._vars_chain[2]
+
+    def store_in_context(self, context: dict, value):
```
sfc-gh-cgorrie, line 35:
```suggestion
    def store_in_context(self, context: dict, value: typing.Any) -> None:
```

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,196 @@
...
+        """
+        Takes a generic context dict to modify, and a value
+
+        Traverse through the multi-level dictionary to the location where this variables goes to.
+        Sets this location to the content of value.
```
sfc-gh-cgorrie, line 40: We should mention that this mutates the object rather than returning a new one, even though it's obvious from the signature.

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,196 @@
...
+
+def _build_dependency_graph(
+    env: Environment, all_vars: set[Variable], context
+) -> Graph[Variable]:
+    dependencies_graph = Graph[Variable]()
+    for variable in all_vars:
+        dependencies_graph.add(Node[Variable](key=variable.get_key(), data=variable))
+
+    for variable in all_vars:
+        if variable.is_env_var() and variable.get_env_var_name() in os.environ:
```
sfc-gh-cgorrie, line 120: Nit: I feel like any sort of env dict should be able to be passed in, rather than having to use `os.environ`. Maybe a future refactor when we need it, but it could also simplify testing.

src/snowflake/cli/api/utils/graph.py
```diff
@@ -0,0 +1,68 @@
...
+
+T = TypeVar("T")
+
+
+@dataclass
+class Node(Generic[T]):
+    key: str
+    data: T
+    neighbors: set[Node[T]] = field(default_factory=set)
+    status: Optional[str] = None
```
sfc-gh-cgorrie, line 16: Suggest that you use an Enum instead of a raw string, and instead of None being the default have it be "UNVISITED".

src/snowflake/cli/api/utils/dict_utils.py
```diff
@@ -0,0 +1,53 @@
...
+    Takes 2 dictionaries as input: original and override.
+
+    For every key in the override dictionary, override the same key
+    in the original dictionary, or create a new one if it doesn't exist.
+
+    If the override value and the original value are both dictionaries,
+    instead of overriding, recursively call this function to merge the keys of the sub-dictionaries.
+    """
+    if not isinstance(override_values, dict) or not isinstance(original_values, dict):
+        return
```
sfc-gh-cgorrie, line 15: Raise a `ValueError` instead of returning immediately.

src/snowflake/cli/api/utils/dict_utils.py
```diff
@@ -0,0 +1,53 @@
+from __future__ import annotations
+
+
+def deep_merge_dicts(original_values: dict, override_values: dict):
```
sfc-gh-cgorrie, line 4:
```suggestion
def deep_merge_dicts(original_values: dict, override_values: dict) -> None:
```

src/snowflake/cli/api/utils/dict_utils.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    Takes 2 dictionaries as input: original and override.
+
+    For every key in the override dictionary, override the same key
+    in the original dictionary, or create a new one if it doesn't exist.
```
sfc-gh-cgorrie, line 9:
```suggestion
    in the original dictionary, or create a new one if the key is not present.
```

src/snowflake/cli/api/utils/dict_utils.py
```diff
@@ -0,0 +1,53 @@
...
+
+    If the override value and the original value are both dictionaries,
+    instead of overriding, recursively call this function to merge the keys of the sub-dictionaries.
```
sfc-gh-cgorrie, line 12:
```suggestion
    instead of overriding, this function recursively calls itself to merge the keys of the sub-dictionaries.
```

src/snowflake/cli/api/utils/dict_utils.py
```diff
@@ -0,0 +1,53 @@
+from __future__ import annotations
+
+
+def deep_merge_dicts(original_values: dict, override_values: dict):
+    """
+    Takes 2 dictionaries as input: original and override.
```
sfc-gh-cgorrie, line 6:
```suggestion
    Takes 2 dictionaries as input: original and override. The original dictionary is modified.
```

sfc-gh-cgorrie (commented):
Added some thoughts.

src/snowflake/cli/api/project/definition_manager.py
sfc-gh-turbaszek: I think we should be documenting it

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,196 @@
...
+
+        Traverse through the multi-level dictionary to the location where this variables goes to.
+        Sets this location to the content of value.
+
+        Example: vars chain contains ['ctx', 'env', 'x'], and context is {}, and value is 'val'.
+        At the end of this call, context content will be: {'ctx': {'env': {'x': 'val'}}}
+        """
+        current_dict_level = context
+        for i, var in enumerate(self._vars_chain):
+            if i == len(self._vars_chain) - 1:
```
sfc-gh-turbaszek, line 47:
```suggestion
            if i == chain_len:
```
Consider introducing variable

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,196 @@
...
+                current_dict_level[var] = value
+            else:
+                current_dict_level.setdefault(var, {})
+                current_dict_level = current_dict_level[var]
+
+    def read_from_context(self, context):
```
sfc-gh-turbaszek, line 53:
```suggestion
    def read_from_context(self, context: dict):
```
Should consider defining `Context = Dict[str, Any]` so we can use such typing in many places?

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,196 @@
...
+                not isinstance(current_dict_level, dict)
+                or key not in current_dict_level
+            ):
+                raise UndefinedError(
+                    f"Could not find template variable {self.get_key()}"
+                )
+            current_dict_level = current_dict_level[key]
+
+        value = current_dict_level
+        if value == None or isinstance(value, dict) or isinstance(value, list):
```
sfc-gh-turbaszek, line 74:
```suggestion
        if value is None or isinstance(value, (dict, list)):
```

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,196 @@
...
+    ast = env.parse(template_str)
+    return _get_referenced_vars(ast)
+
+
+def _build_dependency_graph(
+    env: Environment, all_vars: set[Variable], context
+) -> Graph[Variable]:
+    dependencies_graph = Graph[Variable]()
+    for variable in all_vars:
+        dependencies_graph.add(Node[Variable](key=variable.get_key(), data=variable))
```
sfc-gh-turbaszek, line 117: What is `Node[Variable]`?

src/snowflake/cli/api/project/definition_manager.py
sfc-gh-melnacouzi: It does not need. I would like to keep it in case the implementation changes in the future, and to document for users that this is cached (without having to read internal functions)

src/snowflake/cli/api/project/definition_manager.py
sfc-gh-melnacouzi: we can use this in the project file, and yes, we will be documenting it.

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-melnacouzi: removed yaml from the logic.

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,163 @@
...
+    # now that we determined value of all referenced vars, use these resolved values as a fresh context to resolve project file later
+    final_context = {}
+    node: TemplateGraphNode
+    for node in dependencies_graph.get_all_nodes():
+        node.store_in_context(final_context)
+
+    # resolve combined project file based on final context
+    template = jinja_env.from_string(pdf_yaml_str)
+    rendered_template = template.render(final_context)
+    rendered_definition = yaml.load(rendered_template, Loader=yaml.loader.BaseLoader)
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: as discussed in slack, most people (You, cam) are against it, and I think it makes sense to disallow it.
We can always allow it back if customer request it.

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-melnacouzi: updated - should be cleaner now

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,196 @@
...
+
+    dependencies_graph = _build_dependency_graph(
+        jinja_env, referenced_vars, pdf_context
+    )
+
+    dependencies_graph.dfs(
+        visit_action=lambda node: _render_graph_node(jinja_env, node)
+    )
+
+    # now that we determined the values of all tempalted vars,
```
sfc-gh-bdufour, line 186: typo: templated

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+def _build_dependency_graph(
+    env: Environment, all_vars: set[Variable], context
+) -> Graph[Variable]:
+    dependencies_graph = Graph[Variable]()
+    for variable in all_vars:
+        dependencies_graph.add(Node[Variable](key=variable.get_key(), data=variable))
+
+    for variable in all_vars:
```
sfc-gh-bdufour, line 121: this way to do overrides is cleaner than the prev version, so thanks for that, it's much easier to reason about.

tests/api/utils/test_dict_utils.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def test_merge_dicts_empty():
+    test_dict = {}
+    deep_merge_dicts(test_dict, {})
+    assert test_dict == {}
+
+
+def test_merge_dicts_recursive_map():
+    test_dict = {"a": "a1", "b": "b1", "c": {"d": "d1", "e": "e1"}}
```
sfc-gh-bdufour, line 11: nit: woah that's hard to read. Guessing the formatted messed with what would otherwise be a very clear test. How about cheating then and parsing the dict from a multi-line JSON structure? It's really hard to do the merge in my head based on the single-line structures here.

tests/api/utils/test_dict_utils.py
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    visited_elements = []
+
+    def visit_action(element):
+        visited_elements.append(element)
+
+    deep_traverse(test_struct, visit_action)
+
+    assert visited_elements == [
```
sfc-gh-bdufour, line 42: just to confirm, the visiting order is part of the contract, yes?

tests/api/utils/test_graph.py
```diff
@@ -0,0 +1,55 @@
+from __future__ import annotations
+
+import pytest
+from snowflake.cli.api.utils.graph import Graph, Node
```
sfc-gh-bdufour, line 4: hmm, what's the coverage of the graph code here? Looks like you missed a few parts of the contract.

tests/api/utils/test_definition_rendering.py
```diff
@@ -0,0 +1,303 @@
...
+    definition = {
+        "definition_version": "1.1",
+        "env": env,
+    }
+    with pytest.raises(UndefinedError) as err:
+        render_definition_template(definition)
+    assert msg in str(err.value)
+
+
+@mock.patch.dict(os.environ, {}, clear=True)
```
sfc-gh-bdufour, line 276: can we also add a test that clearly documents what happens when one tries to inject complex YAML through a variable? e.g. can I add artifact mappings through env vars, etc? We either allow it or we don't, but it should be documented as it is part of the implied contract here.

tests/api/utils/test_dict_utils.py
sfc-gh-melnacouzi: yea - it's very annoying. Let me see if there is a way to skip the formatting on this, otherwise, I'll use Json as you suggested.

src/snowflake/cli/api/commands/flags.py
```diff
@@ -499,6 +501,7 @@ def _callback(project_path: Optional[str]):
 
         cli_context_manager.set_project_definition(project_definition)
         cli_context_manager.set_project_root(project_root)
+        cli_context_manager.set_template_context(dm.template_context)
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: it's currently used for SQL expansion, and could be used for expanding templates in other types of files in the future.

src/snowflake/cli/api/project/definition.py
```diff
@@ -14,32 +15,47 @@
...
-            project.update_from_dict(overrides)
+            deep_merge_dicts(definition, overrides)
 
-    return project
+    return definition
+
+
+def load_project(paths: List[Path]) -> ProjectProperties:
+    """
+    Loads project definition, optionally overriding values. Definition values
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: That was the original text for loading project definition. I believe it means snowflake.local.yml optional overrides.

tests/api/utils/test_definition_rendering.py
sfc-gh-melnacouzi: added a test. We just treat the env as a string. We never yaml expand it.

tests/api/utils/test_graph.py
sfc-gh-melnacouzi: added more coverage.

tests/api/utils/test_dict_utils.py
sfc-gh-melnacouzi: it doesn't have to be in the future. I changed this check to ignore the order.

tests/api/utils/test_dict_utils.py
sfc-gh-melnacouzi: Can be avoided by using # fmt: off / #fmt: on

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,196 @@
...
+        Example: vars chain contains ['ctx', 'env', 'x'], and context is {}, and value is 'val'.
+        At the end of this call, context content will be: {'ctx': {'env': {'x': 'val'}}}
+        """
+        current_dict_level = context
+        for i, var in enumerate(self._vars_chain):
+            if i == len(self._vars_chain) - 1:
+                current_dict_level[var] = value
+            else:
+                current_dict_level.setdefault(var, {})
+                current_dict_level = current_dict_level[var]
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: if you don't mind, I think it's more readable this way.

src/snowflake/cli/api/utils/graph.py
```diff
@@ -0,0 +1,68 @@
...
+        from_node.neighbors.add(to_node)
+
+    def _dfs_visit(self, node: Node[T], visit_action):
+        if node.status == "VISITED":
+            return
+
+        node.status = "VISITING"
+        for neighbour_node in node.neighbors:
+            if neighbour_node.status == "VISITING":
+                raise ClickException("Cycle detected")
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: cycle detection can be part of a general-purpose graph, but it shouldn't be enforced (no cycles) by default. I tried different things, and optional cycle detection/action seems to be the cleanest so far.

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-melnacouzi: I will keep it as is for now - I have the another task that might involve this kind of refactor, so I will try to do that in there.

src/snowflake/cli/api/project/definition_manager.py
sfc-gh-cgorrie: Given it's likely a small-ish dict, I'm ok with the memory duplication here in favour of a stable contract.

src/snowflake/cli/api/exceptions.py
```diff
@@ -16,6 +16,14 @@ class MissingConfiguration(ClickException):
     pass
 
 
+class CycleDetectedError(ClickException):
```
sfc-gh-bdufour, line 19: nit: feels a little out of place here, in the sense that this exception is rather specific to one part of the logic, but I defer to the CLI team ultimately for this one.

src/snowflake/cli/api/utils/definition_rendering.py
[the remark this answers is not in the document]
sfc-gh-bdufour: as you wish

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+def render_definition_template(definition: dict[str, Any]) -> dict[str, Any]:
+    if "definition_version" not in definition or Version(
+        definition["definition_version"]
+    ) < Version("1.1"):
+        return definition
+
+    jinja_env = get_snowflake_cli_jinja_env()
+    pdf_context = {CONTEXT_KEY: definition}
```
sfc-gh-bdufour, line 204: nit: `project_context`? PDF is not a well-known acronym outside our immediate team.

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,237 @@
...
+        referenced_vars.update(_get_referenced_vars_from_str(jinja_env, element))
+
+    traverse(definition, visit_action=find_any_template_vars)
+
+    dependencies_graph = _build_dependency_graph(
+        jinja_env, referenced_vars, pdf_context
+    )
+
+    def on_cycle_action():
+        raise CycleDetectedError("Cycle detected in templating variables")
```
sfc-gh-bdufour, line 218: it would be useful for a user to know where the cycle is. If even one of the variables is part of the output, they'd know where to start looking. This is a bit too opaque.

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,237 @@
...
+    # use these resolved values as a fresh context to resolve definition
+    final_context: dict = {}
+    for node in dependencies_graph.get_all_nodes():
+        node.data.add_to_context(final_context)
+
+    traverse(
+        definition,
+        update_action=lambda val: _render_dict_element(jinja_env, final_context, val),
+    )
+    deep_merge_dicts(definition, {"env": dict(os.environ)})
```
sfc-gh-bdufour, line 235: this is modifying the input param, then returning it. In general we should really avoid modifying input params. The signature for this method is misleading, you see in: dict, out: dict, and you don't assume it will modify the input. At the very least we should document that in a docstring (this is a public API for this module essentially, so it should very much have a docstring). However, I think the general pattern we're using here is somewhat risky from a maintenance POV.

src/snowflake/cli/api/utils/graph.py
```diff
@@ -0,0 +1,79 @@
...
+        from_node = self.get(from_node_key)
+        to_node = self.get(to_node_key)
+        from_node.neighbors.add(to_node)
+
+    @staticmethod
+    def _dfs_visit(
+        nodes_status: dict[str, VisitStatus],
+        node: Node[T],
+        visit_action,
+        on_cycle_action,
```
sfc-gh-bdufour, line 55: these are both (required) callable params, let's add type annotations. Same for the dfs entry point.

sfc-gh-bdufour (commented):
Looks good, but needs integ testing and likely at least one entry in the manual test plan before being queued up for release.

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-melnacouzi: the return type of this is None. It also has a docstring explaining it.

sfc-gh-bdufour: I was talking about `render_definition_template`

sfc-gh-melnacouzi: I thought about cloning it, but then I run about the risk of missing something from the original. Let me know what you think.

sfc-gh-melnacouzi: actually copy.deepcopy might do just that. Let me try it out.

tests/api/utils/test_definition_rendering.py
```diff
@@ -185,7 +185,7 @@ def test_resolve_variables_error_on_cycle(definition):
     with pytest.raises(CycleDetectedError) as err:
         render_definition_template(definition)
 
-    assert err.value.message == f"Cycle detected in templating variables"
+    assert err.value.message.startswith("Cycle detected in templating variable ")
```
sfc-gh-bdufour, line 188: Is it not deterministic?

sfc-gh-melnacouzi: No. Depends on which node you start with in DFS.

sfc-gh-bdufour (approved):

src/snowflake/cli/api/exceptions.py
sfc-gh-turbaszek: If usage of the exception is self-contained to a module / package I would move the exception there.

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,251 @@
...
+
+def _get_referenced_vars_from_str(
+    env: Environment, template_str: str
+) -> set[TemplateVar]:
+    ast = env.parse(template_str)
+    return _get_referenced_vars(ast, template_str)
+
+
+def _build_dependency_graph(
+    env: Environment, all_vars: set[TemplateVar], context: dict[str, Any]
```
sfc-gh-turbaszek, line 144:
```suggestion
    env: Environment, all_vars: set[TemplateVar], context: Context
```
Where `Context = dict[str, Any]`

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -0,0 +1,251 @@
...
+    for child_node in ast_node.iter_child_nodes():
+        all_referenced_vars.update(
+            _get_referenced_vars(child_node, template_value, current_attr_chain)
+        )
+
+    return all_referenced_vars
+
+
+def _get_referenced_vars_from_str(
+    env: Environment, template_str: str
```
sfc-gh-turbaszek, line 137: A lot of those method require env as input. Would it make sense to implement `class TemplatedEnvironment(Environment)`  that would encapsulate the logic

sfc-gh-turbaszek (dismissed):

sfc-gh-bdufour (approved):

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-bdufour: update: it worked, resolving.

sfc-gh-turbaszek (approved):

PR #1138, opened 2024-05-28
Correctly handle read-only files when cleaning up in `snow app list-templates`

sfc-gh-bdufour:
### Pre-review checklist

### Changes description

SNOW-1449514 - `snow app list-templates` fails on Windows due to permission errors. This change corrects the handling of the temporary git repo used by this command and fixes the bug.


tests_integration/nativeapp/test_templates.py
```diff
@@ -1,12 +1,12 @@
...
+def test_list_templates_no_options_success(runner):
     args = ["app", "list-templates"]
-    result = runner.invoke(args)
+    result = runner.invoke_json(args)
 
     assert result.exit_code == 0
-    assert result.output == snapshot
+    templates = result.json
+    assert len(templates) > 0
+    assert "basic" in [t["template"] for t in templates]
```
sfc-gh-bdufour, line 12: The snapshot test was failing due to a different number of columns. In general I don't think updating the README or adding more templates to the repo should cause the tests to fail, so I reduced the checks to just making sure there's a basic template in the output. Thoughts?

sfc-gh-cgorrie: The change seems reasonable, but I can imagine a bug that only returns the first template. I'd be more comfortable with asserting `streamlit-python` and `streamlit-java` are there too, though I don't think we need to be exhaustive here as we add more templates.

sfc-gh-cgorrie (dismissed):
Windows loves exclusive file access 😅

sfc-gh-bgoel (approved):

sfc-gh-cgorrie (approved):

PR #1161, opened 2024-06-05
[SNOW-1462584] Implement setup script validation

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Calls `system$validate_native_app_setup()` to validate the setup script in the stage on `snow app deploy`, `snow app run`, and the new `snow app validate`. The validation is done server-side, all we need to do is raise an error and show messages if the validation failed.

For `snow app validate`, we use a scratch stage to avoid clobbering the app's dev stage in between deploys. This stage is deleted after every validation run and on `snow app teardown` to avoid leaving behind any garbage.

### Scenarios
App has warning but no errors (exit code 0):
image

App has validation errors (exit code 1):
image

Deploy-time validation error (exit code 1, notice it's using the app's stage and not the scratch stage):
image

Skip validation at deploy time (exit code 0):
image

JSON output (exit code 0 regardless of status):
image


src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -571,3 +575,39 @@ def deploy(
...
+    def validate(self):
+        """Call system$validate_native_app_setup() to validate deployed Native App setup script."""
+        cc.step(f"Validating Native App setup script.")
+        stage_name = StageManager.get_standard_stage_prefix(self.stage_fqn)
+        try:
+            cursor = self._execute_query(
+                f"call system$validate_native_app_setup('{stage_name}')"
+            )
+        except ProgrammingError as err:
+            if "does not exist or not authorized" in err.msg:
```
sfc-gh-bgoel, line 588: Since the App Pkg does not exist error is a specific error, we should possibly tighten the constraint on this condition with an `and` on the error number. 

Once case I'm thinking of but is not necessarily true today: "does not exist or not authorized" could be thrown if some setup script is missing or permissions change that setup script cannot be accessed.

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -571,3 +575,39 @@ def deploy(
...
+            if not cursor.rowcount:
+                raise SnowflakeSQLExecutionError()
+            result_data = json.loads(cursor.fetchone()[0])
+
+            # First print warnings, regardless of the outcome of validation
+            for warning in result_data.get("warnings", []):
+                cc.warning(_validation_item_to_str(warning))
+
+            # Then raise an exception if validation failed
+            if result_data["status"] == "FAIL":
```
sfc-gh-bgoel, line 601: Is `status` always an existing key in `result_data`? or can we get a `KeyError` here?

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -285,3 +285,17 @@ def app_deploy(
...
+@with_project_definition("native_app")
+def app_validate(**options):
+    """
+    Validates a deployed Native App's setup script.
+    """
+    manager = NativeAppManager(
+        project_definition=cli_context.project_definition,
+        project_root=cli_context.project_root,
+    )
+    manager.validate()
```
sfc-gh-bgoel, line 300: There are three commands that can create an application package today: `snow app deploy`, `snow app run`, `snow app version create`. These involve creating package, uploading files to stage, and then some additional steps. Should we add `manager.validate()` to these three commands as well once they finish the second step of upload? We can save user's time by erroring out early before they have to wait further for Snowflake operations. 

Can discuss as a team.

tests_integration/nativeapp/test_validate.py
```diff
@@ -0,0 +1,45 @@
...
+from tests_integration.test_utils import (
+    pushd,
+)
+
+USER_NAME = f"user_{uuid.uuid4().hex}"
+TEST_ENV = generate_user_env(USER_NAME)
+
+
+@pytest.mark.integration
+def test_nativeapp_validate(runner, snowflake_session):
```
sfc-gh-bgoel, line 15: For sake of completeness, we should add one more test on a failing script.

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-fcampbell: It's always set, and defaults to SUCCESS: https://github.com/snowflakedb/snowflake/blob/d88930c088fb783c3b3a23acd4f6f1a6d776628c/GlobalServices/src/main/java/com/snowflake/metadata/dictionary/NativeAppsValidationResult.java#L130. It gets set to FAIL on the first error: https://github.com/snowflakedb/snowflake/blob/d88930c088fb783c3b3a23acd4f6f1a6d776628c/GlobalServices/src/main/java/com/snowflake/metadata/dictionary/NativeAppsValidationResult.java#L156-L158

src/snowflake/cli/plugins/nativeapp/commands.py
sfc-gh-fcampbell: Yeah that can definitely be done, this method should be callable anywhere. Should I do this in this PR or in a followup?

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-fcampbell: yeah it's always set in the response. It defaults to SUCCESS and gets flipped to FAIL on the first recorded error

src/snowflake/cli/plugins/nativeapp/commands.py
sfc-gh-bgoel: If it is as straightforward as adding `manager.validate()` to the three methods, then we should do it here and mention the new validation step in the three commands in the release notes.

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-fcampbell: if a permission issue hides the visibility of the app package, wouldn't that be functionally equivalent to it not existing, from the PoV of the user? I can update the exception's error message to add "or is not authorized" (i'll add the errno check too)

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -285,3 +285,17 @@ def app_deploy(
     return MessageResult(
         f"Deployed successfully. Application package and stage are up-to-date."
     )
+
+
+@app.command("validate", requires_connection=True)
+@with_project_definition("native_app")
+def app_validate(**options):
```
sfc-gh-bdufour, line 292: A few thoughts on the expected behaviour for this:
- do we target the current local version of the setup script, prior to full deployment, or the version that's already deployed? As I argued offline, I think we offer more value if it's the former. We could also let the user choose, but we'd have to think of a good mechanism for that.
- what happens when the user specifies `--format json` for the `validate` command? I can see tools being built on this functionality.
- should we validate as part of other commands too, e.g. `snow app deploy` or `run`? I'm not sure if GS runs validation by default ATM, but it would good for the user to see the output in other flows.

sfc-gh-fcampbell: 1 and 3 are addressed in the design doc. For 2, I'm thinking of just returning the raw response and skip the whole abort behaviour. Do we support running `snow app deploy` or `snow app run` with `--format json` too?

tests_integration/nativeapp/test_validate.py
sfc-gh-fcampbell: added

src/snowflake/cli/api/project/schemas/native_app/native_app.py
```diff
@@ -32,6 +32,10 @@ class NativeApp(UpdatableModel):
         title="Identifier of the stage that stores the application artifacts.",
         default="app_src.stage",
     )
+    scratch_stage: Optional[str] = Field(
+        title="Identifier of the stage that stores temporary scratch data used by the Snowflake CLI.",
+        default="app_src.stage_snowflake_cli_scratch",
```
sfc-gh-fcampbell, line 37: I got an earlier comment about making this a template, but template rendering actually happens on the parsed `snowflake.yml` before resolving this `default` value, so we can't use templates here.

sfc-gh-cgorrie: @sfc-gh-melnacouzi we have multiple defaults we could now phrase like templates -- source stage, application name, and package name. What's the right way for us to proceed here?

@sfc-gh-fcampbell as the originator of the comment you're talking about, this is not a blocker for this PR.

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -335,35 +344,39 @@ def sync_deploy_root_with_stage(
...
         recursive: bool,
+        stage_fqn: str,
         local_paths_to_sync: List[Path] | None = None,
     ) -> DiffResult:
         """
         Ensures that the files on our remote stage match the artifacts we have in
         the local filesystem.
 
         Args:
+            bundle_map: the artifact mapping computed during the `bundle` step. Required when local_paths_to_sync is
```
sfc-gh-bdufour, line 355: that's my bad, thanks for fixing it!

sfc-gh-bdufour: It's always required now BTW, so we can remove the 2nd part of the sentence

sfc-gh-fcampbell: updated

tests_integration/nativeapp/test_validate.py
```diff
@@ -0,0 +1,73 @@
...
+        ["app", "init", project_name],
+        env=TEST_ENV,
+    )
+    assert result.exit_code == 0, result.output
+
+    with pushd(Path(os.getcwd(), project_name)):
+        # Create invalid SQL file
+        Path("app/empty.sql").touch()
+        Path("app/setup_script.sql").write_text(
+            "EXECUTE IMMEDIATE FROM './empty.sql';\n"
```
sfc-gh-bdufour, line 53: Wondering if that could be flaky. What about having so obviously invalid SQL in there instead?

tests_integration/nativeapp/test_deploy.py
```diff
@@ -102,9 +102,13 @@ def test_nativeapp_deploy(
...
         # deploy --prune removes remote-only files
-        ["app deploy --prune", ["stage/manifest.yml"], ["stage/README.md"]],
+        [
+            "app deploy --prune --no-validate",
+            ["stage/manifest.yml"],
+            ["stage/README.md"],
+        ],
         # deploy removes remote-only files (--prune is the default value)
-        ["app deploy", ["stage/manifest.yml"], ["stage/README.md"]],
+        ["app deploy --no-validate", ["stage/manifest.yml"], ["stage/README.md"]],
```
sfc-gh-bdufour, line 111: Did you have to add this everywhere to avoid touching all the snapshots?

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -564,12 +579,75 @@ def deploy(
...
+        for warning in validation_result.get("warnings", []):
+            cc.warning(_validation_item_to_str(warning))
+
+        # Then raise an exception if validation failed
+        if validation_result["status"] == "FAIL":
+            messages = [
+                _validation_item_to_str(error)
+                for error in validation_result.get("errors", [])
+            ]
+            raise SetupScriptFailedValidation(messages)
```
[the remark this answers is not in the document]
sfc-gh-fcampbell: there was no specific reason, but looking back I like it like that since it separates errors and warnings, and it makes the errors really obvious by printing them inside the ASCII box art (see the screenshot in the PR description). I'm not opposed to changing it if it's a blocker

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -564,12 +579,75 @@ def deploy(
...
             )
 
+        if validate:
+            self.validate(use_scratch_stage=False)
+
         return diff
+
+    def validate(self, use_scratch_stage: bool = False):
+        """Validates Native App setup script SQL."""
+        cc.step(f"Validating Snowflake Native App setup script.")
```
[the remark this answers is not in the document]
sfc-gh-fcampbell: yeah it could be considered a phase, but the output looks a little wonky since the diff log uses `"\n.join` on a bunch of strings, so they don't come out indented:

tests_integration/nativeapp/test_deploy.py
sfc-gh-fcampbell: no, it's because these examples aren't valid apps (some don't upload the manifest, some don't upload the setup script, etc). Those operations are valid by themselves, but since the files on the stage at that point don't make a valid app, the validation function fails. Since it's not the goal of the test, I omitted validation here to restore the original behaviour

tests_integration/nativeapp/test_validate.py
sfc-gh-fcampbell: fair enough, I'll change it to something less flaky

src/snowflake/cli/api/project/schemas/native_app/native_app.py
sfc-gh-bdufour: Agreed, let's go with this for now, we can introduce the templates later on.

src/snowflake/cli/plugins/nativeapp/commands.py
sfc-gh-bdufour: All commands support it, but run and deploy don't really output anything worth including in JSON. Validate is a little different.

src/snowflake/cli/plugins/nativeapp/manager.py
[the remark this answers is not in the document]
sfc-gh-bdufour: The diff output is bad, I have a fix locally that I will submit in the next few days. So it's not a really good comparison.

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -578,12 +592,75 @@ def deploy(
...
+        finally:
+            if use_scratch_stage:
+                cc.step(f"Dropping stage {self.scratch_stage_fqn}.")
+                with self.use_role(self.package_role):
+                    self._execute_query(
+                        f"drop stage if exists {self.scratch_stage_fqn}"
+                    )
+
+
+def _validation_item_to_str(item: dict[str, str | int]):
```
sfc-gh-bdufour, line 662: For the user's convenience, IMHO we should include the line number and filename too. IIRC these are both present in the blob we get back from the validation API.

tests_integration/nativeapp/test_deploy.py
sfc-gh-bdufour: That's fair, resolving.

src/snowflake/cli/plugins/nativeapp/manager.py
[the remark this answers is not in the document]
sfc-gh-bdufour: I think we'll want to control how we format these messages, and this way of outputting them will prevent us from doing that. I'll do a pass on the overall output this week or next week, it's long overdue. We should sync on that, but I'm fine if validate becomes slightly inconsistent (but better) while the other commands catch up.

IMHO we keep the box small and generic here (something like "Validation failed, re-run with --no-validate to skip this check"). We output errors the same way as we do warnings, but maybe we make them visually stand out in the output (maybe we use shell colors to highlight errors?)

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-fcampbell: They're already included in the error/warning message, and where they aren't it's because they aren't known (see the "App has validation errors" screenshot in the description)

src/snowflake/cli/plugins/nativeapp/manager.py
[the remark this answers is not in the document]
sfc-gh-fcampbell: I updated it to print the messages before the error: 
image

src/snowflake/cli/plugins/nativeapp/manager.py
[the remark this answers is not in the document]
sfc-gh-fcampbell: changed it to a phase

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-bdufour: oh right, I remember now, thanks.

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -578,12 +592,77 @@ def deploy(
...
+        """Validates Native App setup script SQL."""
+        with cc.phase(f"Validating Snowflake Native App setup script."):
+            validation_result = self.get_validation_result(use_scratch_stage)
+
+            # First print warnings, regardless of the outcome of validation
+            for warning in validation_result.get("warnings", []):
+                cc.warning(_validation_item_to_str(warning))
+
+            # Then print errors
+            for warning in validation_result.get("errors", []):
```
sfc-gh-bdufour, line 620: nit: warning -> error

sfc-gh-bdufour (dismissed):

sfc-gh-bdufour (approved):

PR #1162, opened 2024-06-06
[NADE] Remove symlinking to instead copy artifacts recursively to the deploy root

sfc-gh-bgoel:
### Pre-review checklist

### Changes description
An internal-only change. We observed a number of issues with retaining the symlinking behavior for the `snow app bundle` step, especially for directories and their expected behavior. One of the recent ones was discovered during code generation during the `snow app bundle` step, where a directory can be symlinked but its files are not symlinks, i.e. `file.is_symlink()` is `False`. However, if you try to modify the file in the deploy root, it also modifies source. As a result, we need to remove the symlinking behavior. 

We still retain the glob resolution as it currently is, for the copy uses a resolved path.


src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -456,7 +452,9 @@ def build_bundle(
     for (absolute_src, absolute_dest) in bundle_map.all_mappings(
         absolute=True, expand_directories=False
     ):
-        symlink_or_copy(absolute_src, absolute_dest, deploy_root=deploy_root)
+        copy_artifacts_to_deploy_root(
```
sfc-gh-bgoel, line 455: There is a choice here:
1. `bundle_map.all_mappings(..., expand_directories=True) which will ensure that `absolute_src` and `absolute_dest` are files only, and there `copy_artifacts_to_deploy_root` will only copy files.
2. Leave `bundle_map.all_mappings` as is, and therefore `absolute_src` can be a directory, and we rely on SecurePath.copy` from `copy_artifacts_to_deploy_root` to honor its contract to always perform a recursive copy.

Inviting opinions from the team here. Either way, we need to remove the symlinking, reasons described in the preamble.

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bgoel, line 1: ### Comment 1
There is a decision to be made in `build_bundle`:
1. `bundle_map.all_mappings(..., expand_directories=True) which will ensure that `absolute_src` and `absolute_dest` are files only, and therefore `symlink_or_copy` only deals with files.
2. Leave `bundle_map.all_mappings` as is, and therefore `absolute_src` can be a directory, and `symlink_or_copy` will use the recursive part of the conditional.

Inviting opinions from the team here. Either way, `symlink_or_copy` should be able to deal with any type of path - file or dir.

### Comment 2
There are no existing unit tests for just the original `symlink_or_copy`, it is tested as part of `build_bundle` in unit tests. But this has long been a painful "bug" for us, so I have introduced unit tests that capture most cases. More on it in comment 3.

### Comment 3
This implementation will not detect any cycles that may occur if a user has incorrectly created symlinks for their source code in the project root. In that case, the function will end up looping indefinitely. I don't believe Path or os offer cycle detection either, so either we can align with that for consistency and leave it to our user to create symlinks reliably, or I can introduce a simple set to track for visited sources.

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -416,17 +415,32 @@ def symlink_or_copy(
...
     resolved_deploy_root = deploy_root.resolve()
-    if resolved_deploy_root not in resolved_dst.parents:
+    dst_is_deploy_root = resolved_deploy_root == resolved_dst
+    if (not dst_is_deploy_root) and (resolved_deploy_root not in resolved_dst.parents):
         raise NotInDeployRootError(dest_path=dst, deploy_root=deploy_root, src_path=src)
 
-    if overwrite:
+    if (not dst_is_deploy_root) and overwrite:
+        # TODO Verify: We do not want to delete if dst_is_deploy_root as it may contain artifacts from other runs of bundle map
+        # But this will treat deploy_root as a special dst, as compared to another dst directory that may also have other artifacts from a prev run
```
sfc-gh-bgoel, line 427: FYI ^ Also inviting opinions here.

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -416,17 +415,32 @@ def symlink_or_copy(
...
+        except OSError:
+            ssrc.copy(dst)
+    else:
+        # 1. Create an empty shell directory in the deploy root
+        dst.mkdir(exist_ok=True)
+        # 2. Get all children of the current directory, files and sub-directories
+        for child in sorted(absolute_src.iterdir()):
+            child_path = Path(absolute_src, child.name)
+            dest_path = Path(dst, child.name)
+            symlink_or_copy(src=child_path, dst=dest_path, deploy_root=deploy_root)
```
sfc-gh-bdufour, line 443: Is this vulnerable to infinite loops if a subdir is a symlink pointing back to an earlier dir? AFAIK os.walk handles that just fine, but I'm worried that this recursive logic wouldn't.

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bdufour: When do we rely on overwrite here ? We clear the deploy root before each bundle, so in principle the changes should be purely additive.

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bdufour: oh, just read this, I have a comment about os.walk being resilient to this, so I'd prefer to use that instead. We can't knowingly introduce code that could loop forever in a CLI tool.

sfc-gh-bgoel: Pushed a change with `os.walk`, but also rebased because this was out of date with main.

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bgoel: IIRC from late last night, I thought of this:
```
- src: foo/baz
   dest: some/dir
- src: foo/bar
   dest: ./
```
So `output/deploy/some/dir` has some contents.
Now `foo/bar` wants to go directly in `output/deploy`, i.e. `dst = output/deploy`.
Since `overwrite` is True by default, not only do we clear `output/deploy` at the beginning of overall bundle, but also carry out `delete(dst)`, meaning that entire `output/deploy` is deleted, including `output/deploy/some/dir`. 
My disclaimer is that I may have an incorrect understanding, I am okay to remove this extra condition if that is indeed the case.

sfc-gh-bdufour: What I am asking is why would overwrite ever need to be true? We should always create directories as needed, and overwrite files. No?

sfc-gh-bdufour: The issue here is that with my changes, all directories could behave that way, not just deploy root.

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -400,12 +401,9 @@ def delete(path: Path) -> None:
         spath.rmdir(recursive=True)  # remove dir and all contains
 
 
-def symlink_or_copy(
-    src: Path, dst: Path, deploy_root: Path, makedirs=True, overwrite=True
-) -> None:
+def symlink_or_copy(src: Path, dst: Path, deploy_root: Path, makedirs=True) -> None:
```
sfc-gh-bdufour, line 404: why keep makedirs though? We don't ever set it to false, and it muddies the contract (+ requires extra testing). How about we just delete it and remove the tests that covered that unnecessary behaviour?

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -416,17 +414,34 @@ def symlink_or_copy(
...
-    try:
-        os.symlink(src, dst)
-    except OSError:
-        ssrc.copy(dst)
+        try:
+            os.symlink(absolute_src, dst)
+        except OSError:
+            ssrc.copy(dst)
+    else:
+        # 1. Create an empty shell directory in the deploy root
```
sfc-gh-bdufour, line 432: nit: there is no step 2
other nit: it's not an empty shell, we just copy the directory, contents and all. It's just that the contents is symlinks. But that's still not empty.

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -416,17 +414,34 @@ def symlink_or_copy(
...
+        dst.mkdir(exist_ok=True)
+        for root, _, files in sorted(os.walk(absolute_src, followlinks=True)):
+            relative_root = Path(root).relative_to(absolute_src)
+            absolute_root_in_deploy = Path(dst, relative_root)
+            absolute_root_in_deploy.mkdir(parents=True, exist_ok=True)
+            for file in files:
+                absolute_file_in_project = Path(absolute_src, relative_root, file)
```
sfc-gh-bdufour, line 439: nit: that's just Path(root, file)

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -416,17 +414,34 @@ def symlink_or_copy(
...
+                absolute_file_in_deploy = Path(absolute_root_in_deploy, file)
+                try:
+                    os.symlink(absolute_file_in_project, absolute_file_in_deploy)
+                except OSError:
+                    shutil.copy(absolute_file_in_project, absolute_file_in_deploy)
```
sfc-gh-bdufour, line 444: we should call this recursively, since you have checks that'd be skipped here due to duplicated symlinking logic.

sfc-gh-bdufour: to clarify: we should call `symlink_or_copy` again and not repeat the symlink logic that lives in the other branch of the if condition.

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bgoel: I want to keep this simple by keeping it explicit, I get different outcomes depending on how I create this path, which can get too confusing.

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bgoel: Oh okay so back to the previous approach, but this time with os.walk. Making the change now.

sfc-gh-bdufour (approved):

PR #1170, opened 2024-06-06
[SNOW-1371918] Make snow app deploy behave more intuitively

sfc-gh-bdufour:
### Pre-review checklist

### Changes description

`snow app deploy`, when invoked with directories, current a matching artifact deployment rule for each parameter. This is not very intuitive. For example, when a user specifies:

```
- src: app/*
- dest: ./
```
in their `snowflake.yml`, `snow app deploy -r app` results in an error. This change makes the command deploy all files that have been changed within the specified directory instead.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -51,7 +51,7 @@
 * Improved error message in `snow spcs image-registry login` when docker is not installed.
 * Improved detection of conflicts between artifact rules for native application projects
 * Fixed URL generation for applications, streamlits, and notebooks that use a quoted identifier with spaces.
-* `snow app deploy` now scans specified directories for changes even if there is no artifact rule matching that directory
+* Passing a directory to `snow app deploy` will now deploy any of its subfolders specified in the application's artifact rules
 
 
 # v2.4.0
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -51,7 +51,7 @@
 * Improved error message in `snow spcs image-registry login` when docker is not installed.
 * Improved detection of conflicts between artifact rules for native application projects
 * Fixed URL generation for applications, streamlits, and notebooks that use a quoted identifier with spaces.
-* Passing a directory to `snow app deploy` will now deploy any of its subfolders specified in the application's artifact rules
+* Passing a directory to `snow app deploy` will now deploy any contained file or subfolder specified in the application's artifact rules
 
 
 # v2.4.0
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -22,6 +22,7 @@
 ## New additions
 
 ## Fixes and improvements
+* Passing a directory to `snow app deploy` will now deploy any contained file or subfolder specified in the application's artifact rules
 
 # v2.5.0
 ## Backward incompatibility
@@ -51,8 +52,6 @@
 * Improved error message in `snow spcs image-registry login` when docker is not installed.
 * Improved detection of conflicts between artifact rules for native application projects
 * Fixed URL generation for applications, streamlits, and notebooks that use a quoted identifier with spaces.
-* Passing a directory to `snow app deploy` will now deploy any contained file or subfolder specified in the application's artifact rules
-
 
 # v2.4.0
 ## Backward incompatibility
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -51,6 +51,8 @@
 * Improved error message in `snow spcs image-registry login` when docker is not installed.
 * Improved detection of conflicts between artifact rules for native application projects
 * Fixed URL generation for applications, streamlits, and notebooks that use a quoted identifier with spaces.
+* `snow app deploy` now scans specified directories for changes even if there is no artifact rule matching that directory
+
 
 # v2.4.0
 ## Backward incompatibility
--- a/src/snowflake/cli/plugins/nativeapp/commands.py
+++ b/src/snowflake/cli/plugins/nativeapp/commands.py
@@ -300,7 +300,7 @@ def app_deploy(
             recursive = False
 
     if has_files and prune:
-        raise ClickException("--prune cannot be used when files are also specified")
+        raise ClickException("--prune cannot be used when paths are also specified")
 
     manager = NativeAppManager(
         project_definition=cli_context.project_definition,
--- a/src/snowflake/cli/plugins/nativeapp/commands.py
+++ b/src/snowflake/cli/plugins/nativeapp/commands.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 
 import logging
 from pathlib import Path
+from textwrap import dedent
 from typing import List, Optional
 
 import typer
@@ -269,7 +270,7 @@ def app_teardown(
 def app_deploy(
     prune: Optional[bool] = typer.Option(
         default=None,
-        help=f"""Whether to delete specified files from the stage if they don't exist locally. If set, the command deletes files that exist in the stage, but not in the local filesystem. This option cannot be used when files are specified.""",
+        help=f"""Whether to delete specified files from the stage if they don't exist locally. If set, the command deletes files that exist in the stage, but not in the local filesystem. This option cannot be used when paths are specified.""",
     ),
     recursive: Optional[bool] = typer.Option(
         None,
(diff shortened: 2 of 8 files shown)


RELEASE-NOTES.md
```diff
@@ -11,6 +11,7 @@
   * `snow notebook create` proving an option to create a Snowflake Notebook from a file on stage.
 
 ## Fixes and improvements
+* `snow app deploy` now scans specified directories for changes even if there is no artifact rule matching that directory
```
sfc-gh-bdufour, line 14: Any thoughts on how to make that clearer, but not too low level? I'm not super happy with this phrasing, but that's the best I got ATM.

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -390,6 +389,17 @@ def sync_deploy_root_with_stage(
             for resolved_path in resolved_paths_to_sync:
                 verify_exists(resolved_path)
                 deploy_paths = bundle_map.to_deploy_paths(resolved_path)
+                if not deploy_paths:
```
sfc-gh-cgorrie, line 392: There's another way of implementing this where we change what `to_deploy_paths` gives back and do the prefixing in there. I'm fine with the way you've done it, but let's call out explicitly in that docstring that it requires an exact match and prefixing will not be done.

sfc-gh-bdufour: which docstring are you referring to? `to_deploy_paths`? Its contract remains specific to valid src/dest mappings from the PDF. You're right, I could have relaxed that, but I preferred to make the bundle map the source of truth for the actual mapping, and leave any kind of fuzzy matching to what is essentially domain logic. I might want to revisit in the future (and push the fuzzy stuff into bundle map) if we ever need it elsewhere though.

sfc-gh-bdufour: Assuming I understood correctly, I've updated the docstring in the next rev. PTAL.

sfc-gh-cgorrie: You understood my comment exactly!

RELEASE-NOTES.md
sfc-gh-cgorrie: Perhaps...
```suggestion
* Passing a directory to `snow app deploy` will now deploy any of its subfolders specified in the application's artifact rules
```

tests_integration/nativeapp/test_deploy.py
```diff
@@ -421,3 +421,103 @@ def test_nativeapp_deploy_path_with_no_mapping(
...
+                env=TEST_ENV,
+            )
+            assert result.exit_code == 0
+
+            stage_files = runner.invoke_with_connection_json(
+                ["stage", "list-files", f"{package_name}.{stage_name}"],
+                env=TEST_ENV,
+            )
+
+            assert contains_row_with(stage_files.json, {"name": "stage/README.md"})
```
sfc-gh-cgorrie, line 516: I'm confused why this works, considering you just removed the local source of `stage/README.md` and called deploy with `--prune`. Can you explain it?

sfc-gh-bdufour: Great catch, I really thought I had written the flipped condition here, which would indeed make the test fail. What's happening is that prune + file arguments doesn't really work. Once a file is deleted, we lose track of why it was on the stage in the first place. We see that stage/README.md is present, but we filter it out because we have no idea where it came from. I see three possible ways to handle that:

1. Disallow `--prune` when arguments are passed in to deploy. It doesn't work intuitively, it will only prune things that match the local path anyway. At least we'd disallow it until we fix the behaviour, likely by doing some bookkeeping locally to remember mappings for deleted files.
2. Have `--prune` always prune _all_ matches, regardless of the arguments provided. That's potentially surprising to users, since `deploy -r --prune app` doesn't limit the pruning to the specified directory.
3. Have `--prune` work as it does now, which is that if you map artifacts to the same directory in the deploy root (e.g. src -> src), then prune will work as you expect, otherwise it will have no effect (as shown here, since app -> .)

In all cases, I think we need to fix the behaviour. We've talked about the long term solution before in the context of auto-reload, we have a design for it that should be revisited. In the meantime though, I prefer solution 1, but I'm open to arguments for the other options as well. Thoughts?

sfc-gh-cgorrie: The first approach makes the most sense to me. Let's make the command error out when `--prune` and files/folders are given, and then update the documentation:

- if we are saying `snow app deploy` is essentially an alias for `snow app deploy --prune -r .` we'll need to drop that `.`
- document the fact that one dest folder can be populated from multiple source folders, which is why `--prune` cannot work with files

sfc-gh-bdufour: Agreed, the alias thing would now become aspirational, when we fix the prune logic. Will make the necessary changes here and resubmit then, I think it makes sense to bundle all the changes in one PR.

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -268,18 +270,24 @@ def app_teardown(
...
         default=None,
         show_default=False,
-        help=f"""Paths, relative to the the project root, of files you want to upload to a stage. The paths must match one of the artifacts src pattern entries in snowflake.yml. If unspecified, the command syncs all local changes to the stage.""",
+        help=dedent(
+            f"""
+            Paths, relative to the the project root, of files or directories you want to upload to a stage. If a file is
+            specified, it must match one of the artifacts src pattern entries in snowflake.yml. If a directory is
+            specified, it will be searched for subfolders or files to deploy based on artifacts src pattern entries. If
+            unspecified, the command syncs all local changes to the stage."""
+        ).strip(),
```
sfc-gh-bgoel, line 290: iirc, the reason why this only says file or directory, and not a glob pattern, is because a user's shell should automatically expand that and pass that in to this arg as a list. Is that correct? Do we still need to clarify that here?

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -298,6 +306,9 @@ def app_deploy(
         if recursive is None:
             recursive = False
 
+    if has_paths and prune:
```
sfc-gh-bgoel, line 309: Is all the above a good candidate for expanding the policies we wrote earlier for capturing combinations of flags? Not a holdup for this PR though.

tests_integration/nativeapp/test_deploy.py
```diff
@@ -445,3 +440,202 @@ def test_nativeapp_deploy_path_with_no_mapping(
...
+):
+    project_name = "myapp"
+    project_dir = "app root"
+    result = runner.invoke_json(
+        ["app", "init", project_dir, "--name", project_name],
+        env=TEST_ENV,
+    )
+    assert result.exit_code == 0
+
+    with pushd(Path(os.getcwd(), project_dir)):
```
sfc-gh-bgoel, line 459: Nit, fixture `temporary_working_directory` already returns a path, so you should be able to use it directly instead of `os.getcwd()`.

src/snowflake/cli/plugins/nativeapp/commands.py
sfc-gh-bdufour: That message has nothing to do with globbing. Any path specified as argument could be a file or a directory.

src/snowflake/cli/plugins/nativeapp/commands.py
sfc-gh-bdufour: I don't think that falls under a policy. It's just mutually exclusive flags that we'll end up allowing in the future I think, but there needs significant work before we can do that.

tests_integration/nativeapp/test_deploy.py
sfc-gh-bdufour: I can make that change, but then I'll do it throughout the file. I was just keeping with the pattern established by other tests.

sfc-gh-bgoel (dismissed):
LGTM, replied to open threads.

src/snowflake/cli/plugins/nativeapp/commands.py
sfc-gh-bdufour: ah. Yes they can, but this PR doesn't change that, it was true before as well. We it does change is what happens if some/dir is specified. Basically the path -> artifact rule mapping is just being relaxed here, nothing more.

tests_integration/nativeapp/test_deploy.py
sfc-gh-bdufour: Tried it, some tests failed. I don't feel highly motivated to dig deeper :)

sfc-gh-cgorrie (approved):

PR #1173, opened 2024-06-07
[SNOW-1465372] Strip version numbers from the Snowpark lib during codegen

sfc-gh-bdufour:
### Pre-review checklist

### Changes description

The Snowpark library sometimes appends its current version to the reported package dependencies. This is too strict for native apps, since they are expected to outlast Snowpark library upgrades. This change normalizes the Snowpark library package and removes that version if it's present. In the case where a user specified that version, we'd strip it as well, but arguably that is correct behaviour for this specific package due to strict backwards compatibility requirements.

### DIFF
--- a/src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
+++ b/src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
@@ -417,7 +417,7 @@ def generate_create_sql_ddl_statement(
         )
 
     if extension_fn.packages:
-        create_query += f"\nPACKAGES=({', '.join(extension_fn.packages)})"
+        create_query += f"\nPACKAGES=({', '.join(ensure_all_string_literals([pkg.strip() for pkg in extension_fn.packages]))})"
 
     if extension_fn.external_access_integrations:
         create_query += f"\nEXTERNAL_ACCESS_INTEGRATIONS=({', '.join(ensure_all_string_literals(extension_fn.external_access_integrations))})"
--- a/src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
+++ b/src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
@@ -44,8 +44,9 @@ DEFAULT_TIMEOUT = 30
 TEMPLATE_PATH = Path(__file__).parent / "callback_source.py.jinja"
 SNOWPARK_LIB_NAME = "snowflake-snowpark-python"
 SNOWPARK_LIB_REGEX = re.compile(
+    # support PEP 508, even though not all of it is supported in Snowflake yet
     rf"'{SNOWPARK_LIB_NAME}\s*((<|<=|!=|==|>=|>|~=|===)\s*[a-zA-Z0-9_.*+!-]+)?'"
-)  # support PEP 508, even though not all of it is supported in Snowflake yet
+)
 STAGE_PREFIX = "@"
 
 
--- a/src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
+++ b/src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
@@ -43,7 +43,9 @@ from snowflake.cli.plugins.stage.diff import to_stage_path
 DEFAULT_TIMEOUT = 30
 TEMPLATE_PATH = Path(__file__).parent / "callback_source.py.jinja"
 SNOWPARK_LIB_NAME = "snowflake-snowpark-python"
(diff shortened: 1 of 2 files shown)


src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -266,14 +269,12 @@ def _normalize(
...
-        for pkg in extension_fn.packages:
-            if snowpark_lib_regex.fullmatch(ensure_string_literal(pkg.strip())):
-                snowpark_lib_found = True
-                break
-        if not snowpark_lib_found:
-            extension_fn.packages.append(SNOWPARK_LIB_NAME)
+        extension_fn.packages = [
+            self._normalize_package_name(pkg) for pkg in extension_fn.packages
+        ]
+        snowpark_lib_name = ensure_string_literal(SNOWPARK_LIB_NAME)
```
sfc-gh-bgoel, line 275: Seems like `ensure_string_literal(SNOWPARK_LIB_NAME)` is how the package is used everytime, as opposed to just `SNOWPARK_LIB_NAME`. Should line 45 then just be `SNOWPARK_LIB_NAME = ensure_string_literal("snowflake-snowpark-python")` so we don't have to keep track of the string representation?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -266,14 +269,12 @@ def _normalize(
...
-        snowpark_lib_regex = re.compile(SNOWPARK_LIB_REGEX)
-        for pkg in extension_fn.packages:
-            if snowpark_lib_regex.fullmatch(ensure_string_literal(pkg.strip())):
-                snowpark_lib_found = True
-                break
-        if not snowpark_lib_found:
-            extension_fn.packages.append(SNOWPARK_LIB_NAME)
+        extension_fn.packages = [
+            self._normalize_package_name(pkg) for pkg in extension_fn.packages
+        ]
```
sfc-gh-fcampbell, line 274: this is replacing all package names with the normalized version, which the current code doesn't do. Is that intentional?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -266,14 +269,12 @@ def _normalize(
...
+        snowpark_lib_name = ensure_string_literal(SNOWPARK_LIB_NAME)
+        if snowpark_lib_name not in extension_fn.packages:
```
sfc-gh-bgoel, line 276: (The packages list is not expected to be long, so this is not really an issue.) 
We're doing two linear operations here: list comprehension and the contains check. Should `_normalize_package_name` perhaps take care of both in one iteration, similar to the `snowpark_lib_found` approach on the left?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -266,14 +269,12 @@ def _normalize(
...
+            extension_fn.packages.append(snowpark_lib_name)
```
sfc-gh-fcampbell, line 277: This is adding the normalized version of `SNOWPARK_LIB_NAME` to the package list, when before we were adding `SNOWPARK_LIB_NAME` itself. Is it just a coincidence that this doesn't change the quoting in the snapshot output?

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bdufour: Just found it to be cleaner like that. As you said, this is expected to be a very short list, so I'm ok with the repeated traversal.

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bdufour: You missed the use in defining SNOWPARK_LIB_REGEX I think

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bdufour: Yes, this is done so I can reliably look for the snowpark lib package in the next step. We always did that as a last step (which I didn't change in this PR because we have tests that verify that this happens as a safety net).

src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-bdufour: No it's not, we always normalized in the end. This change is moving the normalization earlier to simplify the rest of the logic.

sfc-gh-fcampbell (approved):

sfc-gh-bgoel (approved):
LGTM!

PR #1174, opened 2024-06-07
Propagate bundle map to all uses of the deployment logic

sfc-gh-bdufour:
### Pre-review checklist

### Changes description

This is a pure refactoring that is intended to prepare for a future code change. The general idea is that whenever we perform a bundle operation, we should propagate the BundleMap produced to the deployment logic so that it can be queried. This will be useful when trying to reason about changes files (during the diff operation) later on. Currently the bundle map is only required when specific paths are deployed. With this change, it's always a requirement argument.

### Testing

- Updated the existing tests for the API change. However, since the bundle map itself is not used yet outside of the current logic that requires it, no new tests need to be added just yet.
- Manually ran through the NADE test plan to make sure nothing was accidentally broken.

### DIFF
--- a/src/snowflake/cli/plugins/nativeapp/commands.py
+++ b/src/snowflake/cli/plugins/nativeapp/commands.py
@@ -181,8 +181,9 @@ def app_run(
         project_definition=cli_context.project_definition,
         project_root=cli_context.project_root,
     )
-    processor.build_bundle()
+    bundle_map = processor.build_bundle()
     processor.process(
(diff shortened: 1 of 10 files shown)


sfc-gh-cgorrie (approved):
LGTM. I did not do any manual testing.

sfc-gh-fcampbell (approved):

sfc-gh-turbaszek (approved):

PR #1190, opened 2024-06-11
Prevent accidentally clobbering artifacts in native app projects

sfc-gh-bdufour:
### Pre-review checklist

### Changes description

Artifact rules could allow artifacts to be clobbered in certain cases. This change introduces two fixes that, combined, prevent these invalid rules from being accepted. First, additional checks are introduced to verify overlap between rules. Second, in order to make sure valid cases are processed correctly, the logic has been modified to keep track of individual insertion order for (src, dest) pairs. This way, we can guarantee that the deploy root will contain the correct files after the bundle step.

### Testing

- Added unit tests to cover the new behaviour
- Manually invoked the CLI using the cases fixed here to verify the deploy root directory. Some of the behaviour that becomes allowed as part of this change will require a follow-up change to remove dir symlinking in order to be enabled, however. Currently, adding new files to a previously-deployed dir can be disallowed because the directory is a symlink back into the source directory.

### DIFF
--- a/src/snowflake/cli/plugins/nativeapp/artifacts.py
+++ b/src/snowflake/cli/plugins/nativeapp/artifacts.py
@@ -18,7 +18,7 @@ import itertools
 import os
 from pathlib import Path
 from textwrap import dedent
-from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, Iterable, Iterator, List, Optional, Tuple, Union
 
 from click.exceptions import ClickException
 from snowflake.cli.api.constants import DEFAULT_SIZE_LIMIT_MB
@@ -101,6 +101,109 @@ class NotInDeployRootError(ClickException):
 ArtifactPredicate = Callable[[Path, Path], bool]
 
 
+class _ArtifactPathMap:
+    """
+    A specialized version of an ordered multimap used to keep track of artifact
+    source-destination mappings. The mapping is bidirectional, so it can be queried
+    by source or destination paths. All paths manipulated by this class must be in
+    relative, canonical form (relative to the project or deploy roots, as appropriate).
+    """
+
+    def __init__(self, project_root: Path):
+        self._project_root = project_root
+
+        # All (src,dest) pairs in inserting order, for iterating
+        self.__src_dest_pairs: List[Tuple[Path, Path]] = []
+        # built-in dict instances are ordered as of Python 3.7
+        self.__src_to_dest: Dict[Path, List[Path]] = {}
+        self.__dest_to_src: Dict[Path, List[Path]] = {}
+        self._dest_is_dir: Dict[Path, bool] = {}
+
+    def put(self, src: Path, dest: Path, dest_is_dir: bool) -> None:
+        """
+        Adds a new source-destination mapping pair to this map, if necessary.
+
+        Arguments:
+            src {Path} -- the source path, in canonical form.
+            dest {Path} -- the destination path, in canonical form.
+            dest_is_dir {bool} -- whether the destination path is a directory.
+        """
+        # Both paths should be in canonical form
+        assert not src.is_absolute()
+        assert not dest.is_absolute()
+
+        absolute_src = self._project_root / src
+
+        current_sources = self.__dest_to_src.get(dest, [])
+        src_is_dir = absolute_src.is_dir()
+        if dest_is_dir:
+            # directory -> directory
+            # Check that dest is currently unmapped
+            current_is_dir = self._dest_is_dir.get(dest, False)
+            if current_is_dir:
+                # mapping to an existing directory is not allowed
+                raise TooManyFilesError(dest)
+        else:
+            # file -> file
+            # Check that there is no previous mapping for the same file.
+            if current_sources and src not in current_sources:
+                # There is already a different source mapping to this destination
+                raise TooManyFilesError(dest)
+
+        if src_is_dir:
+            # mark all subdirectories of this source as directories so that we can
+            # detect accidental clobbering
+            for (root, _, files) in os.walk(absolute_src, followlinks=True):
+                canonical_subdir = Path(root).relative_to(absolute_src)
+                canonical_dest_subdir = dest / canonical_subdir
+                self._update_dest_is_dir(canonical_dest_subdir, is_dir=True)
+                for f in files:
+                    self._update_dest_is_dir(canonical_dest_subdir / f, is_dir=False)
+
+        # make sure we check of dest_is_dir consistency regardless of whether the
+        # insertion happened. This update can fail, so we need to do it first to
+        # avoid applying partial updates to the underlying data storage.
+        self._update_dest_is_dir(dest, dest_is_dir)
+
+        dests = self.__src_to_dest.setdefault(src, [])
+        srcs = self.__dest_to_src.setdefault(dest, [])
+        if dest not in dests:
+            dests.append(dest)
+            srcs.append(src)
+            self.__src_dest_pairs.append((src, dest))
+
+    def get_sources(self, dest: Path) -> Iterable[Path]:
+        """
+        Returns all source paths associated with the provided destination path, in insertion order.
+        """
+        return self.__dest_to_src.get(dest, [])
+
+    def get_destinations(self, src: Path) -> Iterable[Path]:
+        """
+        Returns all destination paths associated with the provided source path, in insertion order.
+        """
+        return self.__src_to_dest.get(src, [])
+
+    def __iter__(self) -> Iterator[Tuple[Path, Path]]:
+        """
+        Returns all (source, destination) pairs known to this map, in insertion order.
+        """
+        return iter(self.__src_dest_pairs)
+
+    def _update_dest_is_dir(self, dest: Path, is_dir: bool) -> None:
+        current_is_dir = self._dest_is_dir.get(dest, None)
+        if current_is_dir is not None and current_is_dir != is_dir:
+            raise ArtifactError(
+                "Conflicting type for destination path: {canonical_dest}"
+            )
+
+        parent = dest.parent
+        if parent != dest:
+            self._update_dest_is_dir(parent, True)
+
+        self._dest_is_dir[dest] = is_dir
+
+
 class BundleMap:
     """
     Computes the mapping between project directory artifacts (aka source artifacts) to their deploy root location
@@ -110,9 +213,7 @@ class BundleMap:
     def __init__(self, *, project_root: Path, deploy_root: Path):
         self._project_root: Path = resolve_without_follow(project_root)
         self._deploy_root: Path = resolve_without_follow(deploy_root)
-        self._src_to_dest: Dict[Path, List[Path]] = {}
-        self._dest_to_src: Dict[Path, List[Path]] = {}
-        self._dest_is_dir: Dict[Path, bool] = {}
+        self._artifact_map = _ArtifactPathMap(project_root=self._project_root)
 
     def deploy_root(self) -> Path:
         return self._deploy_root
@@ -140,7 +241,7 @@ class BundleMap:
             )
 
         if self._deploy_root in absolute_src.parents:
-            # ignore this item since it's in the deploy root. This can happen if the bundle map is create
+            # ignore this item since it's in the deploy root. This can happen if the bundle map is created
             # after the bundle step and a project is using rules that are not sufficiently constrained.
             # Since the bundle step starts with deleting the deploy root, we wouldn't normally encounter this situation.
             return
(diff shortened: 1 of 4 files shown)


src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -101,6 +101,109 @@ def __init__(
...
+            srcs.append(src)
+            self.__src_dest_pairs.append((src, dest))
+
+    def get_sources(self, dest: Path) -> Iterable[Path]:
+        """
+        Returns all source paths associated with the provided destination path, in insertion order.
+        """
+        return self.__dest_to_src.get(dest, [])
+
+    def get_destinations(self, src: Path) -> Iterable[Path]:
```
sfc-gh-bgoel, line 181: For 181 and 175, the type for values in each of these dicts is defined as `List[Path]` instead of more generic `Iterable`. Do we expect any other type of `Iterable` to be returned from these?

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -101,6 +101,109 @@ def __init__(
...
+            # mark all subdirectories of this source as directories so that we can
+            # detect accidental clobbering
+            for (root, _, files) in os.walk(absolute_src, followlinks=True):
+                canonical_subdir = Path(root).relative_to(absolute_src)
+                canonical_dest_subdir = dest / canonical_subdir
+                self._update_dest_is_dir(canonical_dest_subdir, is_dir=True)
+                for f in files:
+                    self._update_dest_is_dir(canonical_dest_subdir / f, is_dir=False)
+
+        # make sure we check of dest_is_dir consistency regardless of whether the
```
sfc-gh-bgoel, line 163: typo: "off"

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -101,6 +101,109 @@ def __init__(
...
+        # insertion happened. This update can fail, so we need to do it first to
+        # avoid applying partial updates to the underlying data storage.
+        self._update_dest_is_dir(dest, dest_is_dir)
```
sfc-gh-bgoel, line 166: Not sure I understand this. Shouldn't `dest` already be included in this dict when doing `os.walk` above the very first time?

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -140,34 +241,22 @@ def _add(self, src: Path, dest: Path, map_as_child: bool) -> None:
...
             canonical_dest = canonical_dest / canonical_src.name
-            dest_is_dir = src.is_dir()
-
-        # Verify that multiple files are not being mapped to a single file destination
-        current_sources = self._dest_to_src.setdefault(canonical_dest, [])
-        if not dest_is_dir:
-            # the destination is a file
-            if (canonical_src not in current_sources) and len(current_sources) > 0:
-                raise TooManyFilesError(dest)
+            dest_is_dir = absolute_src.is_dir()
```
sfc-gh-bgoel, line 255: To clarify, we are ruling out if a user provided something like this?
```
src: some/dir
dest: path/to/file
```
This is a mismatch and should throw an error?

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -215,36 +304,47 @@ def add(self, mapping: PathMapping) -> None:
...
+            src {Path} -- the source path
+            dest {Path} -- the destination path
+            absolute {bool} -- when True, all mappings will be yielded as absolute paths
+            expand_directories {bool} -- when True, child mappings are yielded if the source path is a directory.
+            predicate {ArtifactPredicate} -- when specified, only mappings satisfying this predicate will be yielded.
+        """
         canonical_src = self._canonical_src(src)
-        canonical_dests = self._src_to_dest.get(canonical_src)
-        assert canonical_dests is not None
+        canonical_dest = self._canonical_dest(dest)
```
sfc-gh-bgoel, line 328: Also unclear to me. Are we losing information here by only getting the one `canonical_dest` instead of the list of the values like in the original map?

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -268,9 +368,10 @@ def all_mappings(
         Returns:
           An iterator over all matching deployed artifacts.
         """
-        for src in self._src_to_dest.keys():
-            for deployed_src, deployed_dest in self._mappings_for_source(
+        for src, dest in self._artifact_map:
```
sfc-gh-bgoel, line 371: `self._artifact_map` is of type `_ArtifactPathMap`. Did you mean
```
for src, dest in self._artifact_map.__src_dest_pairs`:
```

sfc-gh-bdufour: No, `_ArtifactPathMap` defines `__iter__` for that very reason.

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bdufour: If I return a list, I should make it immutably or make a copy of it so it doesn't get modified outside of this class. I didn't think the overhead of the copy was justified, so I went for Iterable instead to say "you can look but not modify".

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bdufour: The invariant here is that the "is dir" map is a shadow copy of the tree we're putting together in the deploy root. It should have one entry for each dir or file (recursively too) that was "claimed" by any rule we've previously processed. By calling this unconditionally here, we detect odd cases where a rule is adding a path as a file, and later on we decide to treat it as a dir.

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bdufour: this particular line is a bug fix, it only worked because project dir == cwd before. What this logic does is deal with src -> dest vs src -> dest/. In the first case, we're renaming src to dest in the deploy root. In the second case, we're creating dest/src in the deploy root. So this particular condition is updating the dest to add that level of nesting.

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bdufour: No, before the logic was:

```
for src in all sources:
   for dest in all src -> dest mappings:
        <process>
```

but now it's

```
for src -> dest in all mappings:
     <process>
```

it's really the same thing, I just moved the iteration logic around a little.

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bdufour: good eye!

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bgoel: Got it, covering any edge case.

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bgoel: Okay, I did not know `Iterable` also does that too, was only aware of `Tuple` or `FrozenSet` offering immutability.

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bgoel: "project dir == cwd" assumption being key, got it.

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bgoel: Missed that, thank you!

sfc-gh-bgoel (approved):
LGTM, I understand this much better now.

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bdufour: it doesn't _do_ anything to be clear. It's just giving any human (or IDE) a hint that the value returned shouldn't be used as a mutable list (although it is, and could be mistakenly used that way). Since we control the entire logic, it felt OK to me.

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -101,6 +101,114 @@ def __init__(
...
+    def __init__(self, project_root: Path):
+        self._project_root = project_root
+
+        # All (src,dest) pairs in inserting order, for iterating
+        self.__src_dest_pairs: List[Tuple[Path, Path]] = []
+        # built-in dict instances are ordered as of Python 3.7
+        self.__src_to_dest: Dict[Path, List[Path]] = {}
+        self.__dest_to_src: Dict[Path, List[Path]] = {}
+
+        # This dictionary accumulates keys for each directory or file to be  created in
```
sfc-gh-cgorrie, line 121: The nittiest nit
```suggestion
        # This dictionary accumulates keys for each directory or file to be created in
```

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -101,6 +101,114 @@ def __init__(
...
+
+        current_sources = self.__dest_to_src.get(dest, [])
+        src_is_dir = absolute_src.is_dir()
+        if dest_is_dir:
+            # directory -> directory
+            # Check that dest is currently unmapped
+            current_is_dir = self._dest_is_dir.get(dest, False)
+            if current_is_dir:
+                # mapping to an existing directory is not allowed
+                raise TooManyFilesError(dest)
```
sfc-gh-cgorrie, line 150: We should change the docstring on `TooManyFilesError` to be multiple files / directories (right now it just says multiple files)

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -101,6 +101,114 @@ def __init__(
...
+        # Both paths should be in canonical form
+        assert not src.is_absolute()
+        assert not dest.is_absolute()
+
+        absolute_src = self._project_root / src
+
+        current_sources = self.__dest_to_src.get(dest, [])
+        src_is_dir = absolute_src.is_dir()
+        if dest_is_dir:
+            # directory -> directory
```
sfc-gh-cgorrie, line 145: I don't think this comment is right -- `src_is_dir` could be `False`

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -101,6 +101,114 @@ def __init__(
...
+        """
+        return self.__src_to_dest.get(src, [])
+
+    def __iter__(self) -> Iterator[Tuple[Path, Path]]:
+        """
+        Returns all (source, destination) pairs known to this map, in insertion order.
+        """
+        return iter(self.__src_dest_pairs)
+
+    def _update_dest_is_dir(self, dest: Path, is_dir: bool) -> None:
```
sfc-gh-cgorrie, line 198: Suggested docstring:
```
Recursively marks seen destination paths as either files or folders, raising
an error if any inconsistencies from previous invocations of this method are encountered.
```
Should also mention `dest` must be a canonical relative path, could use the assert as well.

tests/nativeapp/test_artifacts.py
```diff
@@ -470,6 +476,76 @@ def test_bundle_map_disallows_absolute_dest(bundle_map):
         bundle_map.add(PathMapping(src="app", dest=str(absolute_dest)))
 
 
+def test_bundle_map_disallows_clobbering_parent_directories(bundle_map):
+    # one level of nesting
+    with pytest.raises(TooManyFilesError):
+        bundle_map.add(PathMapping(src="snowflake.yml", dest="./app/"))
+        # Adding a new rule to populate ./app/ from an existing directory. This would
+        # clobber the output of the previous rule, so it's disallowed
+        bundle_map.add(PathMapping(src="./app", dest="./"))
```
sfc-gh-cgorrie, line 485: Is using `./` in an src peeking into the deploy root here? I'm not fully understanding this test or how things get clobbered, it seems to me that this test simply copies `snowflake.yml -> app/snowflake.yml` and `app/* -> ./*`.

sfc-gh-cgorrie (commented):
Some docstring suggestions as well as an overarching question about the tests. Can we now map deploy root paths to deploy root paths? Was this possible before?

tests/nativeapp/test_artifacts.py
sfc-gh-bdufour: This rule says "in the deploy root, first copy snowflake.yml to a directory named `app`, then create a directory named `app` and make it identical to `app` from the project root." So in principle snowflake.yml could be lost. If the rules are flipped it's OK though, because first we create the directory, then we add a file to it.

sfc-gh-bdufour: To be 100% clear, if the rule was app/* -> app, then it'd be fine too as long as snowflake.yml doesn't get multiple matches.

sfc-gh-cgorrie: Ah yes, missed the fact that dest is `./` and not `.`. Makes sense

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bdufour: yes but then it's an error, will rephrase anyway because it is confusing.

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -101,6 +101,126 @@ def __init__(
...
+        # Both paths should be in canonical form
+        assert not src.is_absolute()
+        assert not dest.is_absolute()
+
+        absolute_src = self._project_root / src
+
+        current_sources = self.__dest_to_src.get(dest, [])
+        src_is_dir = absolute_src.is_dir()
+        if dest_is_dir:
+            assert src_is_dir  # file -> directory is not possible here given how rules are processed
```
sfc-gh-cgorrie, line 145: Surprising to me, given that the `artifacts:` rules themselves allow this, e.g. `README.md -> ./app/`. I assume the code that calls `put` is responsible for concatenating the filename with the dest directory? We should call it out in the docstring.

sfc-gh-bdufour: Added, LMK

sfc-gh-bgoel: +1 on this check.

sfc-gh-bgoel (approved):
LGTM.

sfc-gh-cgorrie (approved):

PR #1204, opened 2024-06-14
Improve the output of stage diffs in `snow app` commands

sfc-gh-bdufour:
### Pre-review checklist

### Changes description

Snow app commands that deploy to a stage (e.g. deploy and run) use a diff output that is very difficult to read. As a result, a user could easily miss unwanted uploads. It also just generally leads to poor UX. This change introduces some changes to the console that enable better diff output, as well as a revamp of the diff output itself.

Full list of high-level changes included here:
* Improved the overall stage diff output for all native app commands and snow stage diff (currently a hidden command).
* Added `to_deploy_path` to BundleMap APIs
* Modified BundleMap's internal data structures to account for the fact that dest -> src is 1:1, not 1:*
* Added indented() as a new console API
* Removed the restriction on using markup in console.message, and used it to display colors in the diff output (similar to git, we use red to show local or remote paths that are "dirty", i.e. not fully deployed yet.
* Added the ability for `snow stage diff` to output JSON (since I was modifying the command anyway)
* Created a new `tests_common` directory that can be used to promote test fixture / utility reuse across various kinds of tests (e.g. unit and integ)

### Testing

* Added both unit and integ tests to cover the new behaviour. Because the output is platform-specific, I've decided to monkey-patch Path output to ensure that only POSIX-style separators ('/') are used. This introduces a tiny blind spot, but makes cross-platform test authoring much easier. I like the tradeoff here, but open to discussion.
* Manually tested the modified commands (deploy, run and diff) to make sure the output was correct. Manually tested with apps using code generation to make sure the case of no source mapping was handled correctly.
* Verified that no color output is used when stdout is redirected (no extra work necessary, Rich handles this automatically).

### Sample Output

`snow app deploy` with all categories present:
deploy_diff_output

`snow app run` publishing a new app with generated code present:
app_run_diff_output

### DIFF
--- a/src/snowflake/cli/api/console/abc.py
+++ b/src/snowflake/cli/api/console/abc.py
@@ -68,8 +68,11 @@ class AbstractConsole(ABC):
 
     @contextmanager
     @abstractmethod
-    def indented(self) -> Iterator[Callable[[str], None]]:
-        """A context manager for indenting phase or step messages"""
+    def indented(self):
+        """
+        A context manager for temporarily indenting messages and warnings. Phases and steps cannot be used in indented blocks,
+        but multiple indented blocks can be nested (use sparingly).
+        """
 
     @abstractmethod
     def step(self, message: str):
--- a/src/snowflake/cli/api/console/abc.py
+++ b/src/snowflake/cli/api/console/abc.py
@@ -66,6 +66,11 @@ class AbstractConsole(ABC):
     ) -> Iterator[Callable[[str], None]]:
         """A context manager for organising steps into logical group."""
 
+    @contextmanager
+    @abstractmethod
+    def indented(self) -> Iterator[Callable[[str], None]]:
+        """A context manager for indenting phase or step messages"""
+
     @abstractmethod
     def step(self, message: str):
         """Displays a message to output."""
--- a/src/snowflake/cli/api/console/console.py
+++ b/src/snowflake/cli/api/console/console.py
@@ -15,7 +15,7 @@
 from __future__ import annotations
 
 from contextlib import contextmanager
-from typing import Callable, Iterator, Optional
+from typing import Optional
 
 from rich.style import Style
 from rich.text import Text
@@ -88,11 +88,14 @@ class CliConsole(AbstractConsole):
                 self._print(self._format_message(exit_message, Output.PHASE))
 
     @contextmanager
-    def indented(self) -> Iterator[Callable[[str], None]]:
-        """A context manager for indenting phase or step messages"""
+    def indented(self):
+        """
+        A context manager for temporarily indenting messages and warnings. Phases and steps cannot be used in indented blocks,
+        but multiple indented blocks can be nested (use sparingly).
+        """
         self._extra_indent += 1
         try:
-            yield self.message
+            yield
         finally:
             self._extra_indent -= 1
 
--- a/src/snowflake/cli/api/console/console.py
+++ b/src/snowflake/cli/api/console/console.py
@@ -15,7 +15,7 @@
 from __future__ import annotations
 
 from contextlib import contextmanager
-from typing import Optional
+from typing import Callable, Iterator, Optional
 
 from rich.style import Style
 from rich.text import Text
@@ -44,22 +44,27 @@ class CliConsole(AbstractConsole):
     """
 
     _indentation_level: int = INDENTATION_LEVEL
+    _extra_indent: int = 0
     _styles: dict = {
         "default": "",
         Output.PHASE: PHASE_STYLE,
         Output.STEP: STEP_STYLE,
-        Output.INFO: INFO_STYLE,
+        Output.INFO: None,
         Output.IMPORTANT: IMPORTANT_STYLE,
     }
 
     def _format_message(self, message: str, output: Output) -> Text:
         """Wraps message in rich Text object and applies formatting."""
         style = self._styles.get(output, "default")
-        text = Text(message, style=style)
+        if style is not None:
+            text = Text(message, style=style)
+        else:
+            text = Text.from_markup(message)
 
+        current_indent = self._extra_indent
         if self.in_phase and output in {Output.STEP, Output.INFO, Output.IMPORTANT}:
-            text.pad_left(self._indentation_level)
-
+            current_indent += 1
+        text.pad_left(current_indent * self._indentation_level)
         return text
 
     @contextmanager
@@ -67,6 +72,10 @@ class CliConsole(AbstractConsole):
         """A context manager for organising steps into logical group."""
         if self.in_phase:
             raise CliConsoleNestingProhibitedError("Only one phase allowed at a time.")
+        if self._extra_indent > 0:
+            raise CliConsoleNestingProhibitedError(
+                "Phase cannot be used in an indented block."
+            )
 
         self._print(self._format_message(enter_message, Output.PHASE))
         self._in_phase = True
@@ -78,11 +87,24 @@ class CliConsole(AbstractConsole):
             if exit_message:
                 self._print(self._format_message(exit_message, Output.PHASE))
 
+    @contextmanager
+    def indented(self) -> Iterator[Callable[[str], None]]:
+        """A context manager for indenting phase or step messages"""
+        self._extra_indent += 1
+        try:
+            yield self.message
+        finally:
+            self._extra_indent -= 1
+
     def step(self, message: str):
         """Displays a message to output.
 
         If called within a phase, the output will be indented.
         """
+        if self._extra_indent > 0:
+            raise CliConsoleNestingProhibitedError(
+                "Step cannot be used in an indented block."
+            )
         text = self._format_message(message, Output.STEP)
         self._print(text)
 
--- a/src/snowflake/cli/api/output/types.py
+++ b/src/snowflake/cli/api/output/types.py
@@ -27,6 +27,15 @@ class CommandResult:
         raise NotImplementedError()
 
 
+class EmptyResult(CommandResult):
+    def __init__(self):
+        pass
+
+    @property
+    def result(self):
+        return ""
+
+
 class ObjectResult(CommandResult):
     def __init__(self, element: t.Dict):
         self._element = element
--- a/src/snowflake/cli/api/output/types.py
+++ b/src/snowflake/cli/api/output/types.py
@@ -27,15 +27,6 @@ class CommandResult:
         raise NotImplementedError()
 
 
-class EmptyResult(CommandResult):
-    def __init__(self):
-        pass
-
-    @property
-    def result(self):
-        return ""
-
-
 class ObjectResult(CommandResult):
     def __init__(self, element: t.Dict):
         self._element = element
--- a/src/snowflake/cli/plugins/nativeapp/artifacts.py
+++ b/src/snowflake/cli/plugins/nativeapp/artifacts.py
@@ -406,7 +406,7 @@ class BundleMap:
     def to_deploy_paths(self, src: Path) -> List[Path]:
         """
         Converts a source path to its corresponding deploy root path. If the input path is relative to the project root,
-        paths relative to the deploy root are returned. If the input path is absolute, absolute path are returned.
+        paths relative to the deploy root are returned. If the input path is absolute, absolute paths are returned.
 
         Note that the provided source path must be part of a mapping. If the source path is not part of any mapping,
         an empty list is returned. For example, if `app/*` is specified as the source of a mapping,
--- a/src/snowflake/cli/plugins/nativeapp/artifacts.py
+++ b/src/snowflake/cli/plugins/nativeapp/artifacts.py
@@ -457,6 +457,33 @@ class BundleMap:
         for src in self._artifact_map.all_sources():
             yield self._to_output_src(src, absolute)
 
+    def to_source_paths(self, dest: Path) -> List[Path]:
+        is_absolute = dest.is_absolute()
+        try:
+            canonical_dest = self._canonical_dest(dest)
+        except ArtifactError:
+            # No mapping possible for the dest path
+            return []
+
+        output_sources: List[Path] = []
+        canonical_srcs = self._dest_to_src.get(canonical_dest)
+        if canonical_srcs is not None:
+            for src in canonical_srcs:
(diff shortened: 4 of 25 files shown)


src/snowflake/cli/api/console/console.py
```diff
@@ -44,29 +44,38 @@ class CliConsole(AbstractConsole):
...
 
     @contextmanager
     def phase(self, enter_message: str, exit_message: Optional[str] = None):
         """A context manager for organising steps into logical group."""
         if self.in_phase:
             raise CliConsoleNestingProhibitedError("Only one phase allowed at a time.")
+        if self._extra_indent > 0:
+            raise CliConsoleNestingProhibitedError(
+                "Phase cannot be used in an indented block."
+            )
```
sfc-gh-cgorrie, line 78: Conceptually, when would I use phase and when would I use indenting as a client of this API?

src/snowflake/cli/api/console/console.py
```diff
@@ -78,11 +87,24 @@ def phase(self, enter_message: str, exit_message: Optional[str] = None):
             if exit_message:
                 self._print(self._format_message(exit_message, Output.PHASE))
 
+    @contextmanager
+    def indented(self) -> Iterator[Callable[[str], None]]:
+        """A context manager for indenting phase or step messages"""
```
sfc-gh-cgorrie, line 92: Docstring seems at odds with the exception raised in phase?

src/snowflake/cli/api/output/types.py
```diff
@@ -27,6 +27,15 @@ def result(self):
         raise NotImplementedError()
 
 
+class EmptyResult(CommandResult):
+    def __init__(self):
+        pass
+
+    @property
+    def result(self):
+        return ""
```
sfc-gh-cgorrie, line 36: Any reason we don't want to return None (or just a bare return)? Does this have to be a string?

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -407,11 +406,15 @@ def all_mappings(
     def to_deploy_paths(self, src: Path) -> List[Path]:
         """
         Converts a source path to its corresponding deploy root path. If the input path is relative to the project root,
-        a path relative to the deploy root is returned. If the input path is absolute, an absolute path is returned.
+        paths relative to the deploy root are returned. If the input path is absolute, absolute path are returned.
```
sfc-gh-cgorrie, line 409:
```suggestion
        paths relative to the deploy root are returned. If the input path is absolute, absolute paths are returned.
```

tests/stage/__snapshots__/test_diff.ambr
```diff
@@ -0,0 +1,197 @@
...
+  '''
+# ---
+# name: test_print_diff_to_console_no_bundlemap.3
+  '''
+  Local changes to be deployed:
+    modified: different
+    modified: dir/different
+    modified: dir/modified
+    modified: modified
+  Deleted paths to be removed from your stage:
```
sfc-gh-cgorrie, line 31: Suggestion for consistency with `git status`: leave a blank line between groups.

tests/stage/__snapshots__/test_diff.ambr
```diff
@@ -0,0 +1,197 @@
+# serializer version: 1
+# name: test_print_diff_to_console_no_bundlemap
+  '''
+  Your stage is up-to-date with your local deploy root
+  
+  '''
+# ---
+# name: test_print_diff_to_console_no_bundlemap.1
+  '''
+  Your stage is up-to-date with your local deploy root
```
sfc-gh-cgorrie, line 10: Suggestion for consistency with `git status`.
```suggestion
  Your stage is up-to-date with your local deploy root.
```

sfc-gh-cgorrie (commented):
Timeboxed review, will come back and review the tests properly

src/snowflake/cli/api/console/console.py
sfc-gh-fcampbell: Seems like `phase()` can't be used within `indented()`, but `indented()` can be used within `phase()`

src/snowflake/cli/api/output/types.py
sfc-gh-fcampbell: How about just allowing commands to return `None`?

sfc-gh-turbaszek: @sfc-gh-fcampbell that's possible today:
https://github.com/snowflakedb/snowflake-cli/blob/4e50d2223352ef95581ab699b96bb90619cf663d/src/snowflake/cli/plugins/spcs/services/commands.py#L201-L203

We haven't had time to refactor this. For logs the solution is to have a `StreamResult`. 

I'm personally inclined to no return approach than having `return Empty`

tests_common/path_utils.py
```diff
@@ -0,0 +1,25 @@
...
+from pathlib import PurePosixPath
+
+import pytest
+
+
+@pytest.fixture
+def print_paths_as_posix():
+    with mock.patch("pathlib.WindowsPath.__str__", autospec=True) as mock_str:
+        mock_str.side_effect = lambda path: str(PurePosixPath(*path.parts))
+        yield mock_str
```
sfc-gh-fcampbell, line 25: could you add the explanation from the PR description as a docstring?

tests_integration/test_stage.py
```diff
@@ -227,3 +227,163 @@ def test_stage_execute(runner, test_database, test_root_path, snapshot):
...
+        file_path.unlink()
+        result = runner.invoke_with_connection(
+            ["stage", "diff", stage_name, str(local_dir)]
+        )
+        assert result.exit_code == 0, result.output
+        assert result.output == snapshot
+
+
+@pytest.mark.integration
+def test_stage_diff_json(runner, snowflake_session, test_database, tmp_path, snapshot):
```
sfc-gh-fcampbell, line 311: this test doesn't use snapshots
```suggestion
def test_stage_diff_json(runner, snowflake_session, test_database, tmp_path):
```

tests_integration/test_stage.py
```diff
@@ -227,3 +227,163 @@ def test_stage_execute(runner, test_database, test_root_path, snapshot):
...
+    another_filename = "another.md"
+    with tempfile.TemporaryDirectory() as td:
+        local_dir = Path(td).resolve()
+        result = runner.invoke_with_connection_json(
+            ["stage", "diff", stage_name, str(local_dir)]
+        )
+        assert result.exit_code == 0, result.output
+        assert result.json == {"modified": [], "added": [], "deleted": []}
+
+        # avoid nesting the files into directories because the snapshots would then become system-dependent
```
sfc-gh-fcampbell, line 337: this test doesn't use snapshots
```suggestion
```

src/snowflake/cli/api/console/console.py
sfc-gh-bdufour: phase vs step predates this PR, but conceptually, a step is a key part of a complex workflow. A phase is a group of related steps. A message is some additional information that's emitted during a step (at least that's how I think of it). Really everything other than phase or step is just a regular message. warning etc are just there for formatting. indented is the new part, it lets you indent messages/warnings. It shouldn't be used too often, but in cases like the diff output I'm adding here, it makes sense to visually separate the diff itself (via indentation) from the other messages within the phases/steps.

src/snowflake/cli/api/console/console.py
sfc-gh-bdufour: This was meant to mean something like "messages issues in steps and phases", but it reads horribly. Will rephrase.

src/snowflake/cli/api/console/console.py
```diff
@@ -78,11 +87,24 @@ def phase(self, enter_message: str, exit_message: Optional[str] = None):
...
+        self._extra_indent += 1
+        try:
+            yield self.message
+        finally:
+            self._extra_indent -= 1
```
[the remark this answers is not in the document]
sfc-gh-bdufour: I specifically don't want to do this because indenting for more than one level should not be a common thing to do. Since the beginning the idea was the keep the API minimal and rather opinionated so that best practices are followed intuitively.

src/snowflake/cli/api/output/types.py
sfc-gh-bdufour: EmptyResult -> None makes sense, yes, happy to make that change if we think that a result should essentially be optional. I didn't go for it here because it's harder to forget then to return something (None being the default RV in Python and all). Not returning anything, to me, was an explicit choice that should be handled carefully for the reasons Tomasz pointed out. In this case, the stage diff command handles JSON formatting specifically, so I was OK with the empty/missing return.

src/snowflake/cli/plugins/stage/diff.py
```diff
@@ -331,3 +289,70 @@ def sync_local_diff_with_stage(
...
+            messages_to_output.append(
+                (
+                    src_dest_pair,
+                    _to_diff_line("added", src_dest_pair[0], src_dest_pair[1]),
+                )
+            )
+
+        with cc.indented() as msg:
+            for key, message in sorted(messages_to_output, key=lambda pair: pair[0]):
+                msg(message)
```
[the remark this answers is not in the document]
sfc-gh-bdufour: Yeah, I was trying to stay with the "phase yields a step" behaviour here, but maybe I should remove it altogether.

src/snowflake/cli/api/output/types.py
sfc-gh-bdufour: Ok, I made the change to None, since it's already a supported use case.

src/snowflake/cli/plugins/stage/diff.py
```diff
@@ -331,3 +289,74 @@ def sync_local_diff_with_stage(
...
+
+        with cc.indented():
+            for key, message in sorted(messages_to_output, key=lambda pair: pair[0]):
+                cc.message(message)
+
+        blank_line_needed = True
+
+    if diff.only_on_stage:
+        prefix = "\n" if blank_line_needed else ""
+        cc.message(f"{prefix}Deleted paths to be removed from your stage:")
```
sfc-gh-fcampbell, line 358: this would break indentation if called inside a nested `indented()` block, how about
```suggestion
        if blank_line_needed:
            cc.message("")
        cc.message("Deleted paths to be removed from your stage:")
```

sfc-gh-fcampbell: I have a larger suggestion of updating `message()` to split lines, pad each line with indentation, then join back, but I feel like it's out of scope

sfc-gh-bdufour: Oh wow, I can't believe I did that, thanks for catching it! As for officially supporting multiple lines, that's something we can consider as we update the rest of the messages. I do agree I've probably piled up enough content in this particular PR already though.

sfc-gh-bgoel (approved):
Left some non-blocking comments, but I agree with the changes in the PR.

src/snowflake/cli/api/console/console.py
sfc-gh-bdufour: It's not really about connections, e.g. bundle uses a phase for codegen but (other than telemetry) is a fully offline operation.

src/snowflake/cli/api/console/abc.py
```diff
@@ -66,6 +66,14 @@ def phase(
     ) -> Iterator[Callable[[str], None]]:
         """A context manager for organising steps into logical group."""
 
+    @contextmanager
+    @abstractmethod
+    def indented(self):
+        """
+        A context manager for temporarily indenting messages and warnings. Phases and steps cannot be used in indented blocks,
```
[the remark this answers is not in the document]
sfc-gh-bdufour: the context manager part. You indent for a set of messages, then you restore the previous indentation.

src/snowflake/cli/plugins/stage/diff.py
```diff
@@ -62,56 +64,12 @@ def has_changes(self) -> bool:
...
-                    *[str(p) for p in self.identical],
-                ]
-            )
-
-        return "\n".join(components)
+    def to_dict(self) -> dict:
+        return {
+            "modified": [str(p) for p in sorted(self.different)],
+            "added": [str(p) for p in sorted(self.only_local)],
+            "deleted": [str(p) for p in sorted(self.only_on_stage)],
```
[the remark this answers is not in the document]
sfc-gh-bdufour: Better UX and deterministic output for snapshots both. We don't really print this structure more than once (and really, why would we?) I figured that requiring the paths to always be sorted is just more difficult to handle in the future as an invariant than just sorting on the way out.

tests/api/console/test_cli_console_output.py
```diff
@@ -85,6 +86,69 @@ def test_phase_is_cleaned_up_on_exception(cli_console):
...
+            with cli_console.indented():
+                cli_console.message("Double indented message")
+                cli_console.warning("Double indented warning")
+            cli_console.message("Message with single indentation again")
+            cli_console.warning("Warning with single indentation again")
+        cli_console.message("No longer indented message")
+        cli_console.warning("No longer indented warning")
+    cli_console.warning("OPS")
+
+    expected_output = dedent(
```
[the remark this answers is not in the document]
sfc-gh-bdufour: Not NADE code, so I was staying with what was used in the other tests for consistency.

src/snowflake/cli/plugins/stage/diff.py
```diff
@@ -331,3 +289,75 @@ def sync_local_diff_with_stage(
...
+    status_prefix = f"[red]{status}[/red]: {padding}"
+
+    return f"{status_prefix}{src_prefix}{dest}"
+
+
+def print_diff_to_console(
+    diff: DiffResult,
+    bundle_map: Optional[BundleMap] = None,
+):
+    if not diff.different and not diff.only_local and not diff.only_on_stage:
```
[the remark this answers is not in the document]
sfc-gh-bdufour: replaced with diff.has_changes()

sfc-gh-bgoel (approved):
LGTM, and acked replies to previous conversations.

sfc-gh-astus (approved):

sfc-gh-fcampbell (approved):

PR #1214, opened 2024-06-17
[SNOW-1486485] Small cleanup to `drop_application`

sfc-gh-fcampbell:
The [no-whitespace diff](https://github.com/snowflakedb/snowflake-cli/pull/1214/files?w=1) can be useful to simplify reviewing this PR

### Pre-review checklist
   * N/A I've described my changes in the section below.

### Changes description
Small code cleanup in `NativeAppTeardownProcessor.drop_application` and related functions to make future PR diffs look cleaner:
- Fixed incorrect type annotation
- Removed redundant `if interactive:` check that was already within an `elif interactive:` block
- Simplified prompt response check using `lower()`
- Simplified logic around prompting the user for drop confirmation when the app isn't owned by the CLI

### DIFF
--- a/src/snowflake/cli/plugins/nativeapp/teardown_processor.py
+++ b/src/snowflake/cli/plugins/nativeapp/teardown_processor.py
@@ -66,7 +66,7 @@ class NativeAppTeardownProcessor(NativeAppManager, NativeAppCommandProcessor):
             cc.message(f"Dropped {object_type} {object_name} successfully.")
 
     def _application_objects_to_str(
-        self, application_objects: ApplicationOwnedObject
+        self, application_objects: list[ApplicationOwnedObject]
     ) -> str:
         """
         Returns a list in an "(Object Type) Object Name" format. Database-level and schema-level object names are fully qualified:
--- a/src/snowflake/cli/plugins/nativeapp/teardown_processor.py
+++ b/src/snowflake/cli/plugins/nativeapp/teardown_processor.py
@@ -143,18 +143,17 @@ class NativeAppTeardownProcessor(NativeAppManager, NativeAppCommandProcessor):
                     f"The following objects are owned by application {self.app_name}:\n{application_objects_str}"
                 )
             elif interactive:
-                if interactive:
-                    user_response = typer.prompt(
-                        f"The following objects are owned by application {self.app_name}:\n{application_objects_str}\n\nWould you like to drop these objects in addition to the application? [y/n/ABORT]",
-                        show_default=False,
-                        default="ABORT",
-                    )
-                    if user_response in ["y", "yes", "Y", "Yes", "YES"]:
-                        cascade = True
-                    elif user_response in ["n", "no", "N", "No", "NO"]:
-                        cascade = False
-                    else:
-                        raise typer.Abort()
+                user_response = typer.prompt(
+                    f"The following objects are owned by application {self.app_name}:\n{application_objects_str}\n\nWould you like to drop these objects in addition to the application? [y/n/ABORT]",
+                    show_default=False,
+                    default="ABORT",
+                ).lower()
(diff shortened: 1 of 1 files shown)


src/snowflake/cli/plugins/nativeapp/teardown_processor.py
```diff
@@ -103,30 +103,28 @@ def drop_application(
 
         # 3. Check if created by the Snowflake CLI
         row_comment = show_obj_row[COMMENT_COL]
-        if row_comment in ALLOWED_SPECIAL_COMMENTS:
-            # No confirmation needed before dropping
-            needs_confirm = False
```
sfc-gh-fcampbell, line 108: `needs_confirm` is never checked again, so we can just remove this whole block and invert the `if` check logic to make it cleaner

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
```diff
@@ -143,18 +141,17 @@ def drop_application(
                     f"The following objects are owned by application {self.app_name}:\n{application_objects_str}"
                 )
             elif interactive:
-                if interactive:
```
sfc-gh-fcampbell, line 146: this `if` check is redundant since it's already within an `elif interactive:` block without `interactive` being mutated in between the checks

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
```diff
@@ -143,18 +141,17 @@ def drop_application(
...
-                    else:
-                        raise typer.Abort()
+                user_response = typer.prompt(
+                    f"The following objects are owned by application {self.app_name}:\n{application_objects_str}\n\nWould you like to drop these objects in addition to the application? [y/n/ABORT]",
+                    show_default=False,
+                    default="ABORT",
+                ).lower()
+                if user_response in ["y", "yes"]:
+                    cascade = True
+                elif user_response in ["n", "no"]:
```
sfc-gh-fcampbell, line 151: we can avoid checking too many casing combinations by only comparing lowercase values

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
```diff
@@ -66,7 +66,7 @@ def drop_generic_object(
             cc.message(f"Dropped {object_type} {object_name} successfully.")
 
     def _application_objects_to_str(
-        self, application_objects: ApplicationOwnedObject
```
sfc-gh-fcampbell, line 69: not sure why mypy didn't flag this one when IntelliJ did /shrug

sfc-gh-bgoel: I thought 3.8 does not allow `list[Type]` yet only `List[Type]` is accepted? But if it does, then great.

sfc-gh-bgoel (approved):
I like the simplicity, lgtm.

src/snowflake/cli/plugins/nativeapp/teardown_processor.py
sfc-gh-fcampbell: we have `from __future__ import annotations` in all files, so annotations are only saved as string literals to be evaluated on-demand. Since the app never reads this particular annotation, it's not an issue (otherwise it would also crash on all the `X | Y` we use instead of `typing.Union`

sfc-gh-fcampbell: (i also tested this manually on 3.8 and it was fine)

sfc-gh-bgoel: Aah yes, I forgot about the `from __future__ import annotations` change to code files. Makes sense.

PR #1216, opened 2024-06-17
[SNOW-1362197] Abort teardown if we don't drop application

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
If we find that the app isn't "owned" by the CLI (i.e. the object's comment isn't what we expect it to be), and the user says that they don't want to drop that app, abort the teardown since dropping the package afterwards would leave the app orphaned and non-functional.

No need to handle `--force` any differently since passing `--force` forces dropping the app anyways. With the current CLI flags, it's no longer possible to create orphan apps.

Denying the drop:
image

Forcing the drop:
image


sfc-gh-bdufour (approved):
LGTM

PR #1223, opened 2024-06-18
[NADE] Introduce order for iterating over files in bundle, fix tests

sfc-gh-bgoel:
### Pre-review checklist

### Changes description
OS Directory traversal is in random order, and hence any side effects of the traversal will also be random. This randomness can cause flakiness in tests that check code paths that cause exceptions. This means we cannot deterministically test for file paths when the code errors early, because each run of the test may have traversed the directory (up until the exception) in a different order than before. 
This PR removes the flakiness of the tests, and introduces a deterministic order for the traversal.

### DIFF
--- a/tests/nativeapp/test_artifacts.py
+++ b/tests/nativeapp/test_artifacts.py
@@ -1161,15 +1161,6 @@ def test_symlink_or_copy_with_no_symlinks_in_project_root(snapshot):
                     dst=Path(deploy_root, "Grand4/Parent3"),
                     deploy_root=deploy_root,
                 )
-            assert not Path(deploy_root, "Grand4/Parent3").is_symlink()
-            assert not Path(deploy_root, "Grand4/Parent3/ChildA").is_symlink()
-            assert Path(deploy_root, "Grand4/Parent3/ChildB.py").is_symlink()
-            assert Path(deploy_root, "Grand4/Parent3/ChildC").is_symlink()
-            assert not Path(
-                deploy_root, "Grand4/Parent3/ChildC/GrandChildA"
-            ).is_symlink()
-
-            assert_dir_snapshot(Path("./output/deploy/Grand4/Parent3"), snapshot)
 
 
 def test_symlink_or_copy_with_symlinks_in_project_root(snapshot):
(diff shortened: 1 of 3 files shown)


tests/nativeapp/test_artifacts.py
```diff
@@ -1161,15 +1161,6 @@ def test_symlink_or_copy_with_no_symlinks_in_project_root(snapshot):
...
                 )
-            assert not Path(deploy_root, "Grand4/Parent3").is_symlink()
-            assert not Path(deploy_root, "Grand4/Parent3/ChildA").is_symlink()
-            assert Path(deploy_root, "Grand4/Parent3/ChildB.py").is_symlink()
-            assert Path(deploy_root, "Grand4/Parent3/ChildC").is_symlink()
-            assert not Path(
-                deploy_root, "Grand4/Parent3/ChildC/GrandChildA"
-            ).is_symlink()
-
-            assert_dir_snapshot(Path("./output/deploy/Grand4/Parent3"), snapshot)
```
sfc-gh-bgoel, line 1172: The lines above are being removed because:
1. The main thing to check is that we received a NotInDeployRootError
2. Even though an exception was raised, there might have been side effects to the filesystem already. Without `sorted`, these side effects can occur in random order, and hence checks from 1164 to 1172 are not guaranteed to always work. 
3. Even with sorted, these checks do not offer additional benefit, because the exception has already been verified. Realistically, a user would run `bundle` again and it will delete the deploy root and create it from scratch again.

tests/nativeapp/test_artifacts.py
sfc-gh-bgoel, line 1: 1. Test `test_symlink_or_copy_raises_error` only deals with one test file in a temporary directory. 
a. This file is first symlinked into the deploy root without an error. 
b. This file is then run through an exception scenario, but it is only used to find its resolved path. Then it errors as it should.
c. The file is unlinked and created in a regular manner within the deploy root. 
d. Hence the test should not have flaky behavior. 

2. Test `test_symlink_or_copy_with_no_symlinks_in_project_root` only had one early exception. 
a. We retain that exception to make sure the correct code path is being tested, but remove the checks around the side effects that occur during the function call.

3. Test `test_symlink_or_copy_with_symlinks_in_project_root` has no such early exceptions, it tests for a successful run of the function.

sfc-gh-bdufour (approved):

PR #1224, opened 2024-06-18
[SNOW-1486485] Fix tearing down an app whose package has already been dropped

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
When a package is dropped, subsequent `snow app teardown` calls fail with “Application is no longer available for use. Please contact the application provider for more details.” since the CLI is still trying to execute `show objects owned by application`. If this query fails with this error, we should ask the user if they want to cascade drop the application without being able to see the objects.

#### Interactive prompted for cascade
image

#### Non-interactive, can't prompt
image

#### Cascade denied
image

#### Cascade allowed
image

#### Interactive prompted for cascade
image

#### Non-interactive, can't prompt
image

#### Cascade denied
image

#### Cascade allowed
image


sfc-gh-cgorrie (approved):
LGTM

PR #1225, opened 2024-06-18
[APPS-41865] Add LSP plugin interface with pygls language server to formalize communication with vs code extension

sfc-gh-klin:
### Pre-review checklist

### Changes description
...

We propose adding code into SnowCLI that allows the VS Code extension to start a JSON-RPC Pygls language server and provides a new interface for “LSP plugins” so that the VS Code extension for Snowflake can make requests to this long-lived language server to invoke commands such as Native Apps commands instead of spawning one-off processes through the command line tool. 

This lets us communicate between the two without parsing command line output. We can also formalize version compatibility rather than have the potential of SnowCLI changes breaking the VS Code extension for Snowflake. 

[Doc here](https://docs.google.com/document/d/1pO46WL4ekGaOXZez-v1StAM2mT70TegHRzPMF1ljV2M/edit)
VS Code extension for Snowflake PR here: https://github.com/snowflakedb/snowflake-vscode-extension/pull/550

TODO:

- get some help writing integration tests to help prevent CLI changes breaking VS Code extension for Snowflake


src/snowflake/cli/plugins/rpc/commands.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    Initializes the RPC LSP language server.
+    """
+    manager = RpcManager()
+
+    return MessageResult(f"RPC LSP began.")
+
+
+@lsp_plugin(
+    name="nade",
```
sfc-gh-cgorrie, line 36: Should this be named something else?

src/snowflake/cli/plugins/rpc/manager.py
```diff
@@ -0,0 +1,124 @@
...
+            conn = connector.connect()
+            return conn
+
+    def create_connection(self) -> Session:
+        session_token = os.environ.get("SESSION_TOKEN")
+        master_token = os.environ.get("MASTER_TOKEN")
+        account = os.environ.get("SF_ACCOUNT")
+        host = os.environ.get("SF_HOST")
+        port = os.environ.get("SF_PORT")
+        protocol = os.environ.get("SF_PROTOCOL")
```
sfc-gh-cgorrie, line 78: Are these all set by e.g. vscode?

src/snowflake/cli/plugins/nativeapp/lsp_commands.py
```diff
@@ -0,0 +1,88 @@
...
+from pygls.server import LanguageServer
+from snowflake.cli.api.output.types import MessageResult
+from snowflake.cli.api.project.definition_manager import DefinitionManager
+from snowflake.cli.plugins.nativeapp.init import nativeapp_init
+from snowflake.cli.plugins.nativeapp.manager import NativeAppManager
+from snowflake.cli.plugins.rpc.manager import LSPPluginContext
+from snowflake.snowpark.dataframe import DataFrame
+
+
+def lsp_plugin(name, version, capabilities):
```
sfc-gh-turbaszek, line 12: Please add typings

src/snowflake/cli/plugins/nativeapp/lsp_commands.py
```diff
@@ -0,0 +1,88 @@
...
+    return _decorator
+
+
+class ConnectionParams(TypedDict):
+    sessionToken: str
+    masterToken: str
+    account: str
+    host: str
+    port: str
+    protocol: str
```
sfc-gh-turbaszek, line 29: How different this is from `_ConnectionContext`?

src/snowflake/cli/plugins/nativeapp/lsp_commands.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+class CreateApplication(TypedDict):
```
sfc-gh-turbaszek, line 32: Why not dataclass?

src/snowflake/cli/plugins/nativeapp/lsp_commands.py
```diff
@@ -0,0 +1,88 @@
...
+)
+def nade_lsp_plugin(server: LanguageServer, ctx: LSPPluginContext):
+    @server.command("createApplication")
+    def create_app(params: list[CreateApplication]):
+        server.show_message_log(repr(params))
+        project = nativeapp_init(
+            path=params[0]["workspacePath"],
+            name=params[0]["applicationName"],
+            git_url=params[0]["templateRepo"],
+            template=params[0]["templateName"],
```
sfc-gh-turbaszek, line 60: Let's avoid using string keys. Imho dataclasses would be a better choice.

src/snowflake/cli/plugins/nativeapp/lsp_commands.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+@lsp_plugin(
+    name="nativeapp",
+    version="0.0.1",
+    capabilities={
+        "createApplication": True,
+        "openApplication": True,
+        "teardownApplication": True,
+    },
```
sfc-gh-turbaszek, line 50: Should this be an enum? We are duplicating the naming here and in `command` decorators

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -162,8 +162,13 @@ class NativeAppManager(SqlExecutionMixin):
     Base class with frequently used functionality already implemented and ready to be used by related subclasses.
     """
 
-    def __init__(self, project_definition: NativeApp, project_root: Path):
-        super().__init__()
+    def __init__(
+        self,
+        project_definition: NativeApp,
+        project_root: Path,
+        connection: SnowflakeConnection = None,
```
sfc-gh-turbaszek, line 169:
```suggestion
        connection: SnowflakeConnection | None = None,
```

src/snowflake/cli/plugins/rpc/commands.py
```diff
@@ -0,0 +1,48 @@
...
+)
+
+app = SnowTyper(
+    name="rpc",
+    help="Manages a RPC server for LSP.",
+)
+
+log = logging.getLogger(__name__)
+
+manager = None
```
sfc-gh-turbaszek, line 20: Let's try to avoid top level instances. Should we consider a factory approach?

src/snowflake/cli/plugins/rpc/manager.py
```diff
@@ -0,0 +1,124 @@
...
+MAX_WORKERS = 5
+
+
+class ConnectionParams(TypedDict):
+    sessionToken: Optional[str]
+    masterToken: Optional[str]
+    account: Optional[str]
+    host: Optional[str]
+    port: Optional[str]
+    protocol: Optional[str]
```
sfc-gh-turbaszek, line 23: Why this definition is duplicated? We already defined it in other file

src/snowflake/cli/plugins/rpc/manager.py
```diff
@@ -0,0 +1,124 @@
...
+                if "251005" in error_message:
+                    raise connector.ProgrammingError(
+                        "Failed to use the active connection from the Snowflake extension. Requires snowflake-python-connector version 3.7.0 or higher."
+                    )
+                else:
+                    raise
+            return conn
+        else:
+            conn = connector.connect()
+            return conn
```
sfc-gh-turbaszek, line 70: Why not reuse cli logic fro building a connection?

src/snowflake/cli/plugins/rpc/manager.py
sfc-gh-turbaszek: Why not use CLI standard for those variables?

src/snowflake/cli/plugins/rpc/manager.py
```diff
@@ -0,0 +1,124 @@
...
+
+class RpcManager(SqlExecutionMixin):
+    """
+    Base class LSP Controller
+    """
+
+    def __init__(self):
+        super().__init__()
+        server = LanguageServer(
+            name="lsp_controller", version="v0.0.1", max_workers=MAX_WORKERS
```
sfc-gh-turbaszek, line 100: I'm really worried in how many places we have version strings.

src/snowflake/cli/plugins/rpc/manager.py
```diff
@@ -0,0 +1,124 @@
...
+        )
+        context = LSPPluginContext()
+
+        plugins = self.discover_lsp_plugins("snowflake.cli.plugins")
+        for plugin_func in plugins:
+            plugin_func(server, context)
+        server.start_io()
+
+    def discover_lsp_plugins(self, package_name):
```
sfc-gh-turbaszek, line 109: This will need security consultation for risk of importing 3rd party modules.

sfc-gh-turbaszek (changes requested):
Left few comments, I'm mostly worried about how much code duplication we have at this moment.

src/snowflake/cli/plugins/rpc/manager.py
```diff
@@ -0,0 +1,124 @@
...
+import os
+import pkgutil
+from typing import Optional, TypedDict
+
+from pygls.server import LanguageServer
+from snowflake import connector
+from snowflake.cli.api.sql_execution import SqlExecutionMixin
+from snowflake.snowpark import Session
+
+MAX_WORKERS = 5
```
sfc-gh-turbaszek, line 14: This should be configurable from config.toml. We probably will need `[cli.lsp]` section.

src/snowflake/cli/plugins/rpc/manager.py
sfc-gh-klin: Yeah, VS Code is setting these. I will switch 'em over to the CLI standard with the "SNOWFLAKE" in front for consistency.

src/snowflake/cli/plugins/nativeapp/lsp_commands.py
sfc-gh-klin: The host/port/protocol part is different, since the Python connector allows for providing those values and I wanted to make sure it'd work for those situations too.

src/snowflake/cli/plugins/nativeapp/lsp_commands.py
sfc-gh-klin: Dataclass makes more sense since the rest of the repository is using it. I can switch over to using that instead.

sfc-gh-klin: I was partly mistaken. The contents that are coming in from the VS Code extension side are all dicts, so I will use TypedDict to indicate that. But past that part, I can use dataclass for the rest.

src/snowflake/cli/plugins/nativeapp/lsp_commands.py
sfc-gh-klin: I've used dataclasses where I am able to, and kept TypedDicts for the interface where the extension communicates with the language server since they are dicts when they come from the extension.

src/snowflake/cli/plugins/rpc/manager.py
sfc-gh-klin: Thanks, I will do that.

src/snowflake/cli/plugins/rpc/manager.py
sfc-gh-klin: I can remove it for now to keep it from distracting from the rest of it.

src/snowflake/cli/plugins/rpc/manager.py
sfc-gh-klin: We may have to have two. One where it's for the Dict coming in from the extension, and another where it is a dataclass used by the other layer.

src/snowflake/cli/plugins/rpc/manager.py
```diff
@@ -0,0 +1,62 @@
...
+                for _, submodule_name, is_pkg in pkgutil.iter_modules(
+                    submodule.__path__
+                ):
+                    module = importlib.import_module(
+                        f"{package_name}.{module_name}.{submodule_name}"
+                    )
+                    for _name, obj in inspect.getmembers(module, inspect.isfunction):
+                        if hasattr(obj, "_lsp_context"):
+                            plugins.append(obj)
+        return plugins
```
sfc-gh-klin, line 62: @sfc-gh-turbaszek I've tried to use Pluggy to replace this discovery mechanism and I am stumped. I tried following a small toy example of it from their [docs](https://pluggy.readthedocs.io/en/stable/#a-toy-example) but I don't understand how the automatic plugin discovery is supposed to work, because it seems like it either needs some mapping the way the command_plugins_loader.py has get_builtin_plugin_name, or it uses the setuptools_entrypoints which I don't think I should need since there's no need for external plugins here.

sfc-gh-klin: It looks like SnowCLI uses the hook markers but then uses its own mechanism to identify the plugins. The LoadedCommandPlugin has a command_spec that gets used in "_load_command_spec" in command_plugins_loader to return the plugin.command_spec(), and the contents of that occur in plugin_spec.py for each plugin. The command_spec() for native apps for example is:
```
@plugin_hook_impl
def command_spec():
    return CommandSpec(
        parent_command_path=SNOWCLI_ROOT_COMMAND_PATH,
        command_type=CommandType.COMMAND_GROUP,
        typer_instance=commands.app.create_instance(),
    )
```
And it seems like the @plugin_hook_impl from the pluggy marker is not really used/needed by the builtin plugin process since the code doesn't use pluggy for discovery of those builtins. e.g., removing @plugin_hook_impl, everything still works. 

So I guess I can follow the pattern there for command_spec but with another spec for the LSP stuff.

src/snowflake/cli/api/plugins/command/__init__.py
```diff
@@ -28,6 +28,8 @@
 
 plugin_hook_spec = pluggy.HookspecMarker(SNOWCLI_COMMAND_PLUGIN_NAMESPACE)
 plugin_hook_impl = pluggy.HookimplMarker(SNOWCLI_COMMAND_PLUGIN_NAMESPACE)
+lsp_plugin_hook_spec = pluggy.HookspecMarker(SNOWCLI_COMMAND_PLUGIN_NAMESPACE)
+lsp_plugin_hook_impl = pluggy.HookimplMarker(SNOWCLI_COMMAND_PLUGIN_NAMESPACE)
```
sfc-gh-cgorrie, line 32: Should these use the same namespace?

sfc-gh-klin: Yeah, they can use the same namespace and just use different hook markers. It seemed simpler to not need to configure another plugin manager for another namespace, since SnowCLI isn't using pluggy for discovery on builtins.

sfc-gh-cgorrie: You also have some other Markers in another file, are they dupes?

sfc-gh-klin: Yes, dupes. Let me figure out which to keep where.

src/snowflake/cli/api/plugins/command/plugin_hook_specs.py
```diff
@@ -12,10 +12,16 @@
...
 
 @plugin_hook_spec
 def command_spec():
     """Command spec"""
     pass
+
+
+@lsp_plugin_hook_spec
+def lsp_spec():
+    """Command spec"""
```
sfc-gh-tkojima, line 26: should update this description

src/snowflake/cli/app/commands_registration/command_plugins_loader.py
```diff
@@ -145,6 +145,20 @@ def _load_command_spec(plugin_name: str, plugin) -> Optional[CommandSpec]:
             )
             return None
 
+    # TODO: Guard against logging of all these exceptions of all modules trying to read lsp spec
```
sfc-gh-tkojima, line 148: can remove this comment

src/snowflake/cli/plugins/rpc/__init__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) 2024 Snowflake Inc.
```
sfc-gh-tkojima, line 1: folder name should be `lsp` and not `rpc`?

src/snowflake/cli/plugins/rpc/commands.py
```diff
@@ -0,0 +1,48 @@
...
+
+from snowflake.cli.api.commands.snow_typer import SnowTyper
+from snowflake.cli.api.output.types import CommandResult, MessageResult
+from snowflake.cli.plugins.rpc.manager import (
+    load_lsp_plugins,
+)
+
+app = SnowTyper(
+    name="rpc",
+    help="Manages a RPC server for LSP.",
```
sfc-gh-tkojima, line 28: > Manages a RPC server for LSP.

This is redundant, an LSP server is always technically an RPC server, so 
"Manages a Snowflake LSP server" is sufficient

src/snowflake/cli/plugins/nativeapp/lsp_commands.py
```diff
@@ -0,0 +1,111 @@
...
+            project_definition=project_definition,
+            project_root=project_root,
+            connection=connection,
+        )
+        if manager.get_existing_app_info():
+            url = manager.get_snowsight_url()
+            return MessageResult(f"{url}")
+        else:
+            return MessageResult(
+                'Snowflake Native App not yet deployed! Please run "snow app run" first.'
```
sfc-gh-cgorrie, line 110: As a matter of style, it would make sense to direct clients to run a different server_command rather than using CLI terminology here.

src/snowflake/cli/plugins/nativeapp/lsp_commands.py
```diff
@@ -0,0 +1,111 @@
...
+
+class ConnectionParams(TypedDict):
+    session_token: str
+    master_token: str
+    account: str
+    connection_name: str
+    params: dict
+
+
+def server_command(server: LanguageServer, command_name: str):
```
sfc-gh-cgorrie, line 51: Should this be in `plugins/lsp` rather than `plugins/nativeapp`?

src/snowflake/cli/plugins/rpc/manager.py
sfc-gh-turbaszek: Is this guaranteed somewhere? Are we sure 3rd party plugins can be loaded?

sfc-gh-turbaszek:
@sfc-gh-klin what's the plan for end to end testing of LSP plugin? How we make sure that cli changes won't break VSC?

src/snowflake/cli/plugins/lsp/commands.py
```diff
@@ -0,0 +1,48 @@
...
+from snowflake.cli.api.commands.snow_typer import SnowTyper
+from snowflake.cli.api.output.types import CommandResult, MessageResult
+from snowflake.cli.plugins.lsp.manager import (
+    load_lsp_plugins,
+)
+
+app = SnowTyper(
+    name="lsp",
+    help="Manages a Snowflake LSP server.",
+)
```
sfc-gh-turbaszek, line 29: Let's make it hidden, at least for know 
```suggestion
    hidden=True,
)
```

src/snowflake/cli/plugins/nativeapp/lsp_commands.py
```diff
@@ -0,0 +1,111 @@
...
+                "lsp_plugin_version": version,
+                "lsp_plugin_capabilities": capabilities,
+            },
+        )
+        return func
+
+    return _decorator
+
+
+class ConnectionParams(TypedDict):
```
sfc-gh-turbaszek, line 43: How does this releates to `ConnectionDict`?

src/snowflake/cli/plugins/lsp/commands.py
sfc-gh-klin: Very well

src/snowflake/cli/plugins/nativeapp/lsp_commands.py
sfc-gh-klin: I deleted the duplicate one, sorry about that.

src/snowflake/cli/plugins/nativeapp/lsp_commands.py
sfc-gh-klin: Yes! I moved the decorator and the parts intended to be used by other plugin areas to make their own lsp_plugins into the lsp/utils.py file. I'm not quite sure utils is the best name for it since other util files in this repo tend to be pretty small non-core utility functions, but the file has a collection of functions for use.

src/snowflake/cli/plugins/rpc/manager.py
sfc-gh-klin: Okay, I think it works out if we have a single version number for the lsp plugin, since that's what from pyglas gets passed into ServerInfo to the client during init.

sfc-gh-klin:
> @sfc-gh-klin what's the plan for end to end testing of LSP plugin? How we make sure that cli changes won't break VSC?

I think having integration tests that start up the LSP server and then tests the interface would help ensure CLI changes don't break the extension. I'm not too well versed with writing integration tests in Python though so any pointers would be appreciated.

src/snowflake/cli/plugins/rpc/manager.py
sfc-gh-klin: Using pluggy seemed more official and in line with the code in the repo than the way I was doing it before. Did you prefer the previous version that didn't use pluggy? In the end SnowCLI has its own mechanism to find the plugins; the plugin.command_spec and plugin.lsp_spec, so it isn't automatic discovery for the internal plugins. That's my current understanding.

src/snowflake/cli/plugins/rpc/manager.py
sfc-gh-klin: The way I am doing it now follows the path of calling load_only_builtin_command_plugins

```
def load_only_builtin_command_plugins() -> List[LoadedCommandPlugin]:
    loader = CommandPluginsLoader()
    loader.register_builtin_plugins()
    return loader.load_all_registered_plugins()

```

"get_builtin_plugin_name_to_plugin_spec" so it's only the stuff in builtin_plugins.py, which restricts it to the internal ones. 

See `load_lsp_plugins` for where it calls the `load_only_builtin_command_plugins`.

sfc-gh-klin:
I tried to set up integration tests but my test is hanging and I'm trying to debug why. Do you guys have tips for how to get it to log some information when running? I'm running with python -m hatch <test marker> to run my one integration test and naively trying to add repr() or logging.getLogger(__name__) and trying to log information doesn't work through that method.

sfc-gh-klin:
🫡

PR #1232, opened 2024-06-19
Added info log about loaded external plugins

sfc-gh-astus:
### Pre-review checklist

### Changes description
* Added info log message about loaded external plugins
* Added test

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -45,6 +45,7 @@
 * The `snow sql` command will show query text before executing it.
 * Improved stage diff output in `snow app` commands
 * Hid the diff from `snow app validate` output since it was redundant
+* Added log into the file with loaded external plugins
 
 # v2.5.0
 ## Backward incompatibility
--- a/src/snowflake/cli/app/__init__.py
+++ b/src/snowflake/cli/app/__init__.py
@@ -15,8 +15,12 @@
 import logging
 import os
 
+from snowflake.cli.app import loggers
+
 # Suppress logging from Snowflake connector
 logging.getLogger("snowflake").setLevel(logging.ERROR)
+
+loggers.create_initial_loggers()
 
 # Restrict permissions of all created files
 os.umask(0o077)
--- a/src/snowflake/cli/app/__init__.py
+++ b/src/snowflake/cli/app/__init__.py
@@ -15,12 +15,8 @@
 import logging
 import os
 
-from snowflake.cli.app import loggers
-
 # Suppress logging from Snowflake connector
 logging.getLogger("snowflake").setLevel(logging.ERROR)
-
-loggers.create_initial_loggers()
 
 # Restrict permissions of all created files
 os.umask(0o077)
--- a/src/snowflake/cli/app/commands_registration/command_plugins_loader.py
+++ b/src/snowflake/cli/app/commands_registration/command_plugins_loader.py
@@ -100,6 +100,10 @@ class CommandPluginsLoader:
         self._loaded_command_paths[
             loaded_plugin.command_spec.full_command_path
         ] = loaded_plugin
+
+        if self._is_external_plugin(loaded_plugin):
+            log.info("Loaded external plugin: %s", plugin_name)
+
         return loaded_plugin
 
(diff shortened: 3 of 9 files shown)


src/snowflake/cli/app/commands_registration/command_plugins_loader.py
```diff
@@ -100,6 +100,10 @@ def _load_new_plugin(
         self._loaded_command_paths[
             loaded_plugin.command_spec.full_command_path
         ] = loaded_plugin
+
+        if self._is_external_plugin(loaded_plugin):
+            log.info("Loaded external plugin: %s", plugin_name)
+
```
sfc-gh-mraba, line 106: Does it make sense to log all loaded plugins?
Just mark them as internal/external.

sfc-gh-mraba (dismissed):

src/snowflake/cli/app/commands_registration/command_plugins_loader.py
sfc-gh-astus: I thought about it and I'm not sure about internal, for now we have around 10 and log them every time you use Snowflake CLI is waste of disc space for me.

src/snowflake/cli/app/__init__.py
```diff
@@ -12,11 +12,11 @@
...
 # limitations under the License.
 
-import logging
 import os
 
-# Suppress logging from Snowflake connector
-logging.getLogger("snowflake").setLevel(logging.ERROR)
+from snowflake.cli.app import loggers
+
+loggers.create_initial_loggers()
```
sfc-gh-pjob, line 19: Did you test that custom log file path is correctly read (that it is set before invoking this line)?

src/snowflake/cli/app/commands_registration/command_plugins_loader.py
sfc-gh-pjob: This log is added only to be sure which plugins were loaded during CLI's execution as kind of audit logs. They are helpful because user can enable/disable external plugins. It's not possible to disable builtin plugins, so IMO there is no point of logging them (especially on INFO level).

sfc-gh-pjob (changes requested):

src/snowflake/cli/app/__init__.py
sfc-gh-astus: Added test.

tests_e2e/test_error_handling.py
```diff
@@ -91,3 +92,40 @@ def test_corrupted_config_in_default_location(
...
+@pytest.mark.e2e
+def test_initial_log_with_loaded_external_plugins_in_custom_log_path(
+    snowcli, temp_dir, isolate_default_config_location, test_root_path
+):
+    default_config = Path(temp_dir) / "config.toml"
+    default_config.write_text(
+        dedent(
+            f"""
+    [cli.logs]
+    path = "{temp_dir}/logs"
```
sfc-gh-pjob, line 106: Maybe I'm wrong but I think this is the same path as the default one in this case (`temp_dir / logs`), so this test isn't testing what it says :)

sfc-gh-turbaszek: FYI this path style is platform dependent, we should use os.path.join

sfc-gh-astus: I changed path and used os.path.join

sfc-gh-pjob (approved):

sfc-gh-pjob (approved):

RELEASE-NOTES.md
```diff
@@ -45,6 +45,7 @@
 * The `snow sql` command will show query text before executing it.
 * Improved stage diff output in `snow app` commands
 * Hid the diff from `snow app validate` output since it was redundant
+* Added log into the file with loaded external plugins
```
sfc-gh-turbaszek, line 48:
```suggestion
* Added log info the file with loaded external plugins
```

sfc-gh-turbaszek, line 48: I'm not sure I fully understand

sfc-gh-turbaszek (approved):

PR #1233, opened 2024-06-19
[NADE][Bugfix] Convert to stagepaths before writing to SQL files

sfc-gh-bgoel:
### Pre-review checklist

### Changes description
On windows, there is currently a bug where appending relative file paths to a sql file that will be uploaded to a Snowflake stage is done incorrectly. 
E.g. if a file `__generated.sql` exists in a path `some\path\__generated\__generated.sql`, then the correct string to append to the sql file should be `EXECUTE IMMEDIATE FROM '/__generated/__generated.sql';`, but the bug currently causes this: `EXECUTE IMMEDIATE FROM '/__generated\__generated.sql';`. This PR is to correct the bug.

### DIFF
--- a/src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
+++ b/src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
@@ -510,7 +510,9 @@ def edit_setup_script_with_exec_imm_sql(
             sql_file_relative_path = sql_file.relative_to(
                 deploy_root
             )  # Path on stage, without the leading slash
-            file.write(f"EXECUTE IMMEDIATE FROM '/{sql_file_relative_path}';\n")
+            file.write(
+                f"EXECUTE IMMEDIATE FROM '/{to_stage_path(sql_file_relative_path)}';\n"
+            )
 
     # Find the setup script in the deploy root.
     setup_file_path = find_setup_script_file(deploy_root=deploy_root)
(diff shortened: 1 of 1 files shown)


src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -510,7 +510,9 @@ def edit_setup_script_with_exec_imm_sql(
             sql_file_relative_path = sql_file.relative_to(
                 deploy_root
             )  # Path on stage, without the leading slash
-            file.write(f"EXECUTE IMMEDIATE FROM '/{sql_file_relative_path}';\n")
+            file.write(
```
sfc-gh-bgoel, line 513: line 231 stays as is, with output like:
`Generated by the Snowflake CLI from cli_gen\accepted.py`
Do we also want to change this comment to be stage path?

sfc-gh-bdufour: "generated from" should be the source file, i.e. a local path in the project root, not a dest path / stage path.

sfc-gh-bdufour:
Needs tests so that we can prevent future regressions.

sfc-gh-bdufour:
> Needs tests so that we can prevent future regressions.

Discussed offline, without E2E tests on Windows we are left with very few options, all of them rather bad. We should focus on getting E2E set up for Windows + backfill the tests. Approving this as is for now.

sfc-gh-bdufour (approved):

PR #1241, opened 2024-06-20
Run E2E tests on windows runners

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
Run e2e tests on windows to cover more cases.

The test matrix is unified for test and build jobs:
- on schedule run on  `[windows, macos, ubuntu] x [3.8, 3.9, 3.10. 3.11, 3.12]`
- else (PR, push)  run on  `[windows, ubuntu] x [3.8, 3.10]`

Decisions taken:
1. Put windows as first in matrix as it's the biggest possible source of errors (we use unix locally)
2. Use fail fast for jobs (if windows fails, there's a bug we need to fix), and use fail fast for tests (after 10 tests).
3. PR/push matrix is uses lowest python and most frequently used one. Macos is excluded and very similar to ubuntu, we don't have much of mac-specific code.


pyproject.toml
```diff
@@ -147,7 +147,7 @@ select = [
 ]
 
 [tool.pytest.ini_options]
-addopts = "-m 'not integration and not performance and not e2e and not spcs and not loaded_modules and not integration_experimental'"
+addopts = "-vv --maxfail=10 -m 'not integration and not performance and not e2e and not spcs and not loaded_modules and not integration_experimental'"
```
sfc-gh-pjob, line 150: `--maxfail=10` - do we want to keep it? Do I correctly understand that in case of less failures, the final result will still a general failure?

tests_e2e/conftest.py
```diff
@@ -50,36 +96,49 @@ def temp_dir():
...
+            yield tmp_dir_path / "Scripts" / "snow.exe"
+        else:
+            yield tmp_dir_path / "bin" / "snow"
 
 
 @pytest.fixture(autouse=True)
 def isolate_default_config_location(monkeypatch, temp_dir):
     monkeypatch.setenv("SNOWFLAKE_HOME", temp_dir)
+    monkeypatch.setenv("_CLI_TESTS_CONTEXT", "True")
+    # monkeypatch.setenv("TERM", "unknown")
```
sfc-gh-pjob, line 124: can we remove this commented line?

pyproject.toml
sfc-gh-turbaszek: Yes. When we discussed this few days ago we agreed that if too many test are failing it means a wide spread problem. Using random order makes this indeterministic but still there's a small optimization for changes that are not well tested.

tests_e2e/conftest.py
sfc-gh-turbaszek: Removed

tests_e2e/conftest.py
```diff
@@ -25,6 +27,50 @@
 
 TEST_DIR = Path(__file__).parent
 
+IS_WINDOWS = platform.system() == "Windows"
+
+
+def _clean_output(text: str):
+    """
+    Replacing util to clean up console output. Typer is using rich.Panel to show the --help content.
```
sfc-gh-astus, line 35: `rich. Panel`

sfc-gh-astus (approved):

sfc-gh-pjob (approved):

.github/workflows/test_e2e.yaml
```diff
@@ -9,22 +9,32 @@ on:
       - "v*" # Push events to matching v*, i.e. v1.0, v20.15.10
     branches:
       - main
+  schedule:
+    - cron: "0 8 * * *"
```
sfc-gh-bdufour, line 13: So that means we're going to trigger these e2e tests on main every morning then? If there a specific even that triggered this change, or we're just being extra careful?

sfc-gh-bdufour (approved):

sfc-gh-jvasquezrojas (approved):

.github/workflows/test_e2e.yaml
sfc-gh-turbaszek: > we're just being extra careful

On scheduled run the matrix is wider (all python, all runners (macos added)). But yeah, it's more or less being extra careful

PR #1242, opened 2024-06-20
Add support for --env overrides from command line for templating

sfc-gh-melnacouzi:
### Pre-review checklist

### Changes description
Add support for CLI argument --env to allow users to overwrite env variables for templating.
User can provide multiple --env variables (e.g. --env x=1 --env y=a)
User can set empty variable (--env x=) or (--env x='')
Order of template resolution priority:
1) --env vars from cli
2) os env variables
3) env variables defined in project definition file.

This does not alter environment variables.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -30,6 +30,10 @@
   * Format of the argument: `--env key1=value1 --env key2=value2`.
   * Overrides `env` variables values when used in templating.
   * Can be referenced in templating through `ctx.env.<key_name>`.
+  * Templating will read env vars in this order of priority (highest priority to lowest priority):
+    * vars from `--env` command line argument.
+    * vars from shell environment variables.
+    * vars from `env` section of project definition file.
 
 ## Fixes and improvements
 * Passing a directory to `snow app deploy` will now deploy any contained file or subfolder specified in the application's artifact rules
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -25,6 +25,11 @@
 * Added `--auto-compress` flag to `snow stage copy` command enabling use of gzip to compress files during upload.
 * Added new `native_app.application.post_deploy` section to `snowflake.yml` schema to execute actions after the application has been deployed via `snow app run`.
   * Added the `sql_script` hook type to run SQL scripts with template support.
+* Added support for `--env` command line arguments for templating.
+  * Available for commands that make use of the project definition file.
+  * Format of the argument: `--env key1=value1 --env key2=value2`.
+  * Overrides `env` variables values when used in templating.
+  * Can be referenced in templating through `ctx.env.<key_name>`.
 
 ## Fixes and improvements
 * Passing a directory to `snow app deploy` will now deploy any contained file or subfolder specified in the application's artifact rules
--- a/src/snowflake/cli/api/cli_global_context.py
+++ b/src/snowflake/cli/api/cli_global_context.py
@@ -227,7 +227,7 @@ class _CliGlobalContextManager:
         self._project_definition = None
         self._project_root = None
         self._project_path_arg = None
-        self._project_env_overrides_args = []
+        self._project_env_overrides_args = {}
         self._typer_pre_execute_commands = []
         self._template_context = None
         self._silent: bool = False
@@ -285,10 +285,12 @@ class _CliGlobalContextManager:
         self._project_path_arg = project_path_arg
 
     @property
-    def project_env_overrides_args(self) -> list[str]:
+    def project_env_overrides_args(self) -> dict[str, str]:
         return self._project_env_overrides_args
 
-    def set_project_env_overrides_args(self, project_env_overrides_args: list[str]):
+    def set_project_env_overrides_args(
+        self, project_env_overrides_args: dict[str, str]
+    ):
         self._project_env_overrides_args = project_env_overrides_args
 
     @property
--- a/src/snowflake/cli/api/cli_global_context.py
+++ b/src/snowflake/cli/api/cli_global_context.py
@@ -16,7 +16,7 @@ from __future__ import annotations
 
 import re
 from pathlib import Path
-from typing import Dict, Optional
+from typing import Callable, Optional
 
 from snowflake.cli.api.exceptions import InvalidSchemaError
 from snowflake.cli.api.output.formats import OutputFormat
@@ -226,6 +226,9 @@ class _CliGlobalContextManager:
         self._experimental = False
         self._project_definition = None
         self._project_root = None
+        self._project_path_arg = None
+        self._project_env_overrides_args = []
+        self._typer_pre_execute_commands = []
         self._template_context = None
         self._silent: bool = False
 
@@ -261,7 +264,7 @@ class _CliGlobalContextManager:
         self._experimental = value
 
     @property
-    def project_definition(self) -> Optional[Dict]:
+    def project_definition(self) -> Optional[ProjectDefinition]:
         return self._project_definition
 
     def set_project_definition(self, value: ProjectDefinition):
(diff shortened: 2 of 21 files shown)


RELEASE-NOTES.md
```diff
@@ -21,6 +21,11 @@
 
 ## New additions
 * Added support for `title` field in Streamlit definition in `snowflake.yml` project file.
+* Added support for `--env` command line arguments for templating.
+  * Available for commands that make use of the project definition file.
+  * Format of the argument: `--env key1=value1 --env key2=value2`.
+  * Overrides `env` variables values when used in templating.
```
sfc-gh-cgorrie, line 27: Does it override system-defined environment variables? I'm guessing yes

sfc-gh-bdufour:
Needs a proper description? That's why the checkbox exists.

RELEASE-NOTES.md
sfc-gh-melnacouzi: no, it does not alter them.

sfc-gh-turbaszek: Can you explain the hierarchy for the inputs?

src/snowflake/cli/api/cli_global_context.py
```diff
@@ -274,13 +277,36 @@ def project_root(self):
...
 
+    @property
+    def project_path_arg(self) -> Optional[str]:
+        return self._project_path_arg
+
+    def set_project_path_arg(self, project_path_arg: str):
+        self._project_path_arg = project_path_arg
+
+    @property
+    def project_env_overrides_args(self) -> list[str]:
```
sfc-gh-turbaszek, line 288: Wouldn't it be better to use some kind of dictionary instead of list?

src/snowflake/cli/api/commands/flags.py
```diff
@@ -499,70 +501,83 @@ def execution_identifier_argument(sf_object: str, example: str) -> typer.Argumen
...
     return typer.Option(
         None,
         "-p",
         "--project",
-        help=f"Path where the {project_name_help} project resides. "
+        help=f"Path where {_get_project_long_name(project_name)} project resides. "
         f"Defaults to current working directory.",
-        callback=_callback,
+        callback=_callback(lambda: project_definition_callback),
         show_default=False,
```
sfc-gh-turbaszek, line 571: I think there are two changes that would be beneficial and would simplify the implementation:
1. Getting rid of `project_name`, it won't be used in v2, will remove the "long name" mapping. Removing it means we need to add proper validation in all commands requiring part of project definition.
2. `is_optional` does not have to be part of `register_project_definition`, we can use it as input for `exists` in `type.Option`. If project is optional the file may not exists and that's fine. If project is required, the file must exists.

Today project definition handling is tightly coupled with "project type" and this is a source of now convoluted code (it was good idea some time ago).

My suggestion would be to first to the above changes in separate PR to decouple refactoring and new features. I'm happy to contribute.

RELEASE-NOTES.md
sfc-gh-melnacouzi: Sure i can add that.

src/snowflake/cli/api/cli_global_context.py
sfc-gh-melnacouzi: these come directly from the cli. I can update the typer callback to transform them before storing them.

src/snowflake/cli/api/commands/flags.py
sfc-gh-melnacouzi: 1. and 2. I do not want to add behavior changes on this.
Did you mean to add 1. and 2. in a separate PR? or move some of the current changes into a separate PR?

sfc-gh-melnacouzi: We had an offline discussion. I will remove project_name altogether as a follow up PR to unblock this feature.

RELEASE-NOTES.md
```diff
@@ -22,6 +22,15 @@
...
 * Added `--auto-compress` flag to `snow stage copy` command enabling use of gzip to compress files during upload.
+* Added support for `--env` command line arguments for templating.
+  * Available for commands that make use of the project definition file.
+  * Format of the argument: `--env key1=value1 --env key2=value2`.
+  * Overrides `env` variables values when used in templating.
+  * Can be referenced in templating through `ctx.env.<key_name>`.
+  * Templating will read env vars in this order of priority (highest priority to lowest priority):
+    * vars from `--env` command line argument.
+    * vars from shell environment variables.
+    * vars from `env` section of project definition file.
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: README for snow CLI seems to be minimal (installation instruction). This will go into public documentation - which I am working on separately.

src/snowflake/cli/api/commands/decorators.py
```diff
@@ -136,10 +146,15 @@ def _options_decorator_factory(
...
+    """
+    execute_before_command_using_new_options executes before command telemetry has been emitted,
+    but after command line options have been populated.
+    """
+
     @wraps(func)
     def wrapper(**options):
         if execute_before_command_using_new_options:
-            execute_before_command_using_new_options()
+            execute_before_command_using_new_options(**options)
```
sfc-gh-turbaszek, line 157: Should it be `_execute_before_command_using_global_options`? @sfc-gh-pjob wdyt about using this for registering the callbacks?

src/snowflake/cli/api/commands/flags.py
```diff
@@ -499,70 +501,86 @@ def execution_identifier_argument(sf_object: str, example: str) -> typer.Argumen
...
+            project_root = None
+            template_context = {
+                CONTEXT_KEY: {
+                    "env": ProjectEnvironment(
+                        default_env={}, override_env=env_overrides_args
+                    )
+                }
+            }
+        else:
+            raise
```
sfc-gh-turbaszek, line 533: Can we make this using if's instead of raising and handling exception?

src/snowflake/cli/api/commands/decorators.py
sfc-gh-melnacouzi: it's not restricted to global options. it gets all the parameters of the current command.

src/snowflake/cli/api/commands/flags.py
sfc-gh-melnacouzi: yes - refactored.

src/snowflake/cli/api/commands/decorators.py
sfc-gh-pjob: @sfc-gh-turbaszek I'm not sure what exactly is the subject of your question.

Maybe I miss sth in the changes but I think changes in this function `_options_decorator_factory` are only cleanup changes and aren't responsible for the effect of this PR. @sfc-gh-melnacouzi please correct me if I'm wrong. And I think these changes are good changes but maybe not necessary should be part of this PR (but IMO they can be).

@sfc-gh-turbaszek I think all the magic goes via `typer_pre_execute.py` and I'm not sure about keeping functions to run before commands in our context but IMO it looks OK and simple to use.

RELEASE-NOTES.md
```diff
@@ -22,6 +22,15 @@
 ## New additions
 * Added support for `title` field in Streamlit definition in `snowflake.yml` project file.
 * Added `--auto-compress` flag to `snow stage copy` command enabling use of gzip to compress files during upload.
+* Added support for `--env` command line arguments for templating.
```
sfc-gh-turbaszek, line 25: Can I override streamlit name? Probably by doing something like `--env="streamlit.name=foo"`?

src/snowflake/cli/api/commands/flags.py
```diff
@@ -499,70 +500,74 @@ def execution_identifier_argument(sf_object: str, example: str) -> typer.Argumen
...
 
     return typer.Option(
-        None,
-        "-p",
-        "--project",
-        help=f"Path where Snowflake project resides. Defaults to current working directory.",
-        callback=_callback,
+        [],
+        "--env",
+        help="String in format of key=value. Overrides env variables used for templating.",
```
sfc-gh-turbaszek, line 569:
```suggestion
        help="String in format of key=value. Overrides variables from env section used for templating.",
```

src/snowflake/cli/api/commands/decorators.py
sfc-gh-turbaszek: My concern with this approach is the implicit order of callbacks. Callbacks will be registered in the order of parsing/evaluating commands, this is a very deep implementation detail imho.

RELEASE-NOTES.md
sfc-gh-melnacouzi: no. Overrides only apply to env section.

sfc-gh-turbaszek (approved):

sfc-gh-pjob (dismissed):

sfc-gh-cgorrie (commented):
@sfc-gh-melnacouzi quick question -- what happens if the same env var is provided twice by two subsequent `--env` args? e.g.

`snow app run --env ABC=123 --env ABC=456`

What will `ABC` be set to -- `456`? Or will this be an error?

sfc-gh-turbaszek (dismissed):

sfc-gh-turbaszek (approved):

sfc-gh-bdufour (approved):

PR #1244, opened 2024-06-20
App post-deploy hook: sql scripts

sfc-gh-gbloom:
### Pre-review checklist

### Changes description

Adding a `application.post_deploy` section to `snowflake.yml` for users to be able to run SQL scripts after a dev-mode application object has been created/upgraded (after running `snow app run`).

These post-deploy hooks are not bundled into the application package and will not have an effect on consumers.

The first supported post-deploy hook type is `sql_script`:
```yml
definition_version: 1
native_app:
  name: myapp
  artifacts:
    ...
  package:
    ...
  application:
    post_deploy:
      - sql_script: scripts/post_deploy1.sql
      - sql_script: scripts/post_deploy2.sql
```

The script paths are relative to project root.

By default, the scripts use the project warehouse, or the connection warehouse. If a connection database has been set, it will be used by default as well.

Template variables in the SQL scripts are expanded:
```sql
SELECT &{ ctx.native_app.name };
SELECT &{ ctx.env.foo };
```

### Test

Manually tested running `snow app run` on a new application with 2 `post_deploy` scripts (similar to the configuration above). Verified that both scripts were executed on the first deployment when the application is created, and on the following run when the application is upgraded. Also verified that template variables are expanded correctly.

For the integration test, added a similar flow where each script writes to a temporary `post_deploy_log` table to verify both scripts are executed both on application creation and upgrade.


tests/__snapshots__/test_help_messages.ambr
```diff
@@ -7105,79 +7105,6 @@
   ╰──────────────────────────────────────────────────────────────────────────────╯
   
   
-  '''
-# ---
-# name: test_help_messages[streamlit.'\x07']
```
sfc-gh-fcampbell, line 7110: why is this deleted?

src/snowflake/cli/api/project/schemas/native_app/application.py
```diff
@@ -40,3 +46,7 @@ class Application(UpdatableModel):
         title="Whether to enable debug mode when using a named stage to create an application object",
         default=True,
     )
+    post_deploy: Optional[List[ApplicationPostDeployHook]] = Field(
+        title="Steps that will be executed after the application object is created/upgraded",
+        default=None,
+    )
```
sfc-gh-fcampbell, line 52: can we make this an empty list instead of None?

tests/__snapshots__/test_help_messages.ambr
sfc-gh-gbloom: Looks like it was recently added in #1238 -- seems like a mistake to me. It's also removed when running `pytest --snapshot-update` on `main`.
@sfc-gh-turbaszek Can you confirm?

RELEASE-NOTES.md
```diff
@@ -22,6 +22,10 @@
 ## New additions
 * Added support for `title` field in Streamlit definition in `snowflake.yml` project file.
 * Added `--auto-compress` flag to `snow stage copy` command enabling use of gzip to compress files during upload.
+* Added new `native_app.application.post_deploy` section to `snowflake.yml` schema to run SQL scripts after the application has been deployed.
```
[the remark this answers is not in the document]
sfc-gh-gbloom: Clarified in the following bullets

src/snowflake/cli/api/project/schemas/native_app/application.py
sfc-gh-gbloom: It was my intuition at first too, but it's not consistent with other similar existing properties. This way we will also lose the notion of the user not specifying any value and specifying an empty list.
Any strong arguments for not keeping it `None`?

RELEASE-NOTES.md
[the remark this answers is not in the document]
sfc-gh-bdufour: These bullets feel way, way too detailed for release notes. We should amend the Snowflake CLI docs for native apps instead, as this change extends the project definition schema anyway and it's required anyway.

src/snowflake/cli/api/project/schemas/native_app/application.py
```diff
@@ -40,3 +46,7 @@ class Application(UpdatableModel):
         title="Whether to enable debug mode when using a named stage to create an application object",
         default=True,
     )
+    post_deploy: Optional[List[ApplicationPostDeployHook]] = Field(
+        title="Steps that will be executed after the application object is created/upgraded",
```
sfc-gh-bdufour, line 50: Steps -> Actions maybe?

RELEASE-NOTES.md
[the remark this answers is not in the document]
sfc-gh-bdufour: I'd just consider saying something like "[...] has been deployed via `snow app run`"

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -264,6 +264,16 @@ def app_role(self) -> str:
         else:
             return self._default_role
 
+    @cached_property
+    def app_post_deploy_hooks(self) -> Optional[str]:
```
sfc-gh-bdufour, line 268: doesn't it return a list of them?

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -141,6 +142,34 @@ def _create_dev_app(self, diff: DiffResult) -> None:
...
+                self._execute_queries(sql_script)
+            except ProgrammingError as err:
+                generic_sql_error_handler(err)
+
+    def _execute_post_deploy_hooks(self):
+        post_deploy_script_hooks = self.app_post_deploy_hooks
+        if post_deploy_script_hooks:
+            for hook in post_deploy_script_hooks:
+                if hook.sql_script:
+                    cc.step(
```
sfc-gh-bdufour, line 164: consistency: consider using a phase here with one step per script. That's what we do for codegen.

tests/nativeapp/test_post_deploy.py
```diff
@@ -0,0 +1,74 @@
...
+    with project_directory("napp_post_deploy") as project_dir:
+        processor = _get_run_processor(str(project_dir))
+
+        processor._execute_post_deploy_hooks()  # noqa SLF001
+
+        assert mock_execute_query.mock_calls == [
+            mock.call("use warehouse MockWarehouse"),
+            mock.call("use database MockDatabase"),
+            mock.call("use warehouse MockWarehouse"),
+            mock.call("use database MockDatabase"),
```
sfc-gh-bdufour, line 56: I assume we repeat these for each script because we're worried the script can issue a use warehouse/database command and we need to reset? Worth documenting IMHO (in the source)

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-gbloom: Yep, good catch!

tests/nativeapp/test_post_deploy.py
sfc-gh-gbloom: Yep, I wish there was a `WITH` SQL statement.
I'll add a description to that method.

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -141,6 +143,38 @@ def _create_dev_app(self, diff: DiffResult) -> None:
...
+                    self._execute_query(f"use database {self._conn.database}")
+                sql_script = snowflake_sql_jinja_render(content=sql_script)
+                self._execute_queries(sql_script)
+            except ProgrammingError as err:
+                generic_sql_error_handler(err)
+
+    def _execute_post_deploy_hooks(self):
+        post_deploy_script_hooks = self.app_post_deploy_hooks
+        if post_deploy_script_hooks:
+            with cc.phase("Executing application post-deploy actions..."):
```
sfc-gh-bdufour, line 168: nit: lose the '...' for consistency with other commands.

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -141,6 +143,38 @@ def _create_dev_app(self, diff: DiffResult) -> None:
...
+                for hook in post_deploy_script_hooks:
+                    if hook.sql_script:
+                        cc.step(f"Executing SQL script: {hook.sql_script}")
+                        self._execute_sql_script(hook.sql_script)
+                    else:
+                        raise ValueError(
+                            f"Unsupported application post-deploy hook type: {hook}"
+                        )
```
sfc-gh-bdufour, line 176: Consider validating upfront so we don't spend time/credits executing something just to have the whole flow fail later on. Better CX that way.

tests/__snapshots__/test_help_messages.ambr
sfc-gh-bdufour: What's the latest on this?

tests/nativeapp/test_post_deploy.py
```diff
@@ -0,0 +1,94 @@
...
+                select bar;
+                """
+                )
+            ),
+            mock.call("-- app post-deploy script (2/2)\n"),
+        ]
+
+
+@mock_connection()
+def test_missing_sql_script(
```
sfc-gh-bdufour, line 85: Seems like we're missing a few tests to cover the entire contract. For example, you raise a ValueError if the sql_script key isn't present, but I don't see it covered here. Same with covering the fact that warehouse/db activation are optional. There might be others, please make sure the entire contract is accurately represented here.

RELEASE-NOTES.md
```diff
@@ -22,6 +22,7 @@
 ## New additions
 * Added support for `title` field in Streamlit definition in `snowflake.yml` project file.
 * Added `--auto-compress` flag to `snow stage copy` command enabling use of gzip to compress files during upload.
+* Added new `native_app.application.post_deploy` section to `snowflake.yml` schema to run SQL scripts after the application has been deployed via `snow app run`.
```
sfc-gh-melnacouzi, line 25: we should mention sql_script as well.

src/snowflake/cli/api/project/schemas/native_app/application.py
```diff
@@ -23,6 +23,12 @@
 )
 
 
+class ApplicationPostDeployHook(UpdatableModel):
```
sfc-gh-melnacouzi, line 26: have you tried this with multiple options to see if it works when we expand it in the future?

sfc-gh-melnacouzi: you might have to add @root_validator to ensure that one and only one of the attributes is defined.

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-gbloom: This validation is in addition to the schema validation which catches non-supported hook types on command execution:
```
During evaluation of ApplicationPostDeployHook in project definition following errors were encountered:
Extra inputs are not permitted. You provided field '('non_existing_hook',)' with value 'scripts/post_install.sql' that is not supported in given version.
```

src/snowflake/cli/api/project/schemas/native_app/application.py
sfc-gh-gbloom: Yes, that's a good suggestion. I think we should add this once there's actually a second attribute, because now it would be an unreachable code path.

Tried to add a second attribute locally and verified it works, including the validation:

```python
class ApplicationPostDeployHook(UpdatableModel):
    sql_script: Optional[str] = Field(
        title="SQL file path relative to the project root", default=None
    )
    grant: Optional[str] = Field(
        title="Grant a permission", default=None
    )

    @model_validator(mode='before')
    @classmethod
    def ensure_exactly_one_hook_type(cls, hook):
        supported_keys = cls.model_fields.keys()
        actual_keys = hook.keys()
        previously_found_key = None
        for key in actual_keys:
            if key in supported_keys:
                if previously_found_key:
                    raise ValueError(f"Only one of the following keys can be specified: {supported_keys}")
                else:
                    previously_found_key = True
        if not previously_found_key:
            raise ValueError(f"One of the following keys must be specified: {supported_keys}")

        return hook
```

```yml
# Works
post_deploy: 
  - sql_script: scripts/post_install.sql

# Works
post_deploy: 
  - grant: grant

# Error
post_deploy: 
  - sql_script: scripts/post_install.sql
  - grant: grant

# Error
post_deploy: 
  - something: else
```

tests/__snapshots__/test_help_messages.ambr
sfc-gh-turbaszek: I can't replicate this on current main. Nevertheless I will change test so we are treating unused snapshots as error.

src/snowflake/cli/api/project/schemas/native_app/application.py
sfc-gh-melnacouzi: But you can use this to ensure there is exactly one attribute in use. In this case, if someone doesn't enter one attribute (sql_script for example), it should fail (even though it says optional).

tests/__snapshots__/test_help_messages.ambr
sfc-gh-gbloom: Removed the snapshot update from this PR, but this help section is still removed for me on main after running `--snapshot-update`

src/snowflake/cli/api/project/schemas/native_app/application.py
sfc-gh-fcampbell: nah, I was just trying to avoid None as much as possible, but differentiating between no list provided vs an explicit empty list makes sense

src/snowflake/cli/api/project/schemas/native_app/application.py
sfc-gh-gbloom: Good suggestion. Added custom validation to catch these 2 cases:
```yml
post_deploy: 
  - {}
```
```yml
post_deploy: 
  -
```

src/snowflake/cli/api/project/schemas/native_app/application.py
```diff
@@ -14,15 +14,47 @@
...
+    @model_validator(mode="before")
+    @classmethod
+    def ensure_exactly_one_hook_type(cls, hook):
+        supported_keys = cls.model_fields.keys()
+        supported_keys_str = ", ".join(supported_keys)
+        if type(hook) == str:
+            raise ValueError(
+                f"Hooks must be dictionaries with one of the following keys: {supported_keys_str}"
+            )
+        actual_keys = hook.keys()
```
sfc-gh-melnacouzi, line 40: can't we add the keys to a set, and ensure that the set size is 1?

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-bdufour: So this is dead code?

src/snowflake/cli/api/project/schemas/native_app/application.py
sfc-gh-gbloom: Hooks can have more than one key. Each hook type has one "primary" key we use to determine its type. For example:
```yml
post_deploy: 
  - sql_script: scripts/post_install.sql
  - grant: execute task
    on: account
```

sfc-gh-melnacouzi: I see. If it can support multiple keys in the future, this will need to be done ad-hoc. Let's discuss offline. I think it might be best to use different classes for different types of hooks.

sfc-gh-melnacouzi: As per our offline suggestion, I agree this is not necessary now. To make this work, we would need something like this:

```
    post_deploy: List[Union[SqlScript, SystemScript]] = Field(
        title="Post application setup ",
        default=None,
        validation_alias="post_deploy",
    )
```
AND
we need to remove ValidationError to SchemaValidationError in UpdatableModel because this is messing with the Union logic.

sfc-gh-melnacouzi (approved):

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-gbloom: In a way. It's an additional protection layer: someone could change the schema without updating the logic. Something could also change the definition object dynamically.

Maybe it's personal preference, but I'm not comfortable with having an `if` condition and assume the `else` flow is not reachable.

PR #1245, opened 2024-06-20
When running `snow app run` against an application whose package was dropped, offer to automatically tear it down

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
When an app package has been dropped but the app still exists, the app can't be interacted with. In this case, running `snow app run` fails to upgrade the app to the latest staged files.

Since the app is effectively unusable at this point, the only option is to drop it. If the app owns any external resources (e.g. compute pools), the drop must use `cascade`. Unfortunately, since the app is unusable, `show objects owned by application` throws an error, so we have to handle that case and ask the user if they still want to drop the app without knowing what the objects are. If they say yes, we retry the drop with `cascade`.


src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -493,6 +493,20 @@ def get_objects_owned_by_application(self) -> List[ApplicationOwnedObject]:
             ).fetchall()
             return [{"name": row[1], "type": row[2]} for row in results]
 
+    def _application_objects_to_str(
```
sfc-gh-fcampbell, line 496: move from `NativeAppTeardownProcessor`

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -186,11 +198,34 @@ def get_existing_version_info(self, version: str) -> Optional[dict]:
...
+                    generic_sql_error_handler(err)
+                cc.warning(
+                    "The application owns other objects but they could not be determined."
+                )
+            what_to_drop = "the existing application and its owned objects"
+        else:
+            what_to_drop = "the existing application object"
+
+        user_prompt = (
+            f"Do you want the Snowflake CLI to drop {what_to_drop} and recreate it?"
```
sfc-gh-gbloom, line 227: > Do you want the Snowflake CLI to drop the existing application and its owned objects and recreate it?

Nit: maybe clarify that "it" means the application object, so that there will no expectation for the other dropped objects to be necessarily recreated.

tests_integration/nativeapp/test_init_run.py
```diff
@@ -418,3 +418,79 @@ def test_nativeapp_init_from_repo_with_single_template(
...
+            assert result.exit_code == 0, result.output
+
+            if force_flag:
+                command = ["app", "run", "--force"]
+                _input = None
+            else:
+                command = ["app", "run", "--interactive"]  # show prompt in tests
+                _input = "y\n"  # yes to drop app
+            result = runner.invoke_with_connection(command, input=_input, env=TEST_ENV)
+            assert result.exit_code == 0, result.output
```
sfc-gh-gbloom, line 476: Maybe also verify the expected state? The command succeeding might not be enough

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -493,6 +493,20 @@ def get_objects_owned_by_application(self) -> List[ApplicationOwnedObject]:
...
+        """
+        Returns a list in an "(Object Type) Object Name" format. Database-level and schema-level object names are fully qualified:
+        (COMPUTE_POOL) POOL_NAME
+        (DATABASE) DB_NAME
+        (SCHEMA) DB_NAME.PUBLIC
+        ...
+        """
+        return "\n".join(
+            [f"({obj['type']}) {obj['name']}" for obj in application_objects]
+        )
```
sfc-gh-bdufour, line 508: This would be a good candidate for the new `indented` console API I think, similar to the stage diffs.

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -186,11 +198,34 @@ def get_existing_version_info(self, version: str) -> Optional[dict]:
...
+                if application_objects := self.get_objects_owned_by_application():
+                    application_objects_str = self._application_objects_to_str(
+                        application_objects
+                    )
+                    cc.message(
+                        f"The following objects are owned by application {self.app_name} and need to dropped:\n{application_objects_str}"
+                    )
+            except ProgrammingError as err:
+                if err.errno != 93079 and ERROR_MESSAGE_093079 not in err.msg:
+                    generic_sql_error_handler(err)
```
sfc-gh-bdufour, line 218: nit: Could you please add a comment here about why this specific error is handled differently? Just for general readability/maintainability in the future.

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-bdufour: +1, when the "and its owned objects" form is used, this becomes a bit ambiguous.

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-fcampbell: This is actually prompting me to ask about general error handling. I've noticed a few places where we check both the numeric error code and the contents of the message. That seems a little brittle to me, in case the user-facing message changes. Do we not trust the numeric error code enough to have a local set of constants for each code?
i.e.
```python
if err.errno == APPLICATON_NO_LONGER_AVAILABLE
```
or something like that?

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-fcampbell: i felt that since `it` is singular it can only refer to the application and not the objects, but I can change it

tests_integration/nativeapp/test_init_run.py
sfc-gh-fcampbell: updated

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-fcampbell: updated the wording

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -186,11 +198,34 @@ def get_existing_version_info(self, version: str) -> Optional[dict]:
...
         """
-        user_prompt = "Do you want the Snowflake CLI to drop the existing application object and recreate it?"
+        if cascade:
+            try:
+                if application_objects := self.get_objects_owned_by_application():
+                    application_objects_str = self._application_objects_to_str(
+                        application_objects
+                    )
+                    cc.message(
+                        f"The following objects are owned by application {self.app_name} and need to dropped:\n{application_objects_str}"
```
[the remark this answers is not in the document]
sfc-gh-fcampbell: fixed

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-bdufour: Sure

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-bdufour: I would think that it's sufficient, yes. Maybe someone else from the team has context?

sfc-gh-bdufour (approved):

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-cgorrie: I think just error number is sufficient, especially if we properly annotate what the constant means (i.e. `ERROR_NO_APP_NO_LONGER_AVAILABLE` is much better than `ERROR_MESSAGE_093079`).

sfc-gh-cgorrie: I don't think we should be introspecting error messages because it's possible some user-provided stuff leaks through there.

sfc-gh-fcampbell: ok sounds good. given that this PR is already approved and there are other places we do this, i'll make it a separate PR to fix them all

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -109,7 +117,11 @@ def _create_dev_app(self, diff: DiffResult) -> None:
                     return
 
                 except ProgrammingError as err:
-                    generic_sql_error_handler(err)
+                    if err.errno not in UPGRADE_RESTRICTION_CODES:
+                        generic_sql_error_handler(err)
+                    else:
+                        cc.warning(err.msg)
```
sfc-gh-cgorrie, line 123: I understand here we're going to print whatever gets sent back from us from the backend -- what is the user going to see if the app has been orphaned this way? Is it enough to understand what's happening, or is it just going to be something like `Application is not available` which is potentially confusing @sfc-gh-fcampbell?

sfc-gh-fcampbell: The message could be one of a few depending on the reason for the upgrade being denied. In the case that this PR is addressing, it's "Application is no longer available for use. Please contact the application provider for more details.".

To keep things consistent, I lifted this code from https://github.com/snowflakedb/snowflake-cli/blob/38a4d680fe62913940cb85fd4e8e1735e0eb3d70/src/snowflake/cli/plugins/nativeapp/run_processor.py#L328-L333, which handles the same case when trying to upgrade an application using a release directive. As part of the error messaging improvements I'll do (https://github.com/snowflakedb/snowflake-cli/pull/1245#discussion_r1654931887), I'll make this a more friendly message (and in a separate improvement I'll reduce the code duplication going on here)

sfc-gh-fcampbell: yeah we could improve this message to explain what happened, but since the actual response from Snowflake is a bit vague on the details, I wasn't sure how much we want to reveal to the user. Is it OK for us to tell them that the package was dropped?

PR #1260, opened 2024-06-26
Project Definition v2 - schema validation

sfc-gh-gbloom:
### Pre-review checklist

### Changes description

1. Added a feature flag that enables specifying `definition_version: 2` in `snowflake.yml`.
2. Added the temporary hidden command `snow ws validate` as a first step to validate a minimal syntax subset of the new schema.
3. Created `get_project_definition()` to get a `ProjectDefinition` of the defined version
4. Created schemas for `ApplicationEntity` and `ApplicationPackageEntity`

### Test

Manually tested that:
* When the feature flag is disabled, `definition_version: 2` is not recognized: `Version 2 is not supported. Supported versions: 1, 1.1`
* Versions `1` and `1.1` work the same way as they did previously with the feature flag enabled and disabled
* The `ws` command is not listed in `snow --help`
* When the feature flag is enabled, verified `snow ws validate` correctly pass and fail with different configurations with `definition_version: 2`
* When the feature flag is enabled, `snow app run` with `definition_version: 2` fails: `No native_app project definition found in None`


src/snowflake/cli/plugins/workspace/commands.py
```diff
@@ -0,0 +1,38 @@
...
+from snowflake.cli.api.commands.flags import (
+    project_type_option,
+)
+from snowflake.cli.api.commands.snow_typer import SnowTyper
+from snowflake.cli.api.output.types import MessageResult
+
+ws = SnowTyper(
+    name="ws",
+    hidden=True,
+    help="Manages workspaces in Snowflake.",
```
sfc-gh-cgorrie, line 28: Having given some though to the "workspace" term it's probably safer to do something like this:
```suggestion
    help="Deploy and interact with snowflake.yml-based entities",
```

tests/nativeapp/codegen/test_compiler.py
```diff
@@ -27,7 +27,7 @@
 
 @pytest.fixture()
 def test_proj_def():
-    return ProjectDefinition(
```
sfc-gh-cgorrie, line 30: Can we mess around with the export so we can keep `ProjectDefinition(...)` everywhere even though it's now different based on the feature flag?

PR #1262, opened 2024-06-26
Add bundle root for generation

sfc-gh-bdufour:
### Pre-review checklist

### Changes description

Introducing `bundle_root` as a new project property for native apps. The bundle root is similar to the deploy root, but is going to be used during code generation to store temporary artifacts. For example, an upcoming feature will require setting up a virtual environment for the user so that Python code can be appropriately sandboxed. That environment will be created in the bundle root.

The bundle root change it itself was very small. Most of this PR actually involves wiring the bundle root through to the compiler. Instead of adding one more parameter to the compiler and processors, I felt it was time to take a step change and recheck the contract to the processors. So I extracted the data portion of the native app manager, and am I using it as a kind of "preserve whole object" refactoring. This way, processors can access the information they need without having to thread it through the set of calls every time we need a new piece of data.

I have avoided updating the release notes for now, as this should be part of the actual feature entry when it gets released. The bundle root is, at the moment, unused by any logic, so it doesn't make sense to announce it as something new by itself.

### Testing

* Executed the manual test plan to make sure there were no regressions
* Added unit tests tests to reflect the change in APIs. Integ and E2E tests are unaffected by the refactoring
* Manually checked for regressions in the code generation feature


src/snowflake/cli/plugins/nativeapp/project_model.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def __init__(
+        self,
+        project_definition: NativeApp,
+        project_root: Path,
+    ):
+        self._project_definition = project_definition
+        self._project_root = resolve_without_follow(project_root)
+
+    @property
```
sfc-gh-bdufour, line 59: All of these were just moved verbatim from the manager class

src/snowflake/cli/plugins/nativeapp/project_model.py
```diff
@@ -0,0 +1,182 @@
...
+    ApplicationPostDeployHook,
+)
+from snowflake.cli.api.project.schemas.native_app.native_app import NativeApp
+from snowflake.cli.api.project.schemas.native_app.path_mapping import PathMapping
+from snowflake.cli.api.project.util import extract_schema, to_identifier
+from snowflake.cli.plugins.nativeapp.artifacts import resolve_without_follow
+from snowflake.connector import DictCursor
+
+
+def current_role() -> str:
```
sfc-gh-bgoel, line 37: I believe this can be used incorporated in https://github.com/snowflakedb/snowflake-cli/blob/851bfdbda0cb4a9521c46fc4c17273adb36fbd36/src/snowflake/cli/api/sql_execution.py#L91 to reduce the duplicate logic in that context manager.

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -168,135 +163,94 @@ class NativeAppManager(SqlExecutionMixin):
 
     def __init__(self, project_definition: NativeApp, project_root: Path):
         super().__init__()
-        self._project_root = project_root
-        self._project_definition = project_definition
+        self._project = NativeAppProjectModel(
```
sfc-gh-bgoel, line 166: nit: Since we're renaming some stuff, we can make this explicit that this is not the entire project, but only the `na_project`

src/snowflake/cli/plugins/nativeapp/codegen/artifact_processor.py
```diff
@@ -36,18 +35,12 @@ def __init__(self, processor_name: str):
...
-        project_definition: NativeApp,
-        project_root: Path,
-        deploy_root: Path,
-        generated_root: Path,
+        project: NativeAppProjectModel,
         **kwargs,
     ) -> None:
-        assert project_root.is_absolute()
-        assert deploy_root.is_absolute()
-        assert generated_root.is_absolute()
```
sfc-gh-bgoel, line 50: I saw that the asserts have been moved from here to the SnowparkAnnotationProcessor class. Do we not want to have this assert at the base class level?

sfc-gh-bgoel (commented):
1. I don't see the use of `bundle_root` yet, except to define its property method. I am guessing this is intentional and will be used with the new feature's code.
2. I am not sure I understand the example supplied for `bundle_root`, let's catch up offline to dive into it.

sfc-gh-bdufour:
> 1. I don't see the use of `bundle_root` yet, except to define its property method. I am guessing this is intentional and will be used with the new feature's code.
> 2. I am not sure I understand the example supplied for `bundle_root`, let's catch up offline to dive into it.

1. Correct, that's all explained in the PR description. I'm already building on top of this PR.
2. What example are you referring to?

src/snowflake/cli/plugins/nativeapp/codegen/artifact_processor.py
sfc-gh-bdufour: So far the Snowpark processor relies heavily on that invariant, but I didn't think it was a property of _all_ processors. In any case, I made it a strong invariant now, so the assertions are really redundant.

src/snowflake/cli/plugins/nativeapp/project_model.py
sfc-gh-bdufour: Spotted that too, but we don't have ownership where it should go, so I'd prefer to split it into its own submission to avoid noise for the CLI team.

sfc-gh-bgoel (dismissed):
Caught up offline to go over the `bundle_root` introduction in more detail. The new addition + refactor LGTM. Acking failing IT.

src/snowflake/cli/api/project/schemas/native_app/native_app.py
```diff
@@ -34,6 +34,10 @@ class NativeApp(UpdatableModel):
     artifacts: List[Union[PathMapping, str]] = Field(
         title="List of file source and destination pairs to add to the deploy root",
     )
+    bundle_root: Optional[str] = Field(
```
sfc-gh-cgorrie, line 37: I'm not 100% sold on bundle root, but I can't think of a better name right now...

src/snowflake/cli/plugins/nativeapp/project_model.py
sfc-gh-cgorrie: TBH I think it's worth getting their sign-off here, it's so generally useful. But we can always make a one-off PR afterwards that moves it.

sfc-gh-cgorrie (approved):
Approved though I'm not completely sold on the name.

src/snowflake/cli/plugins/nativeapp/project_model.py
sfc-gh-bdufour: Yeah, I don't think the current_role() calls belongs there, I was always planning to move it due to it being generally useful. There's a location that has all these default wh/role/etc helpers, I'm planning to move it there too.

sfc-gh-fcampbell (approved):
LGTM pending integration test success

PR #1270, opened 2024-06-28
Snow 1465196 snow init

sfc-gh-pczajka:
### Pre-review checklist

### Changes description
Implementation of template rendering for project initialisation, according to https://docs.google.com/document/d/1OPlxhPmdilYLPkI6-ixmgz7zD0TfflsmulFLlx4WkYE/

**Scope:**
 * snow init command working with local and remote templates
 * design of template.yml
 * working unit tests
 
**Not in scope**
 * refactor of existing init commands - to be done when template repo will become available
 * integration tests - will be done after testing templates from this PR appear in main


src/snowflake/cli/api/project/schemas/template.py
```diff
@@ -0,0 +1,66 @@
...
+from typing import Any, List, Optional
+
+from pydantic import BaseModel, Field
+from snowflake.cli.api.secure_path import SecurePath
+
+
+class TemplateVariableType(Enum):
+    STRING = "string"
+    INTEGER = "int"
+    FLOAT = "float"
```
sfc-gh-turbaszek, line 27:
```suggestion
    STRING = str
    INTEGER = int
    FLOAT = float
```
Won't it work?

sfc-gh-turbaszek: Let's also add support for booleans. For booleans we should support [y/N] prompts by using `typer.confirm` instead of `typer.prompt`

src/snowflake/cli/api/project/schemas/template.py
```diff
@@ -0,0 +1,66 @@
...
+    )
+    prompt: Optional[str] = Field(title="Prompt message for the variable", default=None)
+    default: Optional[Any] = Field(title="Default value of the variable", default=None)
+
+
+class Template(BaseModel):
+    minimum_cli_version: Optional[str] = Field(
+        None, title="Minimum version of Snowflake CLI supporting this template"
+    )
+    rendered_files: List[str] = Field(title="List of files to be rendered", default=[])
```
sfc-gh-turbaszek, line 51:
```suggestion
    file_with_templates: List[str] = Field(title="List of files to be rendered", default=[])
```
WDYT? `rendered_files` sounds like something already rendered

src/snowflake/cli/api/project/schemas/template.py
```diff
@@ -0,0 +1,66 @@
...
+    variables: List[TemplateVariable] = Field(
+        title="List of variables to be rendered", default=[]
+    )
+
+    def __init__(self, template_root: SecurePath, **kwargs):
+        super().__init__(**kwargs)
+        self._validate_files_exist(template_root)
+
+    def _validate_files_exist(self, template_root: SecurePath) -> None:
```
sfc-gh-turbaszek, line 60: Can this be a filed validator instead of doing this in constructor?

src/snowflake/cli/api/utils/rendering.py
```diff
@@ -93,6 +95,21 @@ def get_sql_cli_jinja_env():
     )
 
 
+def get_template_cli_jinja_env(template_root: SecurePath) -> Environment:
```
sfc-gh-turbaszek, line 98: I would suggest moving this into `plugins.init` module to avoid too many templatings in one place. WDYT?

src/snowflake/cli/plugins/init/commands.py
```diff
@@ -0,0 +1,238 @@
...
+
+
+def _fetch_local_template(
+    template_source: SecurePath, path: Optional[str], destination: SecurePath
+) -> SecurePath:
+    """Copies local template to [dest] and returns path to the template root.
+    Ends with an error of the template does not exist."""
+
+    template_source.assert_exists()
+    template_origin = template_source
```
sfc-gh-turbaszek, line 69: Is this line neccessary?

src/snowflake/cli/plugins/init/commands.py
```diff
@@ -0,0 +1,238 @@
...
+    template_source: Optional[str] = SourceOption,
+    variables: Optional[List[str]] = VariablesOption,
+    no_interactive: bool = NoInteractiveOption,
+    **options,
+) -> CommandResult:
+    """
+    Creates project from template.
+    """
+    if SecurePath(path).exists():
+        raise FileExistsError(f"{path} already exists")
```
sfc-gh-turbaszek, line 201: Can't we delegate this to typer.Option?

src/snowflake/cli/plugins/init/commands.py
```diff
@@ -0,0 +1,238 @@
...
+
+    variables_from_flags = {
+        v.key: v.value
+        for v in parse_key_value_variables(variables if variables else [])
```
sfc-gh-turbaszek, line 205: Let's move the if else to `parse_key_value_variables` implementation

src/snowflake/cli/plugins/init/commands.py
```diff
@@ -0,0 +1,238 @@
...
+            f" by template is {required_version}. Please upgrade before continuing"
+        )
+
+
+def _is_remote_source(source: str) -> bool:
+    exists_as_file = SecurePath(source).exists()
+    whitelisted_prefix = any(
+        source.startswith(prefix) for prefix in ["git@", "http://", "https://"]
+    )
+    return (not exists_as_file) and whitelisted_prefix
```
sfc-gh-turbaszek, line 185: Consider making this callback on `SourceOption` also I think we can simplify it to
```py
if SecurePath(source).exists():
   return False
if  any(
        source.startswith(prefix) for prefix in ["git@", "http://", "https://"]):
     return True
raise ClickException(...)
```

WDYT?

src/snowflake/cli/plugins/init/commands.py
```diff
@@ -0,0 +1,238 @@
...
+
+
+def _validate_cli_version(required_version: str) -> None:
+    from packaging.version import parse
+    from snowflake.cli.__about__ import VERSION
+
+    if parse(required_version) > parse(VERSION):
+        raise ClickException(
+            f"Snowflake CLI version ({VERSION}) is too low - minimum version required"
+            f" by template is {required_version}. Please upgrade before continuing"
```
sfc-gh-turbaszek, line 176:
```suggestion
            f" by template is {required_version}. Please upgrade before continuing."
```

src/snowflake/cli/plugins/init/commands.py
```diff
@@ -0,0 +1,238 @@
...
+    # TODO: during nativeapp refactor get rid of this dependency
+    from snowflake.cli.plugins.nativeapp.utils import shallow_git_clone
+
+    shallow_git_clone(url, to_path=destination.path)
+    if path:
+        template_root = destination / path
+    else:
+        # remove .git directory not to copy it to the template
+        template_root = destination
+        git_rmtree((template_root / ".git").path)
```
sfc-gh-turbaszek, line 98: Can you explain this part?

src/snowflake/cli/api/project/schemas/template.py
sfc-gh-astus: `files_with_templates`? 😄

src/snowflake/cli/api/project/schemas/template.py
sfc-gh-pczajka: I thought about that, but they create problems with rendering. By default jinja renders True/False, but this works only for Python files - for example json will have syntax errors
WDYT?

src/snowflake/cli/api/project/schemas/template.py
```diff
@@ -0,0 +1,66 @@
...
+        for file in self.rendered_files:
+            full_path = template_root / file
+            if not full_path.exists():
+                raise FileNotFoundError(f"Template does not have file {file}")
+            if full_path.is_dir():
+                raise IsADirectoryError(f"{file} is a directory")
```
sfc-gh-astus, line 66: Maybe throw full path instead of only dir name?

src/snowflake/cli/api/project/schemas/template.py
sfc-gh-pczajka: Unfortunately not - field validator is a class method, and I need an information about template_root on this one.
I've tried model_validators as well, but they seemed to have a problem with setting up undeclared argument in constructor.
Calling that deliberately in constructor is by far easiest solution I found

src/snowflake/cli/plugins/init/commands.py
sfc-gh-pczajka: This supports the case when whole repository is a single template. (compatibility with native app features)

src/snowflake/cli/api/utils/rendering.py
sfc-gh-pczajka: I had a different thought - to have all template definitions is one place :D
IMO it would be convenient for maintenance - we know which '<* ... *>' go with which context.

If we want to move this function closer to its plugin, I'd suggest doing the same with the rest of them

src/snowflake/cli/api/project/schemas/template.py
sfc-gh-pczajka: `files_to_render` :D?

src/snowflake/cli/api/project/schemas/template.py
sfc-gh-pczajka: > Won't it work?
Unfortunately :(

```
3 validation errors for Template
variables.1.type
  Input should be <class 'str'>, <class 'int'> or <class 'float'> [type=enum, input_value='int', input_type=str]
    For further information visit https://errors.pydantic.dev/2.7/v/enum
```

src/snowflake/cli/api/project/schemas/template.py
sfc-gh-pczajka: This way the user would see a whole path from '/' through magical temporary directory `tmp123` user have no idea about to the file inside a template. What I wanted to achieve was:
```
# template.yml
files_to_render:
  - a_good_file.py
  - actually_a_dir
  
# error message:
"actually_a_dir" is a directory
```

I renamed "file" to "path_in_template" to make the code clear

src/snowflake/cli/plugins/init/commands.py
sfc-gh-pczajka: Changed to
```py
template_origin = template_source / path if path else template_source
```

src/snowflake/cli/plugins/init/commands.py
sfc-gh-pczajka: I simplified it to
```py
is_remote = any(
     template_source.startswith(prefix) for prefix in ["git@", "http://", "https://"]
)
```
Validation whether template exists or not happens in `_fetch_remote/local_template`

src/snowflake/cli/plugins/init/commands.py
sfc-gh-pczajka: done

src/snowflake/cli/plugins/init/commands.py
sfc-gh-pczajka: Done!

src/snowflake/cli/api/project/schemas/template.py
```diff
@@ -0,0 +1,70 @@
...
+        return {
+            TemplateVariableType.STRING: str,
+            TemplateVariableType.INTEGER: int,
+            TemplateVariableType.FLOAT: float,
+        }[self]
+
+
+class TemplateVariable(BaseModel):
+    name: str = Field(..., title="Variable identifier")
+    type: Optional[TemplateVariableType] = Field(  # noqa: A003
```
sfc-gh-turbaszek, line 40: Do we need enum at all? `TemplateVariableType = Union[str, float, int]`? Is there any value added in the enum?

src/snowflake/cli/api/project/schemas/template.py
sfc-gh-turbaszek: Thanks!

src/snowflake/cli/plugins/init/commands.py
sfc-gh-turbaszek: Sounds good, please add this comment to the code for future 🙆‍♂️

src/snowflake/cli/api/project/schemas/template.py
sfc-gh-pczajka: I've just checked - If we do that, TemplateVariable.type would expect basically any value of typ str, int or float. We wanted choice field between str/int/float/None
Enum does validation for us

src/snowflake/cli/plugins/init/commands.py
sfc-gh-pczajka: done!

src/snowflake/cli/api/utils/rendering.py
sfc-gh-pczajka: Nice idea - this way they still would be close and easy to find/differentiate, but we won't mix context in a single file

sfc-gh-pczajka: Refactor rendering module: https://github.com/snowflakedb/snowflake-cli/pull/1270/commits/18c86b2af3b367038e7b2516f96bc698e46bf562

Refactor `snow init` code: https://github.com/snowflakedb/snowflake-cli/pull/1270/commits/448ebd8069d22af33f9be158229ca5eafb763142

src/snowflake/cli/plugins/init/commands.py
```diff
@@ -0,0 +1,241 @@
...
+
+def _fetch_remote_template(
+    url: str, path: Optional[str], destination: SecurePath
+) -> SecurePath:
+    """Downloads remote repository template to [dest],
+    and returns path to the template root.
+    Ends with an error of the template does not exist."""
+    from git import rmtree as git_rmtree
+
+    # TODO: during nativeapp refactor get rid of this dependency
```
sfc-gh-cgorrie, line 102: Is this native app refactor planned already? My understanding of the plan is that we will keep "snow app init" for a while until:

1. we confirm that "snow init" supports all of our use cases
2. definition_version: 2 is out

Then we'll re-create all of the templates in the new repo with definition_version: 2. Does this align with the plan?

src/snowflake/cli/plugins/init/commands.py
```diff
@@ -0,0 +1,241 @@
...
+def init(
+    path: str = PathArgument,
+    template: Optional[str] = TemplateOption,
+    template_source: Optional[str] = SourceOption,
+    variables: Optional[List[str]] = VariablesOption,
+    no_interactive: bool = NoInteractiveOption,
+    **options,
+) -> CommandResult:
+    """
+    Creates project from template.
```
sfc-gh-cgorrie, line 199:
```suggestion
    Creates project directory from template.
```

tests/test_init.py
```diff
@@ -0,0 +1,356 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import json
+from contextlib import contextmanager
+from pathlib import Path
+
+import pytest
```
sfc-gh-cgorrie, line 19: Can we include a test that demonstrates the ability to pull the name of the directory we created, as well as converting it into an identifier? 

See https://github.com/snowflakedb/snowflake-cli/blob/main/src/snowflake/cli/plugins/nativeapp/init.py#L323 and the lines after it -- we allow the user to set the project name variable if desired, but if it's not provided we pull it from the path name and turn it into an identifier with a set of rules that prevents users from hitting rough edges (e.g we'll convert `-` to `_` to prevent users from accidentally opting into quoted identifiers).

sfc-gh-cgorrie (commented):
Some thoughts from the NADE point-of-view

src/snowflake/cli/api/rendering/project_templates.py
```diff
@@ -0,0 +1,51 @@
...
+from snowflake.cli.api.secure_path import SecurePath
+
+_PROJECT_TEMPLATE_START = "<!"
+_PROJECT_TEMPLATE_END = "!>"
+
+
+def get_template_cli_jinja_env(template_root: SecurePath) -> Environment:
+    _random_block = "___very___unique___block___to___disable___logic___blocks___"
+    return env_bootstrap(
+        Environment(
```
sfc-gh-melnacouzi, line 30: nit:
if the context data is nested dict, I recommend using IgnoreAttrEnvironment. Otherwise, this will pick up dict specific attributes in nested dicts. 

This comment is only important if you support nest dicts or lists. Since you only have one level of dict, Jinja handles this properly. In this case, using IgnoreAttrEnvironment is less important, but might remove some code duplication if we only pass template_start/template_end to the env_bootstrap function.

src/snowflake/cli/api/rendering/project_templates.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def render_template_files(
+    template_root: SecurePath, files_to_render: List[str], data: Dict[str, Any]
+) -> None:
+    """Override all listed files with their rendered version."""
+    jinja_env = get_template_cli_jinja_env(template_root)
+    for path in files_to_render:
+        jinja_template = jinja_env.get_template(path)
+        rendered_result = jinja_template.render(**data)
```
sfc-gh-melnacouzi, line 49: i recommend try/except here on TemplateSyntaxError, and throw InvalidTemplate (ClickException) with message and lineno. It provides nicer and more actionable output for users (e.g. when forgetting to close <!)

src/snowflake/cli/plugins/init/commands.py
```diff
@@ -0,0 +1,241 @@
...
+    """
+    variables_from_flags = {
+        v.key: v.value for v in parse_key_value_variables(variables)
+    }
+    is_remote = any(
+        template_source.startswith(prefix) for prefix in ["git@", "http://", "https://"]  # type: ignore
```
sfc-gh-melnacouzi, line 205: have we tested this with private git repo (with ssh keys set-up)?

tests/test_init.py
sfc-gh-pczajka: The identifier is not a case for some projects - for example for Snowpark, where we have bunch of different objects defined in one `snowflake.yml`

This particular feature (as well as `--name` flag) will be preserved on `snow app init` side.

src/snowflake/cli/api/rendering/project_templates.py
sfc-gh-pczajka: Thanks for the context! Changed

src/snowflake/cli/plugins/init/commands.py
sfc-gh-pczajka: Yes, I've created private repository with dummy template - works as intended

src/snowflake/cli/plugins/init/commands.py
sfc-gh-pczajka: Moved discussion to Slack

src/snowflake/cli/api/project/schemas/template.py
sfc-gh-pczajka: Changed to `Literal["string", "int", "float"]`, validation works well

tests/test_init.py
sfc-gh-pczajka: I've created `to_snowflake_identifier` jinja filter, which can be used in templates to convert strings into Snowflake identifiers

src/snowflake/cli/api/rendering/project_templates.py
sfc-gh-pczajka: Nice one! done

sfc-gh-melnacouzi (dismissed):
lgtm, but I see some tests are failing.

sfc-gh-turbaszek (approved):

sfc-gh-turbaszek (approved):

sfc-gh-cgorrie (approved):

PR #1275, opened 2024-07-02
Add future support for project definition v2

sfc-gh-gbloom:
### Pre-review checklist

### Changes description

This is the first PR in a series of changes to enable support for `definition_version: 2` in `snowflake.yml`.
The actual v2 schema and entities implementation will be added as separate PRs.

1. Added a feature flag: `enable_pdf_v2`
2. Added the temporary hidden command `snow ws validate` as a v2 validation entry point
3. Refactoring `ProjectDefinition`:
  a. Up until now we only had a single definition version, so `ProjectDefinition` simply extended `_DefinitionV11`. Now that we have 2 versions, `get_project_definition()` should be used to get the correct definition class depending on the specified version.
  b. `ProjectDefinition` was also used as a type that implied v1. Now this type refers to the base class instead.
  c. Changed the doc generation commands to explicitly use v1

#### Valid schema
```yml
definition_version: 2
entities: {}
```
```
$ snow ws validate
Project definition is valid.
```

#### Invalid schema
```yml
definition_version: 2
abc: def
```
```
$ snow ws validate

During evaluation of _DefinitionV20 in project definition following errors were encountered:
Your project definition is missing following fields: ('entities',)
Extra inputs are not permitted. You provided field '('abc',)' with value 'def' that is not supported in given version.
```

#### No interference with the existing behavior

* When the feature flag is disabled, `definition_version: 2` is not recognized: `Version 2 is not supported. Supported versions: 1, 1.1`
* Versions `1` and `1.1` work the same way as they did previously with the feature flag enabled and disabled
* The `ws` command is not listed in `snow --help`
* When the feature flag is enabled, `snow app run` with `definition_version: 2` fails: `No native_app project definition found in None`


src/snowflake/cli/api/project/schemas/project_definition.py
```diff
@@ -45,7 +46,7 @@ class ProjectProperties:
     project_context: Context
 
 
-class _BaseDefinition(UpdatableModel):
+class ProjectDefinition(UpdatableModel):
```
sfc-gh-cgorrie, line 49: I think we should keep the `_` to show that this shouldn't be imported directly. A docstring to explain that this is a base class and not actually a class we can call the initializer on would be good too.

src/snowflake/cli/api/project/schemas/project_definition.py
```diff
@@ -95,23 +96,45 @@ def _convert_env(
...
-        if not isinstance(data, dict):
-            return
-        if version := str(data.get("definition_version")):
-            version_model = _version_map.get(version)
-            if not version_model:
-                raise ValueError(
-                    f"Unknown schema version: {version}. Supported version: {_supported_version}"
-                )
-            version_model(**data)
+def get_project_definition(**data):
```
sfc-gh-cgorrie, line 127: Suggest something like `build_project_definition` as we're not getting anything, we're calling an initializer. Also a docstring to say what kinds of things should use this.

src/snowflake/cli/api/project/schemas/project_definition.py
```diff
@@ -95,23 +96,45 @@ def _convert_env(
...
+    if not isinstance(data, dict):
+        return
+    if FeatureFlag.ENABLE_PDF_V2.is_enabled():
```
sfc-gh-cgorrie, line 130: To confirm, we cannot put this into the version map at the outer level of this file because feature flags won't be enabled yet, correct?

src/snowflake/cli/app/dev/docs/project_definition_docs_generator.py
```diff
@@ -39,7 +39,7 @@ def generate_project_definition_docs(root: SecurePath):
 
     root.mkdir(exist_ok=True)
     list_of_sections = model_json_schema(
-        ProjectDefinition, schema_generator=ProjectDefinitionGenerateJsonSchema
+        _DefinitionV11, schema_generator=ProjectDefinitionGenerateJsonSchema
```
sfc-gh-cgorrie, line 42: It's a code smell to me to import something that's `_`-prefixed as in Python this indicates the item is internal to the file. WDYT?

src/snowflake/cli/api/project/schemas/project_definition.py
sfc-gh-gbloom: I generally agree, but `ProjectDefinition` is also used for type annotations -- it's not only internal to this file. I'll add a docstring to clarify that this is the base class.

sfc-gh-cgorrie: Sounds good to me. We can keep the invariant that anything with `_` is not intended to be imported outside of the file. How about `ProjectDefinitionBase`?

sfc-gh-gbloom: I think signatures using it might be confusing. For example here, it's not the base class that we're returning:
```python
class _CliGlobalContextAccess:
    @property
    def project_definition(self) -> ProjectDefinition:
        return self._manager.project_definition
```

Instead, I created an alias type and kept `_ProjectDefinitionBase` "internal":
```python
ProjectDefinition = _DefinitionV10 | _DefinitionV11 | _DefinitionV20
```

src/snowflake/cli/api/project/schemas/project_definition.py
sfc-gh-gbloom: Looks like they are parsed when this executes for the first time! Moved it to the top level.

src/snowflake/cli/app/dev/docs/project_definition_docs_generator.py
sfc-gh-gbloom: I agree, but the code here specifically tests for v1.1. It's just that before my change it was done explicitly.
I'll let the CLI team to review this area.

src/snowflake/cli/api/project/schemas/project_definition.py
sfc-gh-gbloom: No, it works at the top level as well, but if the config file is corrupted it will cause the CLI to crash with an uncaught exception instead of a pretty error box.

src/snowflake/cli/api/feature_flags.py
```diff
@@ -48,3 +48,4 @@ class FeatureFlag(FeatureFlagMixin):
     ENABLE_STREAMLIT_EMBEDDED_STAGE = BooleanFlag(
         "ENABLE_STREAMLIT_EMBEDDED_STAGE", False
     )
+    ENABLE_PDF_V2 = BooleanFlag("ENABLE_PDF_V2", False)
```
sfc-gh-bdufour, line 51: This is a customer-facing flag, let's avoid acronyms

src/snowflake/cli/app/dev/docs/project_definition_docs_generator.py
sfc-gh-astus: I agree with @sfc-gh-cgorrie

src/snowflake/cli/plugins/workspace/commands.py
```diff
@@ -0,0 +1,34 @@
...
+)
+
+
+@ws.command(requires_connection=True)
+@with_project_definition()
+def validate(
+    **options,
+):
+    """Validates the project definition file."""
+    return MessageResult("Project definition is valid.")
```
sfc-gh-astus, line 34: I'm not sure about this command. Do you need it now? Returns only static message.

src/snowflake/cli/app/dev/docs/project_definition_docs_generator.py
sfc-gh-jvasquezrojas: Can we change def `_get_version_map()` to be externally accesible? Renaming it as `get_version_map()`. we could use that function to generate the docs for multiple versions and to avoid explicitly accessing to file clases

src/snowflake/cli/plugins/workspace/commands.py
sfc-gh-gbloom: It might look silly at first, but yes. It uses `@with_project_definition()`, so the message will be printed only if the schema is valid.
Once we will add the actual logic we will be able to remove `snow ws validate` and replace it with other commands such as `snow ws deploy`, but for now we need something to only trigger the schema validation.

sfc-gh-fcampbell:
```suggestion
def validate(
    **options,
):
    """Validates the project definition file."""
    # If we get to this point, @with_project_definition() has already validated the PDF
    return MessageResult("Project definition is valid.")
```

tests/project/test_config.py
```diff
@@ -145,6 +148,31 @@ def test_fields_are_parsed_correctly(project_definition_files, snapshot):
...
+    with pytest.raises(SchemaValidationError) as err:
+        build_project_definition(**{"definition_version": "2", "entities": {}})
+    assert "Version 2 is not supported" in str(err.value)
+
+
+def mock_config_key(key, value):
+    def get_config_side_effect_func(*args, **kwargs):
+        if kwargs["key"] == key:
+            return value
+        return kwargs["default"]
```
sfc-gh-fcampbell, line 167: for requests other than the mocked key, shouldn't we dispatch to the original `get_config_value` function instead of returning the default?

src/snowflake/cli/app/dev/docs/project_definition_docs_generator.py
sfc-gh-fcampbell: If we want to keep only generating docs for 1.1, I don't see the harm in making the definition classes public

sfc-gh-gbloom: Removed the `_` prefix from the definition classes and the `get_version_map()` function

tests/project/test_config.py
sfc-gh-gbloom: Done. Also moved this code to the test utils folder.

sfc-gh-jvasquezrojas (dismissed):

src/snowflake/cli/api/project/schemas/project_definition.py
```diff
@@ -102,23 +104,53 @@ def _convert_env(
...
-                raise ValueError(
-                    f"Unknown schema version: {version}. Supported version: {_supported_version}"
-                )
-            version_model(**data)
+ProjectDefinition = Union[DefinitionV10, DefinitionV11, DefinitionV20]
 
 
-_version_map = {"1": _DefinitionV10, "1.1": _DefinitionV11}
-_supported_version = tuple(_version_map.keys())
+def get_version_map():
```
sfc-gh-bdufour, line 152: nit: _get_version_map? It was private before. Unless I missed something, it's still only used here.

sfc-gh-bdufour (approved):

src/snowflake/cli/api/project/schemas/project_definition.py
sfc-gh-bdufour: Ah, just saw the comments above, ignore this, it was intentional.

src/snowflake/cli/api/project/schemas/project_definition.py
```diff
@@ -102,23 +104,53 @@ def _convert_env(
...
-        if not isinstance(data, dict):
-            return
-        if version := str(data.get("definition_version")):
-            version_model = _version_map.get(version)
-            if not version_model:
-                raise ValueError(
-                    f"Unknown schema version: {version}. Supported version: {_supported_version}"
-                )
-            version_model(**data)
+ProjectDefinition = Union[DefinitionV10, DefinitionV11, DefinitionV20]
```
sfc-gh-melnacouzi, line 149: How would the rest of the code use this? Will we be adding a translation layer?

sfc-gh-gbloom: It's only used for type checking. If we have a function with this signature:
```python
def project_definition() -> ProjectDefinition
```
The type checker will understand the types it can hold:
```python
definition = get_definition()
definition.definition_version # str | int
definition.native_app # NativeApp | None
```
Up until now when we used `ProjectDefinition` as a type it implied `DefinitionV11`, which was wrong (it could be `DefinitionV10`), but now that we have `DefinitionV20` I wanted to make it clear that it could be any version.
If some code path expects a specific version, it should narrow the type to that version.

src/snowflake/cli/api/project/schemas/project_definition.py
```diff
@@ -102,23 +104,53 @@ def _convert_env(
         return ProjectEnvironment(default_env=(env or {}), override_env={})
 
 
-class ProjectDefinition(_DefinitionV11):
-    def __init__(self, **kwargs):
-        super().__init__(**kwargs)
-        self._validate(kwargs)
+class DefinitionV20(_ProjectDefinitionBase):
+    entities: Dict = Field(
```
sfc-gh-turbaszek, line 108: Is there any option to make this `Dict[str, Entity]`, `Entity` being a pydantic base for future enitites?

sfc-gh-gbloom: This is just a basic schema at this point. Let's discuss the actual pydantic schemas in the [next PR](https://github.com/snowflakedb/snowflake-cli/pull/1280).

src/snowflake/cli/api/project/schemas/project_definition.py
sfc-gh-melnacouzi: let's take native_app as example. Are we expecting to see a lot of this in the business logic?
```
if definition.native_app:
   native_app = definition.native_app
else:
  native_app = translate(definition.entity.something['native_app']) # just a random structure
```
Or will it automatically be handled by a translation layer?

sfc-gh-melnacouzi: To be honest, I think it would be clearer if we have DefinitionV1 and DefinitionV2. DefinitionV1 can still have the same inheritance structure as now (DefinitionV1 -> DefinitionV11 -> DefinitionV10 -> Base) and (DefinitionV2 -> DefinitionV20 -> Base).

The rest of the code can then use DefinitionV1 throughout. ProjectDefinition by itself as a combination of V1 and V2 is just confusing. We can use it in places that handle both (Like render_definition) but the rest of the code should be using DefinitionV1 exclusively.

sfc-gh-gbloom: We will handle it in future PRs, but in general the business logic layer shouldn't check for definition versions. We will have a version translation layer, so the underlying code will work as is (v1) until we refactor it to use v2 and flip the translation direction. For simple commands that don't require much refactoring we can move to v2->v1 translation directly.

We can discuss the migration plan f2f, but in any case the version narrowing will be done in a single place, and the logic will be guaranteed to receive a specific definition version.

The change in this PR doesn't really affect the existing code in terms of type checking:
```python
def app_bundle():
    manager = NativeAppManager(
        # project_definition.native_app is guaranteed to be NativeApp
        # because of @with_project_definition check at runtime
        project_definition=cli_context.project_definition.native_app
    )
    ...
```

sfc-gh-gbloom: After discussing f2f, I created these types:
```python
ProjectDefinitionV1 = Union[DefinitionV10, DefinitionV11]
ProjectDefinitionV2 = DefinitionV20
ProjectDefinition = Union[ProjectDefinitionV1, ProjectDefinitionV2]
```
And changed `DefinitionManager` to explicitly return `ProjectDefinitionV1`.

sfc-gh-melnacouzi (approved):

sfc-gh-turbaszek (approved):

PR #1276, opened 2024-07-02
Remove project section check outside with_project_definition decorator

sfc-gh-melnacouzi:
### Pre-review checklist

### Changes description
Refactor with_project_definition so that section checks are done outside of the decorator code.
- visible changes:
- 1) when snowflake.yml is not found, the message is generic (e.g. Snowflake project instead of the Native App project)
- 2) when section is not found, use project_root in the message (absolute path of snowflake.yml folder) instead of the value of the optional -p param.

Release notes not updated because there are no significant changes from the point of view of users.

This refactor was requested by  @sfc-gh-turbaszek earlier. We might be introducing more code duplication with this refactor, so I would like to confirm with @sfc-gh-turbaszek if this is still needed.


sfc-gh-bdufour:
What's the motivation behind this change? There's definitely a tradeoff happening here.

src/snowflake/cli/plugins/nativeapp/version/commands.py
```diff
@@ -137,6 +150,12 @@ def drop(
     Drops a version defined in your application package. Versions can either be passed in as an argument to the command or read from the `manifest.yml` file.
     Dropping patches is not allowed.
     """
+
+    if cli_context.project_definition.native_app is None:
+        raise NoProjectDefinitionError(
+            project_type="native_app", project_file=cli_context.project_root
+        )
```
sfc-gh-turbaszek, line 157: How about introducing a helper
`assert_project_type(project_type="native_app")`

this would reduce the code duplication. WDYT?

sfc-gh-turbaszek:
> What's the motivation behind this change? There's definitely a tradeoff happening here.

Introduction of env support made all commands using project file to require full project definition. To do so we already changed `with_project_definition`. Now, the proposed change make sure that the decorator provides the definition (full, not a single part like app or streamlit). It's also opens the code for further changes related to snowflake.yml V2.

src/snowflake/cli/plugins/nativeapp/version/commands.py
sfc-gh-melnacouzi: updated

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -145,13 +146,19 @@ def app_list_templates(**options) -> CommandResult:
...
     **options,
 ) -> CommandResult:
     """
     Prepares a local folder with configured app artifacts.
     """
+
+    if cli_context.project_definition.native_app is None:
+        raise NoProjectDefinitionError(
+            project_type="native_app", project_file=cli_context.project_root
+        )
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: added util function

sfc-gh-fcampbell (approved):

sfc-gh-turbaszek (approved):

PR #1279, opened 2024-07-03
Add warning when templating is used in v1

sfc-gh-melnacouzi:
### Pre-review checklist

### Changes description
Add warning when templating is used in v1
- we try to parse the template variables, if we find any, we warn.
- if Jinja throws an error, it means templating Syntax was attempted, but was not fully correct. We will also warn in that case.

### DIFF
--- a/src/snowflake/cli/api/utils/definition_rendering.py
+++ b/src/snowflake/cli/api/utils/definition_rendering.py
@@ -254,8 +254,8 @@ def _get_referenced_vars_in_definition(
 
 def _template_version_warning():
     cc.warning(
-        "Ignoring template pattern in 'snowflake.yml' project definition file. "
-        "Update definition_version to 1.1 or later to enable template expansion."
+        "Ignoring template pattern in project definition file. "
+        "Update 'definition_version' to 1.1 or later in snowflake.yml to enable template expansion."
     )
 
 
--- a/src/snowflake/cli/api/utils/definition_rendering.py
+++ b/src/snowflake/cli/api/utils/definition_rendering.py
@@ -254,7 +254,7 @@ def _get_referenced_vars_in_definition(
 
 def _template_version_warning():
     cc.warning(
-        "Ignoring template pattern in project definition file.\n"
+        "Ignoring template pattern in project definition file. "
         "Update project definition version to 1.1 or later to enable template expansion."
     )
 
--- a/src/snowflake/cli/api/utils/definition_rendering.py
+++ b/src/snowflake/cli/api/utils/definition_rendering.py
@@ -252,10 +252,10 @@ def _get_referenced_vars_in_definition(
     return referenced_vars
 
 
-def _template_version_warning(version):
+def _template_version_warning():
     cc.warning(
-        f"Possible use of templating in project definition file.\n"
-        f"Templating is not supported in current definition_version: {version}"
+        "Ignoring template pattern in project definition file.\n"
+        "Update project definition version to 1.1 or later to enable template expansion."
     )
 
 
@@ -296,10 +296,10 @@ def render_definition_template(
                 template_env, definition
             )
             if referenced_vars:
-                _template_version_warning(definition["definition_version"])
+                _template_version_warning()
         except Exception:
             # also warn on Exception, as it means the user is incorrectly attempting to use templating
-            _template_version_warning(definition["definition_version"])
+            _template_version_warning()
 
         project_definition = ProjectDefinition(**original_definition)
         project_context[CONTEXT_KEY]["env"] = environment_overrides
--- a/src/snowflake/cli/api/utils/definition_rendering.py
+++ b/src/snowflake/cli/api/utils/definition_rendering.py
@@ -17,8 +17,9 @@ from __future__ import annotations
 import copy
 from typing import Any, Optional
 
-from jinja2 import Environment, nodes
+from jinja2 import Environment, TemplateSyntaxError, nodes
 from packaging.version import Version
+from snowflake.cli.api.console import cli_console as cc
 from snowflake.cli.api.exceptions import CycleDetectedError, InvalidTemplate
 from snowflake.cli.api.project.schemas.project_definition import (
     ProjectDefinition,
@@ -49,7 +50,13 @@ class TemplatedEnvironment:
 
     def get_referenced_vars(self, template_value: Any) -> set[TemplateVar]:
         template_str = str(template_value)
-        ast = self._jinja_env.parse(template_str)
+        try:
+            ast = self._jinja_env.parse(template_str)
+        except TemplateSyntaxError as e:
+            raise InvalidTemplate(
+                f"Error parsing template from project definition file. Value: '{template_str}'. Error: {e}"
+            ) from e
+
         return self._get_referenced_vars(ast, template_str)
 
     def _get_referenced_vars(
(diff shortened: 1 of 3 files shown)


src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -232,6 +239,26 @@ def _validate_env_section(env_section: dict):
...
+
+    traverse(definition, visit_action=find_any_template_vars)
+
+    return referenced_vars
+
+
+def _template_version_warning(version):
+    cc.warning(
+        f"Possible use of templating in project definition file.\n"
+        f"Templating is not supported in current definition_version: {version}"
```
sfc-gh-bdufour, line 258: In such messages, we should favour the resolution so that the user knows what the action should be. Something like "Ignoring template pattern in file <file>, update the project definition version to 1.1 or later in snowflake.yml to enable template expansion".

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -259,25 +286,29 @@ def render_definition_template(
...
     ) < Version("1.1"):
+        try:
+            referenced_vars = _get_referenced_vars_in_definition(
+                template_env, definition
+            )
+            if referenced_vars:
+                _template_version_warning(definition["definition_version"])
+        except Exception:
+            # also warn on Exception, as it means the user is incorrectly attempting to use templating
+            _template_version_warning(definition["definition_version"])
```
sfc-gh-bdufour, line 302: Can you explain this? Generic exception -> version warning seems a bit odd at first glance. I'm worried we could end up masking something real in the future, for one.

sfc-gh-melnacouzi: It's the same code that runs on both v1 and v1.1. If there is an issue hidden, it should appear with v1.1 templating.
We do not want to surface any accidental exceptions (if we miss to list an exception).

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-melnacouzi: Updated the message.

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -232,6 +239,26 @@ def _validate_env_section(env_section: dict):
...
+    def find_any_template_vars(element):
+        referenced_vars.update(template_env.get_referenced_vars(element))
+
+    traverse(definition, visit_action=find_any_template_vars)
+
+    return referenced_vars
+
+
+def _template_version_warning():
+    cc.warning(
```
sfc-gh-bdufour, line 256: Don't add line breaks, it becomes brittle when we're using indentation. Same line will do. Soon I hope to make this API a bit better.

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -232,6 +239,26 @@ def _validate_env_section(env_section: dict):
...
+        "Ignoring template pattern in project definition file.\n"
+        "Update project definition version to 1.1 or later to enable template expansion."
```
sfc-gh-bdufour, line 258: Why did you end up deciding not to mention snowflake.yml? I thought it would be useful for a user to be guided to the solution.

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-bdufour: Ok, so _get_referenced_vars_in_definition is 1.1+, so really if it fails you're assuming that the user was indeed using templates and should be warned about the PDF version? Is this because they could have an invalid template or something like that?

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -232,6 +239,26 @@ def _validate_env_section(env_section: dict):
...
+
+    def find_any_template_vars(element):
+        referenced_vars.update(template_env.get_referenced_vars(element))
+
+    traverse(definition, visit_action=find_any_template_vars)
+
+    return referenced_vars
+
+
+def _template_version_warning():
```
sfc-gh-bdufour, line 255: I still think that with non-PDF files templating coming soon, this should take in the file name and use that instead of a generic "project definition file" description. This way when we template more files we can have uniform error messages. Thoughts?

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-melnacouzi: what if they're using snowflake.local.yml?

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-melnacouzi: yes exactly

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-melnacouzi: This code is special for project definition file (due to the chicken and egg issue with it). The rest of templating will use simpler templating and they can produce their own messages. Once we move to pdf v2, we can then easily have access to the file name since we won't have 2 merged files anymore. For now, project definition file might be a good compromise (since it matches what's in the documentation - https://docs.snowflake.com/en/developer-guide/snowflake-cli-v2/project-definitions/about)

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-melnacouzi: Replaced them with spaces.

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-cgorrie: I think we can count on the user to understand that changes that apply to `snowflake.yml` can also apply to `snowflake.local.yml`. Having `definition_version` specified here may help too as they can simply search for that string in their project directory.

sfc-gh-bdufour: @sfc-gh-cgorrie My thoughts exactly

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-bdufour: I was after not changing the pattern of the message for the very next feature we'll implement. I generally prefer specific file names in warnings/errors as opposed to vague descriptions that require bouncing to the documentation and back. I won't insist though, do as you see fit.

sfc-gh-bdufour (approved):
Approving despite minor comments so that we can check this in in time for 2.6.0. While my preference would be to resolve the comments with this PR, we can fine tune the error messages as a follow up if we end up running out of time for this release.

tests/api/utils/test_definition_rendering.py
```diff
@@ -83,6 +85,86 @@ def test_no_resolve_in_version_1():
...
+    assert result == {
+        "ctx": {
+            "definition_version": "1",
+            "native_app": {"name": "test_source_<% ctx.env.A %>", "artifacts": []},
+            "env": ProjectEnvironment({}, {}),
+        }
+    }
+    warning_mock.assert_called_once_with(
+        "Ignoring template pattern in 'snowflake.yml' project definition file. "
+        "Update definition_version to 1.1 or later to enable template expansion."
```
sfc-gh-bdufour, line 106: In this context snowflake.yml vs snowflake.local.yml does matter though. If you're saying something like "update the version in your snowflake.yml", it's OK to assume the user understands override. If you say "I found templates in file 'snowflake.yml'" but it's really "snowflake.local.yml", then it's misleading, as the name is expected to be a specific file on the FS. If you can't get the specific file name without significant surgery, I'd do something like:

```
Ignoring template pattern in project definition file. Update 'definition_version' to 1.1 or later in snowflake.yml to enable template expansion.
```

sfc-gh-melnacouzi: updated

sfc-gh-bdufour (approved):

sfc-gh-turbaszek (approved):

sfc-gh-astus (approved):

PR #1280, opened 2024-07-03
Workspaces entities implementation: application package and application

sfc-gh-gbloom:
### Pre-review checklist

### Changes description
...


src/snowflake/cli/api/project/schemas/entities/application_entity.py
```diff
@@ -0,0 +1,51 @@
...
+        validation_alias=AliasChoices("from"),
+        title="An application package this entity should be created from",
+    )
+    debug: Optional[bool] = Field(
+        title="Whether to enable debug mode when using a named stage to create an application object",
+        default=True,
+    )
+
+
+class ApplicationFromField(UpdatableModel):
```
sfc-gh-cgorrie, line 48: In the future, we should put the generic on this field type, e.g.
```
from_: EntityReferenceField[ApplicationPackageEntity]
```

src/snowflake/cli/api/project/schemas/entities/application_entity.py
```diff
@@ -0,0 +1,51 @@
...
+    EntityType,
+    TargetField,
+)
+from snowflake.cli.api.project.schemas.updatable_model import (
+    UpdatableModel,
+)
+
+
+class ApplicationEntity(EntityBase):
+    entity_type: ClassVar[str] = EntityType.APPLICATION
```
sfc-gh-cgorrie, line 34: Possible to have `ClassVar[EntityType]`?

tests_integration/workspaces/test_validate_schema.py
```diff
@@ -0,0 +1,34 @@
...
+from unittest import mock
+
+
+@pytest.mark.integration
+@mock.patch.dict(
+    os.environ,
+    {
+        "SNOWFLAKE_CLI_FEATURES_ENABLE_PROJECT_DEFINITION_V2": os.environ.get(
+            "SNOWFLAKE_CLI_FEATURES_ENABLE_PROJECT_DEFINITION_V2", "true"
+        ),
```
sfc-gh-cgorrie, line 26: Would we ever want to override this to be `false` for this test?

src/snowflake/cli/api/project/schemas/project_definition.py
```diff
@@ -18,16 +18,35 @@
...
 from snowflake.cli.api.project.schemas.snowpark.snowpark import Snowpark
 from snowflake.cli.api.project.schemas.streamlit.streamlit import Streamlit
 from snowflake.cli.api.project.schemas.updatable_model import UpdatableModel
 from snowflake.cli.api.utils.models import ProjectEnvironment
 from snowflake.cli.api.utils.types import Context
 
+_v2_entity_types_map = {
+    EntityType.APPLICATION.value: ApplicationEntity,
+    EntityType.APPLICATION_PACKAGE.value: ApplicationPackageEntity,
+}
```
sfc-gh-cgorrie, line 48: Next up for us is to figure out how these pydantic models map to classes with e.g. deploy, drop, bundle methods

tests_integration/workspaces/test_validate_schema.py
sfc-gh-gbloom: To test that v2 is not recognized by default? I have a unit test that covers it.

sfc-gh-cgorrie: Yeah, that's fair, what I'm wondering is why this isn't just `"SNOWFLAKE_CLI_FEATURES_ENABLE_PROJECT_DEFINITION_V2": "true"`

sfc-gh-cgorrie: Why would we need to override the test with an env var?

src/snowflake/cli/api/project/schemas/entities/application_entity.py
sfc-gh-gbloom: Changed

src/snowflake/cli/api/project/schemas/entities/application_package_entity.py
```diff
@@ -0,0 +1,50 @@
...
+from pydantic import Field
+from snowflake.cli.api.project.schemas.entities.common import (
+    EntityBase,
+    EntityType,
+)
+from snowflake.cli.api.project.schemas.native_app.package import DistributionOptions
+from snowflake.cli.api.project.schemas.native_app.path_mapping import PathMapping
+
+
+class ApplicationPackageEntity(EntityBase):
```
sfc-gh-cgorrie, line 28: There are some properties not accounted for here, like `scratch_stage`, `generated_root`, and `bundle_root`. What's the plan for those?

sfc-gh-gbloom: Added these as well. While we maintain both versions we will have to remember to update v2 schema when we make v1 changes.

sfc-gh-cgorrie (dismissed):

pyproject.toml
```diff
@@ -143,6 +143,9 @@ select = [
   "FA100",  # Missing from __future__ import annotations
   "W605",   # Invalid escape sequences
 ]
+ignore = [
+  "A003"    # Class attribute shadowing a Python builtin
+]
```
sfc-gh-gbloom, line 148: Ignoring flake8 A003 and removing some `noqa: A003` comments. This prevents us from adding a `type` field on Entity classes. Using an alias prevents us from using discriminators (not supported by Pydantic).

sfc-gh-cgorrie (dismissed):
Maybe we want to restrict ignoring `A003` to just the entities dir, but it's OK to me as is.

pyproject.toml
sfc-gh-turbaszek: I don't like it, it's a known source of errors. It's not also limited to `type` but also other builtins. 

I think `noqa: A003` on single line in model (well, every model) is safer.

src/snowflake/cli/api/project/schemas/entities/application_package_entity.py
```diff
@@ -0,0 +1,61 @@
...
+        default="output/deploy/",
+    )
+    generated_root: Optional[str] = Field(
+        title="Subdirectory of the deploy root where files generated by the Snowflake CLI will be written.",
+        default="__generated/",
+    )
+    stage: Optional[str] = Field(
+        title="Identifier of the stage that stores the application artifacts.",
+        default="app_src.stage",
+    )
```
sfc-gh-turbaszek, line 50: Should we consider introducing `IdentifierField` that would run validation that the passed value is proper SQL identifier? What's more it may be useful to cast it to `FQN` instance for further reference.

I think same can be applied to `name` and any other field being a SQL identifier. Later when we add support for schema and database we can use model validators to validate that for example there's no FQN `name` passed together with database and schema.

src/snowflake/cli/api/project/schemas/entities/application_package_entity.py
```diff
@@ -0,0 +1,61 @@
...
+    scratch_stage: Optional[str] = Field(
+        title="Identifier of the stage that stores temporary scratch data used by the Snowflake CLI.",
+        default="app_src.stage_snowflake_cli_scratch",
+    )
+    distribution: Optional[DistributionOptions] = Field(
+        title="Distribution of the application package created by the Snowflake CLI",
+        default="internal",
+    )
+    manifest: str = Field(
```
sfc-gh-turbaszek, line 59:
```suggestion
    manifest: Path = Field(
```
Or `SecurePath`.

src/snowflake/cli/api/project/schemas/entities/common.py
```diff
@@ -0,0 +1,87 @@
...
+)
+from snowflake.cli.api.project.schemas.updatable_model import (
+    IdentifierField,
+    UpdatableModel,
+)
+
+
+class EntityType(Enum):
+    APPLICATION = "application"
+    APPLICATION_PACKAGE = "application package"
```
sfc-gh-turbaszek, line 34: We have those strings as literal on entities classes and we have them here. Should we consider using this Enum as source of truth? Or should we use entities to build Enum / something?

Taking into account that we are constructing `Union[ApplicationEntity, ApplicationPackageEntity]` and
```py
_v2_entity_types_map = {
    EntityType.APPLICATION.value: ApplicationEntity,
    EntityType.APPLICATION_PACKAGE.value: ApplicationPackageEntity,
}
```

It means that adding new entity requires it to be registered in at least two places.


it may be wise to have

```py
KNOWN_ENTITIES = [
  ApplicationEntity,
  ApplicationPackageEntity
]

ENTITIES_TYPE = Union[*KNOWN_ENTITIES]
_v2_entity_types_map = {get_args(eval(inspect.get_annotations(e)["type"]))[0]: e for e in KNOWN_ENTITIES}
```

The dict comprehension is a bit of a magic, but probably there's something smarter we can do.

src/snowflake/cli/api/project/schemas/entities/common.py
```diff
@@ -0,0 +1,87 @@
...
+
+class MetaField(UpdatableModel):
+    warehouse: Optional[str] = IdentifierField(
+        title="Warehouse used to run the scripts", default=None
+    )
+    role: Optional[str] = IdentifierField(
+        title="Role to use when creating the entity object",
+        default=None,
+    )
+    post_deploy: Optional[List[ApplicationPostDeployHook]] = Field(
```
sfc-gh-turbaszek, line 45: Is this common for all entities?

src/snowflake/cli/api/project/schemas/entities/application_package_entity.py
```diff
@@ -0,0 +1,61 @@
...
+from snowflake.cli.api.project.schemas.native_app.package import DistributionOptions
+from snowflake.cli.api.project.schemas.native_app.path_mapping import PathMapping
+
+
+class ApplicationPackageEntity(EntityBase):
+    type: Literal["application package"]
+    name: str = Field(
+        title="Name of the application package created when this entity is deployed"
+    )
+    artifacts: List[Union[PathMapping, str]] = Field(
```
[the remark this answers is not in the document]
sfc-gh-astus: `bundle_root`, `deploy_root`, `generated_root` should be `Path` too? Will defaults work on Windows?

src/snowflake/cli/api/project/schemas/project_definition.py
```diff
@@ -105,8 +125,55 @@ def _convert_env(
...
+    )
+
+    @model_validator(mode="before")
+    @classmethod
+    def apply_defaults(cls, data: Dict) -> Dict:
+        """
+        Applies default values that exist on the model but not specified in yml
+        """
+        if "defaults" in data:
+            if "entities" in data:
```
sfc-gh-astus, line 139: if "defaults" in data and "entities" in data:

pyproject.toml
sfc-gh-gbloom: That's fair. I tried to avoid having `noqa` on each entity type. Reverted.

src/snowflake/cli/api/project/schemas/entities/application_package_entity.py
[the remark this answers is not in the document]
sfc-gh-gbloom: Updated all of these to `Path`.
The models here basically mirror the [existing NA schema](https://github.com/snowflakedb/snowflake-cli/blob/main/src/snowflake/cli/api/project/schemas/native_app/native_app.py). I tried to align with the existing types to make it easier to work with the existing code, but we will gradually improve these when we start refactoring the underlying code (probably when we switch to v1->v2 conversion).
For v1 we have additional path validations and transformations downstream, but I agree that it's better to validate the fields earlier here if possible.

src/snowflake/cli/api/project/schemas/entities/common.py
sfc-gh-gbloom: Yep

src/snowflake/cli/api/project/schemas/entities/common.py
sfc-gh-gbloom: Yes, that's what I tried at first but I couldn't make pydantic understand enums as discriminators and reverted to literal strings.

I'm not sure what I think about this dict comprehension, but I get the idea. I'll try to come up with a cleaner way to have a single source of truth.

src/snowflake/cli/api/project/schemas/entities/entities.py
```diff
@@ -0,0 +1,30 @@
...
+)
+from snowflake.cli.api.project.schemas.entities.application_package_entity import (
+    ApplicationPackageEntity,
+)
+
+Entity = Union[ApplicationEntity, ApplicationPackageEntity]
+
+ALL_ENTITIES = [*get_args(Entity)]
+
+v2_entity_types_map = {e.get_type(): e for e in ALL_ENTITIES}
```
sfc-gh-turbaszek, line 30: Nice!

src/snowflake/cli/api/project/schemas/project_definition.py
```diff
@@ -105,8 +117,54 @@ def _convert_env(
...
+                    if target not in entities:
+                        raise ValueError(f"No such target: {target}")
+                    else:
+                        # Validate the target type
+                        target_cls = entity.from_.__class__.model_fields["target"]
+                        target_type = target_cls.annotation.__args__[0]
+                        actual_target_type = entities[target].__class__
+                        if target_type and target_type is not actual_target_type:
+                            raise ValueError(
+                                f"Target type mismatch. Expected {target_type.__name__}, got {actual_target_type.__name__}"
```
sfc-gh-turbaszek, line 161: I imagine that there will more checks on the enitites should we consider something like
```suggestion
            # TODO Automatically detect TargetFields to validate
            if entity.type == ApplicationEntity.get_type():
               self._validate_application_targets(entity)
```
WDYT?

src/snowflake/cli/api/project/schemas/entities/application_package_entity.py
```diff
@@ -0,0 +1,63 @@
...
+        title="Folder at the root of your project where artifacts necessary to perform the bundle step are stored.",
+        default="output/bundle/",
+    )
+    deploy_root: Optional[Path] = Field(
+        title="Folder at the root of your project where the build step copies the artifacts",
+        default="output/deploy/",
+    )
+    generated_root: Optional[Path] = Field(
+        title="Subdirectory of the deploy root where files generated by the Snowflake CLI will be written.",
+        default="__generated/",
```
sfc-gh-turbaszek, line 47: Should default be a Path or conversion happens automagicaly?

sfc-gh-gbloom: Automagically, but made it `Path` here too to make it clearer

src/snowflake/cli/api/project/schemas/project_definition.py
sfc-gh-gbloom: I split it into 2 methods to have slightly better separation, but we will need to refactor deeper when we find a generic way for this validation. Will address it in a dedicated PR.

sfc-gh-turbaszek (approved):

PR #1281, opened 2024-07-03
Allow debug_mode to be uncontrolled in `snow app run`

sfc-gh-cgorrie:
### Pre-review checklist

### Changes description
Changes the default project definition `native_app.application.debug` value to be `None` rather than `True`. This allows the user's changes to debug_mode on the deployed application object to not be clobbered on upgrade. The previous behavior can be opted back into by explicitly setting `debug: true` in `snowflake.yml`. Users that do not interact with debug mode will not experience any change in behavior.


RELEASE-NOTES.md
```diff
@@ -48,6 +48,7 @@
 * Added log into the file with loaded external plugins
 * Warn users if they attempt to use templating with project definition version 1
 * Improved output and format of Pydantic validation errors
+* The `snow app run` command will no longer set `debug_mode=true` on upgrade unless explicitly set in `snowflake.yml`
```
sfc-gh-bdufour, line 51: nit: maybe something like

```
The `snow app run` command will no longer override the debug mode on during an application upgrade unless [...]
```

Just to make it clear what "debug_mode=true" means in case a customer encountered that behaviour in the past (which is somewhat likely I'd bet)

sfc-gh-bdufour (dismissed):

sfc-gh-bdufour (approved):

PR #1285, opened 2024-07-04
De-dupe app create / upgrade path in NativeAppRunProcessor

sfc-gh-cgorrie:
### Pre-review checklist

### Changes description
Previously, `_create_dev_app` and `upgrade_app` duplicated a lot of each others' code. This refactor merges the two codepaths and introduces the `SameAccountInstallMethod` class to encapsulate some of the per-installation-type logic.

### DIFF
--- a/src/snowflake/cli/plugins/nativeapp/run_processor.py
+++ b/src/snowflake/cli/plugins/nativeapp/run_processor.py
@@ -65,6 +65,13 @@ UPGRADE_RESTRICTION_CODES = {
 }
 
 
+def was_app_created_by_cli(show_app_row: dict):
+    return (
+        show_app_row[COMMENT_COL] in ALLOWED_SPECIAL_COMMENTS
+        and show_app_row[VERSION_COL] == LOOSE_FILES_MAGIC_VERSION
+    )
+
+
 class NativeAppRunProcessor(NativeAppManager, NativeAppCommandProcessor):
     def __init__(self, project_definition: NativeApp, project_root: Path):
         super().__init__(project_definition, project_root)
@@ -91,9 +98,7 @@ class NativeAppRunProcessor(NativeAppManager, NativeAppCommandProcessor):
             if show_app_row:
 
                 # Check if not created by Snowflake CLI or not created using "files on a named stage" / stage dev mode.
-                if show_app_row[COMMENT_COL] not in ALLOWED_SPECIAL_COMMENTS or (
-                    show_app_row[VERSION_COL] != LOOSE_FILES_MAGIC_VERSION
-                ):
+                if not was_app_created_by_cli(show_app_row):
                     raise ApplicationAlreadyExistsError(self.app_name)
 
                 # Check for the right owner
--- a/src/snowflake/cli/plugins/nativeapp/run_processor.py
+++ b/src/snowflake/cli/plugins/nativeapp/run_processor.py
@@ -51,6 +51,9 @@ from snowflake.cli.plugins.nativeapp.manager import (
     generic_sql_error_handler,
 )
 from snowflake.cli.plugins.nativeapp.policy import PolicyBase
+from snowflake.cli.plugins.nativeapp.project_model import (
+    NativeAppProjectModel,
+)
 from snowflake.cli.plugins.stage.manager import StageManager
 from snowflake.connector import ProgrammingError
 from snowflake.connector.cursor import DictCursor, SnowflakeCursor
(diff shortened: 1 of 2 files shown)


src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -342,27 +308,28 @@ def upgrade_app(
...
+                with self.use_role(self.package_role):
                     self._execute_query(
-                        f"grant install on application package {self.package_name} to role {self.app_role}"
+                        f"grant install, develop on application package {self.package_name} to role {self.app_role}"
+                    )
+                    self._execute_query(
+                        f"grant usage on schema {self.package_name}.{self.stage_schema} to role {self.app_role}"
+                    )
+                    self._execute_query(
+                        f"grant read on stage {self.stage_fqn} to role {self.app_role}"
```
sfc-gh-cgorrie, line 319: Merging these into one `_execute_queries` is possible, but it removes the ability to test these are being executed with the correct role (we can only assert they were executed at _some_ point).

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -311,24 +276,25 @@ def upgrade_app(
...
-                    if using_clause:
+                    if install_method.is_dev_mode:
                         # if debug_mode is present (controlled), ensure it is up-to-date
                         if self.debug_mode is not None:
                             self._execute_query(
                                 f"alter application {self.app_name} set debug_mode = {self.debug_mode}"
                             )
+
+                    # hooks always executed after a create or upgrade
+                    self._execute_post_deploy_hooks()
```
sfc-gh-cgorrie, line 297: @sfc-gh-gbloom post-deploy hooks weren't being executed on upgrade -- intentional?

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -283,17 +253,12 @@ def drop_application_before_upgrade(
...
                 generic_sql_error_handler(err)
 
-    def upgrade_app(
+    def create_or_upgrade_app(
         self,
         policy: PolicyBase,
-        is_interactive: bool,
-        version: Optional[str] = None,
-        patch: Optional[int] = None,
+        install_method: SameAccountInstallMethod,
```
sfc-gh-cgorrie, line 259: I agonized for a bit over defaulting this to either unversioned_dev or release_directive, but decided to make the call sites explicit instead. It does add to the test noise somewhat but they're more explicit now.

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -65,102 +68,69 @@
...
 
-                except ProgrammingError as err:
-                    if err.errno not in UPGRADE_RESTRICTION_CODES:
-                        generic_sql_error_handler(err)
-                    else:
-                        cc.warning(err.msg)
-                        self.drop_application_before_upgrade(policy, is_interactive)
+        if self._requires_created_by_cli:
+            if show_app_row[COMMENT_COL] not in ALLOWED_SPECIAL_COMMENTS or (
+                show_app_row[VERSION_COL] != LOOSE_FILES_MAGIC_VERSION
```
sfc-gh-cgorrie, line 122: My hot take: I don't think we need this `VERSION_COL` check anymore. The error handler that handles cross-installation-method errors will capture this later and allow the user to action on it, rather than having to just seen an exception here and bailing out

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-gbloom: Are you sure they weren't? On upgrade they should be executed [here](https://github.com/snowflakedb/snowflake-cli/blob/main/src/snowflake/cli/plugins/nativeapp/run_processor.py#L117), and the flow should be covered by [this test](https://github.com/snowflakedb/snowflake-cli/blob/main/tests_integration/nativeapp/test_init_run.py#L448-L465).

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-bdufour: Sounds good, get rid of it.

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-bdufour: yeah that sounds like a fair choice, I'm not convinced there is a suitable default either.

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-cgorrie: That function only applied to dev mode apps. The previous `upgrade_app` method was used for versioned dev and from-release-directive modes.

sfc-gh-cgorrie: Would have expected it to be right before this return statement: https://github.com/snowflakedb/snowflake-cli/blob/main/src/snowflake/cli/plugins/nativeapp/run_processor.py#L332

sfc-gh-bdufour (approved):

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -311,24 +276,25 @@ def upgrade_app(
             # 3. If existing application is found, perform a few validations and upgrade the application object.
             if show_app_row:
 
-                # We skip comment check here, because prod/pre-existing application objects may not be created by the Snowflake CLI.
-                # Check for the right owner
-                ensure_correct_owner(
-                    row=show_app_row, role=self.app_role, obj_name=self.app_name
-                )
+                install_method.ensure_app_usable(self._na_project, show_app_row)
```
sfc-gh-fcampbell, line 279: at first glance, I'm not a huge fan of this API; it seems like `SameAccountInstallMethod` is kind of just a grab bag of tools that needs to be driven externally in just the right way rather than a class that encapsulates some behaviour of its own. However, I don't want to spend time bikeshedding what's still a net positive for dev velocity, so we should just feel free to revisit this design if it proves to be inflexible in the future

sfc-gh-fcampbell (approved):

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-cgorrie: You're right; I think we can push the CRUD logic up into this class and potentially merge it with `NativeAppProjectModel` to become the `Application` entity in PDFv2 CC @sfc-gh-gbloom

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-gbloom: Good catch, these flows were not covered

sfc-gh-cgorrie: I will keep parity in this PR and fix in a subsequent one.

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-cgorrie: Will be addressed in a subsequent PR.

PR #1286, opened 2024-07-04
Allow cross-upgrade to unversioned mode in "snow app run"

sfc-gh-cgorrie:
### Pre-review checklist

### Changes description
Previously, attempts to `snow app run` when the application object already exists but was installed from a version (e.g via `--version` or `--from-release-directive`) would fail with an error. Now, we hook into the existing `drop_application_before_upgrade` logic and allow cross-upgrades as we already could in the other direction (unversioned to versioned).

### DIFF
--- a/src/snowflake/cli/plugins/nativeapp/exceptions.py
+++ b/src/snowflake/cli/plugins/nativeapp/exceptions.py
@@ -41,7 +41,7 @@ class ApplicationCreatedExternallyError(ClickException):
 
     def __init__(self, name: str):
         super().__init__(
-            f'A application object "{name}" not created by Snowflake CLI already exists in the account.'
+            f'An application object "{name}" not created by Snowflake CLI already exists in the account.'
         )
 
 
--- a/src/snowflake/cli/plugins/nativeapp/exceptions.py
+++ b/src/snowflake/cli/plugins/nativeapp/exceptions.py
@@ -36,12 +36,12 @@ class ApplicationPackageDoesNotExistError(ClickException):
         )
 
 
-class ApplicationAlreadyExistsError(ClickException):
+class ApplicationCreatedExternallyError(ClickException):
     """An application object not created by Snowflake CLI exists with the same name."""
 
     def __init__(self, name: str):
         super().__init__(
-            f'A application object "{name}" not created in development mode using files on a named stage already exists in the account.'
+            f'A application object "{name}" not created by Snowflake CLI already exists in the account.'
         )
 
 
--- a/src/snowflake/cli/plugins/nativeapp/run_processor.py
+++ b/src/snowflake/cli/plugins/nativeapp/run_processor.py
@@ -245,6 +245,8 @@ class NativeAppRunProcessor(NativeAppManager, NativeAppCommandProcessor):
                 )
                 raise typer.Exit(1)
         try:
+            cascade_msg = " (cascade)" if cascade else ""
+            cc.step(f"Dropping application object {self.app_name}{cascade_msg}.")
             cascade_sql = " cascade" if cascade else ""
             self._execute_query(f"drop application {self.app_name}{cascade_sql}")
         except ProgrammingError as err:
--- a/src/snowflake/cli/plugins/nativeapp/run_processor.py
+++ b/src/snowflake/cli/plugins/nativeapp/run_processor.py
@@ -41,13 +41,12 @@ from snowflake.cli.plugins.nativeapp.artifacts import BundleMap
 from snowflake.cli.plugins.nativeapp.constants import (
     ALLOWED_SPECIAL_COMMENTS,
     COMMENT_COL,
-    LOOSE_FILES_MAGIC_VERSION,
     PATCH_COL,
     SPECIAL_COMMENT,
     VERSION_COL,
 )
 from snowflake.cli.plugins.nativeapp.exceptions import (
-    ApplicationAlreadyExistsError,
+    ApplicationCreatedExternallyError,
     ApplicationPackageDoesNotExistError,
 )
 from snowflake.cli.plugins.nativeapp.manager import (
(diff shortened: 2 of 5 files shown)


src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -242,6 +239,8 @@ def drop_application_before_upgrade(
                 )
                 raise typer.Exit(1)
         try:
+            cascade_msg = " (cascade)" if cascade else ""
+            cc.step(f"Dropping application object {self.app_name}{cascade_msg}.")
```
sfc-gh-cgorrie, line 243: Opportunistic fix; noticed there was no CLI output when an application is dropped during the upgrade flow.

tests/nativeapp/test_run_processor.py
```diff
@@ -323,8 +321,6 @@ def test_create_dev_app_create_new_w_missing_warehouse_exception(
     [
         ("dummy", LOOSE_FILES_MAGIC_VERSION),
         ("dummy", "dummy"),
-        (SPECIAL_COMMENT, "dummy"),
-        (SPECIAL_COMMENT_OLD, "dummy"),
```
sfc-gh-cgorrie, line 327: We no longer error out on pure version mismatch.

tests/nativeapp/test_run_processor.py
```diff
@@ -1300,6 +1296,63 @@ def test_upgrade_app_fails_upgrade_restriction_error(
...
+    )
+
+    run_processor = _get_na_run_processor()
+    with pytest.raises(typer.Exit):
+        result = run_processor.create_or_upgrade_app(
+            policy=DenyAlwaysPolicy(),
+            is_interactive=False,
+            install_method=SameAccountInstallMethod.unversioned_dev(),
+        )
+        assert result.exit_code == 1
```
sfc-gh-melnacouzi, line 1352: I am having hard time understanding what this test is doing. Shouldn't we expect this to succeed since we now allow versioned to unversioned upgrade? If so, why are we checking for exit_code 1?

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -118,11 +117,9 @@ def ensure_app_usable(self, app: NativeAppProjectModel, show_app_row: dict):
         """Raise an exception if we cannot proceed with install given the pre-existing application object"""
 
         if self._requires_created_by_cli:
-            if show_app_row[COMMENT_COL] not in ALLOWED_SPECIAL_COMMENTS or (
```
sfc-gh-melnacouzi, line 121: should we also add release notes on this?

tests/nativeapp/test_run_processor.py
sfc-gh-bdufour: Looks this is set up as a negative test (see the fixture creation), but like Michel, I wonder about the absence of a positive test here.

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-bdufour: Weird that no snapshot needed to be updated. Gap in IT coverage?

tests/nativeapp/test_run_processor.py
sfc-gh-cgorrie: Yeah, this is to make sure transitioning to an unversioned app goes through the drop-on-upgrade flow which raises this typer exit. A positive test makes sense; I'll add an integration one.

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-cgorrie: It's a relaxation of a case that the user is unlikely to get into. Part of it is covered by the release note (can now cross-upgrade to unversioned) but part of it is not (can now upgrade an unversioned app that was not created by SnowCLI). Happy to add that as a second release note if it makes sense!

src/snowflake/cli/plugins/nativeapp/exceptions.py
```diff
@@ -36,12 +36,12 @@ def __init__(self, name: str):
...
 
 
-class ApplicationAlreadyExistsError(ClickException):
+class ApplicationCreatedExternallyError(ClickException):
     """An application object not created by Snowflake CLI exists with the same name."""
 
     def __init__(self, name: str):
         super().__init__(
-            f'A application object "{name}" not created in development mode using files on a named stage already exists in the account.'
+            f'A application object "{name}" not created by Snowflake CLI already exists in the account.'
```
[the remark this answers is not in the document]
sfc-gh-cgorrie: Hmm, we had Marketing come in and ask us to change everything to this terminology before PuPr IIRC (CC @sfc-gh-bgoel)

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-cgorrie: Yeah, I think by doing the x-upgrade test as a positive parameterized test we'll get a ton more coverage here. But also I think that integration tests don't do snapshot testing unless you ask? I'll check

src/snowflake/cli/plugins/nativeapp/exceptions.py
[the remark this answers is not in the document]
sfc-gh-bgoel: IIRC, the distinction is as follows:
1. `application object` to refer to the same-account application instance, no listings flow.
2. `snowflake native app` for the framework and the cross-account application instance. 
Apparently, `snowflake native apps` is also not accepted which causes problems in sentences with plurals, but I had flagged this to then PMM poc and it might have changed since.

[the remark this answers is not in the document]
sfc-gh-cgorrie: Ha, I completely missed the fact you're just saying it should be `An` instead of `A`. Will fix after I get tests passing :joy:

tests_integration/nativeapp/test_init_run.py
```diff
@@ -423,18 +422,35 @@ def test_nativeapp_init_from_repo_with_single_template(
 
 # Tests that application post-deploy scripts are executed by creating a post_deploy_log table and having each post-deploy script add a record to it
 @pytest.mark.integration
-def test_nativeapp_app_post_deploy(runner, snowflake_session, project_directory):
+@pytest.mark.parametrize("is_versioned", [True, False])
+def test_nativeapp_app_post_deploy(
+    runner, snowflake_session, project_directory, is_versioned
+):
```
sfc-gh-cgorrie, line 428: When merging the logic, we noticed that post-deploy scripts weren't being called post-upgrade for versioned / release directive install modes. So I parameterized this test to confirm they are now working.

tests/nativeapp/test_run_processor.py
```diff
@@ -1300,6 +1296,98 @@ def test_upgrade_app_fails_upgrade_restriction_error(
...
+@mock_connection()
+def test_versioned_app_upgrade_to_unversioned(
+    mock_conn,
+    mock_get_existing_app_info,
+    mock_execute,
+    temp_dir,
+    mock_cursor,
+):
+    """
+    Ensure that attempting to an upgrade from a versioned dev mode
```
sfc-gh-fcampbell, line 1310:
```suggestion
    Ensure that attempting to upgrade from a versioned dev mode
```

sfc-gh-melnacouzi (approved):

PR #1292, opened 2024-07-08
[NADE][Bugfix] Use resolved app warehouse for regionless query

sfc-gh-bgoel:
### Pre-review checklist

### Changes description
As part of fetching the URL where the Native App is live, the CLI needs to execute a SQL query which requires the use of a warehouse. The `application.warehouse` property in the manager class resolves this to either be the app warehouse if found, or the one derived from the connection. Since this URL is pertaining to the app instance, we have decided to execute this query by only activating app warehouse.


src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -254,6 +257,26 @@ def app_post_deploy_hooks(self) -> Optional[List[ApplicationPostDeployHook]]:
     def debug_mode(self) -> bool:
         return self.na_project.debug_mode
 
+    @contextmanager
+    def use_warehouse(self, new_wh: Optional[str]):
```
sfc-gh-bgoel, line 261: This is similar to the `use_role` contextmanager we also have, but that is part of the parent `SqlExecutionMixin` class as it could be widely used. But the `use_warehouse` is only being used in this new change. Hence keeping it here. But open to:
1. Bumping this up to the parent
2. Creating a generic context manager `use_object` similar to the `self.use` method on line 273.

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -254,6 +257,26 @@ def app_post_deploy_hooks(self) -> Optional[List[ApplicationPostDeployHook]]:
...
+        """
+        Switches to a different warehouse for a while, then switches back.
+        This is a no-op if the requested warehouse is already active, or the requested warehouse is not available.
+        """
+        wh_result = self._execute_query(
+            f"select current_warehouse()", cursor_class=DictCursor
+        ).fetchone()
+        prev_wh = wh_result["CURRENT_WAREHOUSE()"]
+        is_different_wh = (new_wh is not None) and (new_wh.lower() != prev_wh.lower())
```
sfc-gh-melnacouzi, line 270: do we need to pay any consideration for quoted/unquoted identifiers?

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -254,6 +257,26 @@ def app_post_deploy_hooks(self) -> Optional[List[ApplicationPostDeployHook]]:
...
+        if is_different_wh:
+            self._log.debug("Assuming different warehouse: %s", new_wh)
```
sfc-gh-melnacouzi, line 272: nit: maybe something like "Switching to warehouse x" or "Assuming warehouse x"... 'different' seems a bit off.

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -254,6 +257,26 @@ def app_post_deploy_hooks(self) -> Optional[List[ApplicationPostDeployHook]]:
     def debug_mode(self) -> bool:
         return self.na_project.debug_mode
 
+    @contextmanager
+    def use_warehouse(self, new_wh: Optional[str]):
+        """
+        Switches to a different warehouse for a while, then switches back.
+        This is a no-op if the requested warehouse is already active, or the requested warehouse is not available.
```
sfc-gh-melnacouzi, line 264: If the requested warehouse is not available, shouldn't we throw an error?

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-bgoel: I don't believe so. The counterpart `use_role` and the subcall to `use` also do not use it.

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-bgoel: Also similar to the `use_role` contextmanager. If we change it here, we'd have to change it there too. I can add that in for both of them. Just FYI, that will also require SnowCLI team's approval, so it might introduce a bit of a delay with their business hours.

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-bgoel: I will clarify the comment, but the distinction here is:
1. If the CLI is able to determine the app warehouse or a warehouse from the connection context, then use that, in that order.
2. If nothing can be found, then error out as usual.

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-melnacouzi: I am just thinking about the usecase where you have 2 different warehouses with the same chars: 'Warehouse' and 'warehousE'. It's a very corner case, but if it's ignored in other places (like role), maybe we can leave it out of scope for this PR.

sfc-gh-melnacouzi (approved):

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-cgorrie: Can we re-use an existing utility function--or extract it out from somewhere--to check if the two identifiers are equal? I think these are the "identifier as string in resultset" so the rules are slightly different. Also, does this code currently fail if `prev_wh is None`?

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -254,6 +257,30 @@ def app_post_deploy_hooks(self) -> Optional[List[ApplicationPostDeployHook]]:
...
+        if is_different_wh:
+            self._log.debug(
+                "Temporarily switching to a different warehouse: %s", new_wh
+            )
+            self.use(object_type=ObjectType.WAREHOUSE, name=new_wh)
+        try:
+            yield
+        finally:
+            if is_different_wh:
+                self._log.debug("Switching back to the original warehouse: %s", prev_wh)
```
sfc-gh-cgorrie, line 281:
```suggestion
                self._log.debug("Switching back to warehouse: %s", prev_wh)
```

sfc-gh-cgorrie (commented):
Can you also audit other potential gaps we might have missed -- like post-deploy scripts or package scripts? We should use this new `use_warehouse` primitive everywhere it makes sense.

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -254,6 +257,30 @@ def app_post_deploy_hooks(self) -> Optional[List[ApplicationPostDeployHook]]:
...
+                self.use(object_type=ObjectType.WAREHOUSE, name=prev_wh)
```
sfc-gh-cgorrie, line 282: Also, what happens here if `prev_wh is None`?

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-bdufour: +1, this comparison seems wrong, it needs to consider quoted vs unquoted IDs at least (not to mention the case sensitivity param, but I'm OK ignoring that)

sfc-gh-bgoel: Added the built-in `to_identifier` utility in the new commit, up for review.

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -145,15 +145,14 @@ def _execute_sql_script(self, sql_script_path):
         """
         with open(sql_script_path) as f:
             sql_script = f.read()
-            try:
-                if self.application_warehouse:
-                    self._execute_query(f"use warehouse {self.application_warehouse}")
```
sfc-gh-bgoel, line 150: `_execute_sql_script` is only called from the context of creating/updating an app, where the warehouse has already been activated. Hence removed this additional query from here. But if folks feel strongly about this, it can be added back.

tests/nativeapp/test_package_scripts.py
```diff
@@ -175,17 +197,26 @@ def test_package_scripts_w_warehouse_access_exception(
...
     native_app_manager = _get_na_manager(str(working_dir))
 
     with pytest.raises(ProgrammingError) as err:
         native_app_manager._apply_package_scripts()  # noqa: SLF001
 
-    assert "Please grant usage privilege on warehouse to this role." in err.value.msg
+    assert (
+        "Could not use warehouse MockWarehouse. Object does not exist, or operation cannot be performed."
+        in err.value.msg
+    )
```
sfc-gh-bgoel, line 222: This error message is now different because instead of calling `use warehouse` directly which uses our `generic_error_handler`, it calls the method `self.use` provided by the CLI which has its own error handling.

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -254,6 +258,69 @@ def app_post_deploy_hooks(self) -> Optional[List[ApplicationPostDeployHook]]:
...
+        wh_result = self._execute_query(
+            f"select current_warehouse()", cursor_class=DictCursor
+        ).fetchone()
+        # If user has an assigned default warehouse, prev_wh will contain a value even if the warehouse is suspended.
+        prev_wh = wh_result["CURRENT_WAREHOUSE()"]
+
+        if new_wh is not None:
+            new_wh_id = to_identifier(new_wh)
+
+        if prev_wh is None and new_wh is None:
```
sfc-gh-melnacouzi, line 277: this logic is a bit too complicated.
1) Why do we need to stay with prev_wh if new_wh is not defined? is it a valid use case that new_wh is None (especially when dealing with snow app run)? Shouldn't we just error?

2) too many combinations. I suggest to simplify as follows if possible:

a) if new_wh is None, error out.
b) proceed to switch to new_wh if different from prev_wh.
c) yield
d) if prev_wh was None, do nothing, otherwise, switch back to it.

sfc-gh-bgoel: Some concerns raised in the previous comments:
1. It is possible that prev_wh is None, especially in the case when no default warehouse is assigned to a user, and hence `current_warehouse()` may return None. I do not have a way of verifying this as I do not have permissions to set up a new user in an account, but I do believe this is a case we need to account for since the bug ticket was created for a scenario like this when there is no fallback.
2. It is possible that new_wh is None. In our case, `package_warehouse` or `application_warehouse` could be None if user has not provided that warehouse in their snowflake.yml, config.toml and there is no default warehouse. In that case, the connection context would return a None for this property.
3. Given the two above, we do need to tackle their nullity accordingly.
But I am open to hearing if the assumptions in 1 and/or 2 are incorrect and therefore do not need to be handled.

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -254,6 +258,69 @@ def app_post_deploy_hooks(self) -> Optional[List[ApplicationPostDeployHook]]:
...
+        If there is no default warehouse in the account, it will throw an error.
+        """
+        wh_result = self._execute_query(
+            f"select current_warehouse()", cursor_class=DictCursor
+        ).fetchone()
+        # If user has an assigned default warehouse, prev_wh will contain a value even if the warehouse is suspended.
+        prev_wh = wh_result["CURRENT_WAREHOUSE()"]
+
+        if new_wh is not None:
+            new_wh_id = to_identifier(new_wh)
```
sfc-gh-cgorrie, line 275: I think this `to_identifier` belongs in e.g. `nativeapp.project_model`, as that's where the YAML strings are turned into identifiers for other parts of the project definition.

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-cgorrie: I'm with Michel, I think we can simplify it as suggested. Sent you a way to determine if warehouse is not set.

src/snowflake/cli/plugins/nativeapp/project_model.py
```diff
@@ -105,16 +105,16 @@ def stage_schema(self) -> Optional[str]:
     @cached_property
     def package_warehouse(self) -> Optional[str]:
         if self.definition.package and self.definition.package.warehouse:
-            return self.definition.package.warehouse
+            return to_identifier(self.definition.package.warehouse)
         else:
-            return cli_context.connection.warehouse
+            return to_identifier(cli_context.connection.warehouse)
```
sfc-gh-cgorrie, line 110: `cli_context.connection.warehouse` might be `None`, in which case we don't want to pipe it through `to_identifier`.

src/snowflake/cli/plugins/nativeapp/project_model.py
```diff
@@ -105,16 +105,16 @@ def stage_schema(self) -> Optional[str]:
...
 
     @cached_property
     def application_warehouse(self) -> Optional[str]:
         if self.definition.application and self.definition.application.warehouse:
-            return self.definition.application.warehouse
+            return to_identifier(self.definition.application.warehouse)
         else:
-            return cli_context.connection.warehouse
+            return to_identifier(cli_context.connection.warehouse)
```
sfc-gh-cgorrie, line 117: `cli_context.connection.warehouse` might be `None`, in which case we don't want to pipe it through `to_identifier`.

src/snowflake/cli/plugins/nativeapp/run_processor.py
sfc-gh-cgorrie: Make sure to change the docstring to reflect this change. I agree it should not be here because it might be for the application _or_ for the application package.

sfc-gh-cgorrie (commented):
My requests: account for `None` in the project model warehouse properties, and simplify the use warehouse context manager as suggested by Michel.

sfc-gh-melnacouzi (approved):

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -267,89 +267,85 @@ def create_or_upgrade_app(
...
 
             # 1. Need to use a warehouse to create an application object
-            try:
-                if self.application_warehouse:
-                    self._execute_query(f"use warehouse {self.application_warehouse}")
-            except ProgrammingError as err:
-                generic_sql_error_handler(
-                    err=err, role=self.app_role, warehouse=self.application_warehouse
-                )
+            with self.use_warehouse(self.application_warehouse):
```
sfc-gh-cgorrie, line 270: (This applies to all use_warehouse callsites)
I'm still concerned about the case when the passed in warehouse is `None`, which happens when neither `snowflake.yml` nor the connection configuration gives you a warehouse. What you'll get is the fatal `ClickException` thrown that warehouse cannot be `None`. I don't see a test that covers that.

Also, is that error actionable by the user? By throwing it inside `use_warehouse` we don't have the context of why we need the warehouse that would let us throw a better exception.

I have a proposal here:
1. Make `use_warehouse` take `str`, not `Optional[str]`
2. Refactor `self.application_warehouse` -> `self.use_application_warehouse` and throw the error _there_ if there's no warehouse we can resolve --> the error message becomes a lot more meaningful
3. Same for `self.package_warehouse`
4. Add tests that exercise the new exceptions

This refactor assumes we never need to just get the name of either warehouse. We may need both the existing + `use_` versions.

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -218,10 +218,34 @@ def stage_schema(self) -> Optional[str]:
...
+            )
+        )
+
     @property
     def application_warehouse(self) -> Optional[str]:
         return self.na_project.application_warehouse
 
+    def use_application_warehouse(self) -> str:
+        if self.application_warehouse:
+            return self.application_warehouse
```
sfc-gh-cgorrie, line 239: Rather than return a string, I would expect this to be:
```
if self.application_warehouse:
  with self.use_warehouse(self.application_warehouse):
    yield
else:
  raise ...
```

sfc-gh-cgorrie (approved):
LGTM

sfc-gh-melnacouzi (approved):

sfc-gh-astus (approved):

sfc-gh-mraba (approved):

PR #1298, opened 2024-07-09
Allow adding a version to a package without CLI magic comment

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Prompts the user when running `snow app version create` against a package that was not created by the CLI. The standard `--interactive/--no-interactive` and `--force` flags apply as usual.

### DIFF
--- a/src/snowflake/cli/plugins/nativeapp/version/version_processor.py
+++ b/src/snowflake/cli/plugins/nativeapp/version/version_processor.py
@@ -34,6 +34,7 @@ from snowflake.cli.plugins.nativeapp.artifacts import (
 )
 from snowflake.cli.plugins.nativeapp.constants import VERSION_COL
 from snowflake.cli.plugins.nativeapp.exceptions import (
+    ApplicationPackageAlreadyExistsError,
     ApplicationPackageDoesNotExistError,
 )
 from snowflake.cli.plugins.nativeapp.manager import (
@@ -64,12 +65,14 @@ def check_index_changes_in_git_repo(
         # Check if the repo has any changes, including untracked files
         if repo.is_dirty(untracked_files=True):
             cc.warning(
-                "Changes detected in the git repository. (Rerun your command with --skip-git-check flag to ignore this check)"
+                "Changes detected in the git repository. "
+                "(Rerun your command with --skip-git-check flag to ignore this check)"
             )
             repo.git.execute(["git", "status"])
 
             user_prompt = (
-                "You have local changes in this repository that are not part of a previous commit. Do you still want to continue?",
+                "You have local changes in this repository that are not part of a previous commit. "
+                "Do you still want to continue?"
             )
             if not policy.should_proceed(user_prompt):
                 if is_interactive:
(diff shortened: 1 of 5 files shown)


src/snowflake/cli/plugins/nativeapp/version/version_processor.py
```diff
@@ -64,12 +65,14 @@ def check_index_changes_in_git_repo(
...
         if repo.is_dirty(untracked_files=True):
             cc.warning(
-                "Changes detected in the git repository. (Rerun your command with --skip-git-check flag to ignore this check)"
+                "Changes detected in the git repository. "
+                "(Rerun your command with --skip-git-check flag to ignore this check)"
             )
             repo.git.execute(["git", "status"])
 
             user_prompt = (
-                "You have local changes in this repository that are not part of a previous commit. Do you still want to continue?",
```
sfc-gh-fcampbell, line 72: this was actually a tuple is disguise:
image

src/snowflake/cli/plugins/nativeapp/version/version_processor.py
```diff
@@ -64,12 +65,14 @@ def check_index_changes_in_git_repo(
...
+                "You have local changes in this repository that are not part of a previous commit. "
```
sfc-gh-fcampbell, line 74: updated to reduce line length, message doesn't change

src/snowflake/cli/plugins/nativeapp/version/version_processor.py
```diff
@@ -64,12 +65,14 @@ def check_index_changes_in_git_repo(
...
+                "Do you still want to continue?"
```
sfc-gh-fcampbell, line 75: updated to reduce line length, message doesn't change

tests/nativeapp/test_version_create_processor.py
```diff
@@ -682,3 +684,94 @@ def test_process_existing_release_directives_w_existing_version_two(
...
+)
+@mock.patch(
+    f"{VERSION_MODULE}.{CREATE_PROCESSOR}.get_existing_release_directive_info_for_version",
+    return_value=None,
+)
+@mock.patch(
+    f"{VERSION_MODULE}.{CREATE_PROCESSOR}.get_existing_version_info", return_value=None
+)
+@mock.patch(f"{VERSION_MODULE}.{CREATE_PROCESSOR}.add_new_version", return_value=None)
+@mock.patch(f"snowflake.cli.plugins.nativeapp.policy.{TYPER_CONFIRM}")
```
sfc-gh-bdufour, line 712: Possibly controversial opinion here: this is way too many mocks. I think the way we write tests makes them brittle when we refactor the code, and is just too much work overall. Possible solutions I can think of: 1) leave this kind of test to integ tests, 2) invest in setting up a good fixture that is used by most tests so that all the patching is done in one place (with reasonable defaults) so that each test only has to override the behaviour it needs. Thoughts? Could be a good PL topic too.

sfc-gh-bdufour (approved):

tests/nativeapp/test_version_create_processor.py
sfc-gh-fcampbell: yeah, i agree there's a lot of copypasta of mocks going on. I think there's still some value in unit tests (especially since they're faster and easier to run locally), so I think your second suggestion is easily doable with `autouse` fixtures at the file or package level

PR #1308, opened 2024-07-11
[SNOW-1528866] Re-enable some unit tests on Windows (1/N)

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Re-enables a few unit tests on Windows. Issues fixed:
- Gitignored all dirs that start with `.venv`, so we can create a separate `.venv-win` for example
- Keep file iteration sorted to generate deterministic snapshots
- Always emit POSIX paths in snapshots
- Generate test TOML files using `tomlkit` instead of raw strings since windows paths like `C:\Users\..` contain `\U`, which is interpreted as a Unicode escape (dumping using `tomlkit` escapes these properly)
- Cleanup logging handlers between each test to close log file handle
- Update tests with hardcoded `/` in paths to use `os.path.join` or `os.sep`
- Some just worked already (/shrug)


src/snowflake/cli/plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -222,7 +224,7 @@ def process(
                         file.write("\n")
                     insert_newline = True
                     file.write(
-                        f"-- Generated by the Snowflake CLI from {relative_py_file}\n"
+                        f"-- Generated by the Snowflake CLI from {relative_py_file.as_posix()}\n"
```
sfc-gh-fcampbell, line 227: this will emit a posix path on Windows. IMO it's OK because this path is in the deploy root, which technically represents stage paths, which use forward slashes. It also matches with other paths like those in `IMPORTS`

tests/nativeapp/codegen/snowpark/__snapshots__/test_python_processor.ambr
```diff
@@ -545,8 +545,8 @@
 # name: test_process_with_collected_functions.1
```
sfc-gh-fcampbell, line 545: the iteration order happened to get reversed when adding `sorted()`, so the data in the snapshot here is reversed

tests/nativeapp/utils.py
```diff
@@ -153,11 +153,13 @@ def assert_dir_snapshot(root: Path, os_agnostic_snapshot) -> None:
...
-    for path in all_paths:
+    for path in sorted(all_paths):
         if path.is_file():
-            snapshot_contents = f"===== Contents of: {path} =====\n"
+            snapshot_contents = f"===== Contents of: {path.as_posix()} =====\n"
             snapshot_contents += path.read_text(encoding="utf-8")
-            assert snapshot_contents == os_agnostic_snapshot
+            assert (
+                snapshot_contents == os_agnostic_snapshot
+            ), f"\nExpected:\n{os_agnostic_snapshot}\nGot:\n{snapshot_contents}"
```
sfc-gh-fcampbell, line 162: pytest doesn't show expected vs actual if the `assert` statement isn't in the test function itself, so let's show it here

sfc-gh-turbaszek (dismissed):

sfc-gh-bdufour (approved):

sfc-gh-bdufour (approved):

sfc-gh-turbaszek (approved):

PR #1313, opened 2024-07-12
[SNOW-1528866] Re-enable some unit tests on Windows (2/N)

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Updating more unit tests to work on Windows. These ones were concerned with differences between venvs on Unix vs Windows. Some tests also just passed when enabled, and others use symlinks so they remain disabled at the test level on Windows.

### DIFF
--- a/tests/nativeapp/codegen/test_sandbox.py
+++ b/tests/nativeapp/codegen/test_sandbox.py
@@ -21,9 +21,6 @@ import snowflake.cli.plugins.nativeapp.codegen.sandbox as sandbox
 
 from tests_common import IS_WINDOWS
 
-if IS_WINDOWS:
-    pytest.skip("Requires further refactor to work on Windows", allow_module_level=True)
-
 PYTHON_SCRIPT = """
 import sys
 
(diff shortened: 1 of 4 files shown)


sfc-gh-turbaszek (approved):

tests/nativeapp/codegen/test_sandbox.py
```diff
@@ -49,8 +46,15 @@ def mock_environ():
 
 
 @pytest.fixture
-def fake_venv_root(temp_dir):
-    venv_root = Path(temp_dir) / "venv"
+def fake_venv_root(fake_venv_root_unix, fake_venv_root_win32):
+    if IS_WINDOWS:
+        return fake_venv_root_win32
+    return fake_venv_root_unix
```
sfc-gh-bdufour, line 52: Can you explain what the actual issue was? These tests were set up to mimic the file/dir structure, but really shouldn't have relied on the actual OS behaviour. As updated, we seem to lose the ability to run across operating systems, which was part of the original intention.

sfc-gh-fcampbell: the path that the CLI looks for is different based on the value of `sys.platform`: https://github.com/snowflakedb/snowflake-cli/blob/a0ad54f90b62173857a8c6d7df8c2f7f6b089d26/src/snowflake/cli/plugins/nativeapp/codegen/sandbox.py#L106-L109

sfc-gh-fcampbell: the reason these tests were failing was because we were creating a temporary fake venv with `<root>/bin/python` when the CLI was looking for `<root>/Scripts/python.exe` in some of the tests that don't patch `sys.platform`

sfc-gh-bdufour: Ok, that makes sense now, thanks.

sfc-gh-bdufour (approved):

PR #1330, opened 2024-07-17
Add support for defaults in project definition file templating

sfc-gh-melnacouzi:
### Pre-review checklist

### Changes description
Add support for project definition file defaults in templating.
- 3 new template functions introduced: fn.id_concat, fn.id_to_str, fn.str_to_id.
- Defaults will be extracted from Pydantic model and automatically be available for templating (in definition file or in SQL, or other places that uses ctx)
- Defaults can include templates themselves

To achieve this feature, we have to load project definition file into Pydantic twice. Once to get the defaults (while skipping validation on templated values), and second time, to fully parse all the values and validations.
In order to skip validation on templated value, I had to:
1) Pass in a context to Pydantic indicating that I am in the first round of parsing, so ignore any templates you see.
2) Wrap all the Pydantic fields with a custom validator that can shortcircuit the validation when templates are encountered. Note that we can't fully disable Pydantic validation as it won't be able to traverse the model recursively anymore.

### Testing:
- Added unit tests for each of the new template functions added to ensure correctness and input validation.
- Added unit tests verifying that field_validation is ignored on templates fields, and not-ignored otherwise.
- Verified default fields are populated through unit tests.
- Ensured existing tests are still passing.
- Tested manually with invalid input to new functions.
- Tested manually that default fields are available.


src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -305,11 +352,19 @@ def render_definition_template(
...
 
-    default_env = definition.get("env", {})
-    _validate_env_section(default_env)
+    definition = _add_defaults_to_definition(definition)
+    project_context = {CONTEXT_KEY: definition}
+
+    _validate_env_section(definition.get("env", {}))
+
+    # add available templating functions
+    project_context["id_concat"] = TemplatingFunctions.id_concat
```
sfc-gh-melnacouzi, line 365: These need alignment - they're currently top level functions, so users can do:
id_concat('ctx.env.x', 'ctx.env.y')...

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -305,11 +352,19 @@ def render_definition_template(
...
+    project_context["to_id"] = TemplatingFunctions.to_id
+    project_context["to_str"] = TemplatingFunctions.to_str
```
sfc-gh-melnacouzi, line 367: to_id and to_str are not currently used - they could be useful, but we can hold off on introducing them yet.

tests/project/test_util.py
```diff
@@ -231,3 +233,53 @@ def test_to_string_literal(raw_string, literal):
 )
 def test_escape_like_pattern(raw_string, escaped):
     assert escape_like_pattern(raw_string) == escaped
+
+
+@pytest.mark.parametrize(
+    "id1, id2, concatenated_value",
+    [
```
sfc-gh-melnacouzi, line 240: we need agreement on this. For now, I only add quotes in the result if one of the 2 ids originally had quotes.

sfc-gh-astus (approved):

src/snowflake/cli/api/project/schemas/updatable_model.py
```diff
@@ -14,17 +14,125 @@
...
+        _initial_context.reset(token)
+
 
 class UpdatableModel(BaseModel):
     model_config = ConfigDict(validate_assignment=True, extra="forbid")
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(**kwargs)
+    def __init__(self, /, **data: Any) -> None:
+        self.__pydantic_validator__.validate_python(
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: This suggestion came from Pydantic: https://docs.pydantic.dev/latest/concepts/validators/#using-validation-context-with-basemodel-initialization

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -262,6 +276,38 @@ def _template_version_warning():
...
+        return identifier_to_str(a)
+
+
+def _add_defaults_to_definition(definition: Definition):
+    with context({"includes_templates": True}):
+        # pass a flag to Pydantic to skip validation for templated scalars
+        # populate the defaults
+        project_definition = build_project_definition(**definition)
+
+    return project_definition.model_dump(
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: I use useAlias=true, and make sure the aliases use alias field.

src/snowflake/cli/api/project/util.py
```diff
@@ -183,6 +195,34 @@ def get_env_username() -> Optional[str]:
...
+        quoted_found = True
+
+    if is_valid_quoted_identifier(identifier2):
+        identifier2 = identifier_to_str(identifier2)
+        quoted_found = True
+
+    if quoted_found:
+        return quote_identifier(f"{identifier1}{identifier2}")
+
+    return f"{identifier1}{identifier2}"
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: That's one of the points i wanted to discuss. If neither of the 2 identifiers were quoted, I did not add quotes on purpose, for the purpose of templating.
Note that the rest of the logic for application name/package name will add to_identifier to it anyways, so it's not a problem from that perspective. Let me know what you think.

src/snowflake/cli/api/project/schemas/native_app/package.py
```diff
@@ -45,11 +47,18 @@ class Package(UpdatableModel):
...
                 "package.scripts field should contain unique values. Check the list for duplicates and try again"
             )
         return input_list
+
+
+class PackageV11(Package):
+    # Templated defaults only supported in v1.1+
+    name: Optional[str] = IdentifierField(
+        title="Name of the application package created when you run the snow app run command",
+        default=f"<% id_concat(ctx.native_app.name, '_pkg_{get_sanitized_username()}') %>",
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: Could you please elaborate on that? This default will end up in documentation as is? and is there a way to overwrite the docs?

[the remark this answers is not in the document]
sfc-gh-melnacouzi: might not be a bad idea to expose sanitized_username as a command.

src/snowflake/cli/api/project/schemas/updatable_model.py
```diff
@@ -14,17 +14,125 @@
...
+                    cls._add_validator(field)
+
+    @classmethod
+    def _add_validator(cls, field_name: str):
+        def _validator(cls, value, handler):
+            return handler(value)
+
+        setattr(
+            cls,
+            f"_template_validate_{field_name}",
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: no, I am just trying to avoid conflicts with names developers might choose. The more complex the better.

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-melnacouzi: will use fn as the name space.

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -262,6 +276,38 @@ def _template_version_warning():
...
+    def to_id(a):
+        return to_identifier(a)
+
+    @staticmethod
+    def to_str(a):
+        return identifier_to_str(a)
+
+
+def _add_defaults_to_definition(definition: Definition):
+    with context({"includes_templates": True}):
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: how does skip_validation_on_templates sound?

src/snowflake/cli/api/project/schemas/updatable_model.py
```diff
@@ -14,17 +14,125 @@
...
+
+    @classmethod
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__(**kwargs)
+
+        # Collect all the Pydantic fields
+        # from all the classes in the inheritance chain
+        all_fields = {}
+        for class_ in cls.__mro__:
+            if hasattr(class_, "__annotations__"):
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: It's less forgiving in terms of class orders because it evaluates the types. Not a big deal, but I need to loop over the hierarchy to collect field values as well, so I kept it this way - please check updated version.

tests/project/test_util.py
sfc-gh-melnacouzi: yes corret.

src/snowflake/cli/api/project/util.py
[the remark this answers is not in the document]
sfc-gh-melnacouzi: updated so it checks the input, and ensure it always outputs a valid identifier.

src/snowflake/cli/api/project/schemas/native_app/package.py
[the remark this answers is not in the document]
sfc-gh-melnacouzi: As discussed offline, we will keep this internal for now, and expose it in the future if needed.

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -281,14 +327,15 @@ def render_definition_template(
 
     # start with an environment from overrides and environment variables:
     override_env = context_overrides.get(CONTEXT_KEY, {}).get("env", {})
+
+    # default_env not resolved yet, because it could be templated
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: updated.

src/snowflake/cli/api/utils/definition_rendering.py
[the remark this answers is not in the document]
sfc-gh-melnacouzi: refactored that part.

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -338,7 +393,12 @@ def on_cycle_action(node: Node[TemplateVar]):
         update_action=lambda val: template_env.render(val, final_context),
     )
 
-    definition["env"] = ProjectEnvironment(default_env, override_env)
+    default_env = definition.get("env", {})
     project_context[CONTEXT_KEY] = definition
     project_definition = build_project_definition(**definition)
+    # Overwrite the env in the Project Context to automatically handle overwrites
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: rephrased.

sfc-gh-turbaszek (dismissed):

src/snowflake/cli/api/project/schemas/updatable_model.py
```diff
@@ -14,17 +14,147 @@
...
+        """
+        This method will collect all the Pydantic annotations for the class
+        currently being initialized (any subclass of `UpdatableModel`).
+
+        It will add a field validator wrapper for every Pydantic field
+        in order to skip validation when templates are found.
+
+        It will apply this to all Pydantic fields, except for fields
+        marked as `EntityTypeField`. These will be skipped because
+        Pydantic does not support validators for discriminator field types.
```
sfc-gh-bdufour, line 109: nit: Would it be help for clarity then to call some kind if "is_discriminator_field" abstraction that could be extended later if needed?

src/snowflake/cli/api/project/schemas/updatable_model.py
```diff
@@ -14,17 +14,147 @@
...
+        """
+
+        super().__init_subclass__(**kwargs)
+
+        field_annotations = {}
+        field_values = {}
+        # Go through the inheritance classes in reverse order
```
sfc-gh-bdufour, line 116: nit: explain why reverse order here please (I remember from our conversation because it was only last week, but I don't trust I'll immediately remember later).

src/snowflake/cli/api/project/schemas/updatable_model.py
```diff
@@ -14,17 +14,147 @@
...
+        """
+
+        def validator_skipping_templated_str(cls, value, handler, info: ValidationInfo):
+            if _is_templated(info, value):
+                return value
+            return handler(value)
+
+        setattr(
+            cls,
+            f"_field_validator_with_random_name_to_avoid_name_conflict_{field_name}",
```
sfc-gh-bdufour, line 155: nit: it's 100% deterministic and not random at all :)

src/snowflake/cli/api/project/util.py
```diff
@@ -88,6 +88,11 @@ def is_valid_object_name(name: str, max_depth=2, allow_quoted=True) -> bool:
     return re.fullmatch(pattern, name) is not None
 
 
+def quote_identifier(identifier: str) -> str:
+    # double quote the identifier
+    return '"' + identifier.replace('"', '""') + '"'
```
sfc-gh-bdufour, line 93: technically what's passed in is not an identifier, it's a string. This function is rather dangerous as written, from the calling side we could end up quoting a quoted ID and it would look fine. How about to_quoted_identifier(value: str), and if the value is already quoted, do nothing, otherwise quote it? It would be a bit safer.

src/snowflake/cli/api/project/util.py
```diff
@@ -183,6 +195,34 @@ def get_env_username() -> Optional[str]:
...
 
 
+DEFAULT_USERNAME = "unknown_user"
+
+
+def get_sanitized_username():
+    return clean_identifier(get_env_username() or DEFAULT_USERNAME)
+
+
+def concat_identifiers(identifiers: list[str]) -> str:
```
sfc-gh-bdufour, line 205: nit: why not * identifiers so we can use the variadic version in internal logic too?

src/snowflake/cli/api/utils/templating_functions.py
```diff
@@ -0,0 +1,85 @@
...
+
+from snowflake.cli.api.exceptions import InvalidTemplate
+from snowflake.cli.api.project.util import (
+    concat_identifiers,
+    identifier_to_str,
+    to_identifier,
+)
+
+
+class TemplatingFunctions:
```
sfc-gh-bdufour, line 27: add doc strings to all of these. They'll need to make their way to the public docs, so we need to be rather thorough here.

sfc-gh-bdufour, line 27: Why do these even live in a class? Feels like overhead with no benefit.

src/snowflake/cli/api/utils/templating_functions.py
```diff
@@ -0,0 +1,85 @@
...
+
+def get_templating_functions():
+    """
+    Returns a dictionary with all the functions available for templating
+    """
+    templating_functions = {
+        func: getattr(TemplatingFunctions, func)
+        for func in dir(TemplatingFunctions)
+        if callable(getattr(TemplatingFunctions, func)) and not func.startswith("_")
+    }
```
sfc-gh-bdufour, line 83: Publicly exposed functions should be exported explicitly, by name. Something like:
```
 {
    "foo": foo,
    "bar": bar,
}
```

would be a lot clearer (and would make aliases trivial too if we ever need them). Or take a look at the @public scheme I'm using in the new repo, but IMHO it's a bit overkill here.

tests/api/project/schemas/test_updatable_model.py
sfc-gh-bdufour, line 1: With what's here are we confident that we can accept a version bump for Pydantic with very low risk of breaking anything?

tests/testing_utils/fixtures.py
```diff
@@ -429,6 +430,11 @@ def native_app_extension_function(
     return NativeAppExtensionFunction(**native_app_extension_function_raw_data)
 
 
+@pytest.fixture
+def current_sanitized_username():
+    return get_sanitized_username()
```
sfc-gh-bdufour, line 435: why? Can't we just call the function when we need the value?

tests/project/test_util.py
sfc-gh-bdufour, line 1: you added more public methods than are tested here, ensure 100% coverage

tests/testing_utils/fixtures.py
sfc-gh-melnacouzi: remove this whole thing. Having username (template default) accessed at class build time causes issues with integration tests as well. I've updated the PR to add fn.get_username() so this evaluated during templating. We don't need to access this in testing anymore.

src/snowflake/cli/api/project/util.py
sfc-gh-melnacouzi: does variadic version have any advantages? I thought this looks cleaner for use internally.

src/snowflake/cli/api/utils/templating_functions.py
sfc-gh-melnacouzi: From one of your previous comments, I got the idea that we want to avoid defining these in 2 places. 
- In this method, you only need to add the new function to this class.
- Alternatively, we can define the functions separately and manually in a map:
```
def id_to_str(arg):
  ....


- Alternatively, we can do the following:
templating_functions = {
  'id_to_str': id_to_str
}
```

But then there is a bit of boilerplate as we would have to define/reference the function in 2 places. Not a big deal, but let me know what you think.

src/snowflake/cli/api/utils/templating_functions.py
sfc-gh-melnacouzi: COPY OF ANOTHER COMMENT:

From one of your previous comments, I got the idea that we want to avoid defining these in 2 places.

In this method, you only need to add the new function to this class.
Alternatively, we can define the functions separately and manually in a map:
```
def id_to_str(arg):
  ....


- Alternatively, we can do the following:
templating_functions = {
  'id_to_str': id_to_str
}
```
But then there is a bit of boilerplate as we would have to define/reference the function in 2 places. Not a big deal, but let me know what you think.

src/snowflake/cli/api/project/schemas/updatable_model.py
sfc-gh-melnacouzi: updated.

src/snowflake/cli/api/project/util.py
sfc-gh-melnacouzi: good point. Updated.

tests/api/project/schemas/test_updatable_model.py
sfc-gh-melnacouzi: I wouldn't say there is no risk (as with every version bump), but all the logic I used is documented within Pydantic, so the risk of changing is low:
- Passing global context feature is documented on their website.
- Validator orders is documented on their website.
- model_fields is part of their documentation.
Based on that, I believe the risk is low.

src/snowflake/cli/api/project/util.py
sfc-gh-bdufour: subjective, let's leave as is

src/snowflake/cli/api/utils/templating_functions.py
sfc-gh-bdufour: Subjective. Will get behind this, although I would have probably gone with your alternative approach if I had done this work myself.

sfc-gh-bdufour (approved):

sfc-gh-turbaszek (approved):

PR #1332, opened 2024-07-17
Support project definition v2 in "snow app" commands: bundle, validate, deploy

sfc-gh-gbloom:
### Pre-review checklist

### Changes description

* Added the `@nativeapp_definition_v2_to_v1` decorator to `snow app` commands: `bundle`, `validate` and `deploy`
* Updated the `_pdf_v2_to_v1()` logic to support `artifacts`, `bundle_root`, `generated_root`, `deploy_root`

### Test

Parameterized the fixture of the existing tests to run on both v1 and v2 projects.
Created the `override_snowflake_yml_artifacts()` utility for the tests to modify `snowflake.yml` for both v1 and v2.

### DIFF
--- a/src/snowflake/cli/plugins/nativeapp/v2_conversions/v2_to_v1_decorator.py
+++ b/src/snowflake/cli/plugins/nativeapp/v2_conversions/v2_to_v1_decorator.py
@@ -15,7 +15,8 @@
 from __future__ import annotations
 
 from functools import wraps
-from typing import Any, Dict, Optional
+from pathlib import Path
+from typing import Any, Dict, Optional, Union
 
 from click import ClickException
 from snowflake.cli.api.cli_global_context import cli_context, cli_context_manager
@@ -25,10 +26,23 @@ from snowflake.cli.api.project.schemas.entities.application_entity import (
 from snowflake.cli.api.project.schemas.entities.application_package_entity import (
     ApplicationPackageEntity,
 )
+from snowflake.cli.api.project.schemas.native_app.path_mapping import PathMapping
 from snowflake.cli.api.project.schemas.project_definition import (
     DefinitionV11,
     DefinitionV20,
 )
+
+
+def _convert_v2_artifact_to_v1_dict(
+    v2_artifact: Union[PathMapping, Path]
+) -> Union[Dict, str]:
+    if type(v2_artifact) == PathMapping:
+        return {
+            "src": v2_artifact.src,
+            "dest": v2_artifact.dest,
+            "processors": v2_artifact.processors,
+        }
+    return str(v2_artifact)
 
 
 def _pdf_v2_to_v1(v2_definition: DefinitionV20) -> DefinitionV11:
@@ -57,8 +71,13 @@ def _pdf_v2_to_v1(v2_definition: DefinitionV20) -> DefinitionV11:
 
     # NativeApp
     pdfv1["native_app"]["name"] = "Auto converted NativeApp project from V2"
-    pdfv1["native_app"]["artifacts"] = app_package_definition.artifacts
+    pdfv1["native_app"]["artifacts"] = [
+        _convert_v2_artifact_to_v1_dict(a) for a in app_package_definition.artifacts
+    ]
     pdfv1["native_app"]["source_stage"] = app_package_definition.stage
+    pdfv1["native_app"]["bundle_root"] = str(app_package_definition.bundle_root)
+    pdfv1["native_app"]["generated_root"] = str(app_package_definition.generated_root)
+    pdfv1["native_app"]["deploy_root"] = str(app_package_definition.deploy_root)
 
     # Package
     pdfv1["native_app"]["package"] = {}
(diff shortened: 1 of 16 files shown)


src/snowflake/cli/plugins/nativeapp/v2_conversions/v2_to_v1_decorator.py
```diff
@@ -25,12 +26,25 @@
...
 from snowflake.cli.api.project.schemas.project_definition import (
     DefinitionV11,
     DefinitionV20,
 )
 
 
+def _convert_v2_artifact_to_v1_dict(
+    v2_artifact: Union[PathMapping, Path]
+) -> Union[Dict, str]:
+    if type(v2_artifact) == PathMapping:
```
sfc-gh-fcampbell, line 39: the idiomatic python way is to use `isinstance` (which handles subclasses)
```suggestion
    if isinstance(v2_artifact, PathMapping):
```

tests_integration/nativeapp/test_bundle.py
```diff
@@ -29,56 +29,76 @@
...
+        assert Path(deploy_root, "manifest.yml").is_file()
+        assert Path(deploy_root, "setup_script.sql").is_file()
+        assert Path(deploy_root, "README.md").is_file()
 
-    project_root = Path(temporary_working_directory, project_name)
-    deploy_root = Path(project_root, "output", "deploy")
-    assert Path(deploy_root, "manifest.yml").is_file()
-    assert Path(deploy_root, "setup_script.sql").is_file()
-    assert Path(deploy_root, "README.md").is_file()
+        return project_root, runner, definition_version
```
sfc-gh-fcampbell, line 51: if we change this to a generator, the effect of `@enable_definition_v2_feature_flag` should persist during the execution of the test using this fixture, so we can remove `@enable_definition_v2_feature_flag` from each test function
```suggestion
        yield project_root, runner, definition_version
```

tests_integration/nativeapp/test_bundle.py
```diff
@@ -29,56 +29,76 @@
...
+                  artifacts:
+                {indent(artifacts_section, "                  ")}
+                  deploy_root: {deploy_root}
+                """
+        f.write(dedent(file_content))
 
 
 # Tests that we copy files/directories directly to the deploy root instead of creating symlinks.
 @pytest.mark.integration
+@enable_definition_v2_feature_flag
```
sfc-gh-fcampbell, line 84: if my above comment is correct, this can be removed (and in other places in this file)

tests_integration/nativeapp/test_bundle.py
```diff
@@ -29,56 +29,76 @@
...
+    with open("snowflake.yml", "w") as f:
+        if definition_version == "v2":
+            file_content = f"""
+                definition_version: 2
+                entities:
+                  pkg:
+                    type: application package
+                    name: myapp_pkg_<% ctx.env.USER %>
+                    artifacts:
+                {indent(artifacts_section, "                    ")}
```
sfc-gh-fcampbell, line 66:
```suggestion
                {indent(artifacts_section, " " * 20)}
```

sfc-gh-fcampbell: this seems really brittle though, how about using `yaml.dump` to dump a `dict`(and pass `artifacts_section` as a `list`) instead of dealing with raw strings?

tests_integration/nativeapp/test_bundle.py
```diff
@@ -29,56 +29,76 @@
...
+                    manifest: app/manifest.yml
+                    deploy_root: {deploy_root}
+                """
+        else:
+            file_content = f"""
+                definition_version: 1
+                native_app:
+                  name: myapp
+                  artifacts:
+                {indent(artifacts_section, "                  ")}
```
sfc-gh-fcampbell, line 76:
```suggestion
                {indent(artifacts_section, " " * 18)}
```

tests_integration/nativeapp/test_bundle.py
sfc-gh-gbloom: It works, but since this fixture returns a generator now, I have to use it like this:
```diff
- project_root, runner, definition_version = template_setup
+ project_root, runner, definition_version = next(template_setup)
```
Can you think of a python magic I can apply to the fixture so that the fixture consumption will stay the same?

sfc-gh-fcampbell: that shouldn't happen, pytest handles this automatically. what if instead of using `@enable_definition_v2_feature_flag`, you do `with enable_definition_v2_feature_flag:` in the body of the function?

sfc-gh-fcampbell: or make `enable_definition_v2_feature_flag` itself a fixture that yields a mock, like
```python
@pytest.fixture
def enable_definition_v2_feature_flag():
    yield mock.patch.dict(
        os.environ,
        {
            "SNOWFLAKE_CLI_FEATURES_ENABLE_PROJECT_DEFINITION_V2": "true",
        },
    )
```
then you can use it as
```
@pytest.fixture(scope="function", params=["v1", "v2"])
def template_setup(runner, project_directory, request, enable_definition_v2_feature_flag):
```

sfc-gh-gbloom: Awesome, went with the `with enable_definition_v2_feature_flag` approach. Thanks!

sfc-gh-fcampbell: the `return` should stay a `yield` though, and `@enable_definition_v2_feature_flag` can be removed everywhere this fixture is used

sfc-gh-gbloom: Yep, pushed my changes now

sfc-gh-fcampbell (approved):

PR #1336, opened 2024-07-18
Added connectors log_configuration module to reload

sfc-gh-astus:
### Pre-review checklist

### Changes description
Before each test we reload few modules but we missed one module. We had 2 different CONFIG_MANAGER objects, when we were sending the telemetry, we were using "old" CONFIG_MANAGER with default config.toml.

### DIFF
--- a/src/snowflake/cli/api/cli_global_context.py
+++ b/src/snowflake/cli/api/cli_global_context.py
@@ -15,6 +15,7 @@
 from __future__ import annotations
 
 import re
+import warnings
 from pathlib import Path
 from typing import Callable, Optional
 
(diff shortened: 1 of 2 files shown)


src/snowflake/cli/api/cli_global_context.py
```diff
@@ -215,6 +217,13 @@ def _collect_not_empty_connection_attributes(self):
     def _build_connection(self):
         from snowflake.cli.app.snow_connector import connect_to_snowflake
 
+        if IS_WINDOWS:
+            warnings.filterwarnings(
+                action="ignore",
+                message="Bad owner or permissions.*",
+                module="snowflake.connector.config_manager",
+            )
```
sfc-gh-turbaszek, line 225: Please add comment why we do so

sfc-gh-astus: Added

sfc-gh-turbaszek (approved):

PR #1340, opened 2024-07-18
Post_deploy script should run relative to project root, Fixes #1325.

### LINKED ISSUE #1325: SNOW-1540572: When running an app from a given --project path, the post_deploy scripts must be found relative to that path
### SnowCLI version

2.6.1

### Python version

3.11.18

### Platform

macOS-14.2.1-arm64-arm-64bit

### What happened

I am running a `snowflake.yaml` file to start an app, with the `--project` option:
```
> snow app run --project my_path
```

That file specifies a script to run as `post_deploy` but the script is not found, unless I give a full path, starting from the point where I'm invoking the `snow` command.

### Console output

```Shell
The error is 

Executing application post-deploy actions
  Executing SQL script: scripts/post_deploy.sql

[Errno 2] No such file or directory: 'scripts/post_deploy.sql'
```
```


### How to reproduce

This can be reproduced by takin [this test example](https://github.com/snowflakedb/snowflake-cli/tree/8f638f4938ddb2957c011effdbeef173482051c4/tests_integration/test_data/projects/napp_application_post_deploy) and nesting it into a `root` directory.

Then from the `root` directory run:
```
snow app run --project napp_application_post_deploy
```

### PR DESCRIPTION
**Pull request by sfc-gh-melnacouzi**
### Pre-review checklist
   * [] I've described my changes in the release notes.

### Changes description
Post_deploy script should run relative to project root, Fixes #1325

### CODE REVIEW

File: tests_integration/nativeapp/test_init_run.py [hunk 1]
```diff
@@ -492,6 +492,59 @@ def run():
             assert result.exit_code == 0
 
 
+# Tests that application post-deploy scripts are executed even when they are used with --project and project is in another directory
```
sfc-gh-cgorrie, line 495: Is it possible to parametrize the previous test instead using a similar strategy to `is_versioned`?

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -141,13 +145,18 @@ def _execute_sql_script(self, sql_script_path):
         This assumes that a relevant warehouse is already active.
         Consequently, "use database" will be executed first if it is set in definition file or in the current connection.
         """
-        with open(sql_script_path) as f:
-            sql_script = f.read()
+        env = get_sql_cli_jinja_env(
```
sfc-gh-bdufour, line 148: We exploded snowflake_sql_jinja_render, any reason we couldn't modify the existing abstraction instead?

sfc-gh-bdufour: Also, we're diverging between how package scripts are executed and how post-deploy scripts are executed. For package scripts, the env is created differently, and AFAICT if a script fails to expand we guarantee that no script is executed. This doesn't appear to be the case here.

sfc-gh-melnacouzi: I have refactored to reuse some of the existing package script logic, and also make sure that we check for template errors before executing any of the scripts.

tests_integration/nativeapp/test_init_run.py
sfc-gh-melnacouzi: done.

sfc-gh-astus (dismissed):

src/snowflake/cli/plugins/nativeapp/exceptions.py
```diff
@@ -54,18 +54,18 @@ def __init__(self, item: str, expected_owner: str, actual_owner: str):
...
 
-class InvalidPackageScriptError(ClickException):
-    """A referenced package script had syntax error(s)."""
+class InvalidScriptError(ClickException):
+    """A referenced script had syntax error(s)."""
 
     def __init__(self, relpath: str, err: jinja2.TemplateError):
-        super().__init__(f'Package script "{relpath}" is not a valid jinja2 template')
+        super().__init__(f'Script "{relpath}" does not contain a valid template')
         self.err = err
```
sfc-gh-cgorrie, line 69: @sfc-gh-melnacouzi  Despite making the `err` available inside the error, will the user actually see what went wrong? Or will they just get "script does not contain a valid template" and no additional information about the error?

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -561,6 +561,27 @@ def create_app_package(self) -> None:
                 )
             )
 
+    def _expand_script_templates(
+        self, env: jinja2.Environment, jinja_context, scripts: List[str]
+    ):
```
sfc-gh-bdufour, line 566: document the return type of this

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -561,6 +561,27 @@ def create_app_package(self) -> None:
...
+        queued_queries = []
```
sfc-gh-bdufour, line 567: what's a query in this context? The same doesn't sound very meaningful here. Isn't it just script content? I know the name was there before, but it's a good opportunity to clarify.

src/snowflake/cli/plugins/nativeapp/run_processor.py
```diff
@@ -135,36 +139,48 @@ class NativeAppRunProcessor(NativeAppManager, NativeAppCommandProcessor):
...
         """
-        with open(sql_script_path) as f:
-            sql_script = f.read()
-
         try:
-            if self._conn.database:
-                self._execute_query(f"use database {self._conn.database}")
-            sql_script = snowflake_sql_jinja_render(content=sql_script)
-            self._execute_queries(sql_script)
+            if database_name:
```
sfc-gh-bdufour, line 151: nit: `is not None` is the typical way to check this, since we shouldn't make a habit of interpreting what it means for a string to be truthy.

src/snowflake/cli/plugins/nativeapp/exceptions.py
sfc-gh-melnacouzi: Good point. I have updated the message to be more informative.

sfc-gh-bdufour: I believe it's the latter, but we should indeed confirm. Why not fix this here then? It would be informative for a user if we gave them the Jinja error, and while we're fixing bugs, might as well fix all known bugs in the current logic...

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-melnacouzi: updated.

sfc-gh-cgorrie (approved):
Non-blocking, but you could also add snapshot tests for different template errors as what we show users when something goes wrong is somewhat of a contract.

sfc-gh-astus (approved):

PR #1348, opened 2024-07-22
Removed supporting git copy to for user stages

sfc-gh-astus:
### Pre-review checklist

### Changes description
Snowflake doesn't support copy to user stages.

### DIFF
--- a/src/snowflake/cli/plugins/git/manager.py
+++ b/src/snowflake/cli/plugins/git/manager.py
@@ -18,7 +18,12 @@ from pathlib import Path
 from textwrap import dedent
 from typing import List
 
-from snowflake.cli.plugins.stage.manager import StageManager, StagePathParts
+from snowflake.cli.plugins.stage.manager import (
+    USER_STAGE_PREFIX,
+    StageManager,
+    StagePathParts,
+    UserStagePathParts,
+)
 from snowflake.connector.cursor import SnowflakeCursor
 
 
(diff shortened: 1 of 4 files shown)


src/snowflake/cli/plugins/git/manager.py
```diff
@@ -74,6 +75,13 @@ def create(
             query += f"git_credentials = {secret}\n"
         return self._execute_query(query)
 
+    def copy_files(self, source_path: str, destination_path: str) -> SnowflakeCursor:
+        if destination_path.startswith("@~"):
```
sfc-gh-turbaszek, line 79: Let's implement this on `StagePathParts` as `is_user_stage` or somewhere. WDYT?

sfc-gh-jsikorski (approved):

src/snowflake/cli/plugins/git/manager.py
sfc-gh-astus: There is separated UserStagePathParts, so I used it.

src/snowflake/cli/plugins/stage/manager.py
```diff
@@ -83,6 +84,16 @@ def get_full_stage_path(self, path: str):
...
 
+    def get_standard_stage_path(self) -> str:
+        path = self.path
+        return f"@{path}{'/'if self.is_directory and not path.endswith('/') else ''}"
+
+    def get_standard_stage_directory_path(self) -> str:
+        path = self.get_standard_stage_path()
+        if not path.endswith("/"):
+            return path + "/"
+        return path
```
sfc-gh-astus, line 95: I don't know if it's overcomplicated, once we need what user passed, other time we need with / at the end.

sfc-gh-turbaszek (approved):

PR #1349, opened 2024-07-22
Allow commands to return multiple results as a stream

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
As part of implementing `snow app events -f`, we want to be able to stream messages indefinitely. This PR allows commands return a `StreamResult`, which wraps a generator of `CommandResult` objects. When we print a `StreamResult`, we iterate over all of its values and pass them individually through the same printing process.

An example use-case (some code omitted):
```python
@app.command("events", requires_connection=True)
def app_events(**options):
    """Fetches events for this app from the event table configured in Snowflake."""
    manager = NativeAppManager(...)

    def g():
        for event in manager.get_events():
            yield ObjectResult(event)

    return StreamResult(g())
```
This generator-based implementation will allow us to make this command poll for events periodically, yielding new result objects for new events to print.

### DIFF
--- a/src/snowflake/cli/api/output/types.py
+++ b/src/snowflake/cli/api/output/types.py
@@ -57,6 +57,15 @@ class MultipleResults(CommandResult):
         return (element for element in self._elements)
 
 
+class StreamResult(CommandResult):
+    def __init__(self, generator: t.Generator[CommandResult, None, None]):
+        self._generator = generator
+
+    @property
+    def result(self):
+        return self._generator
+
+
 class QueryResult(CollectionResult):
     def __init__(self, cursor: SnowflakeCursor | DictCursor):
         self.column_names = [col.name for col in cursor.description]
--- a/src/snowflake/cli/app/printing.py
+++ b/src/snowflake/cli/app/printing.py
@@ -34,6 +34,7 @@ from snowflake.cli.api.output.types import (
     MessageResult,
     MultipleResults,
     ObjectResult,
+    StreamResult,
 )
 from snowflake.cli.api.sanitizers import sanitize_for_terminal
 
@@ -100,6 +101,12 @@ def print_structured(result: CommandResult):
     """Handles outputs like json, yml and other structured and parsable formats."""
     if isinstance(result, MultipleResults):
         _stream_json(result)
+    elif isinstance(result, StreamResult):
+        # A StreamResult prints each value onto its own line
+        # instead of joining all the values into a JSON array
+        for r in result.result:
+            json.dump(r, sys.stdout, cls=CustomJSONEncoder)
+            print()
     else:
         json.dump(result, sys.stdout, cls=CustomJSONEncoder, indent=4)
     # Adds empty line at the end
@@ -159,7 +166,7 @@ def print_result(cmd_result: CommandResult, output_format: OutputFormat | None =
     output_format = output_format or _get_format_type()
     if is_structured_format(output_format):
         print_structured(cmd_result)
-    elif isinstance(cmd_result, MultipleResults):
+    elif isinstance(cmd_result, (MultipleResults, StreamResult)):
         for res in cmd_result.result:
             print_result(res)
     elif (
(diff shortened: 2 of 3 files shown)


RELEASE-NOTES.md
```diff
@@ -25,6 +25,7 @@
 * Support for Python remote execution via `snow stage execute` and `snow git execute` similar to existing EXECUTE IMMEDIATE support.
 * Added support for autocomplete in `--connection` flag.
 * Added `snow init` command, which supports initializing projects with external templates.
+* Added support for commands that yield multiple results
```
sfc-gh-bdufour, line 28: Not sure if this is required, given it's an implementation detail and not a user feature.

sfc-gh-fcampbell: yeah i can take it out

sfc-gh-bdufour (dismissed):

sfc-gh-astus (approved):

src/snowflake/cli/api/commands/snow_typer.py
```diff
@@ -136,6 +136,10 @@ def process_result(result):
         # We should improve it in future.
         if not result:
             return
+        if isinstance(result, Generator):
+            for r in result:
+                SnowTyper.process_result(r)
+            return
```
sfc-gh-turbaszek, line 142: How different is this from returning `CollectionResult(generator)`?

sfc-gh-fcampbell: A CollectionResult is printed all at once via `list(o.result)`, which won't work for `tail -f` style streaming output. What I want is for lines to be printed as soon as they're generated.

I'm open to editing the result handling code to iterate over the elements of a CollectionResult one at a time instead, but I feel like this approach is simpler since it just requires commands to change `return` to `yield`.

sfc-gh-turbaszek: Hm, would it make sense to make `StreamingResult`? This is a solution we were thinking about for spcs logs streaming. 

The reason I'm concern is inconsistency of commands implementation (return or yield). Hopefully adding `StreamingResult` won't be too complex. WDYT?

sfc-gh-fcampbell: discussed on Slack, updated to return a `StreamResult`

src/snowflake/cli/app/printing.py
```diff
@@ -98,12 +99,21 @@ def is_structured_format(output_format):
...
+    printed_end_line = False
     if isinstance(result, MultipleResults):
         _stream_json(result)
+    elif isinstance(result, StreamResult):
+        # A StreamResult prints each value onto its own line
+        # instead of joining all the values into a JSON array
+        for r in result.result:
+            json.dump(r, sys.stdout, cls=CustomJSONEncoder)
+            print(flush=True)
+            printed_end_line = True
```
sfc-gh-turbaszek, line 111: Won't `_stream_json` work?

sfc-gh-fcampbell: No, it wraps it in a json array, while I want one object per line

sfc-gh-fcampbell: No, it wraps it in a json array, while I want one object per line

sfc-gh-turbaszek (approved):

src/snowflake/cli/api/output/types.py
```diff
@@ -57,6 +57,15 @@ def result(self):
         return (element for element in self._elements)
 
 
+class StreamResult(CommandResult):
+    def __init__(self, generator: t.Generator[CommandResult, None, None]):
+        self._generator = generator
+
+    @property
+    def result(self):
+        return self._generator
```
sfc-gh-turbaszek, line 66: Thanks a lot for doing this! 🚀

src/snowflake/cli/app/printing.py
sfc-gh-fcampbell: piping into JQ does work with multiple objects:
image

and it's possible to collect the output back into an array with `jq -s`:
image

sfc-gh-fcampbell: TBH I think `_stream_json` could be replaced with just
```python
json.dump(list(result.result), sys.stdout, cls=CustomJSONEncoder, indent=4)
```

PR #1352, opened 2024-07-22
SNOW-1544013 Fetch events for app

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Fetches events from the account event table (See #1337). This does not implement filtering, timeboxing, or limiting output (these will come in a followup).

Regular output prints the messages as a log stream, including any ASCII control codes:
image

JSON output prints each message as individual JSON values:
image

This can then be piped to `jq` or equivalent:
image

### DIFF
--- a/src/snowflake/cli/plugins/nativeapp/manager.py
+++ b/src/snowflake/cli/plugins/nativeapp/manager.py
@@ -315,10 +315,10 @@ class NativeAppManager(SqlExecutionMixin):
         )
 
     @cached_property
-    def account_event_table(self) -> str | None:
+    def account_event_table(self) -> str:
         query = "show parameters like 'event_table' in account"
         results = self._execute_query(query, cursor_class=DictCursor)
-        return next((r["value"] for r in results if r["key"] == "EVENT_TABLE"), None)
+        return next((r["value"] for r in results if r["key"] == "EVENT_TABLE"), "")
 
     def verify_project_distribution(
         self, expected_distribution: Optional[str] = None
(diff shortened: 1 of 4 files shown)


src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -89,7 +90,7 @@
 
 def generic_sql_error_handler(
     err: ProgrammingError, role: Optional[str] = None, warehouse: Optional[str] = None
-):
+) -> NoReturn:
```
sfc-gh-bdufour, line 93: nit: doesn't the rest of the code use -> None for this?

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -715,9 +716,25 @@ def get_validation_result(self, use_scratch_stage: bool):
...
     def get_events(self) -> list[dict]:
-        if self.account_event_table is None:
+        if not self.account_event_table:
             raise NoEventTableForAccount()
-        return []
+
+        app_name = (
+            unquote_identifier(self.app_name)
+            if is_valid_quoted_identifier(self.app_name)
+            else self.app_name.upper()
```
sfc-gh-bdufour, line 725: nit: unquote_identifier should already do pretty much all of this, including case normalization

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-fcampbell: `NoReturn` means it never returns (i.e. always raises an exception). This helps the type checker know that any code after a call to this function is unreachable

tests_integration/nativeapp/test_events.py
```diff
@@ -27,7 +27,7 @@
 
 
 @pytest.mark.integration
-def test_app_events(runner, temporary_working_directory):
+def test_app_events_no_event_table(runner, temporary_working_directory):
```
sfc-gh-fcampbell, line 30: I could add a test that creates a new event table, raw inserts events into it, and reads them back, but I'm worried about it being slow and/or flaky. Thoughts?

sfc-gh-bdufour (approved):

PR #1354, opened 2024-07-22
SNOW-1544013 Add interval filtering to `snow app events`

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Adds `--since` and `--until` flags to restrict the time range for events. The query is updated to compare the `timestamp` column against the current time minus an `interval` specified by the user: https://docs.snowflake.com/en/sql-reference/data-types-datetime#interval-constants

### DIFF
--- a/src/snowflake/cli/plugins/nativeapp/commands.py
+++ b/src/snowflake/cli/plugins/nativeapp/commands.py
@@ -383,8 +383,14 @@ def app_validate(**options):
 @with_project_definition()
 @nativeapp_definition_v2_to_v1
 def app_events(
-    since: str = typer.Option(default=""),
-    until: str = typer.Option(default=""),
+    since: str = typer.Option(
+        default="",
+        help="Fetch events that are newer than this time ago, in Snowflake interval syntax.",
+    ),
+    until: str = typer.Option(
+        default="",
+        help="Fetch events that are older than this time ago, in Snowflake interval syntax.",
+    ),
     **options,
 ):
     """Fetches events for this app from the event table configured in Snowflake."""
--- a/src/snowflake/cli/plugins/nativeapp/commands.py
+++ b/src/snowflake/cli/plugins/nativeapp/commands.py
@@ -382,7 +382,11 @@ def app_validate(**options):
 @app.command("events", hidden=True, requires_connection=True)
 @with_project_definition()
 @nativeapp_definition_v2_to_v1
-def app_events(**options):
+def app_events(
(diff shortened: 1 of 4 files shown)


src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -714,17 +714,31 @@ def get_validation_result(self, use_scratch_stage: bool):
...
+        since_clause = (
+            f"and timestamp >= sysdate() - interval '{since_interval}'"
+            if since_interval
+            else ""
+        )
+        until_clause = (
+            f"and timestamp <= sysdate() - interval '{until_interval}'"
+            if until_interval
+            else ""
+        )
```
sfc-gh-fcampbell, line 734: this could also support absolute timestamps in the future, I just wanted a quick and easy way to continue developing and testing but avoid fetching days worth of events by just doing `snow app events --since '1 hour'`

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -382,15 +382,25 @@ def app_validate(**options):
 @app.command("events", hidden=True, requires_connection=True)
 @with_project_definition()
 @nativeapp_definition_v2_to_v1
-def app_events(**options):
+def app_events(
+    since: str = typer.Option(
+        default="",
```
sfc-gh-bdufour, line 387: nit: what happens when we default to ""? Is it from the oldest events in the table? That could be a lot of data, would it make sense to add some kind of numerical limit on the number of events being returned, and default it to unlimited? So `events --since "1 year" --limit 500` would prevent the user from getting too many events displayed before they can kill the command.

sfc-gh-bdufour (approved):

src/snowflake/cli/plugins/nativeapp/commands.py
sfc-gh-fcampbell: yup, `--limit` is coming in https://github.com/snowflakedb/snowflake-cli/pull/1370

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -382,15 +382,25 @@ def app_validate(**options):
...
+        help="Fetch events that are newer than this time ago, in Snowflake interval syntax.",
```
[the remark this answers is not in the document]
sfc-gh-fcampbell: yes it's included, since we use `<=` and `>=`, I just didn't want this help text to get too wordy

PR #1356, opened 2024-07-23
CREATE OR REPLACE for `runtime` and `execute_as_caller`

### LINKED ISSUE #1333: SNOW-1542852: CREATE OR REPLACE when the `runtime` or `execute_as_caller` project file options are changed for a UDF or Stored Procedure during `snow snowpark deploy`
### Description

I would like to force a `CREATE or REPLACE ... ` to be run when the `runtime` or `execute_as_caller` values are modified within a project definition file for a stored procedure or a UDF. I originally thought that's what the `--replace` flag did, but it seems like it tries to be smart and only replace the function if it detects a signature change.

### Context

Currently, when doing a `snow snowpark deploy --replace` command from within a project, if the UDF or Stored Procedure currently exists and the argument signature doesn't change, then the `CREATE or REPLACE ...` command is not run. So if I want to change the runtime or execute_as_caller values, I have to drop the function or stored procedure first and then deploy. For convenience, it would be nice if the CLI did this for me by just running the `CREATE or REPLACE ...`

Here is a sample project file
```yaml
definition_version: 1
snowpark:
  project_name: "my_project"
  stage_name: 'deployment'
  src: "app/"

  procedures:
    - name: my_cool_sproc
      handler: "app_code.handler"
      runtime: "3.10" # I want to be able to change this to 3.11 after already deployed
      signature:
        - name: "arg"
          type: "string"
      returns: variant
      execute_as_caller: True # I want to be able to change this to False after already deployed
```

sfc-gh-jsikorski:
### Pre-review checklist

### Changes description
Fixes #1333 - adds more checks to trigger function/procedure replace when changing `runtime` and `execute_as_caller`


src/snowflake/cli/plugins/snowpark/common.py
```diff
@@ -82,6 +84,22 @@ def check_if_replace_is_required(
...
+        log.info("Runtime versions do not match. Replacing the %s", object_type)
+        return True
+
+    if execute_as_caller is not None and (
+        resource_json.get("execute as", "OWNER")
+        != ("CALLER" if execute_as_caller else "OWNER")
+    ):
+        log.info(
+            "Execute as caller settings do not match. Replacing the %s", object_type
+        )
```
sfc-gh-mraba, line 102: That'a a lot of `if's`.
I suggest we add this for cleanup list in future.

sfc-gh-mraba (approved):

src/snowflake/cli/plugins/snowpark/common.py
sfc-gh-jsikorski: Yeah, i wrote a design doc for it before my leave - but the implementation is still WiP

PR #1370, opened 2024-07-25
SNOW-1544013 Add `--first` and `--last` to `snow app events`

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Adds `--first` and `--last` parameters to `snow app events`. The two parameters are mutually exclusive to avoid generating a query that would return discontinuous results.

This required updating the query to sort in descending order to limit to the last N events, then re-sort in ascending order for printing and to limit to the first M events.

### DIFF
--- a/src/snowflake/cli/plugins/nativeapp/commands.py
+++ b/src/snowflake/cli/plugins/nativeapp/commands.py
@@ -391,19 +391,30 @@ def app_events(
         default="",
         help="Fetch events that are older than this time ago, in Snowflake interval syntax.",
     ),
-    limit: int = typer.Option(
-        default=0, help="Maximum number of latest events to fetch."
+    first: int = typer.Option(
+        default=0, help="Fetch only the first N events. Cannot be used with --last."
+    ),
+    last: int = typer.Option(
+        default=0, help="Fetch only the last N events. Cannot be used with --first."
     ),
     **options,
 ):
     """Fetches events for this app from the event table configured in Snowflake."""
+    if first and last:
+        raise ClickException("--first and --last cannot be used together.")
+
     assert_project_type("native_app")
 
     manager = NativeAppManager(
         project_definition=get_cli_context().project_definition.native_app,
         project_root=get_cli_context().project_root,
     )
-    events = manager.get_events(since, until, limit)
+    events = manager.get_events(
+        since_interval=since,
+        until_interval=until,
+        first=first,
+        last=last,
+    )
     if not events:
         return MessageResult("No events found.")
 
--- a/src/snowflake/cli/plugins/nativeapp/commands.py
+++ b/src/snowflake/cli/plugins/nativeapp/commands.py
@@ -391,6 +391,9 @@ def app_events(
         default="",
         help="Fetch events that are older than this time ago, in Snowflake interval syntax.",
     ),
+    limit: int = typer.Option(
+        default=0, help="Maximum number of latest events to fetch."
+    ),
     **options,
 ):
     """Fetches events for this app from the event table configured in Snowflake."""
@@ -400,7 +403,7 @@ def app_events(
         project_definition=get_cli_context().project_definition.native_app,
         project_root=get_cli_context().project_root,
     )
-    events = manager.get_events(since, until)
+    events = manager.get_events(since, until, limit)
     if not events:
         return MessageResult("No events found.")
 
--- a/src/snowflake/cli/plugins/nativeapp/manager.py
+++ b/src/snowflake/cli/plugins/nativeapp/manager.py
@@ -715,7 +715,10 @@ class NativeAppManager(SqlExecutionMixin):
                     )
 
     def get_events(
-        self, since_interval: str = "", until_interval: str = ""
+        self,
+        since_interval: str = "",
+        until_interval: str = "",
+        limit: int = 0,
     ) -> list[dict]:
         if not self.account_event_table:
             raise NoEventTableForAccount()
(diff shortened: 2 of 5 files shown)


tests/nativeapp/test_manager.py
```diff
@@ -1414,62 +1465,24 @@ def test_get_events_quoted_app_name(
...
-
-@mock.patch(
-    NATIVEAPP_MANAGER_ACCOUNT_EVENT_TABLE,
-    return_value="db.schema.event_table",
-    new_callable=mock.PropertyMock,
-)
-@mock.patch(NATIVEAPP_MANAGER_EXECUTE)
-def test_get_events_interval_filtering(
-    mock_execute, mock_account_event_table, temp_dir, mock_cursor
-):
```
sfc-gh-fcampbell, line 1453: consolidated into `test_get_events`

tests/nativeapp/test_manager.py
```diff
@@ -1353,32 +1353,85 @@ def test_account_event_table_not_set_up(mock_execute, temp_dir, mock_cursor):
...
+    limit,
+    expected_limit_clause,
+):
     create_named_file(
         file_name="snowflake.yml",
         dir_name=temp_dir,
         contents=[mock_snowflake_yml_file],
     )
 
+    events = [dict(TIMESTAMP="2020-01-01T00:00:00Z", VALUE="test")] * 100
```
sfc-gh-fcampbell, line 1417: the actual data doesn't matter since we don't run the SQL for real, but I still wanted a meaningful amount of rows in the response

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -391,6 +391,9 @@ def app_events(
         default="",
         help="Fetch events that are older than this time ago, in Snowflake interval syntax.",
     ),
+    limit: int = typer.Option(
+        default=0, help="Maximum number of latest events to fetch."
```
sfc-gh-bdufour, line 395: not sure about "latest events" here. It's just a standard limit on an ordered query, so if 1000 events match and you specify limit 10, then you'll get the 10 oldest events, no?

sfc-gh-fcampbell: the query was changed to sort descending, take the limit, then sort ascending

sfc-gh-bdufour: Am I the only one who finds that behaviour counter-intuitive? I expect truncation to happen in the natural order the results are presented in. Basically `tail | head -n` behaviour.

sfc-gh-cgorrie: I think limit in ascending order of timestamp makes sense, because if you're looking for a recent event you're likely to use the `--since` arg instead of `--limit`. Though it would be nice to have a way to get the last n events, so I'm not 100% convinced by my own argument.

sfc-gh-fcampbell: my logic is that it doesn't really make sense to get the first N events, especially in the case of a bare `snow app events --limit 100` for example. In this case, I just want to grab the latest events for my app, but not so many that they're so far in the past and no longer relevant.

I can change to `--first <N>` and `--last <N>` if that would make more sense

sfc-gh-bdufour: I do find it surprising that `snow app events --since "2 hours" --limit 100` will not show the oldest events, personally. PL?

sfc-gh-fcampbell: Updated with `--first` and `--last` flags after offline discussion

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -732,14 +739,21 @@ def get_events(
...
+            select * from (
+                select timestamp, value::varchar value
+                from {self.account_event_table}
+                where resource_attributes:"snow.database.name" = '{app_name}'
+                {since_clause}
+                {until_clause}
+                order by timestamp desc
+                {last_clause}
+            ) order by timestamp asc
+            {first_clause}
```
sfc-gh-bdufour, line 755: have you checked what happens to the plans when we do this? Do we need to worry about that cases where it's not strictly required?

sfc-gh-bdufour (approved):

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-fcampbell: I compared
```sql
explain
select *
from build_finetune_db.public.events
where resource_attributes:"snow.database.name" = 'PYTHON_CALLBACKS_FCAMPBELL'
order by timestamp asc
```
and
```sql
explain
select * from (
    select *
    from build_finetune_db.public.events
    where resource_attributes:"snow.database.name" = 'PYTHON_CALLBACKS_FCAMPBELL'
    order by timestamp desc
    -- limit 10
)
order by timestamp asc
```
and the plan is exactly the same:
```
GlobalStats:
    partitionsTotal=812
    partitionsAssigned=2
    bytesAssigned=19659776
Operations:
1:0     ->Result  EVENTS.TIMESTAMP, EVENTS.START_TIMESTAMP, EVENTS.OBSERVED_TIMESTAMP, EVENTS.TRACE, EVENTS.RESOURCE, EVENTS.RESOURCE_ATTRIBUTES, EVENTS.SCOPE, EVENTS.SCOPE_ATTRIBUTES, EVENTS.RECORD_TYPE, EVENTS.RECORD, EVENTS.RECORD_ATTRIBUTES, EVENTS.VALUE, EVENTS.EXEMPLARS  
1:1          ->Sort  EVENTS.TIMESTAMP ASC NULLS LAST  
1:2               ->Filter  (TO_CHAR(GET_PATH(EVENTS.RESOURCE_ATTRIBUTES, '[''snow.database.name'']'))) = 'PYTHON_CALLBACKS_FCAMPBELL'  
1:3                    ->TableScan  BUILD_FINETUNE_DB.PUBLIC.EVENTS  TIMESTAMP, START_TIMESTAMP, OBSERVED_TIMESTAMP, TRACE, RESOURCE, RESOURCE_ATTRIBUTES, SCOPE, SCOPE_ATTRIBUTES, RECORD_TYPE, RECORD, RECORD_ATTRIBUTES, VALUE, EXEMPLARS  {partitionsTotal=812, partitionsAssigned=2, bytesAssigned=19659776}
```

So when there's no `--last`, the `order by timestamp desc` is completely optimized out

sfc-gh-bdufour: Awesome, thanks for checking

PR #1372, opened 2024-07-26
Guess chunk size for multi-part md5sums to prevent unnecessary re-uploads

sfc-gh-cgorrie:
### Pre-review checklist
   * [x] I've updated the manual test document.

### Changes description
By default, Snowflake connectors (e.g. snowflake-connector-python, which we use) defer to automatic md5 calculation that is specific to each cloud storage provider. AWS substitutes its own multi-part chunked etag when files are uploaded in multiple parts, rather than calculating the simple (correct) md5sum for the entire file after all chunks have finished uploading.

This PR adds support for detecting this non-standard "md5sum" and attempts to calculate it locally when encountered by guessing what the chunk size would have been, as well as assuming all chunks (save for the last) are identically-sized. This guessing code is based on the current implementation in snowflake-connector-python, but should work for any files uploaded with drivers using chunk sizes that are exact multiples of a megabyte.

This PR also adds a hidden `snow app diff` command that simply exposes that part of the run/deploy process without uploading anything to the stage.

Working integration test run (marking as skip to save on 200mb upload):
https://github.com/snowflakedb/snowflake-cli/actions/runs/10167229878/job/28119152154?pr=1372


src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -170,6 +175,68 @@ def app_bundle(
...
 
 
+@app.command("hash", hidden=True)
+@with_project_definition()
+@nativeapp_definition_v2_to_v1
+def app_hash(
+    **options,
+) -> CommandResult:
+    """
+    Test command to see what local hashes look like.
```
sfc-gh-cgorrie, line 185: Will delete this before the PR goes out of draft.

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -170,6 +175,68 @@ def app_bundle(
...
+    for path in bundle.all_sources():
+        file_size = os.path.getsize(path)
+        file_hash = compute_md5sum(path, chunk_size if file_size >= threshold else None)
+        is_correct = file_matches_md5sum(path, file_hash)
+        results.append(dict(path=path, hash=file_hash, is_correct=is_correct))
+
+    return CollectionResult(results)
+
+
+@app.command("diff", requires_connection=True, hidden=True)
```
sfc-gh-cgorrie, line 213: Do we want to un-hide as part of this PR?

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -173,6 +178,33 @@ def app_bundle(
...
+    cli_context = get_cli_context()
+    manager = NativeAppManager(
+        project_definition=cli_context.project_definition.native_app,
+        project_root=cli_context.project_root,
+    )
+    bundle_map = manager.build_bundle()
+    diff: DiffResult = compute_stage_diff(
+        local_root=Path(manager.deploy_root), stage_fqn=manager.stage_fqn
+    )
+    if get_cli_context().output_format == OutputFormat.JSON:
```
sfc-gh-fcampbell, line 201: we already have the context in scope
```suggestion
    if cli_context.output_format == OutputFormat.JSON:
```

src/snowflake/cli/plugins/stage/md5.py
```diff
@@ -0,0 +1,150 @@
...
+    if is_md5sum(remote_md5):
+        # regular hash
+        return compute_md5sum(local_file) == remote_md5
+
+    if md5_and_chunks := parse_multipart_md5sum(remote_md5):
+        # multi-part hash (e.g. aws)
+        (_, num_chunks) = md5_and_chunks
+        file_size = os.path.getsize(local_file)
+
+        # chunk size may not be a clean multiple of EST_CHUNK_GRANULARITY_BYTES
```
sfc-gh-fcampbell, line 129: what's `EST_CHUNK_GRANULARITY_BYTES`? is it from the connector lib?

src/snowflake/cli/plugins/stage/md5.py
```diff
@@ -0,0 +1,150 @@
...
+        # try this first then fall back to the usual detection method
+        # at time of writing this would trigger around 80GiB for python connector
```
sfc-gh-fcampbell, line 131: what does "trigger around 80GiB" mean?

src/snowflake/cli/plugins/stage/md5.py
```diff
@@ -0,0 +1,150 @@
...
+        if num_chunks == S3_MAX_PARTS:
+            chunk_size = max(math.ceil(file_size / S3_MAX_PARTS), S3_MIN_PART_SIZE)
+            if compute_md5sum(local_file, chunk_size) == remote_md5:
+                return True
+
+        # try some common chunk size granularities
+        # smaller values create more even chunks; e.g. 13MB => [8MB, 6MB]
+        for granularity in [S3_CHUNK_SIZE, 1024**2]:
```
sfc-gh-fcampbell, line 139: `1024**2` is meant to represent 1 MB right? IMO the meaning is more obvious when using base 2
```suggestion
        for granularity in [S3_CHUNK_SIZE, 2**20]:
```

src/snowflake/cli/plugins/stage/md5.py
```diff
@@ -0,0 +1,150 @@
...
+            granules = 1 + ((file_size - 1) // (num_chunks * granularity))
+            chunk_size = granules * granularity
```
sfc-gh-fcampbell, line 141: i think i'm a bit confused by the math here. what's the difference between granularity and chunk size? why the `-1` on the file size? would you mind adding some code comments explaining what we're calculating here?

tests_integration/nativeapp/test_large_upload.py
```diff
@@ -0,0 +1,138 @@
...
+        if key.lower() == "user":
+            return USER_NAME
+        return original_getenv(key, default)
+
+    with mock.patch.dict(os.environ, {"USER": USER_NAME}):  # for DefinitionManager
+        with mock.patch("os.getenv", side_effect=mock_getenv):  # for ctx.env
+            yield
+
+
+@pytest.mark.skip  # FIXME: requires AWS + python connector to support threshold=<number>
```
sfc-gh-fcampbell, line 58:
```suggestion
@pytest.mark.skip(reason="Requires AWS + python connector to support threshold=<number>")
```
also didn't you mention a few days ago that the feature is supported in the connector but the test user/account has insufficient privileges?

sfc-gh-cgorrie: It's actually not supported in the connector, I thought it might be. There is an account override that overrides the connector properties we could use, but it's not necessarily the right solution. I will bring it up at Snowflake CLI office hours on Thursday.

src/snowflake/cli/plugins/stage/md5.py
sfc-gh-cgorrie: For what it's worth, I find it more confusing that way. I'll introduce a constant.

tests_integration/nativeapp/test_large_upload.py
```diff
@@ -55,7 +55,7 @@ def mock_getenv(key: str, default: str | None = None) -> str | None:
             yield
 
 
-@pytest.mark.skip  # FIXME: requires AWS + python connector to support threshold=<number>
+# @pytest.mark.skip  # FIXME: requires AWS + python connector to support threshold=<number>
```
sfc-gh-fcampbell, line 58: i don't think this meant to make it in

src/snowflake/cli/plugins/stage/md5.py
```diff
@@ -126,21 +127,29 @@ def file_matches_md5sum(local_file: Path, remote_md5: str | None) -> bool:
...
-            if md5 == remote_md5:
+        # Estimates the chunk size the multi-part file must have been uploaded with
+        # by trying chunk sizes that give the most evenly-sized chunks.
+        #
+        # First we'll try the chunk size that's a multiple of S3_CHUNK_SIZE (8mb) from
+        # the python connector that results in num_chunks, then we'll do the same with
+        # a smaller granularity (1mb) that is used by default in some AWS multi-part
+        # upload implementations.
+        #
+        # We're working backwards from num_chunks here becuase it's the only value we know.
```
sfc-gh-fcampbell, line 147:
```suggestion
        # We're working backwards from num_chunks here because it's the only value we know.
```

src/snowflake/cli/plugins/stage/md5.py
```diff
@@ -126,21 +127,29 @@ def file_matches_md5sum(local_file: Path, remote_md5: str | None) -> bool:
...
+        for chunk_size_alignment in [S3_CHUNK_SIZE, ONE_MEGABYTE]:
+            # +1/-1 ensures we don't add an additional chunk if exactly chunk-aligned size
```
sfc-gh-fcampbell, line 149:
```suggestion
            # +1 because we need at least one chunk when file_size < num_chunks * chunk_size_alignment
            # -1 because we don't want to add an extra chunk when file_size is an exact multiple of num_chunks * chunk_size_alignment
```

tests_integration/nativeapp/test_large_upload.py
sfc-gh-cgorrie: Yeah it was, I was paranoid that I broke the test so I unskipped it for now!

sfc-gh-turbaszek (approved):

sfc-gh-fcampbell (approved):

PR #1374, opened 2024-07-26
Decouple project definition object from artifact processors

sfc-gh-gbloom:
### Pre-review checklist

### Changes description

Instead of passing `NativeAppCompiler` the entire project definition object (via `NativeAppProjectModel`), we're passing it now the specific parameters it needs, grouped together as `BundleContext`.

The main reason for this is the upcoming support for project definition v2, which has significantly different structure (it's not enough to pass the entire definition object, we also need to know which entity is being bundled, its dependencies, etc.).

### DIFF
--- a/src/snowflake/cli/plugins/nativeapp/artifacts.py
+++ b/src/snowflake/cli/plugins/nativeapp/artifacts.py
@@ -19,13 +19,27 @@ import os
 from dataclasses import dataclass
 from pathlib import Path
 from textwrap import dedent
-from typing import Any, Callable, Dict, Iterable, Iterator, List, Optional, Tuple, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    Iterator,
+    List,
+    Optional,
+    Tuple,
+    Union,
+)
 
 from click.exceptions import ClickException
 from snowflake.cli.api.constants import DEFAULT_SIZE_LIMIT_MB
 from snowflake.cli.api.project.schemas.native_app.path_mapping import PathMapping
 from snowflake.cli.api.secure_path import SecurePath
 from yaml import safe_load
+
+if TYPE_CHECKING:
+    from snowflake.cli.plugins.nativeapp.project_model import NativeAppProjectModel
 
 
 class DeployRootError(ClickException):
@@ -107,6 +121,17 @@ class BundleContext:
     bundle_root: Path
     deploy_root: Path
     generated_root: Path
+
+    @staticmethod
+    def from_na_project(na: NativeAppProjectModel) -> BundleContext:
+        return BundleContext(
+            na.package_name,
+            na.artifacts,
+            na.project_root,
+            na.bundle_root,
+            na.deploy_root,
+            na.generated_root,
+        )
 
 
 ArtifactPredicate = Callable[[Path, Path], bool]
--- a/src/snowflake/cli/plugins/nativeapp/artifacts.py
+++ b/src/snowflake/cli/plugins/nativeapp/artifacts.py
@@ -16,30 +16,15 @@ from __future__ import annotations
 
 import itertools
 import os
-from dataclasses import dataclass
 from pathlib import Path
 from textwrap import dedent
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Callable,
-    Dict,
-    Iterable,
-    Iterator,
-    List,
-    Optional,
-    Tuple,
-    Union,
-)
+from typing import Any, Callable, Dict, Iterable, Iterator, List, Optional, Tuple, Union
 
 from click.exceptions import ClickException
 from snowflake.cli.api.constants import DEFAULT_SIZE_LIMIT_MB
 from snowflake.cli.api.project.schemas.native_app.path_mapping import PathMapping
 from snowflake.cli.api.secure_path import SecurePath
 from yaml import safe_load
-
-if TYPE_CHECKING:
-    from snowflake.cli.plugins.nativeapp.project_model import NativeAppProjectModel
 
 
 class DeployRootError(ClickException):
@@ -111,27 +96,6 @@ class NotInDeployRootError(ClickException):
         self.dest_path = dest_path
         self.deploy_root = deploy_root
         self.src_path = src_path
-
-
-@dataclass
-class BundleContext:
-    package_name: str
-    artifacts: List[PathMapping]
-    project_root: Path
-    bundle_root: Path
-    deploy_root: Path
-    generated_root: Path
-
-    @staticmethod
-    def from_na_project(na: NativeAppProjectModel) -> BundleContext:
-        return BundleContext(
-            na.package_name,
-            na.artifacts,
-            na.project_root,
-            na.bundle_root,
-            na.deploy_root,
-            na.generated_root,
-        )
 
 
 ArtifactPredicate = Callable[[Path, Path], bool]
--- a/src/snowflake/cli/plugins/nativeapp/artifacts.py
+++ b/src/snowflake/cli/plugins/nativeapp/artifacts.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 
 import itertools
 import os
+from dataclasses import dataclass
 from pathlib import Path
 from textwrap import dedent
 from typing import Any, Callable, Dict, Iterable, Iterator, List, Optional, Tuple, Union
@@ -96,6 +97,16 @@ class NotInDeployRootError(ClickException):
         self.dest_path = dest_path
         self.deploy_root = deploy_root
         self.src_path = src_path
+
+
+@dataclass
+class BundleContext:
+    package_name: str
+    artifacts: List[PathMapping]
+    project_root: Path
+    bundle_root: Path
+    deploy_root: Path
+    generated_root: Path
 
 
 ArtifactPredicate = Callable[[Path, Path], bool]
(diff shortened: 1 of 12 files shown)


src/snowflake/cli/plugins/nativeapp/codegen/artifact_processor.py
```diff
@@ -74,9 +73,17 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 class ArtifactProcessor(ABC):
     def __init__(
         self,
-        na_project: NativeAppProjectModel,
+        package_name: str,
+        project_root: Path,
+        deploy_root: Path,
+        bundle_root: Path,
+        generated_root: Path,
```
sfc-gh-bdufour, line 80: as discussed offline earlier, please don't go back to individual args. I gave a lot of context around the change in https://github.com/snowflakedb/snowflake-cli/pull/1262, feel free to take a look. I'd be OK with passing an object that encapsulates the source of these individual fields, but this is a step backwards that is likely to cause issues soon. In particular, I'm thinking that need a reliable way to find the manifest file, and this will differ in v1 vs v2.

sfc-gh-bdufour: Also, NativeAppProjectModel is already an abstraction that encapsulates the project definition model, so I'm doubly confused about why we need this change in the first place.

sfc-gh-gbloom: `NativeAppProjectModel` is not the right abstraction to support project definition v2: there might be multiple application packages, multiple bundles/stages, etc. That's the main motivation for this change.
What `NativeAppCompiler` and `ArtifactProcessor` really need is a subset of `NativeAppProjectModel` in the v1 flow. For v2 it's slightly different since we need to know which entity is being bundled, and possibly be a little smart about its dependencies.

Grouped the required values as a `BundleContext` data class. We might need to refactor further once we add definition v2 logic, but it should be enough as a first step.

sfc-gh-bdufour: Sure, I understand that, nothing prevents you from reworking the abstraction. You can modify/rename `NativeAppProjectModel ` if you want. You can also create some kind of context object as I suggested in scrum yesterday. That new abstraction would expose the paths that are used now, the artifact rules, etc. In v2 it's going to be tied to a specific entity, so you can just go for that upfront and make it some kind entity-specific context. When v1 is used, then the app project is the entity. Same abstraction, just a special case of it. It would work nicely and be extensible in the future.

tests/nativeapp/codegen/snowpark/test_python_processor.py
```diff
@@ -353,7 +354,17 @@ def test_process_no_collected_functions(
...
-            SnowparkAnnotationProcessor(na_project=project).process(
+            processor = SnowparkAnnotationProcessor(
+                BundleContext(
+                    package_name=project.package_name,
+                    artifacts=project.artifacts,
+                    project_root=project.project_root,
+                    deploy_root=project.deploy_root,
+                    bundle_root=project.bundle_root,
+                    generated_root=project.generated_root,
+                )
```
sfc-gh-cgorrie, line 365: Could introduce static `from_` methods on the `BundleContext` dataclass to prevent duplication / errors

sfc-gh-gbloom: Good suggestion! Added `from_na_project()`

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -16,16 +16,31 @@
...
+)
 
 from click.exceptions import ClickException
 from snowflake.cli.api.constants import DEFAULT_SIZE_LIMIT_MB
 from snowflake.cli.api.project.schemas.native_app.path_mapping import PathMapping
 from snowflake.cli.api.secure_path import SecurePath
 from yaml import safe_load
 
+if TYPE_CHECKING:
+    from snowflake.cli.plugins.nativeapp.project_model import NativeAppProjectModel
```
sfc-gh-cgorrie, line 42: Why do we need to use `if TYPE_CHECKING`?

sfc-gh-cgorrie (commented):
Overall this looks good to me, I'll let @sfc-gh-bdufour approve as he had the most comments before

src/snowflake/cli/plugins/nativeapp/codegen/setup/native_app_setup_processor.py
```diff
@@ -35,19 +35,15 @@
...
 
 
 class NativeAppSetupProcessor(ArtifactProcessor):
-    def __init__(
-        self,
-        na_project: NativeAppProjectModel,
-    ):
-        super().__init__(na_project=na_project)
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
sfc-gh-bdufour, line 46: why change the contract? This will accept anything instead of erroring out if the wrong parameters are passed.

src/snowflake/cli/plugins/nativeapp/artifacts.py
```diff
@@ -98,6 +113,27 @@ def __init__(
...
+class BundleContext:
+    package_name: str
+    artifacts: List[PathMapping]
+    project_root: Path
+    bundle_root: Path
+    deploy_root: Path
+    generated_root: Path
+
+    @staticmethod
+    def from_na_project(na: NativeAppProjectModel) -> BundleContext:
```
sfc-gh-bdufour, line 126: Given that I assume NAPM has to be removed entirely for v2 eventually, would it make sense to extract a bundlecontext from the current NAPM instead duplicating the data as it's happening in this PR?

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-bdufour: +1 that's surprising

sfc-gh-gbloom: It's a way to handle cyclic type dependencies. I wanted to specify the input type of `BundleContext.from_na_project()`, but it created a module dependency (`NativeAppProjectModel` depends on this module).

sfc-gh-bdufour: Why does NAPM depend on BundleContext? I don't see why the cycle needs to exist.

src/snowflake/cli/plugins/nativeapp/codegen/setup/native_app_setup_processor.py
sfc-gh-gbloom: It will fail on the base class if the wrong parameters are passed. I don't mind keeping the existing signature, but it's a duplicate of the base class and doesn't add anything to it.

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-gbloom: It depends on the `artifacts.py` module, not the class. I will extract `BundleContext` to its own file to break the cycle.

src/snowflake/cli/plugins/nativeapp/codegen/setup/native_app_setup_processor.py
sfc-gh-bdufour: Yup, I read that wrong, we can keep as is.

sfc-gh-bdufour (dismissed):

src/snowflake/cli/plugins/nativeapp/artifacts.py
sfc-gh-gbloom: Yeah, I think you're right. I'll move this method to `NativeAppProjectModel.get_bundle_context()`.

src/snowflake/cli/plugins/nativeapp/project_model.py
```diff
@@ -185,3 +186,13 @@ def debug_mode(self) -> Optional[bool]:
...
+
+    def get_bundle_context(self) -> BundleContext:
+        return BundleContext(
+            self.package_name,
+            self.artifacts,
+            self.project_root,
+            self.bundle_root,
+            self.deploy_root,
+            self.generated_root,
+        )
```
sfc-gh-bdufour, line 198: is it possible to pass these by name? If we get the order wrong we might not notice given how the test is written. Before this call and the definition were side by side, but now they're in different files so IMHO ordering is more of a concern.

sfc-gh-bdufour: BTW, this is not really what I meant by "extract a bundlecontext from NAPM". What I had in mind was that BundleContext would become the authority, and NAPM would own a (immutable) bundlecontext to which it would delegate the appropriate property lookups.

tests/nativeapp/test_project_model.py
```diff
@@ -173,3 +174,29 @@ def test_project_model_falls_back_to_current_role(
...
+            PathMapping(src="setup.sql", dest=None),
+            PathMapping(src="README.md", dest=None),
+        ],
+        project_dir,
+        Path(project_dir / "output" / "bundle"),
+        Path(project_dir / "output" / "deploy"),
+        Path(project_dir / "output" / "deploy" / "__generated"),
+    )
+
+    assert actual_bundle_ctx == expected_bundle_ctx
```
sfc-gh-bdufour, line 202: as above, getting worried that the order of values here could become inconsistent. I'd prefer to have the text somehow verify that asking for the deploy root as the right value, etc.

src/snowflake/cli/plugins/nativeapp/project_model.py
sfc-gh-gbloom: Changed to named params.

Yeah I started by doing something similar, but I actually think it's much simpler for NAPM to hold the logic of how to construct a BundleContext, instead of BundleContext being aware of anywhere its being used. WDYT?

sfc-gh-bdufour: sure, I'm fine either way

sfc-gh-bdufour (approved):

PR #1375, opened 2024-07-27
Common error for mutually exclusive flags

sfc-gh-jvasquezrojas:
### Pre-review checklist

### Changes description
- Add OverridableParameter as parent class of OverridableArgument and OverridableOption, it now raise IncompatibleOptionsError instead of ClickException
- Update cortext commands to use OverridableArgument
- Update sql command to raise IncompatibleOptionsError instead of ClickException
[Ticket](https://snowflakecomputing.atlassian.net/browse/SNOW-1482252) 

Error example:
image

### DIFF
--- a/src/snowflake/cli/_plugins/cortex/commands.py
+++ b/src/snowflake/cli/_plugins/cortex/commands.py
@@ -53,7 +53,7 @@ SEARCH_COMMAND_ENABLED = sys.version_info < PYTHON_3_12
 
 SOURCE_EXCLUSIVE_OPTION_NAMES = ["text", "file", "source_document_text"]
 
-ReadableFileOption = OverrideableOption(
+ExclusiveReadableFileOption = OverrideableOption(
     None,
     "--file",
     mutually_exclusive=SOURCE_EXCLUSIVE_OPTION_NAMES,
@@ -64,7 +64,7 @@ ReadableFileOption = OverrideableOption(
     show_default=False,
 )
 
-TextSourceArgument = OverrideableArgument(
+ExclusiveTextSourceArgument = OverrideableArgument(
     mutually_exclusive=SOURCE_EXCLUSIVE_OPTION_NAMES,
 )
 
@@ -120,7 +120,7 @@ def search(
     requires_connection=True,
 )
 def complete(
-    text: Optional[str] = TextSourceArgument(
+    text: Optional[str] = ExclusiveTextSourceArgument(
         default=None,
         help="Prompt to be used to generate a completion. Cannot be combined with --file option.",
         show_default=False,
@@ -130,7 +130,7 @@ def complete(
         "--model",
         help="String specifying the model to be used.",
     ),
-    file: Optional[Path] = ReadableFileOption(
+    file: Optional[Path] = ExclusiveReadableFileOption(
         help="JSON file containing conversation history to be used to generate a completion. Cannot be combined with TEXT argument.",
     ),
     **options,
@@ -169,12 +169,12 @@ def extract_answer(
         help="String containing the question to be answered.",
         show_default=False,
     ),
-    source_document_text: Optional[str] = TextSourceArgument(
+    source_document_text: Optional[str] = ExclusiveTextSourceArgument(
         default=None,
         help="String containing the plain-text or JSON document that contains the answer to the question. Cannot be combined with --file option.",
         show_default=False,
     ),
-    file: Optional[Path] = ReadableFileOption(
+    file: Optional[Path] = ExclusiveReadableFileOption(
         help="File containing the plain-text or JSON document that contains the answer to the question. Cannot be combined with SOURCE_DOCUMENT_TEXT argument.",
     ),
     **options,
@@ -209,12 +209,12 @@ def extract_answer(
     requires_connection=True,
 )
 def sentiment(
-    text: Optional[str] = TextSourceArgument(
+    text: Optional[str] = ExclusiveTextSourceArgument(
         default=None,
         help="String containing the text for which a sentiment score should be calculated. Cannot be combined with --file option.",
         show_default=False,
     ),
-    file: Optional[Path] = ReadableFileOption(
+    file: Optional[Path] = ExclusiveReadableFileOption(
         help="File containing the text for which a sentiment score should be calculated. Cannot be combined with TEXT argument.",
     ),
     **options,
@@ -246,12 +246,12 @@ def sentiment(
     requires_connection=True,
 )
 def summarize(
-    text: Optional[str] = TextSourceArgument(
+    text: Optional[str] = ExclusiveTextSourceArgument(
         default=None,
         help="String containing the English text from which a summary should be generated. Cannot be combined with --file option.",
         show_default=False,
     ),
-    file: Optional[Path] = ReadableFileOption(
+    file: Optional[Path] = ExclusiveReadableFileOption(
         help="File containing the English text from which a summary should be generated. Cannot be combined with TEXT argument.",
     ),
     **options,
@@ -281,7 +281,7 @@ def summarize(
     requires_connection=True,
 )
 def translate(
-    text: Optional[str] = TextSourceArgument(
+    text: Optional[str] = ExclusiveTextSourceArgument(
         default=None,
         help="String containing the text to be translated. Cannot be combined with --file option.",
         show_default=False,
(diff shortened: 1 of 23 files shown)


src/snowflake/cli/api/commands/overridable_parameter.py
```diff
@@ -0,0 +1,143 @@
...
+            return callback(**passed_params)
+
+        return generated_callback
+
+
+class OverrideableArgument(_OverrideableParameter):
+    def get_parameter(
+        self, default: Any = ..., *param_decls: str, **kwargs
+    ) -> typer.models.ArgumentInfo:
+        return typer.Argument(default, *param_decls, **kwargs)
```
sfc-gh-turbaszek, line 134: Is this return type consistent with annotations?

src/snowflake/cli/plugins/sql/commands.py
```diff
@@ -73,6 +74,11 @@ def execute_sql(
     The command supports variable substitution that happens on client-side. Both &VARIABLE or &{ VARIABLE }
     syntax are supported.
     """
+    exclusive_inputs = {"--filename": files, "--query": query, "--stdin": std_in}
+    incompatible_options = [key for key, value in exclusive_inputs.items() if value]
+    if len(incompatible_options) > 1:
+        raise IncompatibleParametersError(incompatible_options)
```
sfc-gh-turbaszek, line 80: Can you share how this error will look like to users?

src/snowflake/cli/api/exceptions.py
```diff
@@ -162,3 +162,12 @@ def __init__(self, part: str, name: str):
...
             f"{part.capitalize()} provided but name '{name}' is fully qualified name."
         )
+
+
+class IncompatibleParametersError(UsageError):
+    def __init__(self, options: list[str]):
+        options_with_quotes = [f"'{option}'" for option in options]
+        comma_separated_options = ", ".join(options_with_quotes[:-1])
+        super().__init__(
+            f"Parameters {comma_separated_options} and {options_with_quotes[-1]} are incompatible."
```
sfc-gh-turbaszek, line 172: Can you share how this error will look like to users?

src/snowflake/cli/plugins/sql/commands.py
```diff
@@ -73,6 +74,11 @@ def execute_sql(
...
+
```
sfc-gh-turbaszek, line 81: Can we handle that on options level?

src/snowflake/cli/plugins/sql/manager.py
```diff
@@ -40,13 +40,7 @@ def execute(
         std_in: bool,
         data: Dict | None = None,
     ) -> Tuple[IsSingleStatement, Iterable[SnowflakeCursor]]:
-        inputs = [query, files, std_in]
         # Check if any two inputs were provided simultaneously
-        if len([i for i in inputs if i]) > 1:
-            raise UsageError(
-                "Multiple input sources specified. Please specify only one."
-            )
-
```
sfc-gh-turbaszek, line 49: This should stay here. Managers methods can be reued outside command context

src/snowflake/cli/api/commands/overridable_parameter.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+# OverrideableOption doesn't work with flags with type List[Any] and default None. Because typer executes the callback
+# function and convert the default value iterating over it, but None is not iterable.
```
sfc-gh-pczajka, line 138:
```suggestion
# OverrideableOption doesn't work with flags with type List[Any] and default None, because typer executes the callback
# function which converts the default value iterating over it, but None is not iterable.
```

src/snowflake/cli/plugins/sql/commands.py
```diff
@@ -24,6 +24,7 @@
     variables_option,
 )
 from snowflake.cli.api.commands.snow_typer import SnowTyperFactory
+from snowflake.cli.api.exceptions import IncompatibleParametersError
```
sfc-gh-pczajka, line 27: Could you add it to other commands as well? I found that `snow object create` has object arguments which are exclusive with `--json object_definition` flag and `snow app events` have conflicting `--first` and `--last`

src/snowflake/cli/plugins/sql/commands.py
sfc-gh-jvasquezrojas: We can, but it will not show the default value for files because typer doesn't support callbacks with type list and default None, there is a discussion in the typer repo https://github.com/tiangolo/typer/discussions/762. 

Instead I added it by this way so it doesn't affect the help option and show the error for incompatible option in sql command.

src/snowflake/cli/api/commands/overridable_parameter.py
sfc-gh-jvasquezrojas: typer.Argument is a function that returns an object of type ArgumentInfo. So it is consistent with this get_parameter function

src/snowflake/cli/api/exceptions.py
sfc-gh-jvasquezrojas: I attached a screenshot to the description

src/snowflake/cli/plugins/sql/commands.py
sfc-gh-turbaszek: sounds good, thanks!

sfc-gh-turbaszek (approved):

sfc-gh-turbaszek (approved):

tests_integration/nativeapp/test_events.py
```diff
@@ -53,9 +53,10 @@ def test_app_events_mutually_exclusive_options(
             ["app", "events", *command],
             env=TEST_ENV,
         )
-        assert result.exit_code == 1, result.output
+
+        assert result.exit_code == 2, result.output
         assert (
-            f"{flag_names[0]} and {flag_names[1]} cannot be used together."
```
sfc-gh-pczajka, line 58: IMHO "cannot be used together" message is more understandable, WDYT?

(I don't want to stop PR with this discussion - feel free to merge if you think otherwise)

sfc-gh-pczajka (approved):

tests_integration/nativeapp/test_events.py
sfc-gh-jvasquezrojas: Yes it is less understandable, I change it to simultaneously instead of together, WDYT?
The full message is like the following:
`Parameters --file and --stdin are incompatible and cannot be used simultaneously.`

src/snowflake/cli/_plugins/cortex/commands.py
```diff
@@ -48,6 +51,23 @@
 
 SEARCH_COMMAND_ENABLED = sys.version_info < PYTHON_3_12
 
+SOURCE_EXCLUSIVE_OPTION_NAMES = ["text", "file", "source_document_text"]
```
sfc-gh-melnacouzi, line 54: Need some documentation for each option.

src/snowflake/cli/_plugins/cortex/commands.py
```diff
@@ -48,6 +51,23 @@
...
+
+ReadableFileOption = OverrideableOption(
```
sfc-gh-melnacouzi, line 56: Need documentation for this option, and TextSourceArgument. When to use them, and what they do.

src/snowflake/cli/_plugins/cortex/commands.py
```diff
@@ -152,14 +169,13 @@ def extract_answer(
         help="String containing the question to be answered.",
         show_default=False,
     ),
-    source_document_text: Optional[str] = typer.Argument(
-        None,
+    source_document_text: Optional[str] = TextSourceArgument(
```
sfc-gh-melnacouzi, line 172: having the name of the argument directly associated with SOURCE_EXCLUSIVE_OPTION_NAMES seems like an antipattern. 
If using (Exclusive)TextSourceArgument and (Exclusive)ReadableFileOption is not enough to ensure exclusivity, then I believe this change is not very useful, as you still need to explicitly specify what fields are exclusive.
Example issues:
- new command that uses new names.
- a command that does not require exclusivity
- command that was using TextSourceArgument (not for the purpose of exclusivity), but suddenly becomes exclusive on rename.

sfc-gh-jvasquezrojas: Thank you!  I rename both arguments to ExclusiveTextSourceArgument and ExclusiveReadableFileOption.

In some cases, the exclusive relationship between parameters is not transitive. Defining exclusiveness solely based on the parameter type can be less useful, as it assumes transitivity at all times.

sfc-gh-turbaszek (approved):

PR #1385, opened 2024-07-29
Add `--follow` to `snow app events`

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Adds `--follow` flag to `snow app events` to enable live-tail of events. The command will poll the `get_events` function every 10 seconds (by default, can be changed with `--follow-interval`), printing new rows on each iteration.

The way we detect new events is to make a second query filtering on rows where the `timestamp >=` the timestamp of the last row from the old results. The bound needs to be inclusive since there could be new events with the same timestamp in the new results (i.e. the `timestamp` column isn't necessarily monotonic). Since the bound is inclusive, the first few events from the new results would be the same events as the last few events of the old results, which means we need to only print the net-new rows from the new results. We do this by finding the point of largest overlap between the end of the old results and the start of the new results, returning the rows that come after.

### DIFF
--- a/src/snowflake/cli/plugins/nativeapp/commands.py
+++ b/src/snowflake/cli/plugins/nativeapp/commands.py
@@ -14,11 +14,12 @@
 
 from __future__ import annotations
 
+import itertools
 import logging
 from enum import Enum
 from pathlib import Path
 from textwrap import dedent
-from typing import List, Optional
+from typing import Generator, Iterable, List, Optional, cast
 
 import typer
 from click import ClickException
@@ -450,6 +451,12 @@ def app_events(
     last: int = typer.Option(
         default=0, help="Fetch only the last N events. Cannot be used with --first."
     ),
+    follow: bool = typer.Option(
+        False,
+        "--follow",
+        "-f",
+        help="Continue polling for events. Implies --last 20 unless overridden.",
+    ),
     **options,
 ):
     """Fetches events for this app from the event table configured in Snowflake."""
@@ -458,26 +465,42 @@ def app_events(
 
     assert_project_type("native_app")
 
+    record_type_names = [r.name for r in record_types]
     manager = NativeAppManager(
         project_definition=get_cli_context().project_definition.native_app,
         project_root=get_cli_context().project_root,
     )
-    events = manager.get_events(
-        since_interval=since,
-        until_interval=until,
-        record_types=[r.name for r in record_types],
-        scopes=scopes,
-        first=first,
-        last=last,
-    )
-    if not events:
-        return MessageResult("No events found.")
+    if follow:
+        if not last:
+            # If we don't have a value for --last, assume 20 so we
+            # at least print something before starting the tail
+            last = 20
+        stream: Iterable[CommandResult] = (
+            EventResult(event)
+            for event in manager.stream_events(
+                last=last,
+                delay_seconds=10,  # configurable?
+                record_types=record_type_names,
+                scopes=scopes,
+            )
+        )
+        # Append a newline at the end to make the CLI output clean when we hit Ctrl-C
+        stream = itertools.chain(stream, [MessageResult("")])
+    else:
+        stream = (
+            EventResult(event)
+            for event in manager.get_events(
+                since_interval=since,
+                until_interval=until,
+                record_types=record_type_names,
+                scopes=scopes,
+                first=first,
+                last=last,
+            )
+        )
 
-    def g():
-        for event in events:
-            yield EventResult(event)
-
-    return StreamResult(g())
+    # Cast the stream to a Generator since that's what StreamResult wants
+    return StreamResult(cast(Generator[CommandResult, None, None], stream))
 
 
 class EventResult(ObjectResult, MessageResult):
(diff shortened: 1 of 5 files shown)


src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -418,34 +424,66 @@ def app_events(
...
-        return MessageResult("No events found.")
-
-    def g():
-        for event in events:
-            yield EventResult(event)
+    if follow:
+        if not last:
+            # If we don't have a value for --last, assume 20 so we
+            # at least print something before starting the tail
+            last = DEFAULT_EVENT_FOLLOW_LAST
```
sfc-gh-gbloom, line 460: Isn't 0 a valid --last value?

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -773,6 +777,52 @@ def get_events(
...
+    last_overlap_found_at = 0
+    while overlap_amount <= min(len(new_events), len(previous_events)):
+        # Check if end of previous_events overlaps with start of new_events
+        # This is O(n) but from the SQL filters and ordering used to generate
+        # both lists of events, we know that the only overlap would be for events
+        # that happened in the exact same microsecond (usually only 1 or 2, rarely more)
+        if previous_events[-overlap_amount:] == new_events[:overlap_amount]:
+            last_overlap_found_at = overlap_amount
+        elif last_overlap_found_at:
+            break
```
sfc-gh-gbloom, line 822: Is it guaranteed that events with the same timestamp arrive in the same order persistently? If not, it might be safer to check all events with the same microsecond and not stop at the first found.

src/snowflake/cli/plugins/nativeapp/commands.py
sfc-gh-bdufour, line 460: Not sure about this, in the sense that if I use `snow app events --since "10 mins" --follow`, I'd expect all events matching my query, then a follow. Sounds like we're preventing that here?

I'm ok with some small number if the user specifies nothing else though, that makes sense. We might want to do this anyway if the user just says `snow app events` without --follow, for consistency.

src/snowflake/cli/plugins/nativeapp/manager.py
```diff
@@ -714,8 +716,8 @@ def get_validation_result(self, use_scratch_stage: bool):
 
     def get_events(
         self,
-        since_interval: str = "",
-        until_interval: str = "",
+        since: str | datetime | None = "",
+        until: str | datetime | None = "",
```
sfc-gh-bdufour, line 720: Alright, I stared at this for too long, so at the risk of asking a dumb question, how do we get a datetime here from the `events` command?

src/snowflake/cli/plugins/nativeapp/commands.py
sfc-gh-fcampbell: yeah using `--last 0` originally didn't make sense, but with `--follow` it does, so i'll update it to use `-1` as the empty value

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-fcampbell: yeah that's a good point, since the query has only `order by timestamp asc` there's no guarantee that the rows with the same timestamp will be ordered the same on the next call. i'll update the logic to handle this

src/snowflake/cli/plugins/nativeapp/commands.py
sfc-gh-fcampbell: good point, i'll rework the UX of the flags when `--follow` is used

src/snowflake/cli/plugins/nativeapp/manager.py
sfc-gh-fcampbell: when we poll for events since the last query, we pass in the timestamp of the last result row from the previous call: https://github.com/snowflakedb/snowflake-cli/blob/a27e8fab746d1271a75e5b094ddc0a32f0c0f91d/src/snowflake/cli/plugins/nativeapp/manager.py#L804, which is automatically parsed as a `datatime` by the python connector

sfc-gh-bdufour: python connector, that's the piece I was missing, thanks!

src/snowflake/cli/plugins/nativeapp/commands.py
```diff
@@ -445,39 +451,79 @@ def app_events(
...
+    ),
+    follow_interval: int = typer.Option(
+        10,
+        help=f"Polling interval in seconds when using the --follow flag.",
     ),
     **options,
 ):
     """Fetches events for this app from the event table configured in Snowflake."""
-    if first and last:
+    if first != -1 and last != -1:
```
sfc-gh-bdufour, line 479: nit: can we use first >= 0 and last >= 0 here? Technically the user could decide to pass in any int, might as well treat anything negative as disabling the option.

sfc-gh-fcampbell: fixed

sfc-gh-bdufour (approved):

PR #1392, opened 2024-07-30
WorkspaceManager and wiring of "snow ws bundle"

sfc-gh-gbloom:
### Pre-review checklist

### Changes description

1. Renamed the entity classes such that `*EntityModel` contain the Pydantic schemas, and the `*Entity` classes contain the actual logic.
2. Added a map to point the schema classes to their corresponding entity classes.
3. Created the `WorkspaceManager` class, similar to the existing pattern.
  a. `perform_action(entity_id, action)`
  b. `entity.supports(action)` -- entities support an action if they implement a method with the same name
4. Added the `snow ws bundle --entity-id=id` command.

### Manual test

Running `snow ws bundle --entity-id=pkg` with the following project definition raises `NotImplementedError("Application package bundle")`:
```yml
definition_version: 2
entities:
  pkg:
    type: application package
    name: myapp_pkg
    artifacts:
      - src: app/*
        dest: ./
    manifest: app/manifest.yml
```


src/snowflake/cli/api/entities/application_entity.py
sfc-gh-bdufour, line 1: Should NADE have code ownership of native app entities?

src/snowflake/cli/plugins/workspace/commands.py
```diff
@@ -33,3 +36,24 @@ def validate(
...
     return MessageResult("Project definition is valid.")
+
+
+@ws.command(requires_connection=True)
+@with_project_definition()
+def bundle(
+    key: str = typer.Option(
+        help=f"""The key of the entity you want to bundle.""",
+    ),
+    **options,
```
sfc-gh-bdufour, line 47: I know it's just temporary, but let's workshop what this "key" is going to be called before the we go too much further. Not sold on 'key' just yet.

src/snowflake/cli/plugins/workspace/manager.py
```diff
@@ -0,0 +1,35 @@
...
+    Entity,
+    v2_entity_model_to_entity_map,
+)
+from snowflake.cli.api.project.schemas.project_definition import DefinitionV20
+
+
+class WorkspaceManager:
+    def __init__(self, project_definition: DefinitionV20, project_root: Path):
+        if project_definition.definition_version != "2":
+            raise InvalidProjectDefinitionVersion20Error(
```
sfc-gh-bdufour, line 15: nit: does it have to be this specific? InvalidProjectDefinitionVersion would do, no? We're publishing exception names for telemetry as of 2.7, let's try to contain the one-off error types.

src/snowflake/cli/plugins/workspace/manager.py
```diff
@@ -0,0 +1,35 @@
...
+                project_definition.definition_version
+            )
+        self._entities_cache: Dict[str, Entity] = {}
+        self._pdf = project_definition
```
sfc-gh-bdufour, line 19: `self._project_definition` please. PDF isn't an intuitive name.

src/snowflake/cli/plugins/workspace/manager.py
```diff
@@ -0,0 +1,35 @@
...
+
+from snowflake.cli.api.exceptions import InvalidProjectDefinitionVersion20Error
+from snowflake.cli.api.project.schemas.entities.entities import (
+    Entity,
+    v2_entity_model_to_entity_map,
+)
+from snowflake.cli.api.project.schemas.project_definition import DefinitionV20
+
+
+class WorkspaceManager:
```
sfc-gh-bdufour, line 12: + doc comments throughout, please

tests/workspace/test_manager.py
```diff
@@ -0,0 +1,77 @@
...
+        _get_ws_manager(pdf_content=mock_snowflake_yml_v1_file)
+
+
+# Test that the same entity instance is returned for the same key
+def test_get_entity_is_cached(temp_dir):
+    ws_manager = _get_ws_manager()
+    pkg1 = ws_manager.get_entity("pkg")
+    pkg2 = ws_manager.get_entity("pkg")
+    app = ws_manager.get_entity("app")
+    assert id(pkg1) == id(pkg2)
```
sfc-gh-bdufour, line 57: nit: I think `pkg1 is pkg2` is a bit more straightforward for these checks. It's a bit more likely to yield a decent error message if it fails.

tests/workspace/utils.py
```diff
@@ -0,0 +1,54 @@
...
+
+from __future__ import annotations
+
+from textwrap import dedent
+
+APP_PACKAGE_ENTITY = (
+    "snowflake.cli.api.entities.application_package_entity.ApplicationPackageEntity"
+)
+
+mock_snowflake_yml_file = dedent(
```
sfc-gh-bdufour, line 23: nit: all caps? It's a constant

src/snowflake/cli/api/entities/application_entity.py
```diff
@@ -0,0 +1,5 @@
+from snowflake.cli.api.entities.common import EntityBase
+
+
+class ApplicationEntity(EntityBase):
```
sfc-gh-bdufour, line 4: + doc comment eventually?

src/snowflake/cli/api/entities/application_package_entity.py
```diff
@@ -0,0 +1,6 @@
+from snowflake.cli.api.entities.common import EntityBase
+
+
+class ApplicationPackageEntity(EntityBase):
```
sfc-gh-bdufour, line 4: ditto

src/snowflake/cli/plugins/workspace/manager.py
```diff
@@ -0,0 +1,35 @@
...
+    def __init__(self, project_definition: DefinitionV20, project_root: Path):
+        if project_definition.definition_version != "2":
```
sfc-gh-bdufour, line 14: we need a more robust version eventually though. Will the manager have to change for an eventual 2.1? I'd expect it usually wouldn't, which means that this check should really be "if major version == 2"

sfc-gh-bdufour (commented):
Looks good overall, some small notes here and there

src/snowflake/cli/api/entities/application_entity.py
sfc-gh-gbloom: Yes, the plan is to eventually move each entity to its place under the existing structure. So these will probably be under `src/snowflake/cli/plugins/nativeapp/v2/`, but it's not finalized yet and for the next couple of PRs it's easier to have these grouped together in a single directory.

src/snowflake/cli/plugins/workspace/commands.py
sfc-gh-gbloom: @sfc-gh-cgorrie Let's discuss the terminology here

src/snowflake/cli/plugins/workspace/manager.py
sfc-gh-gbloom: You're right, changed to a generic `InvalidProjectDefinitionVersionError`

src/snowflake/cli/plugins/workspace/manager.py
sfc-gh-gbloom: Done

tests/workspace/test_manager.py
sfc-gh-gbloom: Done

src/snowflake/cli/plugins/workspace/commands.py
sfc-gh-cgorrie: Considering it's temporary, I'd lean towards `--entity <id>` or `--entity-id <id>`.

sfc-gh-gbloom: Changed to `--entity-id`

src/snowflake/cli/plugins/workspace/manager.py
```diff
@@ -0,0 +1,42 @@
...
+            if entity_id not in self._project_definition.entities:
+                raise ValueError(f"No such entity ID: {entity_id}")
+            entity_model_cls = self._project_definition.entities[entity_id].__class__
+            entity_cls = v2_entity_model_to_entity_map[entity_model_cls]
+            self._entities_cache[entity_id] = entity_cls()
+        return self._entities_cache[entity_id]
+
+    def bundle(self, entity_id: str):
+        entity = self.get_entity(entity_id)
+        if callable(getattr(entity, "bundle", None)):
```
sfc-gh-cgorrie, line 39: @sfc-gh-turbaszek would love to get your thoughts on the way you'd like to dispatch things here. You had suggested a decorator-based approach, e.g.

```python
@action(type=Actions.BUNDLE)
def bundle(self):
    [...]
```

After thinking about it, I realize I don't have strong enough feelings either way, except for if we go for this approach I'd like to encapsulate it, e.g.

```suggestion
        if entity.supports(Action.BUNDLE):
```

src/snowflake/cli/api/project/schemas/entities/streamlit_entity_model.py
```diff
@@ -17,14 +17,14 @@
 from typing import List, Literal, Optional
```
sfc-gh-turbaszek, line 17: Should we put models in one module?

src/snowflake/cli/plugins/workspace/commands.py
```diff
@@ -33,3 +36,24 @@ def validate(
...
+    Prepares a local folder with the configured artifacts of the specified entity.
+    """
+
+    cli_context = get_cli_context()
+    ws = WorkspaceManager(
+        project_definition=cli_context.project_definition,
+        project_root=cli_context.project_root,
+    )
+
+    ws.bundle(entity_id)
```
sfc-gh-turbaszek, line 59: This is `ws bundle <entity_id>` right? I was convinced that we want to have `ws :entity_id bundle`?

src/snowflake/cli/plugins/workspace/manager.py
sfc-gh-turbaszek: My 2 cents:
1. Imho the method have to be decorated as action. We should not assume every instance attribute can be cli action. Especially than not every attribute has to be callable
2. +1 for `entity.supports(Action.BUNDLE)`
4. Why are we implementing `bundle` method here? It shouldn't it be `perform_action(entity_name, action_name)`?

src/snowflake/cli/plugins/workspace/manager.py
```diff
@@ -0,0 +1,42 @@
...
+class WorkspaceManager:
+    """
+    Instantiates entity instances from entity models, providing higher-order functionality on entity compositions.
+    """
+
+    def __init__(self, project_definition: DefinitionV20, project_root: Path):
+        if (
+            project_definition.definition_version != "2"
+            and not project_definition.definition_version.startswith("2.")
+        ):
```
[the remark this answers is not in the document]
sfc-gh-gbloom: Done

src/snowflake/cli/plugins/workspace/manager.py
sfc-gh-gbloom: Changed to the more generic method `perform_action(entity_id, action)` and added `entity.supports(action)`.

For now an entity is considered to support an action if it implements a method with the action name. Will refactor this convention once we introduce the action decorator.

src/snowflake/cli/api/project/schemas/entities/streamlit_entity_model.py
sfc-gh-gbloom: I believe we will eventually move each entity to its place under the existing structure, but for now it's easier to have these grouped together.

src/snowflake/cli/plugins/workspace/commands.py
sfc-gh-gbloom: Yes eventually it's what we'll have, but it requires more infra to support this command syntax. Until it's ready I temporarily used the `--entity-id` argument syntax.

src/snowflake/cli/plugins/workspace/manager.py
sfc-gh-turbaszek: Let's add this comment in the code, so others also know this.

src/snowflake/cli/_plugins/workspace/manager.py
```diff
@@ -0,0 +1,40 @@
...
+            )
+        self._entities_cache: Dict[str, Entity] = {}
+        self._project_definition = project_definition
+
+    def get_entity(self, entity_id: str):
+        if entity_id not in self._entities_cache:
+            if entity_id not in self._project_definition.entities:
+                raise ValueError(f"No such entity ID: {entity_id}")
+            entity_model_cls = self._project_definition.entities[entity_id].__class__
+            entity_cls = v2_entity_model_to_entity_map[entity_model_cls]
```
sfc-gh-turbaszek, line 31: Should we consider implementing something like this on `EntityModelBase`?
```py
model.get_entity_class()
```

In this way we don't have to implement a mapping and we keep the relation defined in one place.
```
class ApplicationEntityModel(EntityModelBase)
   __model_class = ApplicationPackageEntity
```

Later we can change `get_entity_class` to `get_entity` that would return instance of entity build from the model? WDYT?

src/snowflake/cli/_plugins/workspace/manager.py
```diff
@@ -0,0 +1,40 @@
...
+            self._entities_cache[entity_id] = entity_cls()
+        return self._entities_cache[entity_id]
+
+    def perform_action(self, entity_id: str, action: EntityActions):
```
sfc-gh-turbaszek, line 35: Let's add docstring describing what's happening there

src/snowflake/cli/_plugins/workspace/manager.py
```diff
@@ -0,0 +1,40 @@
...
+        entity = self.get_entity(entity_id)
+        if entity.supports(action):
+            getattr(entity, action)()
```
sfc-gh-turbaszek, line 38: Thought: we should probably consider enabling `pre` and `post` execution actions, this can be helpful to abstract operations around single entity.

sfc-gh-astus (approved):

src/snowflake/cli/_plugins/workspace/manager.py
```diff
@@ -0,0 +1,40 @@
...
+)
+from snowflake.cli.api.project.schemas.project_definition import DefinitionV20
+
+
+class WorkspaceManager:
+    """
+    Instantiates entity instances from entity models, providing higher-order functionality on entity compositions.
+    """
+
+    def __init__(self, project_definition: DefinitionV20, project_root: Path):
```
sfc-gh-melnacouzi, line 18: Since __init__ is checking if version is V2, it means users can pass in non-V2 PDF. I suggest we change the signature to ProjectDefinition then.

src/snowflake/cli/_plugins/workspace/manager.py
```diff
@@ -0,0 +1,40 @@
...
+        if not project_definition.meets_version_requirement("2"):
+            raise InvalidProjectDefinitionVersionError(
+                "2.x", project_definition.definition_version
+            )
+        self._entities_cache: Dict[str, Entity] = {}
+        self._project_definition = project_definition
+
+    def get_entity(self, entity_id: str):
+        if entity_id not in self._entities_cache:
```
sfc-gh-melnacouzi, line 27: nit:
to avoid multiple nesting and help with the logic clarity, we could do something like that:
```
if in cache:
  return it.
if not in pdf:
  raise error.
continue with regular logic.
```

src/snowflake/cli/_plugins/workspace/manager.py
sfc-gh-melnacouzi: +1

sfc-gh-gbloom: Yes, that's actually what I started with, but circular deps got in the way and it wasn't trivial to untangle them. Let's do it as a separate PR together with `get_entity()`?

src/snowflake/cli/_plugins/workspace/manager.py
sfc-gh-gbloom: I agree, that's something I think we'll need later on when we introduce compositions. It's probably a bit premature to add these right now (unless you see a concrete use case).

sfc-gh-turbaszek (approved):

sfc-gh-melnacouzi (approved):

PR #1397, opened 2024-07-31
Retry integration test failures automatically

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Integration tests have gotten flaky recently, mostly due to 504 timeouts from Snowflake. Since this isn't limited to a specific test, we can't flag it as `@flaky`, `@retry`, etc. This PR introduces a cross-platform script that will automatically re-run pytest with `--last-failed` so that any tests that fail get one more try. If they pass on the retry, they were most likely flaky; if they fail again, they were most likely not.

In a future PR, I'd like to start tracking the flakiness of individual tests (and its root cause), so we can come up with specific mitigation measures.

### DIFF
--- a/scripts/run_tests_with_retry.py
+++ b/scripts/run_tests_with_retry.py
@@ -5,19 +5,24 @@ from subprocess import run
 test_type = sys.argv[1].title()
 pytest = ["pytest"] + sys.argv[2:]
 
+
+def p(*s):
+    print(*s, file=sys.stderr, flush=True)
+
+
 # Run tests once
 if run(pytest, check=False).returncode == 0:
     sys.exit(0)
 
 # Then run the failed tests once more,
 # if they fail, the failure was most likely legitimate
-print(f"{test_type} tests failed, re-running to detect flakes")
+p(f"{test_type} tests failed, re-running to detect flakes")
 if (ret := run(pytest + ["--last-failed"]).returncode) != 0:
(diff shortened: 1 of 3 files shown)


scripts/run_tests_with_retry.py
```diff
@@ -0,0 +1,24 @@
...
+
+# Then run the failed tests once more,
+# if they fail, the failure was most likely legitimate
+p(f"{test_type} tests failed, re-running to detect flakes")
+if (ret := run(pytest + ["--last-failed"]).returncode) != 0:
+    p(f"{test_type} tests re-run failed")
+    sys.exit(ret)
+
+# If they succeed, they are flaky, we should tell the user
+p(f"{test_type} tests passed during retry, these are most likely flaky")
```
sfc-gh-turbaszek, line 24: Can we easily get a list of flaky test? That would help with marking them as such.

sfc-gh-turbaszek (approved):
One comment, otherwise looks reasonable to me 🚀

sfc-gh-pczajka (approved):

scripts/run_tests_with_retry.py
sfc-gh-pczajka: ➕ 1️⃣

sfc-gh-fcampbell: yeah i wanna do that as a followup. for now we can get the list from the GH actions log

PR #1419, opened 2024-08-06
Add is_empty check to bundle_map in build bundle step

sfc-gh-pjafari:
### Pre-review checklist

### Changes description
Throw an exception during `build_bundle` step if `bundle_map` has no artifacts. 

 Commands that reference build_bundle:
`BundleMap` method `build_bundle` is referenced only in `NativeAppManager`, which is in turn used in the below commands:
`run`, `create`, `bundle`,`diff`, `deploy`, and `validate`. (edit: I had included `drop` in this which was incorrect, it's not even a command :D)


src/snowflake/cli/_plugins/nativeapp/artifacts.py
```diff
@@ -658,6 +667,9 @@ def build_bundle(
     for artifact in artifacts:
         bundle_map.add(artifact)
 
+    if bundle_map.is_empty():
+        raise ArtifactError("No artifacts were found, nothing to do.")
```
sfc-gh-pjafari, line 671: Or I can create a new error class for specificity - e.g. `NoArtifactError`, if we prefer that?

sfc-gh-fcampbell: the error type is fine by me, but I think the message can be a bit more specific
```suggestion
        raise ArtifactError("No artifacts mapping found in project definition, nothing to do.")
```

tests_integration/nativeapp/test_bundle.py
```diff
@@ -285,6 +285,27 @@ def test_nativeapp_bundle_throws_error_on_too_many_files_to_dest(template_setup)
...
+    )
+
+    result = runner.invoke_json(
+        ["app", "bundle"],
+        env=TEST_ENV,
+    )
+    assert result.exit_code == 1
+    assert_that_result_failed_with_message_containing(
+        result,
+        "No artifacts were found, nothing to do.",
```
sfc-gh-fcampbell, line 305:
```suggestion
        "No artifacts mapping found in project definition, nothing to do.",
```

sfc-gh-melnacouzi (approved):
lgtm. It would help though to list all the use cases that use build_bundle() to ensure that this behavior is acceptable.

src/snowflake/cli/_plugins/nativeapp/artifacts.py
```diff
@@ -658,6 +667,11 @@ def build_bundle(
     for artifact in artifacts:
         bundle_map.add(artifact)
 
+    if bundle_map.is_empty():
```
sfc-gh-melnacouzi, line 670: for drop, wasn't part of the requirement to allow it to go through?

sfc-gh-bdufour: Correct. I don't think we should raise this exception in build_bundle, we should return the empty map instead. Individual uses of the bundle map can then check their preconditions as needed.

sfc-gh-pjafari: Though I had mistakenly included `drop` as a command that invokes build_bundle (it's not even a command :')), `teardown` (which is what we mean here) does not invoke `build_bundle`. 

I think it might be fair to say that any command that calls `build_bundle` effectively needs the bundle to not be empty, (and `teardown` does not call build_bundle). Having said that, it should make sense to keep the is_empty check as part of build_bundle. 

If we foresee that we may have a use case for calling build_bundle and allowing an empty bundle, I can add a more specific exception class for  "empty bundle", and then we can catch that specific exception in the caller and ignore it. 

What do you think?

sfc-gh-pjafari: Below two screenshots show:
`run` command with artifacts succeeds. Then artifacts are commented out (empty), `teardown` command succeeds:
run_with
teardown_wo

sfc-gh-melnacouzi (approved):

sfc-gh-fcampbell (approved):

PR #1427, opened 2024-08-08
SNOW-1565541 Implement fetching events shared from consumer account

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Adds `--consumer-org` and `--consumer-account` flags to allow devs to fetch events shared from a specific consumer installation of the app. Since shared events are copied into the provider's event table, everything keeps working the same as for locally installed apps. The only concern is if the organization is set up to send events to a different account using https://docs.snowflake.com/en/developer-guide/native-apps/setting-up-logging-and-events#configure-an-account-to-store-shared-events. We can't really know if this is the case since we're most likely not ORGADMIN, so the best we can do it detect when the account doesn't have an event table set up and alert the user (this was already implemented in #1337).


src/snowflake/cli/_plugins/nativeapp/commands.py
```diff
@@ -475,10 +482,24 @@ def app_events(
...
+    This requires event sharing to be set up to route events to the provider account:
+    https://docs.snowflake.com/en/developer-guide/native-apps/setting-up-logging-and-events
+    """
     if first >= 0 and last >= 0:
         raise ClickException("--first and --last cannot be used together.")
 
+    if (consumer_org and not consumer_account) or (
+        consumer_account and not consumer_org
+    ):
+        raise ClickException(
```
sfc-gh-melnacouzi, line 499: would UsageError() be more accurate in this case?

tests/__snapshots__/test_help_messages.ambr
```diff
@@ -251,48 +251,61 @@
...
-  | --env                      TEXT                   String in format of        |
-  |                                                   key=value. Overrides       |
-  |                                                   variables from env section |
-  |                                                   used for templating.       |
-  | --help             -h                             Show this message and      |
-  |                                                   exit.                      |
+   By default, this command will fetch events generated by an app installed in    
+   the current connection's account. To fetch events generated by an app          
+   installed in a consumer account, use the --consumer-org and --consumer-account 
+   flags. This requires event sharing to be set up to route events to the
```
sfc-gh-melnacouzi, line 257: nit: flags kind of implies it's a boolean... maybe we use arguments or parameters instead?

src/snowflake/cli/_plugins/nativeapp/manager.py
```diff
@@ -811,8 +813,20 @@ def get_events(
         if not self.account_event_table:
             raise NoEventTableForAccount()
 
-        # resource_attributes:"snow.database.name" uses the unquoted/uppercase app name
+        # resource_attributes uses the unquoted/uppercase app and package name
         app_name = unquote_identifier(self.app_name)
+        package_name = unquote_identifier(self.package_name)
```
sfc-gh-melnacouzi, line 818: why are we unquoting these? What if there are unsafe SQL characters (like `.` for example)

src/snowflake/cli/_plugins/nativeapp/commands.py
sfc-gh-fcampbell: good idea, updated all the checks in this command

tests/__snapshots__/test_help_messages.ambr
sfc-gh-fcampbell: updated to "options" since that's what the `--help` uses
image

src/snowflake/cli/_plugins/nativeapp/manager.py
sfc-gh-fcampbell: these aren't being used as identifiers, they're being compared as string literals, e.g. 
```sql
resource_attributes:"snow.application.package.name" = 'MY_PACKAGE.WITH_DOTS'
```

sfc-gh-melnacouzi (approved):

PR #1428, opened 2024-08-08
SNOW-1571915 Add pytest plugin to deflake tests and create a GH issue for flaky tests

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Converts the test runner script to a pytest plugin that retries test failures and marks them as flaky if they pass on the retry. At the end of the pytest run, we open a GitHub issue with failure details for each flaky tests encountered. If the issue already exists, we add a comment so we can track how often this is happening.

When a test flakes, its outcome is changed to `FLAKY` in the pytest output:
image

And it is categorized as `flaky` in the stats:
image

Example of the issue created when a test is flaky:
image

Example of a +1 comment:
image

Sorting issues by most commented to address flakiest tests:
image


tests_integration/conftest.py
```diff
@@ -43,6 +43,7 @@
 
 pytest_plugins = [
     "tests_common",
+    "tests_common.deflake",
```
sfc-gh-fcampbell, line 46: we could also do this for unit tests if we wanted to

.github/workflows/test_trusted.yaml
```diff
@@ -62,6 +62,7 @@ jobs:
 
       - name: Run integration tests
         env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
sfc-gh-fcampbell, line 65: this passes the token, but it's not yet authorized for write access to issues, so the plugin will just log an exception for now

tests_common/deflake.py
```diff
@@ -0,0 +1,341 @@
...
+        self.flaky_tests = 0
+        self.test_run = TestRun()
+
+        self.runner: type[runner] = config.pluginmanager.getplugin("runner")
+        self.test_type: str = config.getoption(TEST_TYPE_OPTION)
+
+        if token := os.getenv("GH_TOKEN"):
+            # Grab the current commit so we can generate absolute URLs
+            rev_parse = run(["git", "rev-parse", "HEAD"], text=True, stdout=PIPE)
+            self.github: GitHub | None = GitHub(token, rev_parse.stdout.strip())
```
sfc-gh-turbaszek, line 52: Have you considered running `gh` in subprocess? If the only thing we want to do is playing with issues then `gh issue edit`

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-github-cli-in-workflows

In this way we may reduce complexity by removing `GitHub` class. And it's a bit safer - the authentication is handled by Github.

tests_common/deflake.py
```diff
@@ -0,0 +1,341 @@
...
+    def pytest_runtest_logstart(self, nodeid: str):
+        # This is called once before each test regardless of the number of retries
+        self.test_run.tests[nodeid] = TestResult(nodeid)
+
+    @pytest.hookimpl(hookwrapper=True)
+    def pytest_runtest_makereport(self, item: Item, call: CallInfo) -> TestReport:
+        # Wrap the creation of the test report for this phase to override the
+        # status to "flaky" if it's a retry and it passed
+        previous_outcomes = item.stash.setdefault(PREVIOUS_OUTCOME_KEY, {})
+        report = (yield).get_result()
```
sfc-gh-turbaszek, line 96: Can you explain this bit of magic? I've never seen something like that 🤔

tests_common/deflake.py
```diff
@@ -0,0 +1,341 @@
...
+        return report
+
+    def pytest_runtest_logreport(self, report: TestReport) -> None:
+        # Record the crash data for the GitHub issue
+        # and set the status of the overall test result
+        test = self.test_run.tests[report.nodeid]
+        phase = getattr(test, report.when)
+        phase.outcome = report.outcome
+
+        if report.outcome == "failed":
```
sfc-gh-turbaszek, line 121: Create constants for failed and flaky

tests_common/deflake.py
sfc-gh-fcampbell: generators in python can receive values after yielding, which is received as the result of the yield. here, pytest is sending an object of type `Result`, which has a `get_result()` method to get the actual return value of the hook being wrapped. 

the specific `report = (yield).get_result()` syntax is just a shorthand for
```python
outcome = yield
report = outcome.get_result()
```

More information on hook wrappers: https://docs.pytest.org/en/7.1.x/how-to/writing_hook_functions.html#hookwrapper-executing-around-other-hooks

tests_common/deflake.py
sfc-gh-fcampbell: i'd still like to have the `GitHub` class to encapsulate all the issue manipulation, and I find it cleaner to deal with HTTP APIs when practical over spawning subprocesses. In terms of security, the GH_TOKEN still needs to be in an env var even when using the `gh` CLI, so I don't think it makes a big difference. If you feel strongly about it I can change it

tests_common/deflake.py
sfc-gh-turbaszek: Thanks! Can we do
```
outcome: Result = yield  # yield nothing, receive result when generator resumes
report = outcome.get_result()
```
I think it will help our future selves

sfc-gh-fcampbell: done

sfc-gh-turbaszek (approved):

PR #1429, opened 2024-08-08
Workspaces application package bundle

sfc-gh-gbloom:
### Pre-review checklist

### Changes description

Implementing the v2 `bundle()` logic of `ApplicationPackageEntity`, enabling `snow ws bundle --entity-id=my_app_pkg`. This is a reuse of the existing v1 `snow app bundle`.

Due to circular dependencies, this change required extracting some of the existing code to dedicated modules:
* `OnErrorType`
* `Variable`
* `parse_key_value_variables`
* `print_diff_to_console`

Additionally, wrapped one of the imports in `cli_global_context.py` with `TYPE_CHECKING`, as breaking this cycle required a significant refactor. See PR comment on the file.

### Manual test

Running `snow ws bundle` with a simple v2 definition, verifying that the deploy directory is created with the correct artifacts:
```
$ snow ws bundle --entity-id=pkg

Bundle generated at /tmp/pdfv2-test/output/deploy
```


src/snowflake/cli/api/cli_global_context.py
```diff
@@ -17,14 +17,16 @@
...
+from typing import TYPE_CHECKING, Callable, Optional
 
 from snowflake.cli.api.exceptions import InvalidSchemaError
 from snowflake.cli.api.output.formats import OutputFormat
-from snowflake.cli.api.project.schemas.project_definition import ProjectDefinition
 from snowflake.connector import SnowflakeConnection
 from snowflake.connector.compat import IS_WINDOWS
 
+if TYPE_CHECKING:
+    from snowflake.cli.api.project.schemas.project_definition import ProjectDefinition
```
sfc-gh-gbloom, line 28: Importing `ProjectDefinition` from this module creates a circular dependency if any of the downstream modules try to `cli_console.warning()` (or step/message/etc).

Up until now we managed to avoid the cycle because the project definition did not have references to the business logic part of the code. For Workspaces we need to map each schema to its entity (`v2_entity_model_to_entity_map`).

This is the import cycle:
```
snowflake.cli.api.cli_global_context ->
snowflake.cli.api.project.schemas.project_definition ->
snowflake.cli.api.project.schemas.entities.entities ->
snowflake.cli.api.entities.application_package_entity ->
snowflake.cli._plugins.nativeapp.codegen.compiler ->
snowflake.cli._plugins.nativeapp.codegen.setup.native_app_setup_processor ->
snowflake.cli._plugins.stage.diff ->
snowflake.cli.api.console ->
snowflake.cli.api.console.console ->
snowflake.cli.api.console.abc ->
snowflake.cli.api.cli_global_context
```

src/snowflake/cli/_plugins/git/commands.py
```diff
@@ -37,6 +36,7 @@
     like_option,
 )
 from snowflake.cli.api.commands.snow_typer import SnowTyperFactory
+from snowflake.cli.api.commands.structs import OnErrorType
```
sfc-gh-bdufour, line 39: nit: not sure about `structs` as a name. `types` or `common` are more in line with existing names IMHO.

src/snowflake/cli/_plugins/workspace/manager.py
```diff
@@ -34,9 +36,10 @@ def get_entity(self, entity_id: str):
             return self._entities_cache[entity_id]
         if entity_id not in self._project_definition.entities:
             raise ValueError(f"No such entity ID: {entity_id}")
-        entity_model_cls = self._project_definition.entities[entity_id].__class__
+        entity_model = self._project_definition.entities[entity_id]
```
sfc-gh-bdufour, line 39: nit: I usually prefer to use `get` and check for `None`. It avoids the double lookup and guards against weird state being introduced elsewhere.

src/snowflake/cli/api/cli_global_context.py
sfc-gh-bdufour: it's a bit surgical to do it this way, but it will do.

src/snowflake/cli/api/commands/utils.py
```diff
@@ -0,0 +1,20 @@
...
+from snowflake.cli.api.commands.structs import Variable
+
+
+def parse_key_value_variables(variables: Optional[List[str]]) -> List[Variable]:
+    """Util for parsing key=value input. Useful for commands accepting multiple input options."""
+    if not variables:
+        return []
+    result: List[Variable] = []
+    if not variables:
+        return result
```
sfc-gh-fcampbell, line 13: i know this is just moved from another module, but we could clean up the duplicated `if not variables` here
```suggestion
    result: List[Variable] = []
    if not variables:
        return result
```

src/snowflake/cli/api/entities/application_package_entity.py
```diff
@@ -1,10 +1,35 @@
...
     """
 
-    def bundle(self):
-        raise NotImplementedError("Application package bundle")
+    _entity_model: ApplicationPackageEntityModel
+
+    def bundle(self, ctx: ActionContext):
+        model = self._entity_model
+        bundle_map = build_bundle(
+            ctx.project_root, Path(model.deploy_root), model.artifacts
```
sfc-gh-fcampbell, line 23: can/should we make `model.deploy_root` a `Path` object? let's avoid keeping paths as raw strings as early as possible

(also for the other path-like strings in that class)

src/snowflake/cli/api/entities/application_package_entity.py
```diff
@@ -1,10 +1,35 @@
...
 
 
 class ApplicationPackageEntity(EntityBase):
     """
     A Native App application package.
     """
 
-    def bundle(self):
-        raise NotImplementedError("Application package bundle")
+    _entity_model: ApplicationPackageEntityModel
```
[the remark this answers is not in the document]
sfc-gh-gbloom: Good suggestion - moved this type to be a generic on the base class

src/snowflake/cli/api/commands/structs.py
```diff
@@ -0,0 +1,17 @@
+from dataclasses import dataclass
+from enum import Enum
+
+
+class OnErrorType(Enum):
```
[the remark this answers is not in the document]
sfc-gh-gbloom: Done

src/snowflake/cli/api/commands/structs.py
```diff
@@ -0,0 +1,17 @@
...
+    BREAK = "break"
+    CONTINUE = "continue"
+
+
+@dataclass
+class Variable:
```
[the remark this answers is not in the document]
sfc-gh-gbloom: Done

src/snowflake/cli/_plugins/git/commands.py
sfc-gh-gbloom: Changed to `common`

src/snowflake/cli/_plugins/workspace/manager.py
sfc-gh-gbloom: Done

src/snowflake/cli/api/commands/utils.py
sfc-gh-gbloom: Done

src/snowflake/cli/api/entities/application_package_entity.py
sfc-gh-gbloom: We started with `Path`, and after a previous debate we decided to keep it as `str` on the model class for now. We can revisit this decision in the future.

src/snowflake/cli/api/entities/common.py
```diff
@@ -1,16 +1,20 @@
 from enum import Enum
+from typing import Generic, TypeVar
 
 
 class EntityActions(str, Enum):
     BUNDLE = "bundle"
 
 
-class EntityBase:
+TEntityModel = TypeVar("TEntityModel")
```
sfc-gh-bdufour, line 9: nit: I think the python standard is to use single letters for type params, most typically `T`.

sfc-gh-bdufour (dismissed):

src/snowflake/cli/api/entities/common.py
sfc-gh-gbloom: Alright, changed to `T`, but personally I find it much less readable when it's not trivial to understand what T is

src/snowflake/cli/api/entities/application_package_entity.py
```diff
@@ -1,10 +1,33 @@
...
 
-class ApplicationPackageEntity(EntityBase):
+class ApplicationPackageEntity(EntityBase[ApplicationPackageEntityModel]):
     """
     A Native App application package.
     """
 
-    def bundle(self):
-        raise NotImplementedError("Application package bundle")
+    def bundle(self, ctx: ActionContext):
```
sfc-gh-turbaszek, line 18: Should we consider `action_bundle` so when we look for action we look for `action_{action}` method? It should make it easier to differentiate between helpers and action method until we have decorators. WDYT?

src/snowflake/cli/api/entities/common.py
```diff
@@ -1,15 +1,26 @@
...
     """
     Base class for the fully-featured entity classes.
     """
 
+    def __init__(self, entity_model: T):
+        self._entity_model = entity_model
+
+    @classmethod
+    def get_entity_model_type(cls) -> T:
+        return get_args(cls.__orig_bases__[0])[0]  # type: ignore[attr-defined]
```
sfc-gh-turbaszek, line 22: Can you add comment what's happening there?

src/snowflake/cli/api/project/schemas/entities/entities.py
```diff
@@ -56,9 +56,5 @@
 
 v2_entity_model_types_map = {e.get_type(): e for e in ALL_ENTITY_MODELS}
 v2_entity_model_to_entity_map: Dict[EntityModel, Entity] = {
-    ApplicationEntityModel: ApplicationEntity,
-    ApplicationPackageEntityModel: ApplicationPackageEntity,
-    StreamlitEntityModel: StreamlitEntity,
-    FunctionEntityModel: FunctionEntity,
-    ProcedureEntityModel: ProcedureEntity,
+    e.get_entity_model_type(): e for e in ALL_ENTITIES
```
sfc-gh-turbaszek, line 59: Thanks!

src/snowflake/cli/api/entities/application_package_entity.py
sfc-gh-gbloom: Sure. Changed the enum value to `BUNDLE = "action_bundle"`

src/snowflake/cli/api/entities/common.py
sfc-gh-gbloom: Done

sfc-gh-turbaszek (dismissed):

src/snowflake/cli/api/entities/common.py
```diff
@@ -1,22 +1,32 @@
...
-    def __init__(self, entity_model: TEntityModel):
+    def __init__(self, entity_model: T):
         self._entity_model = entity_model
 
+    @classmethod
+    def get_entity_model_type(cls) -> T:
+        """
+        Returns the generic model type specified in each entity class.
+        For example:
+            ApplicationEntity(EntityBase[ApplicationEntityModel]
```
sfc-gh-bdufour, line 25: - Malformed, please fix.
- Isn't this method intended to return a class object, not an instance? The signature says it returns an instance, but I'm not sure that's true.
- Not sure I get the example. I expect that when I call this on an ApplicationEntity class, it returns ApplicationEntityModel. Is that right? If so, I don't think the example makes that clear at all.

src/snowflake/cli/api/entities/snowpark_entity.py
```diff
@@ -1,19 +1,27 @@
...
 
-class SnowparkEntity(EntityBase):
+class SnowparkEntity(EntityBase[Generic[T]]):
     pass
 
 
-class FunctionEntity(SnowparkEntity):
+class FunctionEntity(SnowparkEntity[FunctionEntityModel]):
     """
     A single UDF
```
sfc-gh-bdufour, line 18: UDF is not the only kind of function in Snowpark (they have UDTF and UDAF as well)

src/snowflake/cli/api/entities/common.py
sfc-gh-gbloom: Updated

src/snowflake/cli/api/entities/snowpark_entity.py
sfc-gh-gbloom: I'll let @sfc-gh-jsikorski and @sfc-gh-turbaszek address this in a future PR

src/snowflake/cli/_plugins/workspace/commands.py
```diff
@@ -57,4 +58,5 @@ def bundle(
         project_root=cli_context.project_root,
     )
 
-    ws.perform_action(entity_id, EntityActions.BUNDLE)
+    bundle_map: BundleContext = ws.perform_action(entity_id, EntityActions.BUNDLE)
```
sfc-gh-bdufour, line 61: is this right? the bundle action seems to return a bundle map, not a context instance

src/snowflake/cli/_plugins/workspace/manager.py
```diff
@@ -45,6 +48,10 @@ def perform_action(self, entity_id: str, action: EntityActions):
         """
         entity = self.get_entity(entity_id)
         if entity.supports(action):
-            getattr(entity, action)()
+            action_ctx = ActionContext(project_root=self.project_root())
+            return getattr(entity, action)(action_ctx)
```
sfc-gh-bdufour, line 52: if `supports` is a primitive for all entities, shouldn't something like `perform` be defined as well? I'd expect this to look like:
```
action_ctx = ...
if entity.supports(action):
   return entity.perform(action, action_ctx)
```

src/snowflake/cli/api/entities/common.py
```diff
@@ -1,15 +1,32 @@
...
+    def __init__(self, entity_model: T):
+        self._entity_model = entity_model
+
+    @classmethod
+    def get_entity_model_type(cls) -> Type[T]:
+        """
+        Returns the generic model class specified in each entity class.
+
+        For example, the ApplicationEntityModel class will be returned:
+        ApplicationEntity(EntityBase[ApplicationEntityModel]
```
sfc-gh-bdufour, line 26: The wording is super unclear to me. Suggestion: 

"""
For example, calling `ApplicationEntity.get_entity_model_type()` returns `ApplicationEntityModel`
"""

or something like that.

src/snowflake/cli/api/project/schemas/entities/entities.py
sfc-gh-bdufour, line 59: I'm finding it difficult to understand why we have to resort to union types to define the entity and model concepts. One of these days I'd really like it if you could go over it with me.

src/snowflake/cli/_plugins/workspace/manager.py
sfc-gh-gbloom: Changed. (This API is probably temporary and will soon be replaced with action decorators)

src/snowflake/cli/api/entities/common.py
sfc-gh-gbloom: Done

sfc-gh-bdufour (approved):

sfc-gh-turbaszek (approved):

PR #1437, opened 2024-08-12
SNOW-1623250 Add current test name to test resource env var in tests

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Adds the current test name as a suffix on created apps and packages so we can tell which resources are created by which test. The method chosen to inject the name removes the need to pass `env=TEST_ENV` when calling `runner.invoke` and also fixes a bug that was introduced in #1396 due to the use of `--dist-worksteal` and the `TEST_ENV` being a module global (it was possible for a test to drop a resource used by another test concurrently because all tests in a module used the same random identifier).

If necessary for manual queries, the suffix can be retrieved from the `identifier_suffix` fixture.

Also moves the manual setting of the `USER` env var in integration tests to the CLI runner so that we can use `ctx.env.USER` in templates without having to worry about Windows. The default username can be retrieved from the `default_username` fixture if necessary.

### DIFF
--- a/src/snowflake/cli/_plugins/nativeapp/project_model.py
+++ b/src/snowflake/cli/_plugins/nativeapp/project_model.py
@@ -142,7 +142,7 @@ class NativeAppProjectModel:
             # explicitly specified, the default is generated here,
             # so we have to append the test resource suffix here
             name = default_app_package(self.project_identifier)
-            return append_test_resource_suffix(name)
+            return to_identifier(append_test_resource_suffix(name))
 
     @cached_property
     def package_role(self) -> str:
@@ -167,7 +167,7 @@ class NativeAppProjectModel:
             # explicitly specified, the default is generated here,
             # so we have to append the test resource suffix here
             name = default_application(self.project_identifier)
-            return append_test_resource_suffix(name)
+            return to_identifier(append_test_resource_suffix(name))
 
     @cached_property
     def app_role(self) -> str:
--- a/src/snowflake/cli/_plugins/nativeapp/project_model.py
+++ b/src/snowflake/cli/_plugins/nativeapp/project_model.py
@@ -14,7 +14,6 @@
 
 from __future__ import annotations
 
-import os
 from functools import cached_property
 from pathlib import Path
 from typing import List, Optional
@@ -33,13 +32,11 @@ from snowflake.cli.api.project.schemas.native_app.application import (
 from snowflake.cli.api.project.schemas.native_app.native_app import NativeApp
 from snowflake.cli.api.project.schemas.native_app.path_mapping import PathMapping
 from snowflake.cli.api.project.util import (
-    concat_identifiers,
+    append_test_resource_suffix,
     extract_schema,
     to_identifier,
 )
 from snowflake.connector import DictCursor
-
-RESOURCE_SUFFIX_VAR = "SNOWFLAKE_CLI_TEST_RESOURCE_SUFFIX"
 
 
 def current_role() -> str:
@@ -139,10 +136,13 @@ class NativeAppProjectModel:
     @property
     def package_name(self) -> str:
         if self.definition.package and self.definition.package.name:
-            name = self.definition.package.name
+            return to_identifier(self.definition.package.name)
         else:
+            # V1.0 PDF doesn't support templating, so if the identifier isn't
+            # explicitly specified, the default is generated here,
+            # so we have to append the test resource suffix here
             name = default_app_package(self.project_identifier)
-        return concat_identifiers([name, resource_suffix()])
+            return append_test_resource_suffix(name)
 
     @cached_property
     def package_role(self) -> str:
@@ -161,10 +161,13 @@ class NativeAppProjectModel:
     @property
     def app_name(self) -> str:
         if self.definition.application and self.definition.application.name:
-            name = to_identifier(self.definition.application.name)
+            return to_identifier(self.definition.application.name)
         else:
-            name = to_identifier(default_application(self.project_identifier))
-        return concat_identifiers([name, resource_suffix()])
+            # V1.0 PDF doesn't support templating, so if the identifier isn't
+            # explicitly specified, the default is generated here,
+            # so we have to append the test resource suffix here
+            name = default_application(self.project_identifier)
+            return append_test_resource_suffix(name)
 
     @cached_property
     def app_role(self) -> str:
@@ -215,12 +218,3 @@ class NativeAppProjectModel:
             deploy_root=self.deploy_root,
             generated_root=self.generated_root,
         )
-
-
-def resource_suffix() -> str:
-    """
-    A suffix that should be added to account-level Native App resources.
-
-    This is an internal concern that is currently only used in tests.
-    """
-    return os.environ.get(RESOURCE_SUFFIX_VAR, "")
--- a/src/snowflake/cli/api/project/schemas/entities/application_entity_model.py
+++ b/src/snowflake/cli/api/project/schemas/entities/application_entity_model.py
@@ -52,10 +52,5 @@ class ApplicationEntityModel(EntityModelBase):
         )
         with_suffix = append_test_resource_suffix(identifier)
         if isinstance(input_value, Identifier):
-            data = dict(name=with_suffix)
-            if input_value.schema_:
-                data["schema"] = input_value.schema_
-            if input_value.database:
-                data["database"] = input_value.database
-            return Identifier(**data)
+            return input_value.model_copy(update=dict(name=with_suffix))
         return with_suffix
--- a/src/snowflake/cli/api/project/schemas/entities/application_entity_model.py
+++ b/src/snowflake/cli/api/project/schemas/entities/application_entity_model.py
@@ -16,7 +16,7 @@ from __future__ import annotations
 
 from typing import Literal, Optional
 
-from pydantic import Field
+from pydantic import Field, field_validator
 from snowflake.cli.api.project.schemas.entities.application_package_entity_model import (
     ApplicationPackageEntityModel,
 )
@@ -24,9 +24,11 @@ from snowflake.cli.api.project.schemas.entities.common import (
     EntityModelBase,
     TargetField,
 )
+from snowflake.cli.api.project.schemas.identifier_model import Identifier
 from snowflake.cli.api.project.schemas.updatable_model import (
     DiscriminatorField,
 )
+from snowflake.cli.api.project.util import append_test_resource_suffix
 
 
 class ApplicationEntityModel(EntityModelBase):
@@ -39,3 +41,21 @@ class ApplicationEntityModel(EntityModelBase):
         title="Whether to enable debug mode when using a named stage to create an application object",
         default=None,
     )
+
+    @field_validator("identifier")
+    @classmethod
+    def append_test_resource_suffix_to_identifier(
+        cls, input_value: Identifier | str
+    ) -> Identifier | str:
+        identifier = (
+            input_value.name if isinstance(input_value, Identifier) else input_value
+        )
+        with_suffix = append_test_resource_suffix(identifier)
+        if isinstance(input_value, Identifier):
+            data = dict(name=with_suffix)
+            if input_value.schema_:
+                data["schema"] = input_value.schema_
+            if input_value.database:
+                data["database"] = input_value.database
+            return Identifier(**data)
+        return with_suffix
--- a/src/snowflake/cli/api/project/schemas/updatable_model.py
+++ b/src/snowflake/cli/api/project/schemas/updatable_model.py
@@ -160,7 +160,8 @@ class UpdatableModel(BaseModel):
 
         setattr(
             cls,
-            f"_field_validator_with_verbose_name_to_avoid_name_conflict_{field_name}",
+            # Unique name so that subclasses get a unique instance of this validator
+            f"_{cls.__module__}.{cls.__name__}_validate_{field_name}",
             field_validator(field_name, mode="wrap")(validator_skipping_templated_str),
         )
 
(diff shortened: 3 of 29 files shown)


sfc-gh-astus (dismissed):

tests_integration/conftest.py
```diff
@@ -100,6 +104,15 @@ def use_config(self, config_file_name: str) -> None:
     def invoke(self, *a, **kw):
         if "catch_exceptions" not in kw:
             kw.update(catch_exceptions=False)
+
+        # The CLI automatically appends $USER to some created resource identifiers,
+        # let's use this behaviour to add a unique suffix to resources
+        # created/updated/dropped in tests to allow us to run simultaneous instances
+        env = kw.get("env", {}).copy()
+        env["USER"] = (
+            f"{env['USER']}_{self._user_suffix}" if "USER" in env else self._user_suffix
```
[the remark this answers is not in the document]
sfc-gh-fcampbell: outside this class, it's called `identifier_suffix` since it's meant to be a generic concept. inside the class it's called `user_suffix` to hint at how it's used.

I still need to append if USER is passed in since it's adding a random ID and the test name (tests no longer pass in the random ID, and if they pass in a static value for USER, we still need to append the random ID to avoid concurrecy conflicts)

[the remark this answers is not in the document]
sfc-gh-fcampbell: updated: [`d0c7a32` (#1437)](https://github.com/snowflakedb/snowflake-cli/pull/1437/commits/d0c7a32bf8c000ab1693838286cc2d38cfad47fb)

src/snowflake/cli/api/project/util.py
```diff
@@ -183,12 +183,6 @@ def extract_schema(qualified_name: str):
     return None
 
 
-def generate_user_env(username: str) -> dict:
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: I meant we shouldn't add test utils in production code. Don't we have an equivalent util file in tests?

tests_integration/conftest.py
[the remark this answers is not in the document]
sfc-gh-melnacouzi: It's confusing that somehow USER becomes identifier_suffix magically. If you pass in a USER, then USER is also mutated to a different value.
 
1. I understand that you want to make this reusable. In that case, do we need _suffix? it's more like a test_identifier.

2. If I pass in USER in env variables in a test, I do not expect it to be mutated. My test could be focused on testing USER specific logic, and I should be able to do that safely.

src/snowflake/cli/api/project/util.py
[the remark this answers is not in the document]
sfc-gh-fcampbell: it's being removed from prod code :)

tests_integration/conftest.py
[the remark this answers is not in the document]
sfc-gh-melnacouzi: Based on our offline discussion, this is ok for now until we have the long term solution.

sfc-gh-melnacouzi (dismissed):

sfc-gh-turbaszek (approved):

src/snowflake/cli/_plugins/nativeapp/project_model.py
```diff
@@ -139,10 +136,13 @@ def project_identifier(self) -> str:
     @property
     def package_name(self) -> str:
         if self.definition.package and self.definition.package.name:
-            name = self.definition.package.name
+            return self.definition.package.name
```
sfc-gh-melnacouzi, line 139: we need to_identifier() back in that case.

src/snowflake/cli/_plugins/nativeapp/project_model.py
```diff
@@ -139,10 +136,13 @@ def project_identifier(self) -> str:
...
         else:
+            # V1.0 PDF doesn't support templating, so if the identifier isn't
+            # explicitly specified, the default is generated here,
+            # so we have to append the test resource suffix here
             name = default_app_package(self.project_identifier)
```
sfc-gh-melnacouzi, line 144: we need to_identifier here as well, because we don't use concat_identifiers() all the time anymore.

src/snowflake/cli/api/project/schemas/updatable_model.py
```diff
@@ -160,7 +160,8 @@ def validator_skipping_templated_str(cls, value, handler, info: ValidationInfo):
 
         setattr(
             cls,
-            f"_field_validator_with_verbose_name_to_avoid_name_conflict_{field_name}",
+            # Unique name so that subclasses get a unique instance of this validator
+            f"_{cls.__module__}_{cls.__name__}_validate_{field_name}",
```
sfc-gh-melnacouzi, line 164: nit - we can use `.` instead of `_` to separate module and name since __module__ will have . in it anyways.

tests_integration/nativeapp/test_debug_mode.py
```diff
@@ -12,19 +12,17 @@
...
-from snowflake.cli.api.project.util import generate_user_env
 from snowflake.connector.connection import SnowflakeConnection
 from snowflake.connector.errors import ProgrammingError
 
 from tests.project.fixtures import *
 from tests_integration.test_utils import pushd, enable_definition_v2_feature_flag
 
-USER_NAME = f"user_{uuid.uuid4().hex}"
-TEST_ENV = generate_user_env(USER_NAME)
+USER_NAME = os.environ.get("USER", "")
```
sfc-gh-melnacouzi, line 25: this might fail on Windows. I suggest we make it deterministic, and pass a non-blank test user.
If you don't want to pass USER every time, we can default to some default_user value, and pass it as a fixture to the test:
`{project_name}_{default_user}{resource_suffix}`

sfc-gh-fcampbell: removed entirely, `USER` is now automatically passed by the CLI runner unless the test supplies its own

src/snowflake/cli/api/project/schemas/updatable_model.py
sfc-gh-fcampbell: updated

src/snowflake/cli/_plugins/nativeapp/project_model.py
sfc-gh-fcampbell: `append_test_resource_suffix` now does `to_identifier` when it no-ops to make sure it always generates a valid identifier

src/snowflake/cli/_plugins/nativeapp/project_model.py
sfc-gh-fcampbell: added back

sfc-gh-bdufour (approved):

src/snowflake/cli/api/project/schemas/entities/application_entity_model.py
```diff
@@ -39,3 +41,21 @@ class ApplicationEntityModel(EntityModelBase):
...
+        cls, input_value: Identifier | str
+    ) -> Identifier | str:
+        identifier = (
+            input_value.name if isinstance(input_value, Identifier) else input_value
+        )
+        with_suffix = append_test_resource_suffix(identifier)
+        if isinstance(input_value, Identifier):
+            data = dict(name=with_suffix)
+            if input_value.schema_:
+                data["schema"] = input_value.schema_
```
sfc-gh-melnacouzi, line 57: any better way, not to have to remember all fields here? (scenario: someone adds a field in Identifier, but forgets to adjust here)?
- not sure if dumping the whole class dictionary or pydantic model and reconstructing it works (possibility 1)
- or we can introduce some kind of clone function within Identifier.

sfc-gh-bdufour: Can you explain what you mean here? Identifier is [[db.]schema.].resource in SQL. What would be added in the future?

src/snowflake/cli/_plugins/nativeapp/project_model.py
sfc-gh-bdufour: there is no path that doesn't go through `to_identifier` in this logic, you lost me.

sfc-gh-melnacouzi: custom validator uses the append_test_resource_suffix, which would turn 'a.b' application name to '"a.b"' application name even if we are not in test mode (env variable for SUFFIX is not set)

src/snowflake/cli/api/project/schemas/entities/application_entity_model.py
sfc-gh-melnacouzi: I am not sure - but this defeats encapsulation as we are manually cloning an object with all its fields, and if in the future that object evolves, it's easy to miss. It seems a bit like an anti-pattern.

src/snowflake/cli/_plugins/nativeapp/project_model.py
sfc-gh-bdufour: What's the possible negative side-effect? I thought we always have to quote the app/pkg names on use anyway.

sfc-gh-melnacouzi:
```
native_app:
  package:
    name: my.package
  application:
    name: <% ctx.native_app.package.name %>_app
```

app name will then turn out as '"my.package"_app'. 
I know they can use id_concat, but it's counter intuitive. It's ok to do it to defaults, but not when it's specified directly in pdf

src/snowflake/cli/api/project/schemas/entities/application_entity_model.py
sfc-gh-melnacouzi: Since Identifier is a Pydantic model, it's easy to do model_dump and then reload it.

src/snowflake/cli/api/project/schemas/entities/application_entity_model.py
```diff
@@ -39,3 +41,21 @@ class ApplicationEntityModel(EntityModelBase):
...
         default=None,
     )
+
+    @field_validator("identifier")
+    @classmethod
+    def append_test_resource_suffix_to_identifier(
+        cls, input_value: Identifier | str
+    ) -> Identifier | str:
+        identifier = (
+            input_value.name if isinstance(input_value, Identifier) else input_value
```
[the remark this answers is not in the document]
sfc-gh-bdufour: So something like:

```
identifier = Identifier(input_value)  # accepts str or Identifier
return identifier.with_name(append_test_resource_suffix(identifier.name))  # returns a new instance with the name replaced
```

?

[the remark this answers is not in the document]
sfc-gh-melnacouzi: I was thinking 2 separate validators (ordered). 
- It's just that conversion to Identifier might need to apply to all entity types (not only app and package)
- We have to think about templating as well - if users reference ctx.entities.x.identifier in case we modify the structure.

src/snowflake/cli/_plugins/nativeapp/project_model.py
sfc-gh-bdufour: yeah ok that's fair, thanks for explaining. @sfc-gh-fcampbell sounds like a regression we need to address

src/snowflake/cli/api/project/schemas/entities/application_entity_model.py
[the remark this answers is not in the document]
sfc-gh-melnacouzi: oh, yes that would work too. As long as we don't have to copy the fields one by one.

src/snowflake/cli/_plugins/nativeapp/project_model.py
sfc-gh-fcampbell: ah good point, i'll rework the appending part

src/snowflake/cli/api/project/util.py
```diff
@@ -251,3 +248,23 @@ def identifier_to_show_like_pattern(identifier: str) -> str:
...
+    """
+    Append a suffix that should be added to specified account-level resources.
+
+    This is an internal concern that is currently only used in tests
+    to isolate concurrent runs and to add the test name to resources.
+    """
+    suffix = os.environ.get(TEST_RESOURCE_SUFFIX_VAR, "")
+    if identifier_to_str(identifier).endswith(identifier_to_str(suffix)):
+        # If the suffix has already been added, don't add it again
+        return to_identifier(identifier)
```
sfc-gh-melnacouzi, line 263: This part was the issue I mentioned earlier. We shouldn't for it to_identifier here:
```
native_app:
  package:
    name: my.package
  application:
    name: <% ctx.native_app.package.name %>_app
```
should result in application name 'my.package_app', not '"my.package"_app'

sfc-gh-turbaszek (dismissed):

sfc-gh-melnacouzi (approved):

src/snowflake/cli/api/project/schemas/entities/application_entity_model.py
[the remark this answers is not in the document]
sfc-gh-fcampbell: Updated using `return input_value.model_copy(update=dict(name=with_suffix))`

sfc-gh-melnacouzi (approved):

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -370,17 +369,11 @@ def on_cycle_action(node: Node[TemplateVar]):
...
-            exclude_none=True,
-            warnings=False,
-            by_alias=True,
-            round_trip=True,
-        )
-    )
+    # Use the values originally provided by the user as the template context
+    # This intentionally doesn't reflect any field changes made by
+    # validators, to minimize user surprise when templating values
+    project_context[CONTEXT_KEY] = definition
```
sfc-gh-fcampbell, line 375: this undoes a previous change in this PR, restoring to the current behaviour from main

tests/api/utils/test_definition_rendering.py
```diff
@@ -717,7 +717,7 @@ def test_non_str_scalar_with_templates():
                 },
                 "application": {
                     "name": "test_app_username",
-                    "debug": True,
+                    "debug": "true",
```
sfc-gh-fcampbell, line 720: ditto

sfc-gh-turbaszek (approved):

PR #1442, opened 2024-08-13
Refactor app package deploy logic from NativeAppManager to ApplicationPackageEntity

sfc-gh-gbloom:
### Pre-review checklist

### Changes description
This is a preparation PR to enable `snow ws deploy` for application package entities in the [following PR](https://github.com/snowflakedb/snowflake-cli/pull/1461). The end goal is to reuse the same logic in both `NativeAppManager.deploy()` and `ApplicationPackageEntity`/`ApplicationEntity` actions.

#### Extracted methods

The first step is to extract the minimal logic out of `NativeAppManager` to enable `ApplicationPackageEntity.action_deploy()` (which will be very similar to `NativeAppManager.deploy()`). The following methods were moved from `NativeAppManager` and modified to be stateless:

`ApplicationPackageEntity`
- `get_existing_app_pkg_info()`
- `get_app_pkg_distribution_in_snowflake()`
- `verify_project_distribution()`
- `use_package_warehouse()`
- `apply_package_scripts()`
- `create_app_package()`

`utils`
- `generic_sql_error_handler()`
- `ensure_correct_owner()`
- `sync_deploy_root_with_stage()`
- `_get_stage_paths_to_sync()`
- `execute_post_deploy_hooks()`
- `_execute_sql_script()`
- `expand_script_templates()`

Even though they are now just proxies, the methods in `NativeAppManager` were kept with the same signatures in order to not break the interface of the class, as other flows call them and were not refactored yet (teardown, version, etc).

#### SqlExecutor

Added the `SqlExecutor` utility class to be able to execute SQL without having to inherit directly from `SqlExecutionMixin`.

### Manual tests

```
snow app deploy
snow app version create v1
snow app version list
snow app version drop v1
snow app teardown
snow app run
```


src/snowflake/cli/api/entities/application_package_entity.py
```diff
@@ -31,3 +34,20 @@ def action_bundle(self, ctx: ActionContext):
         compiler = NativeAppCompiler(bundle_context)
         compiler.compile_artifacts()
         return bundle_map
+
+    @classmethod
```
sfc-gh-bdufour, line 38: you don't need the cls param, consider making this static instead.

src/snowflake/cli/_plugins/nativeapp/manager.py
```diff
@@ -490,15 +493,11 @@ def get_existing_app_info(self) -> Optional[dict]:
...
-        Check for an existing application package by the same name as in project definition, in account.
-        It executes a 'show application packages like' query and returns the result as single row, if one exists.
-        """
-
-        with self.use_role(self.package_role):
-            return self.show_specific_object(
-                "application packages", self.package_name, name_col=NAME_COL
-            )
+        return ApplicationPackageEntity.get_existing_app_pkg_info(
+            sql_executor=self,
```
sfc-gh-bdufour, line 497: Do we even need an instance to execute the SQL commands? Would we be able to turn the mixin into a stateless utility instead? Or at least I'd try to extract a contract out of it (SqlExecutor?) and leave the mixin out of the equation (but the mixin implements the contract, clearly).

src/snowflake/cli/api/entities/application_package_entity.py
```diff
@@ -31,3 +34,20 @@ def action_bundle(self, ctx: ActionContext):
...
+    def get_existing_app_pkg_info(
+        cls,
+        sql_executor: SqlExecutionMixin,
+        package_name: str,
+        package_role: str,
```
sfc-gh-bdufour, line 43: We'll have to watch the number of arguments these functions eventually get, but I guess since both the entity logic and the manager logic have to invoke this, it's not worth defining a single contract, right?

src/snowflake/cli/api/entities/application_package_entity.py
sfc-gh-gbloom: Done. Some methods are `classmethod` in case they need to call other methods. The rest are now `staticmethod`.

src/snowflake/cli/_plugins/nativeapp/manager.py
sfc-gh-gbloom: Created `SqlExecutor` and a `get_sql_executor()` to avoid using `SqlExecutionMixin` directly and passing an sql_executor instance.

src/snowflake/cli/api/entities/application_package_entity.py
sfc-gh-gbloom: I think it turned out to be reasonable as it is

tests/nativeapp/test_teardown_processor.py
```diff
@@ -759,64 +761,70 @@ def test_drop_package_variable_mistmatch_w_special_comment_auto_drop(
...
+    mock_execute_na,
     auto_yes_param,
     special_comment,
     temp_dir,
     mock_cursor,
 ):
     mock_get_distribution.return_value = "internal"
 
-    side_effects, expected = mock_execute_helper(
+    side_effects_na, expected_na = mock_execute_helper(
```
sfc-gh-bdufour, line 773: why is this test changing so much? This should be a pure refactoring, no?

src/snowflake/cli/api/entities/common.py
```diff
@@ -39,3 +40,21 @@ def perform(self, action: EntityActions, action_ctx: ActionContext):
         Performs the requested action.
         """
         return getattr(self, action)(action_ctx)
+
+
+class SqlExecutor(SqlExecutionMixin):
```
sfc-gh-bdufour, line 45: the mixin has state, no? I'm very reluctant to introduce yet another stateful global var. We could either 1) replace the mixin everywhere, or 2) make the base executor stateless, and make the mixin reuse the base executor.

tests/nativeapp/test_teardown_processor.py
sfc-gh-gbloom: Yes, the logic didn't change. Since not all methods of `NativeAppManager` were refactored yet, the flow in this test executes SQL via both the mixin executor, and the new SQLExecutor. So instead of a single mock, we now need two of them, and expect different calls on each.

src/snowflake/cli/api/entities/common.py
sfc-gh-gbloom: Maybe it would be simpler to override and blank the stateful mixin behavior here? (it's basically a snowpark session)

sfc-gh-bdufour: Wouldn't flipping the inheritance hierarchy basically achieve the same thing? The advantage then is that we can phase out the mixin and delete it over time, no need to change the executor after that.

sfc-gh-gbloom: Yeah, you're right. I'll do it this way.

tests/nativeapp/test_teardown_processor.py
sfc-gh-bdufour: But then we lose fidelity here, since we can't tell if the calls happened in the right order across the mocks. Look into setting up a parent mock and using https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.attach_mock to make sure that we collect all calls across the mocks. This way, the test shouldn't need to be modified.

sfc-gh-gbloom: Nice suggestion! Combined both mocks to a single parent.

src/snowflake/cli/api/entities/common.py
```diff
@@ -39,3 +40,13 @@ def perform(self, action: EntityActions, action_ctx: ActionContext):
         Performs the requested action.
         """
         return getattr(self, action)(action_ctx)
+
+
+_SQL_EXECUTOR: SqlExecutor | None = None
```
sfc-gh-cgorrie, line 45: Because the SqlExecutor has a connection, the extra global is not desirable -- we would then need to refactor all of this code when we allow for scoped CLI context with ContextVar. Can you think of a way to approach this without an additional global, or alternatively piggyback on top of cli context?

sfc-gh-gbloom: There is an optional `SnowflakeConnection` constructor param. If it's not provided, the connection from CliGlobalContext will be used (which is also a singleton). Not sure I understand why it would require a larger refactor, can you explain?

We could also reuse a single `SqlExecutor` instance and provide it a connection (that's what I started with). But it required passing an extra param to each `ApplicationPackageEntity` method, and since the instance is pretty much stateless it was simpler to use a global singleton.

sfc-gh-gbloom: Discussed f2f - I'm removing the global variable, and `get_sql_executor()` will return a new `SqlExecutor` instance every time.

src/snowflake/cli/api/entities/common.py
```diff
@@ -39,3 +40,7 @@ def perform(self, action: EntityActions, action_ctx: ActionContext):
         Performs the requested action.
         """
         return getattr(self, action)(action_ctx)
+
+
+def get_sql_executor() -> SqlExecutor:
```
sfc-gh-cgorrie, line 45:
```suggestion
def get_sql_executor() -> SqlExecutor:
    """ Returns an SQL Executor that uses the connection from the current CLI context """
```

sfc-gh-gbloom: done

sfc-gh-cgorrie (dismissed):

sfc-gh-cgorrie (approved):

sfc-gh-turbaszek (approved):

PR #1443, opened 2024-08-13
Append SNOWFLAKE_CLI_TEST_RESOURCE_SUFFIX value to app and package name

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Prerequisite for #1437. After discussing in-person, we decided to not overwrite the `USER` environment variable in tests and instead add a new `SNOWFLAKE_CLI_TEST_RESOURCE_SUFFIX` env var that will be appended to app package and app identifiers. In #1437, we'll use this variable to generate unique names for these resources so they don't clash during concurrent tests.

### DIFF
--- a/src/snowflake/cli/_plugins/nativeapp/project_model.py
+++ b/src/snowflake/cli/_plugins/nativeapp/project_model.py
@@ -14,6 +14,7 @@
 
 from __future__ import annotations
 
+import os
 from functools import cached_property
 from pathlib import Path
 from typing import List, Optional
@@ -21,18 +22,22 @@ from typing import List, Optional
 from snowflake.cli._plugins.nativeapp.artifacts import resolve_without_follow
 from snowflake.cli._plugins.nativeapp.bundle_context import BundleContext
 from snowflake.cli.api.cli_global_context import get_cli_context
-from snowflake.cli.api.project.definition import (
-    default_app_package,
-    default_application,
-    default_role,
-)
+from snowflake.cli.api.project.definition import DEFAULT_USERNAME, default_role
 from snowflake.cli.api.project.schemas.native_app.application import (
     PostDeployHook,
 )
 from snowflake.cli.api.project.schemas.native_app.native_app import NativeApp
 from snowflake.cli.api.project.schemas.native_app.path_mapping import PathMapping
-from snowflake.cli.api.project.util import extract_schema, to_identifier
+from snowflake.cli.api.project.util import (
+    clean_identifier,
+    concat_identifiers,
+    extract_schema,
+    get_env_username,
+    to_identifier,
+)
 from snowflake.connector import DictCursor
+
+RESOURCE_SUFFIX_VAR = "SNOWFLAKE_CLI_TEST_RESOURCE_SUFFIX"
 
 
 def current_role() -> str:
@@ -129,12 +134,24 @@ class NativeAppProjectModel:
         # sometimes strip out double quotes, so we try to get them back here.
         return to_identifier(self.definition.name)
 
-    @cached_property
+    @property
     def package_name(self) -> str:
+        # If an explicit package name is set, use it and append the resource suffix
+        # In this case, if the suffix is empty we don't append the default suffix
+        # since we want to honor the user's chosen package name
+        suffix = resource_suffix()
         if self.definition.package and self.definition.package.name:
-            return to_identifier(self.definition.package.name)
-        else:
-            return to_identifier(default_app_package(self.project_identifier))
+            return concat_identifiers(
+                [to_identifier(self.definition.package.name), suffix]
+            )
+
+        # If there's no explicit package name set in the project definition,
+        # generate a name for the package from the project identifier and
+        # append the resource suffix
+        # If we don't have a resource suffix specified, use the default one
+        return concat_identifiers(
+            [self.project_identifier, "_pkg", suffix or default_resource_suffix()]
+        )
 
     @cached_property
     def package_role(self) -> str:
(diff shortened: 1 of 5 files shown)


src/snowflake/cli/_plugins/nativeapp/project_model.py
```diff
@@ -129,12 +134,13 @@ def project_identifier(self) -> str:
         # sometimes strip out double quotes, so we try to get them back here.
         return to_identifier(self.definition.name)
 
-    @cached_property
+    @property
```
sfc-gh-fcampbell, line 137: removed caching to avoid caching the suffix in tests, since it will contain a unique ID for each test

RELEASE-NOTES.md
```diff
@@ -31,6 +31,7 @@
   * Currently only supports SQL scripts: `post_deploy: [{sql_script: script.sql}]`
 * Added `snow spcs service execute-job` command, which supports creating and executing a job service in the current schema.
 * Added `snow app events` command to fetch logs and traces from local and customer app installations
+* Added support for `SNOWFLAKE_CLI_RESOURCE_SUFFIX` environment variable to be used as a suffix for Native App package and app names
```
sfc-gh-melnacouzi, line 34: not sure what's the common practice on this. Since this is purely internal for testing purposes, should we skip it from release notes?

src/snowflake/cli/_plugins/nativeapp/project_model.py
```diff
@@ -129,12 +134,15 @@ def project_identifier(self) -> str:
...
     def package_name(self) -> str:
+        suffix = resource_suffix()
         if self.definition.package and self.definition.package.name:
-            return to_identifier(self.definition.package.name)
+            return append_to_identifier(
+                to_identifier(self.definition.package.name), suffix
```
sfc-gh-melnacouzi, line 142: this is a bit tricky, because it assumes suffix is a safe identifier. 
concat_identifiers() might be what you're looking for.

RELEASE-NOTES.md
sfc-gh-fcampbell: i added it cause it could technically be used by end-users. I'm ok leaving it off if we want to consider this an undocumented feature

src/snowflake/cli/_plugins/nativeapp/project_model.py
sfc-gh-fcampbell: oh TIL, will update

RELEASE-NOTES.md
sfc-gh-bdufour: I think the variable is rather misleading since it seems to imply it will cause all resources to be suffixed, but that's not the case. I'd leave this to be undocumented since we're not guaranteeing a user can build workflows on it.

sfc-gh-bdufour: `SNOWFLAKE_CLI_TEST_RESOURCE_SUFFIX` might convey the intention here a little more clearly? We are not in any way saying this should be used in a production setting.

src/snowflake/cli/api/project/definition.py
```diff
@@ -102,16 +104,16 @@ def generate_local_override_yml(
     return project.update_from_dict(local)
 
 
-def default_app_package(project_name: str):
+def resource_suffix() -> str:
+    """A suffix that should be added to account-level resources."""
+    return os.environ.get(SNOWFLAKE_CLI_RESOURCE_SUFFIX_VAR, "")
```
sfc-gh-bdufour, line 109: Should this really be part of the public API? At least we should give it a very descriptive name, I normally worry about making non-prod logic sound so prod-like.

sfc-gh-fcampbell: based on our discussion earlier, it seemed to me that this would indeed be part of the public API (but undocumented). Let's PL this for tomorrow morning

src/snowflake/cli/api/project/definition.py
```diff
@@ -102,16 +102,6 @@ def generate_local_override_yml(
     return project.update_from_dict(local)
 
 
-def default_app_package(project_name: str):
-    user = clean_identifier(get_env_username() or DEFAULT_USERNAME)
-    return append_to_identifier(to_identifier(project_name), f"_pkg_{user}")
```
sfc-gh-fcampbell, line 107: merged into the body of `NativeAppProjectModel.package_name` since this was the only usage of this function (and we needed to separate the addition of the `_pkg` from the addition of the username)

src/snowflake/cli/api/project/definition.py
```diff
@@ -102,16 +102,6 @@ def generate_local_override_yml(
...
-
-
 def default_role():
     conn = get_cli_context().connection
     return conn.role
-
-
-def default_application(project_name: str):
-    user = clean_identifier(get_env_username() or DEFAULT_USERNAME)
-    return append_to_identifier(to_identifier(project_name), f"_{user}")
```
sfc-gh-fcampbell, line 117: merged into the body of `NativeAppProjectModel.app_name` since this was the only usage of this function (and we needed to separate the addition of the `_` from the addition of the username)

src/snowflake/cli/_plugins/nativeapp/project_model.py
```diff
@@ -129,12 +134,24 @@ def project_identifier(self) -> str:
...
+        # If an explicit package name is set, use it and append the resource suffix
+        # In this case, if the suffix is empty we don't append the default suffix
+        # since we want to honor the user's chosen package name
+        suffix = resource_suffix()
         if self.definition.package and self.definition.package.name:
-            return to_identifier(self.definition.package.name)
-        else:
-            return to_identifier(default_app_package(self.project_identifier))
+            return concat_identifiers(
+                [to_identifier(self.definition.package.name), suffix]
```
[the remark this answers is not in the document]
sfc-gh-fcampbell: yeah you're right, it was only needed when i was using `append_to_identifier`

src/snowflake/cli/_plugins/nativeapp/project_model.py
```diff
@@ -129,12 +134,24 @@ def project_identifier(self) -> str:
...
+            )
+
+        # If there's no explicit package name set in the project definition,
+        # generate a name for the package from the project identifier and
+        # append the resource suffix
+        # If we don't have a resource suffix specified, use the default one
+        return concat_identifiers(
```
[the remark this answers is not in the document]
sfc-gh-fcampbell: my original intention was to skip using `$USER` if `$SNOWFLAKE_CLI_TEST_RESOURCE_SUFFIX` is non-empty, but I can just make it a straight append too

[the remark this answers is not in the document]
sfc-gh-melnacouzi: I would vote to append it, that way we keep the rest of the logic intact, and keep this as pure suffix.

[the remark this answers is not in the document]
sfc-gh-fcampbell: sounds good, update

sfc-gh-melnacouzi (approved):

PR #1448, opened 2024-08-16
Update setup processor according to new result schema

sfc-gh-bdufour:
### Pre-review checklist

### Changes description

Updating the setup processor driver to match ToT in the snowflake.app package

### Testing

Since the snowflake.app package is still private, this was tested manually using an env var override. Integ and E2E tests will be added when the feature is closer to PrPr.

### DIFF
--- a/src/snowflake/cli/_plugins/nativeapp/codegen/compiler.py
+++ b/src/snowflake/cli/_plugins/nativeapp/codegen/compiler.py
@@ -14,6 +14,8 @@
 
 from __future__ import annotations
 
+import copy
+import re
 from typing import Dict, Optional
 
 from snowflake.cli._plugins.nativeapp.bundle_context import BundleContext
@@ -34,7 +36,7 @@ from snowflake.cli.api.project.schemas.native_app.path_mapping import (
 )
 
 SNOWPARK_PROCESSOR = "snowpark"
-NA_SETUP_PROCESSOR = "native-app-setup"
+NA_SETUP_PROCESSOR = "native app setup"
 
 _REGISTERED_PROCESSORS_BY_NAME = {
     SNOWPARK_PROCESSOR: SnowparkAnnotationProcessor,
@@ -110,7 +112,15 @@ class NativeAppCompiler:
             # No registered processor with the specified name
             return None
 
-        current_processor = processor_factory(self._bundle_ctx)
+        processor_ctx = copy.copy(self._bundle_ctx)
+        processor_subdirectory = re.sub(r"[^a-zA-Z0-9_$]", "_", processor_name)
+        processor_ctx.bundle_root = (
+            self._bundle_ctx.bundle_root / processor_subdirectory
+        )
+        processor_ctx.generated_root = (
+            self._bundle_ctx.generated_root / processor_subdirectory
+        )
+        current_processor = processor_factory(processor_ctx)
         self.cached_processors[processor_name] = current_processor
 
         return current_processor
--- a/src/snowflake/cli/_plugins/nativeapp/codegen/setup/native_app_setup_processor.py
+++ b/src/snowflake/cli/_plugins/nativeapp/codegen/setup/native_app_setup_processor.py
@@ -14,13 +14,20 @@
 
 from __future__ import annotations
 
+import filecmp
 import json
 import os.path
+import shutil
 from pathlib import Path
 from typing import List, Optional
 
+import yaml
 from click import ClickException
-from snowflake.cli._plugins.nativeapp.artifacts import BundleMap, find_setup_script_file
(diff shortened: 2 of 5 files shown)


src/snowflake/cli/_plugins/nativeapp/codegen/compiler.py
```diff
@@ -34,7 +36,7 @@
 )
 
 SNOWPARK_PROCESSOR = "snowpark"
-NA_SETUP_PROCESSOR = "native-app-setup"
+NA_SETUP_PROCESSOR = "native app setup"
```
sfc-gh-bdufour, line 39: Somehow I keep forgetting that natural strings like this are more natural in YAML.

src/snowflake/cli/_plugins/nativeapp/codegen/compiler.py
```diff
@@ -110,7 +112,15 @@ def _try_create_processor(
             # No registered processor with the specified name
             return None
 
-        current_processor = processor_factory(self._bundle_ctx)
+        processor_ctx = copy.copy(self._bundle_ctx)
+        processor_subdirectory = re.sub(r"[^a-zA-Z0-9_$]", "_", processor_name)
```
sfc-gh-bdufour, line 116: Changed the logic so that each processor uses its own subdirectory for its artifacts, enforced by the framework. Right now each processor was responsible for doing this, so I just centralized the same logic to facilitate the implementation of new processors.

src/snowflake/cli/_plugins/nativeapp/codegen/compiler.py
```diff
@@ -110,7 +112,15 @@ def _try_create_processor(
...
+        processor_ctx.bundle_root = (
+            self._bundle_ctx.bundle_root / processor_subdirectory
```
sfc-gh-melnacouzi, line 118: if each processor gets its own directory, how do we chain processors?

sfc-gh-bdufour: I only made that change for the bundle root (which is the processor's sandbox) and the generated directory, which is similarly processor-specific. The deploy root remains unchanged, as processors should always be reading and writing from/to the deploy root so they can be chained.

sfc-gh-melnacouzi (dismissed):

src/snowflake/cli/_plugins/nativeapp/codegen/setup/native_app_setup_processor.py
```diff
@@ -40,6 +46,16 @@
 DEFAULT_TIMEOUT = 30
 DRIVER_PATH = Path(__file__).parent / "setup_driver.py.source"
 
+log = logging.getLogger(__name__)
```
sfc-gh-pjafari, line 49: wish we had named this `logger` instead of `log` everywhere :')

sfc-gh-pjafari (approved):

src/snowflake/cli/_plugins/nativeapp/codegen/setup/native_app_setup_processor.py
```diff
@@ -40,6 +46,16 @@
...
+
+
+def safe_set(d: dict, *keys: str, **kwargs) -> None:
```
sfc-gh-cgorrie, line 52: I could use a docstring here to understand the intention

sfc-gh-bdufour: fair, done

src/snowflake/cli/_plugins/nativeapp/codegen/setup/native_app_setup_processor.py
```diff
@@ -102,56 +171,68 @@ def _execute_in_sandbox(self, py_files: List[Path]) -> dict:
...
+        with self.edit_file(manifest_path) as f:
+            manifest = yaml.safe_load(f.contents)
+
+            for mod in modifications:
+                for inst in mod.get("instructions", []):
+                    if inst.get("type") == "set":
+                        payload = inst.get("payload")
+                        if payload:
+                            key = payload.get("key")
+                            value = payload.get("value")
```
sfc-gh-melnacouzi, line 215: nit: too much nesting here, would be nice to factor this out if possible.

sfc-gh-melnacouzi (approved):

src/snowflake/cli/_plugins/nativeapp/codegen/setup/native_app_setup_processor.py
sfc-gh-bdufour: sure, I'll keep this in mind when I rework this logic in the near future.

PR #1474, opened 2024-08-22
Added `snow ws migrate` command

sfc-gh-astus:
### Pre-review checklist

### Changes description
Added `snow ws migrate` command that allows to migrate `snowflake.yml` in current directory from V1 to V2.


src/snowflake/cli/_app/cli_app.py
```diff
@@ -234,10 +231,12 @@ def default(
...
-        )
 
     return app
+
+
+# setup_pycharm_remote_debugger_if_provided(
+#     pycharm_debug_library_path="pycharm_debug_library_path",
+#     pycharm_debug_server_host="localhost",
+#     pycharm_debug_server_port=12345,
+# )
```
sfc-gh-turbaszek, line 242: Leftover

sfc-gh-astus: 😮

sfc-gh-astus: Reverted

tests/test_data/projects/migration_streamlit_V2/snowflake.yml
```diff
@@ -0,0 +1,15 @@
+definition_version: 2
```
sfc-gh-turbaszek, line 1: Please add test for migrating snowflake.yml with templates. Templates should not be rendered in outcome file.

sfc-gh-jsikorski: Done

tests/test_data/projects/streamlit_templated_v1/template.yml
```diff
@@ -0,0 +1,12 @@
+files_to_render:
+  - snowflake.yml
+variables:
```
sfc-gh-turbaszek, line 3: Why we need project template for tests?

tests/workspace/__snapshots__/test_manager.ambr
```diff
@@ -0,0 +1,348 @@
...
+        signature: ""
+        returns: string
+  
+  '''
+# ---
+# name: test_if_template_is_not_rendered_during_migration[streamlit_templated_v1]
+  '''
+  definition_version: '2'
+  entities:
+    <! name | to_snowflake_identifier !>:
```
sfc-gh-turbaszek, line 89: We don't support templates as keys. This should be migrated to 
```
entity_1:
  identifier:
    name: <! name | to_snowflake_identifier !>
```

src/snowflake/cli/api/project/definition.py
```diff
@@ -58,15 +61,20 @@ def _get_merged_definitions(paths: List[Path]) -> Optional[Definition]:
...
     Loads project definition, optionally overriding values. Definition values
     are merged in left-to-right order (increasing precedence).
     Templating is also applied after the merging process.
     """
     merged_definitions = _get_merged_definitions(paths)
-    return render_definition_template(merged_definitions, context_overrides or {})
+    if render_env_values:
+        return render_definition_template(merged_definitions, context_overrides or {})
+    else:
+        return raw_project_properties(merged_definitions)
```
sfc-gh-turbaszek, line 77: @sfc-gh-melnacouzi WDYT?

tests/test_data/projects/snowpark_templated_v1/template.yml
```diff
@@ -0,0 +1,11 @@
+files_to_render:
```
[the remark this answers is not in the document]
sfc-gh-jsikorski: Removed

tests/workspace/__snapshots__/test_manager.ambr
sfc-gh-jsikorski: Changed

tests/test_data/projects/streamlit_templated_v1/template.yml
sfc-gh-jsikorski: Changed

src/snowflake/cli/api/project/definition.py
sfc-gh-jsikorski: To add a bit of a context: this was added, as Tomasz pointed out we need to test if templated project files are migrated unchanged from v1 to v2. 
The default behaviour, while loading project file, was to resolve any enviroment variable references with values. This solution is aimed to allow `DefinitionManager` to load and return project definition without processing envs.
As the default value for `render_env_values` is set to True, there should be no change in the function behaviour if called the old way.

src/snowflake/cli/_plugins/workspace/commands.py
```diff
@@ -15,22 +15,53 @@
...
+        pd_v2 = migrate_v1_snowpark_to_v2(pd)
+    else:
+        raise ValueError(
+            "Only Snowpark and Streamlit entities are supported for migration."
+        )
+
+    SecurePath("snowflake.yml").rename("snowflake_V1.yml")
+    with open("snowflake.yml", "w") as file:
+        yaml.dump(
+            pd_v2.model_dump(exclude_unset=True, exclude_none=True, mode="json"), file
```
sfc-gh-melnacouzi, line 59: You probably need by_alias=true, otherwise aliases won't work properly.

src/snowflake/cli/_plugins/snowpark/commands.py
```diff
@@ -463,12 +463,18 @@ def _migrate_v1_snowpark_to_v2(pd: ProjectDefinition):
...
 
-    for entity in [*pd.snowpark.procedures, *pd.snowpark.functions]:
+    for index, entity in enumerate([*pd.snowpark.procedures, *pd.snowpark.functions]):
         identifier = {"name": entity.name}
         if entity.database is not None:
             identifier["database"] = entity.database
         if entity.schema_name is not None:
             identifier["schema"] = entity.schema_name
+
+        if entity.name.startswith("<!") and entity.name.endswith("!>"):
```
sfc-gh-melnacouzi, line 473: I don't get this part. What is it trying to do?

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -386,3 +386,10 @@ def on_cycle_action(node: Node[TemplateVar]):
         default_env=project_context[CONTEXT_KEY].get("env"), override_env=override_env
     )
     return ProjectProperties(project_definition, project_context)
+
+
+def raw_project_properties(definition: Definition) -> ProjectProperties:
+    """
+    Returns the raw project definition data without any templating.
+    """
+    return ProjectProperties(build_project_definition(**definition), {})
```
sfc-gh-melnacouzi, line 395: why is the context empty? This will break all rendering in sql or files

src/snowflake/cli/api/project/definition.py
sfc-gh-melnacouzi: It's not only env. Let's change the name of the boolean to not use "env". Maybe use "render_templates".

src/snowflake/cli/_plugins/workspace/commands.py
```diff
@@ -15,22 +15,53 @@
...
 )
 
 
-@ws.command(requires_connection=True)
+@ws.command()
+def migrate(
+    **options,
+):
+    """Migrates the Snowpark and Streamlit project definition files form V1 to V2."""
+    pd = DefinitionManager().unrendered_project_definition
```
sfc-gh-melnacouzi, line 42: I understand that we are trying to preserve env variables from being expanded, but that's not the only rendering we have. You could refer to variables like `<% ctx.snowpark.project_name %>` anywhere in the project definition file. This will not work anymore in v2 because it's mapped to a different location.

We should at least warn about these templates, because they won't work anymore in v2.

src/snowflake/cli/_plugins/workspace/commands.py
```diff
@@ -15,22 +15,53 @@
...
+
+    SecurePath("snowflake.yml").rename("snowflake_V1.yml")
+    with open("snowflake.yml", "w") as file:
+        yaml.dump(
+            pd_v2.model_dump(exclude_unset=True, exclude_none=True, mode="json"), file
+        )
+    return MessageResult("Project definition migrated to version 2.")
+
+
+@ws.command(requires_connection=True, hidden=True)
```
sfc-gh-melnacouzi, line 64: are we hiding this command on purpose?

src/snowflake/cli/_plugins/snowpark/commands.py
sfc-gh-jsikorski: Solves: https://github.com/snowflakedb/snowflake-cli/pull/1474#discussion_r1731180926

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-jsikorski: But we don`t need to render anything - we're just converting `snowflake.yml`

src/snowflake/cli/_plugins/workspace/commands.py
sfc-gh-jsikorski: @sfc-gh-astus - it's your addition

src/snowflake/cli/_plugins/snowpark/commands.py
sfc-gh-melnacouzi: I am sorry- i need more context. Why is this `<! !>` logic leaking to here? isn't that used for initializing templates? except init command, nothing should be aware of `<! !>` templates.

src/snowflake/cli/_plugins/workspace/commands.py
sfc-gh-melnacouzi: I am not sure. Conversion of these is not straightforward. You would might need a map of their previous location to their new location in pdf v2... it's not something that can quickly be done.

sfc-gh-melnacouzi: If you want to get this out quickly, I suggest:
1) either error out on templating or if templating is used outside of referencing ctx.env stuff...
2) or expand the template first, and provide the expanded template as pdf v2

src/snowflake/cli/_plugins/workspace/commands.py
sfc-gh-jsikorski: Done

src/snowflake/cli/_plugins/snowpark/commands.py
sfc-gh-jsikorski: It s caused by the need to use on of the fields of the V1 entity as a key in V2 entities dict. We don't support the template as the key. So if the field we're using for the key is templated, we change it.

src/snowflake/cli/_plugins/workspace/commands.py
sfc-gh-jsikorski: I went for option 1. Attempt to migrate templated `snowflake.yml` causes error. User has the option to manually override this. In that case, we will log warning, stating that there is no guarantee the migrated file will work.

src/snowflake/cli/_plugins/workspace/commands.py
sfc-gh-astus: Yes.

src/snowflake/cli/_plugins/snowpark/commands.py
sfc-gh-turbaszek: We are mixing snowflake.yml templating and templates for `snow init`. We should only focus on snowflake.yml templating thus `<% ... %>`

src/snowflake/cli/_plugins/streamlit/commands.py
```diff
@@ -204,10 +204,15 @@ def _migrate_v1_streamlit_to_v2(pd: ProjectDefinition):
     if pd.streamlit.database:
         identifier["database"] = pd.streamlit.database
 
+    if pd.streamlit.name.startswith("<!") and pd.streamlit.name.endswith("!>"):
```
sfc-gh-turbaszek, line 207: Same here, why we use `snow init` templates?

sfc-gh-jsikorski: My bad, it was meant to be `"<%"`

src/snowflake/cli/_plugins/workspace/commands.py
```diff
@@ -14,23 +14,73 @@
...
+):
+    """Migrates the Snowpark and Streamlit project definition files form V1 to V2."""
+    pd = DefinitionManager().unrendered_project_definition
+
+    if pd.meets_version_requirement("2"):
+        return MessageResult("Project definition is already at version 2.")
+
+    if "<% ctx." in str(pd):
+        if not accept_templates:
+            raise ValueError(
```
sfc-gh-turbaszek, line 55:
```suggestion
            raise ClickException(
```
This will make the result prettier

sfc-gh-turbaszek (approved):

src/snowflake/cli/_plugins/snowpark/commands.py
sfc-gh-melnacouzi: keys are not allowed to be templated.

src/snowflake/cli/_plugins/workspace/commands.py
```diff
@@ -14,23 +14,74 @@
...
+    ),
+    **options,
+):
+    """Migrates the Snowpark and Streamlit project definition files form V1 to V2."""
+    pd = DefinitionManager().unrendered_project_definition
+
+    if pd.meets_version_requirement("2"):
+        return MessageResult("Project definition is already at version 2.")
+
+    if "<% ctx." in str(pd):
```
sfc-gh-melnacouzi, line 54: that is not sufficient. Space is optional, and it doesn't have to start with ctx. we can have templating functions etc... Checking for "<%" is more accurate.

src/snowflake/cli/_plugins/workspace/commands.py
```diff
@@ -14,23 +14,74 @@
...
+        if not accept_templates:
+            raise ClickException(
+                "Project definition contains templates. They may not be migrated correctly, and require manual migration."
+                "You can try again with --accept-templates  option, to attempt automatic migration."
+            )
+        log.warning(
+            "Your V1 definition contains templates. We cannot guarantee the correctness of the migration."
```
sfc-gh-melnacouzi, line 61: Some things we can add, not verbatim, but along these lines:
- Update any variables referencing the project definition file to point to the new location.

PR #1475, opened 2024-08-22
SNOW-1621834 Cast version to identifier when creating/dropping app versions

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
When running `snow app version create` and `snow app version drop`, wrap the version in `to_identifier()` so users don't have to specify the quotes around version names that aren't valid identifiers. If the name is already quoted, `to_identifier()` doesn't do anything.

### DIFF
--- a/src/snowflake/cli/_plugins/nativeapp/version/version_processor.py
+++ b/src/snowflake/cli/_plugins/nativeapp/version/version_processor.py
@@ -39,10 +39,9 @@ from snowflake.cli._plugins.nativeapp.run_processor import NativeAppRunProcessor
 from snowflake.cli.api.console import cli_console as cc
 from snowflake.cli.api.exceptions import SnowflakeSQLExecutionError
 from snowflake.cli.api.project.schemas.native_app.native_app import NativeApp
-from snowflake.cli.api.project.util import unquote_identifier
+from snowflake.cli.api.project.util import to_identifier, unquote_identifier
 from snowflake.cli.api.utils.cursor import (
     find_all_rows,
-    find_first_row,
 )
 from snowflake.connector import ProgrammingError
 from snowflake.connector.cursor import DictCursor
@@ -121,6 +120,8 @@ class NativeAppVersionCreateProcessor(NativeAppRunProcessor):
         """
         Defines a new version in an existing application package.
         """
+        # Make the version a valid identifier, adding quotes if necessary
+        version = to_identifier(version)
         with self.use_role(self.package_role):
             cc.step(
                 f"Defining a new version {version} in application package {self.package_name}"
@@ -141,6 +142,8 @@ class NativeAppVersionCreateProcessor(NativeAppRunProcessor):
         """
         Add a new patch, optionally a custom one, to an existing version in an application package.
         """
+        # Make the version a valid identifier, adding quotes if necessary
+        version = to_identifier(version)
         with self.use_role(self.package_role):
             cc.step(
                 f"Adding new patch to version {version} defined in application package {self.package_name}"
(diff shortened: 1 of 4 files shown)


src/snowflake/cli/_plugins/nativeapp/version/version_processor.py
```diff
@@ -156,11 +159,7 @@ def add_new_patch_to_version(self, version: str, patch: Optional[int] = None):
                 add_version_query, cursor_class=DictCursor
             )
 
-            show_row = find_first_row(
-                result_cursor,
-                lambda row: row[VERSION_COL] == unquote_identifier(version),
-            )
```
sfc-gh-fcampbell, line 162: we don't actually need this, the query only returns one row

tests/nativeapp/test_version_drop_processor.py
```diff
@@ -96,13 +96,13 @@ def test_process_has_no_existing_app_pkg(mock_get_existing, policy_param, temp_d
 def test_process_no_version_from_user_no_version_in_manifest(
     mock_version_info_in_manifest,
     mock_build_bundle,
-    mock_mismatch,
```
sfc-gh-fcampbell, line 99: not sure why this was named that

tests/nativeapp/test_version_drop_processor.py
```diff
@@ -96,13 +96,13 @@ def test_process_has_no_existing_app_pkg(mock_get_existing, policy_param, temp_d
...
+    mock_distribution,
     mock_get_existing,
     policy_param,
     temp_dir,
 ):
 
-    mock_mismatch.return_Value = "internal"
```
sfc-gh-fcampbell, line 105: this didn't break the test 😅

src/snowflake/cli/_plugins/nativeapp/version/version_processor.py
sfc-gh-cgorrie: We do need it. Because of how the query is made (with like), we can return multiple rows that have the "same" identifier after the like rules are applied. So we need to filter back down to the one we're actually dealing with.

I think `"Name"` and `name` is one such example.

src/snowflake/cli/_plugins/nativeapp/version/version_processor.py
```diff
@@ -328,6 +327,9 @@ def process(
                     "Manifest.yml file does not contain a value for the version field."
                 )
 
+        # Make the version a valid identifier, adding quotes if necessary
+        version = to_identifier(version)
```
sfc-gh-melnacouzi, line 331: any way we can unify this to get_version() in manager or processor so we don't have to remember to do it every time, similar to get_app() or get_package() ?

src/snowflake/cli/_plugins/nativeapp/version/version_processor.py
sfc-gh-fcampbell: the query here is
```python
                    alter application package {self.package_name}
                        add patch {patch if patch else ""} for version {version}
                        using @{self.stage_fqn}
```
I don't see how that issue could happen here

src/snowflake/cli/_plugins/nativeapp/version/version_processor.py
sfc-gh-fcampbell: this also needs to handle the string passed in by the user, it's not part of the project model. I could change this to a CLI callback that casts it at that point though

src/snowflake/cli/_plugins/nativeapp/version/version_processor.py
sfc-gh-cgorrie: Oh sorry, drive-by review. Let me look into it a bit more closely.

sfc-gh-cgorrie: Yep, confirmed, even when there are multiple verisons only the one in question is returned.

sfc-gh-melnacouzi (approved):

PR #1484, opened 2024-08-25
Add mixins

sfc-gh-jsikorski:
### Pre-review checklist

### Changes description
Added mixins to project definition v20, allowing user to add common values to selected entities


src/snowflake/cli/api/project/schemas/project_definition.py
```diff
@@ -194,6 +190,31 @@ def _validate_target_field(
...
+        default=None,
+    )
+
+    @model_validator(mode="before")
+    @classmethod
+    def apply_mixins(cls, data: Dict) -> Dict:
+        """
+        Applies mixins to those entities, whose meta field contains the mixin name.
+        """
+        if "mixins" in data and "entities" in data:
```
sfc-gh-turbaszek, line 204:
```suggestion
        if "mixins" not in data or "entities" not in data:
           return data
```

Less indentation later

src/snowflake/cli/api/project/schemas/project_definition.py
```diff
@@ -194,6 +190,31 @@ def _validate_target_field(
...
+            for entity in data["entities"].values():
+                entity_mixins = entity_mixins_to_list(
+                    entity.get("meta", {}).get("use_mixins")
+                )
+                entity_fields = get_allowed_fields_for_entity(entity)
+                if entity_fields and entity_mixins:
+                    for mixin_name in entity_mixins:
+                        if mixin_name in data["mixins"]:
```
sfc-gh-turbaszek, line 212: Should we raise an error if undefined mixin is used?

src/snowflake/cli/api/project/schemas/project_definition.py
```diff
@@ -194,6 +190,31 @@ def _validate_target_field(
...
+                            for key, value in data["mixins"][mixin_name].items():
+                                if key in entity_fields:
+                                    entity[key] = value
```
sfc-gh-turbaszek, line 215: Will this rise an error if we try to set up unknown field?

src/snowflake/cli/api/project/schemas/project_definition.py
```diff
@@ -222,3 +243,26 @@ def get_version_map():
     if FeatureFlag.ENABLE_PROJECT_DEFINITION_V2.is_enabled():
         version_map["2"] = DefinitionV20
     return version_map
+
+
+def entity_mixins_to_list(entity_mixins: Optional[str | List[str]]) -> List[str]:
```
sfc-gh-turbaszek, line 248: Personally I would make mixins list for simplicity sake.

src/snowflake/cli/api/project/schemas/project_definition.py
```diff
@@ -194,6 +190,31 @@ def _validate_target_field(
         default=None,
     )
 
+    mixins: Optional[Dict[str, Dict]] = Field(
```
sfc-gh-jsikorski, line 193: Currently i use plain dict, as mixins are used before validation, when we manage the raw dict data. 
It's simple, but any misspellings will pas unnoticed.
Another option is, as we have to include any posible field in every entity, to create a class in runtime.  We can get all entites,  extract their fields and create a pydantic class with all those fields set to optional. 
It would allow us to spot any mistaken key in mixins.

tests/test_data/projects/mixins_basic/snowflake.yml
```diff
@@ -0,0 +1,28 @@
...
+    identifier: name
+    returns: string
+    signature:
+    - name: name
+      type: string
+    stage: bar
+    type: function
+mixins:
+  my_mixin:
+    stage: foo
```
sfc-gh-turbaszek, line 28: Do we need that many projects? Would it be possible to use one, complex to define all neccessary tests?

src/snowflake/cli/api/project/schemas/project_definition.py
sfc-gh-jsikorski: Field validator sounds ok, but we apply mixins well before MetaFields are validated. So either we go by design, and check for type, or simplify it and force a list (that can be misleading and pesky, if user has to enter a single mixin as a list).

src/snowflake/cli/api/project/schemas/project_definition.py
sfc-gh-jsikorski: A single mixin can be used to inject data into different entities, this would need some additional logic- possible solution is provided here https://github.com/snowflakedb/snowflake-cli/pull/1484#discussion_r1730784108

src/snowflake/cli/api/project/schemas/project_definition.py
sfc-gh-jsikorski: Done

src/snowflake/cli/api/project/schemas/project_definition.py
sfc-gh-jsikorski: Done

sfc-gh-turbaszek (approved):

Issue #1486, opened 2024-08-26
SNOW-1639627: snow init Fails With Error


sfc-gh-cwiese:
### SnowCLI version

Snowflake CLI version: 2.7.0

### Python version

Python 3.11.9

### Platform

macOS-14.6.1-arm64-arm-64bit

### What happened

`snow init foo` fails with error `Template does not have template.yml file.`. `snow snowpark init foo` succeeds.

### Console output

```Shell
(snow_globe) cwiese@LQ7YYTJ905 snow_globe % snow init foo         
╭─ Error ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Template does not have template.yml file.                                                                                                                                                        │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
(snow_globe) cwiese@LQ7YYTJ905 snow_globe % snow snowpark init foo
DeprecationWarning: The command 'init' is deprecated.
Initialized the new project in foo/
```

### How to reproduce

`snow init foo`

sfc-gh-pczajka:
@sfc-gh-cwiese please use `snow init foo --template example_snowpark`

sfc-gh-cwiese:
@sfc-gh-pczajka that works, however `--template` is not mentioned as a required option in the `--help` output:
```
(snow_globe) cwiese@LQ7YYTJ905 snow_globe % snow init --help
                                                                                                                                                                                                    
 Usage: snow init [OPTIONS] PATH                                                                                                                                                                    
                                                                                                                                                                                                    
 Creates project directory from template.                                                                                                                                                           
                                                                                                                                                                                                    
╭─ Arguments ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ *    path      TEXT  Directory to be initialized with the project. This directory must not already exist [required]                                                                              │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Options ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --template                 TEXT  which template (subdirectory of --template-source) should be used. If not provided, whole source will be used as the template.                                  │
[...]

PR #1488, opened 2024-08-26
More actionable error message for snow init

sfc-gh-pczajka:
### Pre-review checklist

### Changes description
Fixes misleading error message when `snow init` is called with no arguments: https://github.com/snowflakedb/snowflake-cli/issues/1486

### DIFF
--- a/src/snowflake/cli/_plugins/init/commands.py
+++ b/src/snowflake/cli/_plugins/init/commands.py
@@ -132,13 +132,13 @@ def _fetch_remote_template(
     return template_root
 
 
-def _read_template_metadata(template_root: SecurePath) -> Template:
+def _read_template_metadata(template_root: SecurePath, args_error_msg: str) -> Template:
     """Parse template.yml file."""
     template_metadata_path = template_root / TEMPLATE_METADATA_FILE_NAME
     log.debug("Reading template metadata from %s", template_metadata_path.path)
     if not template_metadata_path.exists():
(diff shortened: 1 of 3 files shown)


src/snowflake/cli/_plugins/init/commands.py
```diff
@@ -203,6 +203,9 @@ def init(
     is_remote = any(
         template_source.startswith(prefix) for prefix in ["git@", "http://", "https://"]  # type: ignore
     )
+    args_error_msg = (
+        "Check whether --template and --template_source arguments are correct."
+    )
```
sfc-gh-jsikorski, line 208: Why is it created in command and passed to function as argument?

sfc-gh-pczajka: I wanted to keep it close to actual arguments definition - in case their names will change, no one will remember to edit an error message in function 150 lines above

sfc-gh-jsikorski: So maybe something like:
```
    args_error_msg = (
        "Check whether {TemplateOption.param_decls} and {SourceOption.param_decls} arguments are correct."
    )
```
This should update the message automagically

sfc-gh-pczajka: Neat!
Changed

sfc-gh-jsikorski (approved):

PR #1499, opened 2024-08-27
Skip empty pages directory when deploying streamlit

sfc-gh-astus:
### Pre-review checklist

### Changes description
Added check in Streamlit to skip put on stage when directory is empty.

### DIFF
--- a/src/snowflake/cli/_plugins/streamlit/manager.py
+++ b/src/snowflake/cli/_plugins/streamlit/manager.py
@@ -57,6 +57,10 @@ class StreamlitManager(SqlExecutionMixin):
         stage_manager = StageManager()
         for file in artifacts:
             if file.is_dir():
+                if not any(file.iterdir()):
+                    cli_console.warning(f"Skipping empty directory: {file}")
+                    continue
+
                 stage_manager.put(
                     f"{file.joinpath('*')}", f"{root_location}/{file}", 4, True
                 )
--- a/tests/streamlit/test_commands.py
+++ b/tests/streamlit/test_commands.py
@@ -157,6 +157,58 @@ def test_deploy_only_streamlit_file_no_stage(
 @mock.patch("snowflake.cli._plugins.connection.util.get_account")
 @mock.patch("snowflake.cli._plugins.streamlit.commands.typer")
 @mock.patch("snowflake.connector.connect")
+def test_deploy_with_empty_pages(
+    mock_connector,
+    mock_typer,
+    mock_get_account,
+    mock_cursor,
+    runner,
+    mock_ctx,
+    project_directory,
+):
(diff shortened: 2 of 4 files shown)


src/snowflake/cli/_plugins/streamlit/manager.py
```diff
@@ -57,6 +58,10 @@ def _put_streamlit_files(
         stage_manager = StageManager()
         for file in artifacts:
             if file.is_dir():
+                if not bool(os.listdir(file)):
+                    cli_console.warning(f"Skipping empty directory: {file}")
+                    continue
```
sfc-gh-astus, line 63: Maybe I should add it in `StageManager.put`? WDYT?

src/snowflake/cli/_plugins/streamlit/manager.py
```diff
@@ -57,6 +58,10 @@ def _put_streamlit_files(
         stage_manager = StageManager()
         for file in artifacts:
             if file.is_dir():
+                if not bool(os.listdir(file)):
```
sfc-gh-pczajka, line 61:
```suggestion
                if not any(file.iterdir()):
```
To avoid additional `os` import

tests/streamlit/test_commands.py
```diff
@@ -154,6 +155,58 @@ def test_deploy_only_streamlit_file_no_stage(
...
+                {"CURRENT_ACCOUNT_NAME()": "my_account"},
+            ],
+            columns=["SYSTEM$GET_SNOWSIGHT_HOST()"],
+        )
+    )
+    mock_connector.return_value = ctx
+    mock_get_account.return_value = "my_account"
+
+    with project_directory("streamlit_empty_pages") as directory:
+        os.makedirs(directory / "pages", exist_ok=True)
```
sfc-gh-pczajka, line 184:
```suggestion
        (directory / "pages").mkdir(parents=True, exist_ok=True)
```

src/snowflake/cli/_plugins/streamlit/manager.py
sfc-gh-jsikorski: IMHO it belongs here- it's streamlit logic, not stage put logic- it's the Streamlit manager that should decide what files and dirs to put, and Stage manager should know how to handle them to put'em on stage.

src/snowflake/cli/_plugins/streamlit/manager.py
sfc-gh-jsikorski, line 61: You don't need bool() here. Empty list evaluates as false

sfc-gh-astus: Changed to `any(file.iterdir())`.

PR #1506, opened 2024-08-28
Renamed private-key-path to private-key-file

sfc-gh-astus:
### Pre-review checklist

### Changes description
Renamed private-key-path to private-key-file.
Added private-key-path as alias to keep backward compatibility.


src/snowflake/cli/api/commands/flags.py
```diff
@@ -134,11 +134,12 @@ def _password_callback(value: str):
 
 PrivateKeyPathOption = typer.Option(
     None,
+    "--private-key-file",
     "--private-key-path",
-    help="Snowflake private key path. Overrides the value specified for the connection.",
+    help="Snowflake private key file path. Overrides the value specified for the connection.",
     hide_input=True,
     callback=_callback(
```
sfc-gh-turbaszek, line 141: Should we add deprecation info in callback?

sfc-gh-astus: We don't know which flag was used `"--private-key-path"` is an alias. I tried to add context to callback function but failed.

sfc-gh-turbaszek (approved):

PR #1507, opened 2024-08-28
Workspaces post deploy hooks

sfc-gh-gbloom:
### Pre-review checklist

### Changes description

* Implementing `post_deploy` section in definition v2 for `ApplicationPackageEntity`.
* Added a default warehouse value to `ActionContext` that is read from the connection (matching the functionality in `snow app`).

### DIFF
--- a/src/snowflake/cli/api/entities/application_package_entity.py
+++ b/src/snowflake/cli/api/entities/application_package_entity.py
@@ -26,6 +26,7 @@ from snowflake.cli.api.console.abc import AbstractConsole
 from snowflake.cli.api.entities.common import EntityBase, get_sql_executor
 from snowflake.cli.api.entities.utils import (
     ensure_correct_owner,
+    execute_post_deploy_hooks,
     generic_sql_error_handler,
     render_script_templates,
     sync_deploy_root_with_stage,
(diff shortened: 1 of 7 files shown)


src/snowflake/cli/api/entities/application_package_entity.py
```diff
@@ -111,7 +113,14 @@ def action_deploy(
...
             )
 
-        # TODO Execute post-deploy hooks
+        if model.meta and model.meta.post_deploy:
+            self.execute_post_deploy_hooks(
+                console=ctx.console,
+                project_root=ctx.project_root,
+                post_deploy_hooks=model.meta.post_deploy,
+                package_name=package_name,
+                package_warehouse=model.meta.warehouse or ctx.default_warehouse,
```
sfc-gh-melnacouzi, line 122: is that equivalent to self.package.warehouse?

sfc-gh-melnacouzi (approved):

src/snowflake/cli/api/entities/application_package_entity.py
sfc-gh-gbloom: Yes, it points to `NativeAppProjectModel.package_warehouse` which has [the same logic](https://github.com/snowflakedb/snowflake-cli/blob/main/src/snowflake/cli/_plugins/nativeapp/project_model.py#L101-L109)

PR #1509, opened 2024-08-28
Improve error messages on SQL execution

sfc-gh-melnacouzi:
### Pre-review checklist

### Changes description
By supporting old and new templating syntax, we warn if both old syntax is used, and we error out if both old and new syntax is used at the same time in the same file.

The error messaging for these does not say which file is coming from.
In this PR I do some small refactor while adding the name of the file in the messaging.

I have already added this messaging to templating arbitrary files - this expands it to post_deploy scripts.

### DIFF
--- a/src/snowflake/cli/api/entities/utils.py
+++ b/src/snowflake/cli/api/entities/utils.py
@@ -1,7 +1,7 @@
 import os
 from pathlib import Path
 from textwrap import dedent
-from typing import Any, Callable, Dict, List, NoReturn, Optional
+from typing import Any, List, NoReturn, Optional
 
 import jinja2
 from click import ClickException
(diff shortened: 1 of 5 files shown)


src/snowflake/cli/api/entities/utils.py
```diff
@@ -306,7 +307,10 @@ def render_script_templates(
         script_full_path = SecurePath(project_root) / relpath
         try:
             template_content = script_full_path.read_text(file_size_limit_mb=UNLIMITED)
-            result = render_from_str(template_content, jinja_context)
+            env = override_env or choose_sql_jinja_env_based_on_template_syntax(
+                template_content, reference_name=relpath
+            )
+            result = env.from_string(template_content).render(jinja_context)
```
sfc-gh-pjafari, line 313: Would it make sense to always pass env in to `render_script_template` instead of having it as an optional override?

sfc-gh-pjafari (approved):

src/snowflake/cli/api/entities/utils.py
sfc-gh-melnacouzi: It doesn't work because each file might have different env (based on what syntax is used in each file)

sfc-gh-bdufour (approved):

sfc-gh-turbaszek (approved):

PR #1518, opened 2024-08-29
CLI context as ContextVar: forking, asyncio, connection caching

sfc-gh-cgorrie:
### Pre-review checklist

### Changes description
These changes pave the way for the Language Server implementation by making `cli_global_context` a `ContextVar`, implementing a `fork_cli_context` context manager that allows cloning + modifying the global context within its with statement, and implementing an `OpenConnectionCache` keyed on `_ConnectionContext` objects.

If there is a running asyncio event loop, the connection cache will attempt to close connections when they have not been touched for 10 minutes. If there is no event loop, connections are closed at CLI exit (as before this PR).

The concept of typer pre-execute commands was removed, because it complicates the LSP implementation and it was only used in one place; that functionality is now implemented as the lazily-evaluated `DefinitionManager` in `_CliGlobalContextManager`.

When no connection name is provided by the user and the connection is non-temporary, we resolve the default connection name earlier. Previously it was set back to `CliContextManager` by `connect_to_snowflake`, which was brittle.

`ConnectionContext` and `CliGlobalContextManager` were both changed to dataclasses to simplify implementation. Call sites now use standard attribute setting / `setattr`.

There is now an allowlist of `SUPPORTED_ENV_OVERRIDES` in `snow_connector` rather than this set being implicit (was the set of override keys passed into `connect_to_snowflake`).

### DIFF
--- a/tests_integration/spcs/testing_utils/compute_pool_utils.py
+++ b/tests_integration/spcs/testing_utils/compute_pool_utils.py
@@ -57,7 +57,7 @@ class ComputePoolTestSteps:
         assert result.json, result.output
         assert "status" in result.json
         assert (
-            f"Compute Pool {compute_pool_name.upper()} successfully created."
+            f"Compute pool {compute_pool_name.upper()} successfully created."
             in result.json["status"]  # type: ignore
         )
 


src/snowflake/cli/api/connections.py
```diff
@@ -0,0 +1,349 @@
...
+    and is keyed by ConnectionContext objects, e.g. cache[ctx].execute_string(...).
+    Connections are automatically closed after CONNECTION_CLEANUP_SEC, but
+    are guaranteed to be open (if config is valid) when returned by the cache.
+    """
+
+    connections: dict[str, SnowflakeConnection]
+    cleanup_futures: dict[str, asyncio.TimerHandle]
+
+    CONNECTION_CLEANUP_SEC: float = 10.0 * 60
+
```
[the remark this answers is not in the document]
sfc-gh-cgorrie: No, we don't. It's something I'm looking for CLI's feedback on (@sfc-gh-turbaszek) -- also @sfc-gh-klin would be helpful to get your thoughts too. This timeout says that if it's been 10 minutes since a connection was accessed, we will attempt to close it.

This auto-close will _not_ cancel any in-fight queries, as far as I can tell.

src/snowflake/cli/api/connections.py
```diff
@@ -0,0 +1,357 @@
...
+from snowflake.cli.api.exceptions import InvalidSchemaError
+from snowflake.connector import SnowflakeConnection
+from snowflake.connector.compat import IS_WINDOWS
+
+logger = logging.getLogger(__name__)
+
+schema_pattern = re.compile(r".+\..+")
+
+
+class ConnectionContext:
```
sfc-gh-cgorrie, line 34: Moved from cli_global_context.py

src/snowflake/cli/api/connections.py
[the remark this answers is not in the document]
sfc-gh-klin: In the VS Code extension by default we keep our connection through the Node JS connector open the whole time the user is using the extension unless they change a config setting. But that continuous extension usage connection (to allow users to come back to their session and execute SQL or browse their data) is different to how the connections here are going to be used. Since these connections are being used for execution of SnowCLI commands, if it has been 10 minutes since one has been used, it seems good to close it to save resources. When the user from the extension tries to execute another SnowCLI command, opening a new one or re-using an existing one doesn't make a difference from their perspective--maybe opening a new one takes a bit longer than re-using an existing one but that overhead doesn't seem bad. So this is reasonable.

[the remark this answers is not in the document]
sfc-gh-cgorrie: It's possible to drop the connection cache altogether if you don't think the performance gains are worth the complexity. It's not a correctness thing in this PR, because the session / master token are passed in anyway to ensure the "identity" of the connection is shared between LSP/CLI + Vscode.

src/snowflake/cli/api/connections.py
```diff
@@ -0,0 +1,357 @@
...
+
+    def _touch(self, key: str):
+        """
+        Extend the lifetime of the cached connection at the given key.
+        """
+        if key in self.cleanup_futures:
+            self.cleanup_futures.pop(key).cancel()
+
+        loop = asyncio.get_event_loop()
+        handle = loop.call_later(
```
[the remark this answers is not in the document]
sfc-gh-cgorrie: If there's no active event loop, it's OK that it's not called -- but I'll check for it specifically and log a debug statement.

src/snowflake/cli/api/cli_global_context.py
```diff
@@ -14,249 +14,65 @@
...
-from snowflake.connector.compat import IS_WINDOWS
 
 if TYPE_CHECKING:
+    from snowflake.cli.api.project.definition_manager import DefinitionManager
     from snowflake.cli.api.project.schemas.project_definition import ProjectDefinition
 
-schema_pattern = re.compile(r".+\..+")
-
-
-class _ConnectionContext:
```
sfc-gh-cgorrie, line 33: Moved into connections.py with some changes

src/snowflake/cli/api/cli_global_context.py
```diff
@@ -14,249 +14,75 @@
...
-            diag_allowlist_path=self._diag_allowlist_path,
-            connection_name=self.connection_name,
-            **self._collect_not_empty_connection_attributes(),
-        )
+_CONNECTION_CACHE = OpenConnectionCache()
 
 
 class _CliGlobalContextManager:
+    _definition_manager: DefinitionManager | None
+    _override_project_definition: ProjectDefinition | None
```
sfc-gh-turbaszek, line 37: Shouldn't override be handled by definition manager?

src/snowflake/cli/api/cli_global_context.py
```diff
@@ -14,249 +14,75 @@
...
+        mgr.set_connection_context(self.connection_context.clone())
+        mgr.set_connection_cache(self._connection_cache)
+        mgr._set_definition_manager(self.definition_manager)  # noqa: SLF001
+        mgr.set_enable_tracebacks(self.enable_tracebacks)
+        mgr.set_output_format(self.output_format)
+        mgr.set_verbose(self.verbose)
+        mgr.set_experimental(self.experimental)
+        mgr.set_project_path_arg(self.project_path_arg)
+        mgr.set_project_env_overrides_args(self.project_env_overrides_args.copy())
+        mgr.set_override_project_definition(self.override_project_definition)
```
sfc-gh-turbaszek, line 68: Nit: builder pattern may be interesting here

src/snowflake/cli/api/cli_global_context.py
```diff
@@ -399,19 +274,44 @@ def _should_force_mute_intermediate_output(self) -> bool:
...
+    env: dict[str, str] | None = None,
+    project_is_optional: bool | None = None,
+    project_path: str | None = None,
+) -> Iterator[_CliGlobalContextAccess]:
+    """
+    Forks the global CLI context, making changes that are only visible
+    (e.g. via get_cli_context()) while inside this context manager.
+    """
+    new_manager = _CLI_CONTEXT_MANAGER.get().clone()
+    token = _CLI_CONTEXT_MANAGER.set(new_manager)
```
sfc-gh-turbaszek, line 302: Why it's called token?

src/snowflake/cli/api/commands/typer_pre_execute.py
```diff
@@ -1,26 +0,0 @@
...
-from snowflake.cli.api.cli_global_context import get_cli_context_manager
-
-
-def register_pre_execute_command(command: Callable[[], None]) -> None:
-    get_cli_context_manager().add_typer_pre_execute_commands(command)
-
-
-def run_pre_execute_commands() -> None:
-    for command in get_cli_context_manager().typer_pre_execute_commands:
-        command()
```
sfc-gh-turbaszek, line 26: +1

src/snowflake/cli/api/cli_global_context.py
sfc-gh-cgorrie: Yeah, that's a good idea. Would you like to see that in this PR to get merged, or can we fast-follow?

src/snowflake/cli/api/cli_global_context.py
sfc-gh-cgorrie: https://docs.python.org/3/library/contextvars.html#contextvars.Token

src/snowflake/cli/api/connections.py
```diff
@@ -0,0 +1,366 @@
...
+
+    connections: dict[str, SnowflakeConnection]
+    cleanup_futures: dict[str, asyncio.TimerHandle]
+
+    CONNECTION_CLEANUP_SEC: float = 10.0 * 60
+    """Connections are closed this many seconds after the last time they are accessed."""
+
+    def __init__(self):
+        self.connections = {}
+        self.cleanup_futures = {}
```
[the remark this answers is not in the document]
sfc-gh-cgorrie: Typings are above on lines 294/295

src/snowflake/cli/api/cli_global_context.py
sfc-gh-turbaszek: Can be another PR, but it would add more separation

sfc-gh-cgorrie: I'll see what I can do here!

src/snowflake/cli/_app/snow_connector.py
```diff
@@ -58,6 +55,10 @@ def connect_to_snowflake(
 ) -> SnowflakeConnection:
     if temporary_connection and connection_name:
         raise ClickException("Can't use connection name and temporary connection.")
+    elif not temporary_connection and not connection_name:
+        raise ClickException(
+            "One of connection name and temporary connection is required."
```
sfc-gh-mraba, line 60:
```suggestion
            "One of connection name or temporary connection is required."
```

src/snowflake/cli/api/cli_global_context.py
sfc-gh-mraba: @sfc-gh-turbaszek do you mean override part or whole  project definition?
Somehow we ended with connection and project definition coupled.

sfc-gh-cgorrie:
@sfc-gh-mraba this is the whole project definition being overridden, not just part of it. Let me know if you have more questions!

src/snowflake/cli/_app/snow_connector.py
```diff
@@ -46,6 +44,22 @@
 ENCRYPTED_PKCS8_PK_HEADER = b"-----BEGIN ENCRYPTED PRIVATE KEY-----"
 UNENCRYPTED_PKCS8_PK_HEADER = b"-----BEGIN PRIVATE KEY-----"
 
+# connection keys that can be set using SNOWFLAKE_* env vars
+SUPPORTED_ENV_OVERRIDES = [
```
sfc-gh-cgorrie, line 48: This is necessary because before the possible overrides was implicit in the keys of `_ConnectionContext._collect_not_empty_connection_attributes`, which was a misnomer because it always contained the same keys whether or not they had values. By modifying it to the (assumed) original intention, we needed a list of keys that can be grabbed from `os.environ` somewhere (this is a subset of all keys in `ConnectionContext`).

src/snowflake/cli/_app/snow_connector.py
```diff
@@ -71,24 +89,20 @@ def connect_to_snowflake(
 
     if connection_name:
         connection_parameters = get_connection_dict(connection_name)
-        connection_parameters = get_connection_dict(connection_name)
```
sfc-gh-fcampbell, line 74: is there a side-effect to calling this twice, or was this just a bug?

src/snowflake/cli/api/cli_global_context.py
```diff
@@ -14,337 +14,126 @@
...
 
     @property
-    def project_env_overrides_args(self) -> dict[str, str]:
-        return self._project_env_overrides_args
-
-    def set_project_env_overrides_args(
-        self, project_env_overrides_args: dict[str, str]
-    ):
-        self._project_env_overrides_args = project_env_overrides_args
+    def project_root(self) -> Path | None:
```
sfc-gh-fcampbell, line 92: this always returns a `Path`
```suggestion
    def project_root(self) -> Path:
```

src/snowflake/cli/api/cli_global_context.py
```diff
@@ -399,19 +188,47 @@ def _should_force_mute_intermediate_output(self) -> bool:
         return self._manager.output_format == OutputFormat.JSON
 
 
-_CLI_CONTEXT_MANAGER: _CliGlobalContextManager | None = None
-_CLI_CONTEXT: _CliGlobalContextAccess | None = None
+_INITIAL_CLI_CONTEXT = _CliGlobalContextManager()
+
+_CLI_CONTEXT_MANAGER: ContextVar[_CliGlobalContextManager] = ContextVar(
+    "cli_context", default=_INITIAL_CLI_CONTEXT
```
sfc-gh-fcampbell, line 194: this was only extracted for debugging, we should hide it again
```suggestion
_CLI_CONTEXT_MANAGER: ContextVar[_CliGlobalContextManager] = ContextVar(
    "cli_context", default=_CliGlobalContextManager()
```

src/snowflake/cli/api/connections.py
```diff
@@ -0,0 +1,206 @@
...
+            if key not in field_map:
+                raise KeyError(f"{key} is not a field of {self.__class__.__name__}")
+            setattr(self, key, value)
+
+    def __repr__(self) -> str:
+        """Minimal repr where empty (i.e. None) values have their keys omitted."""
+        items = [
+            f"{k}={repr(v)}"
+            for (k, v) in self.as_nonempty_dict().items()
+            if v is not None
```
sfc-gh-fcampbell, line 81: is this still needed given that we use `as_nonempty_dict()`?

src/snowflake/cli/api/connections.py
```diff
@@ -0,0 +1,206 @@
...
+        """
+        field_map = {field.name: field for field in fields(self)}
+        for (key, value) in updates.items():
+            # ensure key represents a property
+            if key not in field_map:
+                raise KeyError(f"{key} is not a field of {self.__class__.__name__}")
+            setattr(self, key, value)
+
+    def __repr__(self) -> str:
+        """Minimal repr where empty (i.e. None) values have their keys omitted."""
```
sfc-gh-fcampbell, line 77: None and empty aren't the same thing
```suggestion
        """Minimal repr where None values have their keys omitted."""
```

src/snowflake/cli/api/connections.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def validate_and_complete(self):
+        """
+        Ensure we can create a connection from this context.
+        """
+        if not self.temporary_connection and not self.connection_name:
+            self.connection_name = get_default_connection_name()
+
+    def as_nonempty_dict(self) -> dict:
+        return {k: v for (k, v) in asdict(self).items() if v is not None}
```
sfc-gh-fcampbell, line 109: nit, this function is sandwiched between two connection-related ones, can we move it down?

src/snowflake/cli/api/connections.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def clear(self):
+        """Closes all connections and resets the cache to its initial state."""
+        connection_keys = list(self.connections.keys())
+        for key in connection_keys:
+            self._cleanup(key)
+
+        # if any orphaned futures still exist, clean them up too
+        for key in self.cleanup_futures:
+            self.cleanup_futures[key].cancel()
```
sfc-gh-fcampbell, line 163:
```suggestion
        for future in self.cleanup_futures.value():
            future.cancel()
```

src/snowflake/cli/api/connections.py
```diff
@@ -0,0 +1,206 @@
...
+        self.cleanup_futures.clear()
+
+    def _has_open_connection(self, key: str):
+        return key in self.connections
+
+    def _insert(self, key: str, ctx: ConnectionContext):
+        try:
+            self.connections[key] = ctx.build_connection()
+        except Exception:
+            logger.debug("ConnectionCache: failed to connect using {key}; not caching.")
```
sfc-gh-fcampbell, line 173:
```suggestion
            logger.debug(f"ConnectionCache: failed to connect using {key}; not caching.")
```

src/snowflake/cli/api/connections.py
```diff
@@ -0,0 +1,206 @@
...
+        if key in self.cleanup_futures:
+            self.cleanup_futures.pop(key).cancel()
+
+    def _touch(self, key: str):
+        """
+        Extend the lifetime of the cached connection at the given key.
+        """
+        try:
+            loop = asyncio.get_event_loop()
+        except (RuntimeError, DeprecationWarning):
```
sfc-gh-fcampbell, line 186: `DeprecationWarning`s aren't thrown; on earlier python versions this just returns `Non` if there's no loop, so we need to check for that before calling `.call_later()`

src/snowflake/cli/api/connections.py
```diff
@@ -0,0 +1,206 @@
...
+        handle = loop.call_later(
+            self.CONNECTION_CLEANUP_SEC, lambda: self._cleanup(key)
+        )
+        self._cancel_cleanup_future_if_exists(key)
+        self.cleanup_futures[key] = handle
+
+    def _cleanup(self, key: str):
+        """Closes the cached connection at the given key."""
+        if key not in self.connections:
+            logger.debug("Cleaning up connection {key}, but not found in cache!")
```
sfc-gh-fcampbell, line 202:
```suggestion
            logger.debug(f"Cleaning up connection {key}, but not found in cache!")
```

tests/conftest.py
```diff
@@ -69,16 +73,25 @@ def os_agnostic_snapshot(snapshot):
...
-    yield
+    with fork_cli_context():
+        connection_cache = OpenConnectionCache()
+        cli_context_manager = get_cli_context_manager()
+        cli_context_manager.reset()
+        cli_context_manager.verbose = False
+        cli_context_manager.enable_tracebacks = False
+        cli_context_manager.connection_cache = connection_cache
+        config_init(test_snowcli_config)
+        loggers.create_loggers(verbose=False, debug=False)
```
sfc-gh-fcampbell, line 90: there are a few things set here that aren't being set in the fork for integration tests, is that intentional?

tests/api/test_global_cli_context.py
```diff
@@ -0,0 +1,86 @@
...
+    assert mgr.override_project_definition is None
+
+    initial_mgr = _CliGlobalContextManager()
+    for f in fields(mgr):
+        assert getattr(mgr, f.name) == getattr(
+            initial_mgr, f.name
+        ), f"{f.name} was not reset properly"
+
+
+# FIXME: needs asyncio test
```
sfc-gh-fcampbell, line 53: is this still relevant?

tests/api/test_connections.py
```diff
@@ -0,0 +1,102 @@
...
+        "schema",
+        "user",
+        "password",
+        "authenticator",
+        "warehouse",
+        "session_token",
+        "master_token",
+    )
+
+    ctx = ConnectionContext()
```
sfc-gh-fcampbell, line 63: tiny nit, can we call this `old_ctx`?

tests/api/test_connections.py
```diff
@@ -0,0 +1,102 @@
...
+    for key in keys:
+        setattr(ctx, key, "value")
+
+    new_ctx = ctx.clone()
+    assert asdict(new_ctx) == asdict(ctx)
+
+    for key in keys:
+        setattr(new_ctx, key, "values_should_not_appear_in_ctx")
```
sfc-gh-fcampbell, line 71:
```suggestion
        setattr(new_ctx, key, "updated_values_should_not_appear_in_old_ctx")
```

src/snowflake/cli/_app/snow_connector.py
sfc-gh-cgorrie: No side effects, I presume it was a bad merge.

tests/conftest.py
sfc-gh-cgorrie: This preserves the existing behaviour. @sfc-gh-turbaszek any thoughts on differences between context manager setup for unit vs integration tests?

tests/api/test_global_cli_context.py
sfc-gh-cgorrie: Yes, we don't have the pytest-asyncio library yet. That will be introduced in the LSP PR, at which point I'll be able to write tests for the connection caching cleanup methods that are not currently exercised (because nothing creates an event loop, yet).

src/snowflake/cli/api/commands/flags.py
```diff
@@ -303,15 +282,15 @@ def _diag_log_path_callback(path: str):
...
+    callback=_context_callback("output_format"),
     rich_help_panel=_CLI_BEHAVIOUR,
 )
 
 SilentOption = typer.Option(
     False,
     "--silent",
     help="Turns off intermediate output to console.",
-    callback=_callback(lambda: get_cli_context_manager().set_silent),
+    callback=_context_callback("silent"),
```
sfc-gh-turbaszek, line 293: I'm not sold on this one. The setters allowed us to get rid of string variables. With proposed implementation if you change attribute/value name auto-refactor won't work

src/snowflake/cli/api/commands/flags.py
```diff
@@ -485,54 +464,38 @@ def execution_identifier_argument(sf_object: str, example: str) -> typer.Argumen
...
-    cli_context_manager.set_template_context(template_context)
-
-
 def project_definition_option(is_optional: bool):
-    def project_definition_callback(project_path: str) -> None:
-        get_cli_context_manager().set_project_path_arg(project_path)
-        register_pre_execute_command(lambda: register_project_definition(is_optional))
+    def project_path_callback(project_path: str) -> str:
+        get_cli_context_manager().project_path_arg = project_path
+        get_cli_context_manager().project_is_optional = is_optional
```
sfc-gh-turbaszek, line 470:
```suggestion
        ctx_mgr = get_cli_context_manager()
        ctx_mgr.project_path_arg = project_path
        ctx_mgr.project_is_optional = is_optional
```
WDYT?

src/snowflake/cli/_app/snow_connector.py
```diff
@@ -46,6 +44,22 @@
 ENCRYPTED_PKCS8_PK_HEADER = b"-----BEGIN ENCRYPTED PRIVATE KEY-----"
 UNENCRYPTED_PKCS8_PK_HEADER = b"-----BEGIN PRIVATE KEY-----"
 
+# connection keys that can be set using SNOWFLAKE_* env vars
+SUPPORTED_ENV_OVERRIDES = [
+    "account",
+    "user",
+    "password",
+    "authenticator",
+    "private_key_file",
```
sfc-gh-astus, line 53: Will it work for `private_key_path` alias?

src/snowflake/cli/api/cli_global_context.py
```diff
@@ -399,19 +188,45 @@ def _should_force_mute_intermediate_output(self) -> bool:
         return self._manager.output_format == OutputFormat.JSON
 
 
-_CLI_CONTEXT_MANAGER: _CliGlobalContextManager | None = None
-_CLI_CONTEXT: _CliGlobalContextAccess | None = None
+_CLI_CONTEXT_MANAGER: ContextVar[_CliGlobalContextManager] = ContextVar(
+    "cli_context", default=_CliGlobalContextManager()
+)
```
sfc-gh-astus, line 193: Could you change it to lazy init? We had problems with global variables in python connector.

src/snowflake/cli/_app/snow_connector.py
sfc-gh-cgorrie: No, it won't. We lost that ability with this change: https://github.com/snowflakedb/snowflake-cli/pull/1506/files#diff-05f8159f2785119ba7d4251bb727ca88d0b874f88a650847ce5b86adf7ac13abR209

If we want to support both as aliases, I can add that back in here.

src/snowflake/cli/api/commands/flags.py
sfc-gh-cgorrie: Hmm, good point. I can re-add the boilerplate per-callback, but considering that it needs to return the set value as well, it's going to be a lot more code per line, something like...

```python
def callback(silent):
  get_cli_context_manager().silent = silent
  return silent

[...]
  callback=callback,
```

Can you think of a way of reducing it further while directly accessing the attrs?

src/snowflake/cli/_app/snow_connector.py
sfc-gh-cgorrie: @sfc-gh-astus my latest commit supports this alias, and I added a test to ensure that there's no errors when interacting with `_path` vs `_file` in env vars or in the config TOML. Using both is undefined behaviour (which one will "win" is up to iteration order), but allowed.

src/snowflake/cli/_app/snow_connector.py
```diff
@@ -70,37 +99,36 @@ def connect_to_snowflake(
...
-        )
 
     # Apply overrides to connection details
+    # (1) Command line override case
     for key, value in overrides.items():
-        # Command line override case
-        if value:
-            connection_parameters[key] = value
-            continue
+        if value is not None:
```
sfc-gh-cgorrie, line 112: If we ever have boolean connection parameters in `ConnectionContext` this would have failed to set the override when `False`, so just fixing out of caution.

src/snowflake/cli/api/commands/flags.py
```diff
@@ -38,8 +42,18 @@
 
 def _connection_callback(prop: str):
     """Generates a setter for a field on the current context manager's connection context."""
+    if not prop in ConnectionContext.__dataclass_fields__:
```
sfc-gh-fcampbell, line 45: this is probably fine, but the public API for this is `fields(ConnectionContext)`

sfc-gh-fcampbell, line 45:
```suggestion
    if prop not in ConnectionContext.__dataclass_fields__:
```

src/snowflake/cli/api/commands/flags.py
```diff
@@ -48,8 +62,16 @@ def callback(value):
 
 def _context_callback(prop: str):
     """Generates a setter for a field on the current context manager."""
+    if not prop in _CliGlobalContextManager.__dataclass_fields__:
```
sfc-gh-fcampbell, line 65:
```suggestion
    if prop not in _CliGlobalContextManager.__dataclass_fields__:
```

src/snowflake/cli/api/commands/flags.py
sfc-gh-cgorrie: It's slightly different -- fields gives you a list of tuples. This is a dict, which is better from a performance perspective for us.

sfc-gh-fcampbell (approved):

sfc-gh-turbaszek (approved):

PR #1520, opened 2024-08-30
Workspaces application package - validate action and command

sfc-gh-gbloom:
### Pre-review checklist

### Changes description
...


src/snowflake/cli/api/entities/application_package_entity.py
```diff
@@ -161,6 +163,36 @@ def action_drop(
...
+            )
+
+        self.validate_setup_script(
+            console=ctx.console,
+            package_name=package_name,
+            package_role=package_role,
+            stage_fqn=stage_fqn,
+            use_scratch_stage=True,
+            scratch_stage_fqn=model.scratch_stage,
+            deploy_to_scratch_stage_fn=deploy_to_scratch_stage_fn,
```
sfc-gh-melnacouzi, line 192: what's the reason we need to pass in a function to deploy to the stage instead of validate deploying directly?

src/snowflake/cli/api/entities/application_package_entity.py
```diff
@@ -161,6 +163,36 @@ def action_drop(
...
+                stage_fqn=model.scratch_stage,
+            )
+
+        self.validate_setup_script(
+            console=ctx.console,
+            package_name=package_name,
+            package_role=package_role,
+            stage_fqn=stage_fqn,
+            use_scratch_stage=True,
+            scratch_stage_fqn=model.scratch_stage,
```
sfc-gh-melnacouzi, line 191: why do we have to pass both stage_fqn and scratch_stage_fqn. Don't we only use one of the 2?

src/snowflake/cli/_plugins/workspace/commands.py
```diff
@@ -192,3 +181,24 @@ def drop(
         EntityActions.DROP,
         force_drop=force,
     )
+
+
+@ws.command(requires_connection=True)
```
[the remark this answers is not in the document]
sfc-gh-gbloom: Yep, you're right. Hiding this and the other commands as well (except "migrate").

src/snowflake/cli/api/entities/application_package_entity.py
sfc-gh-gbloom: I'll refactor and simply deploy here directly. A few PRs ago it wasn't easy to do, but now I can refactor `deploy()` to a stateless function and reuse it in both `snow app` and `snow ws` flows.

src/snowflake/cli/api/entities/application_package_entity.py
sfc-gh-gbloom: Yes, I considered simplifying it -- that's how it was implemented in `NativeAppManager`. But I prefer keeping the signatures and logic the same to avoid making multiple changes in this refactor.

src/snowflake/cli/api/entities/application_package_entity.py
sfc-gh-gbloom: OK I gave it a try, but the change is still not as trivial as I hoped it to be. I'll follow up with this refactor in a dedicated PR.
The main reason I pass this as a function is that the `validate()` needs to be able to deploy, which requires passing many additional values through multiple call levels. It also requires an additional refactor to make the bundle/deploy methods static, because the validate one is static too.

sfc-gh-melnacouzi (approved):

PR #1526, opened 2024-09-02
Raise error for duplicated keys in project definition

sfc-gh-jsikorski:
### Pre-review checklist

### Changes description
During bug-bash we discovered that project definition files allow duplicated fields. This change causes CLI to quit with an error, when two entities in the PDF have the same name.
Also, i removed some leftover snapshots from recent test cleanup.


src/snowflake/cli/api/project/definition_conversion.py
```diff
@@ -72,6 +73,11 @@ def convert_snowpark_to_v2_data(snowpark: Snowpark) -> Dict[str, Any]:
             else entity.name
         )
 
+        if entity_name in data["entities"]:
+            raise ClickException(
+                f"Entity with name {entity_name} seems to be duplicated. Please rename it, and try again."
```
sfc-gh-pczajka, line 78:
```suggestion
                f"Entity with name {entity_name} seems to be duplicated. Please rename it and try again."
```

src/snowflake/cli/api/project/definition.py
```diff
@@ -90,3 +96,20 @@ def default_role():
...
+    Raises error it there are duplicated keys on the same level in the yaml file
+    """
+    mapping = {}
+
+    for key_node, value_node in node.value:
+        key = loader.construct_object(key_node, deep=deep)
+        value = loader.construct_object(value_node, deep=deep)
+        if key in mapping.keys():
+            raise ClickException(
+                f"While loading the project definition file, duplicate key was found: {key}"
```
sfc-gh-pczajka, line 112: Is it possible to provide a full path? Like `Found duplicated key "entities.test_procedure.handler" while loading project definition file`?

sfc-gh-jsikorski: The custom constructor  is used to construct a single tag, so it has no info about parent nodes.
I can add some variable to track it, but AFAIK it would mean we have to traverse all the nodes while loading each tag, and that would significantly increase time complexity of the loader function

sfc-gh-pczajka (approved):

PR #1536, opened 2024-09-04
SNOW-1643030 Add Native App support to `snow ws migrate` command

sfc-gh-fcampbell:
### Pre-review checklist
   * N/A I've described my changes in the release notes (covered by existing note).

### Changes description
Enables `snow ws migrate` to migrate a v1 or v1.1 Native App project file to v2. Doesn't support migrating package scripts since they use a different template syntax, we'll tackle those in a followup.

### DIFF
--- a/src/snowflake/cli/api/project/definition_conversion.py
+++ b/src/snowflake/cli/api/project/definition_conversion.py
@@ -181,7 +181,9 @@ def convert_native_app_to_v2_data(
         bundle_map = build_bundle(
             project_root, Path(native_app.deploy_root), native_app.artifacts
         )
-        return str(bundle_map.to_project_path(Path("manifest.yml")))
+        # Use a POSIX path to be consistent with other migrated fields
+        # which use POSIX paths as default values
+        return bundle_map.to_project_path(Path("manifest.yml")).as_posix()
 
     package_entity_name = "pkg"
     package = {
--- a/src/snowflake/cli/api/project/definition_conversion.py
+++ b/src/snowflake/cli/api/project/definition_conversion.py
@@ -1,8 +1,13 @@
+from __future__ import annotations
+
 import logging
 from pathlib import Path
 from typing import Any, Dict, Literal, Optional
 
 from click import ClickException
+from snowflake.cli._plugins.nativeapp.artifacts import (
+    build_bundle,
+)
 from snowflake.cli._plugins.snowpark.common import is_name_a_templated_one
 from snowflake.cli.api.constants import (
     DEFAULT_ENV_FILE,
@@ -10,6 +15,9 @@ from snowflake.cli.api.constants import (
     PROJECT_TEMPLATE_VARIABLE_OPENING,
     SNOWPARK_SHARED_MIXIN,
 )
+from snowflake.cli.api.project.schemas.native_app.application import Application
+from snowflake.cli.api.project.schemas.native_app.native_app import NativeApp
+from snowflake.cli.api.project.schemas.native_app.package import Package
 from snowflake.cli.api.project.schemas.project_definition import (
     ProjectDefinition,
     ProjectDefinitionV2,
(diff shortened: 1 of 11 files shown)


src/snowflake/cli/api/project/definition_conversion.py
```diff
@@ -144,6 +159,76 @@ def convert_streamlit_to_v2_data(streamlit: Streamlit):
...
+        )
+        # Use a POSIX path to be consistent with other migrated fields
+        # which use POSIX paths as default values
+        return bundle_map.to_project_path(Path("manifest.yml")).as_posix()
+
+    package_entity_name = "pkg"
+    package = {
+        "type": "application package",
+        "identifier": (
+            native_app.package.name
```
sfc-gh-gbloom, line 192: Should we suffix the identifier values with `_<% ctx.env.USER %>` to keep the v1 behavior?
@sfc-gh-cgorrie WDYT?

tests/workspace/__snapshots__/test_manager.ambr
```diff
@@ -181,6 +181,13 @@
   '''
   definition_version: '2'
   entities:
+    app:
+      from:
+        target: pkg
+      identifier: myapp_app
+      meta:
+        warehouse: app_wh
+      type: application
```
sfc-gh-gbloom, line 190: Kind of a nit, but can we control the order of the keys? Specifically I'd prefer to have "type" to be first and "identifier" second, but if all of the keys could be in the same order as the schema it would be the best.

sfc-gh-gbloom (approved):

sfc-gh-jsikorski (approved):
LGTM

sfc-gh-turbaszek (approved):

src/snowflake/cli/api/project/definition_conversion.py
```diff
@@ -144,6 +159,76 @@ def convert_streamlit_to_v2_data(streamlit: Streamlit):
...
+
+    def _find_manifest():
+        # We don't know which file in the project directory is the actual manifest,
+        # and we can't iterate through the artifacts property since the src can contain
+        # glob patterns. The simplest solution is to bundle the app and find the
+        # manifest file from the resultant BundleMap, since the bundle process ensures
+        # that only a single source path can map to the corresponding destination path
+        bundle_map = build_bundle(
+            project_root, Path(native_app.deploy_root), native_app.artifacts
+        )
```
sfc-gh-pjafari, line 183: The default behaviour is that the `build_bundle` will fail here if there are no artifacts specified. Would that make sense work for this conversion to v2 as well? Or should the conversion happen with the artifact list empty? If the latter, then you might want to catch that error here :)

src/snowflake/cli/api/project/definition_conversion.py
sfc-gh-bdufour: We should discuss this in person I think.

sfc-gh-pjafari (approved):

src/snowflake/cli/api/project/definition_conversion.py
```diff
@@ -144,6 +159,76 @@ def convert_streamlit_to_v2_data(streamlit: Streamlit):
...
+        # Use a POSIX path to be consistent with other migrated fields
+        # which use POSIX paths as default values
+        return bundle_map.to_project_path(Path("manifest.yml")).as_posix()
```
sfc-gh-bdufour, line 186: what happens if the manifest is not found? We should make sure we don't crash in that case.

tests/workspace/__snapshots__/test_manager.ambr
sfc-gh-fcampbell: nah, this is always rendered in alphabetical order no matter the order it's added in the dict

src/snowflake/cli/api/project/definition_conversion.py
sfc-gh-fcampbell: the manifest field is mandatory in v2, i'll make this a user error in a followup

src/snowflake/cli/api/project/definition_conversion.py
sfc-gh-fcampbell: discussed locally, will followup to make this a user error and abort migration

src/snowflake/cli/api/project/definition_conversion.py
sfc-gh-fcampbell: discussed locally, will followup

PR #1540, opened 2024-09-06
Proof-of-concept for "snow ws" command generation

sfc-gh-cgorrie:
### Pre-review checklist

### Changes description

Generates `snow ws @<entity_id>` command groups based on the current directory's `snowflake.yml`.

What's done:
* allows decoupling commands from classes, so that each one can live in its own compilation unit
* defines an annotation DSL (e.g. `HelpText`, `ParameterDeclarations`) that ensures individual commands don't actually rely on click / typer
* generates click / typer commands from those annotations
* allows commands like "version create" to automatically create sub-trees

Things that need addressing:
* deal with position action arguments (e.g. `snow @pkg version create v1` needs to be passed in instead as `--version v1`)
* tests are failing
* commands cannot share option arguments (they all need to re-define them)


src/snowflake/cli/_plugins/workspace/entity_commands.py
```diff
@@ -0,0 +1,102 @@
...
+    for each found entity. Throws a fatal ClickException if templating is used
+    in the basic definition of entities (i.e. the type discriminator field).
+    """
+    dm = DefinitionManager(str(project_root) if project_root is not None else None)
+
+    if not dm.has_definition_file:
+        return
+
+    root = dm.unrendered_project_definition
+    for (target_id, model) in root.entities.items():
```
sfc-gh-turbaszek, line 94: Should we consider `model.get_command_group(target_id)` that would encapsulate L100-102? WDYT?

sfc-gh-cgorrie: I'm not sure I quite understand the suggestion. Can you give a bit more detail?

src/snowflake/cli/_plugins/workspace/entity_commands.py
```diff
@@ -0,0 +1,204 @@
...
+            ws = WorkspaceManager(
+                project_definition=cli_context.project_definition,
+                project_root=cli_context.project_root,
+            )
+
+            # filter out common CLI options our action doesn't take
+            pass_through_options = {
+                k: v for (k, v) in options.items() if k in known_params
+            }
+            ws.perform_action(self.target_id, action, **pass_through_options)
```
sfc-gh-fcampbell, line 141: how do we handle return values? for example, we currently have some command that wrap the return value of the entity action in `MessageResult`, others use `QueryResult`, etc

src/snowflake/cli/_plugins/workspace/entity_commands.py
```diff
@@ -0,0 +1,204 @@
...
+            return MessageResult(
+                f"Successfully performed {action.verb} on {self.target_id}."
+            )
```
sfc-gh-fcampbell, line 144: not all actions want to show a success message (especially those that retrieve data, like `version-list`)

sfc-gh-fcampbell:
Closing this for now due to uncertainty regarding the `snow ws` command. We can use the ideas from this PR in a future re-attempt if necessary.

PR #1546, opened 2024-09-09
SNOW-1653909 Add options for choosing which entities to convert back to v1

sfc-gh-fcampbell:
Use the [no-whitespace diff](https://github.com/snowflakedb/snowflake-cli/pull/1546/files?w=1) for a more pleasant diff-viewing experience

### Pre-review checklist

### Changes description
Adds `--package-entity-id` and `--app-entity-id` options to all commands that use `@nativeapp_definition_v2_to_v1` to allow us to handle v2 definitions that have multiple apps or packages in the old v1 commands.

The new logic will also attempt to determine which app or package to convert if it's non-ambiguous (i.e. there's only one entity of that type) and will infer the package to convert from the application entity's `target` field.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -22,6 +22,7 @@
 
 ## New additions
 * Added templates expansion of arbitrary files for Native Apps through `templates` processor.
+* Added `--package-entity-id` and `--app-entity-id` options to `snow app` commands to allow targeting specific entities when the `definition_version` in `snowflake.yml` is `2` or higher and it contains multiple `application package` or `application` entities.
 
 ## Fixes and improvements
 * Duplicated keys in `snowflake.yml` are now detected and reported.
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -22,7 +22,6 @@
 
 ## New additions
 * Added templates expansion of arbitrary files for Native Apps through `templates` processor.
-* Added `--package-entity-id` and `--app-entity-id` options to `snow app` commands to allow targeting specific entities when the `definition_version` in `snowflake.yml` is `2` or higher and it contains multiple `application package` or `application` entities.
 
 ## Fixes and improvements
 * Duplicated keys in `snowflake.yml` are now detected and reported.
@@ -56,6 +55,7 @@
 * Added support for `<% ... %>` syntax in SQL templating.
 * Support multiple Streamlit application in single snowflake.yml project definition file.
 * Added `snow ws migrate` command to migrate `snowflake.yml` file from V1 to V2.
+* Added `--package-entity-id` and `--app-entity-id` options to `snow app` commands to allow targeting specific entities when the `definition_version` in `snowflake.yml` is `2` or higher and it contains multiple `application package` or `application` entities.
 
 ## Fixes and improvements
 * Fixed problem with whitespaces in `snow connection add` command.
--- a/src/snowflake/cli/_plugins/nativeapp/v2_conversions/v2_to_v1_decorator.py
+++ b/src/snowflake/cli/_plugins/nativeapp/v2_conversions/v2_to_v1_decorator.py
@@ -62,14 +62,12 @@ def _pdf_v2_to_v1(
     app_definition: Optional[ApplicationEntityModel] = None
 
     # Enumerate all application package and application entities in the project definition
-    packages = {}
-    apps = {}
-    for key, entity in v2_definition.entities.items():
-        entity_type = entity.get_type()
-        if entity_type == ApplicationPackageEntityModel.get_type():
-            packages[key] = entity
-        elif entity_type == ApplicationEntityModel.get_type():
-            apps[key] = entity
+    packages: dict[
+        str, ApplicationPackageEntityModel
+    ] = v2_definition.get_entities_by_type(ApplicationPackageEntityModel.get_type())
+    apps: dict[str, ApplicationEntityModel] = v2_definition.get_entities_by_type(
+        ApplicationEntityModel.get_type()
+    )
 
     # Determine the application entity to convert, there can be zero or one
     if app_entity_id:
--- a/src/snowflake/cli/_plugins/nativeapp/v2_conversions/v2_to_v1_decorator.py
+++ b/src/snowflake/cli/_plugins/nativeapp/v2_conversions/v2_to_v1_decorator.py
@@ -14,14 +14,17 @@
 
 from __future__ import annotations
 
+import inspect
 from functools import wraps
 from typing import Any, Dict, Optional, Union
 
+import typer
 from click import ClickException
 from snowflake.cli.api.cli_global_context import (
     get_cli_context,
     get_cli_context_manager,
 )
+from snowflake.cli.api.commands.decorators import _options_decorator_factory
 from snowflake.cli.api.project.schemas.entities.application_entity_model import (
     ApplicationEntityModel,
 )
@@ -48,7 +51,11 @@ def _convert_v2_artifact_to_v1_dict(
     return v2_artifact
 
 
-def _pdf_v2_to_v1(v2_definition: DefinitionV20) -> DefinitionV11:
+def _pdf_v2_to_v1(
+    v2_definition: DefinitionV20,
+    package_entity_id: str = "",
+    app_entity_id: str = "",
+) -> DefinitionV11:
     pdfv1: Dict[str, Any] = {"definition_version": "1.1", "native_app": {}}
 
     app_package_definition: Optional[ApplicationPackageEntityModel] = None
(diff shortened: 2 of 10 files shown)


src/snowflake/cli/_plugins/nativeapp/v2_conversions/v2_to_v1_decorator.py
```diff
@@ -14,14 +14,17 @@
...
 from functools import wraps
 from typing import Any, Dict, Optional, Union
 
+import typer
 from click import ClickException
 from snowflake.cli.api.cli_global_context import (
     get_cli_context,
     get_cli_context_manager,
 )
+from snowflake.cli.api.commands.decorators import _options_decorator_factory
```
sfc-gh-fcampbell, line 27: making this function public would require CLI team review, and I'm not sure we want to do that anyways

src/snowflake/cli/_plugins/nativeapp/v2_conversions/v2_to_v1_decorator.py
```diff
@@ -147,8 +159,32 @@ def wrapper(*args, **kwargs):
...
         if original_pdf.definition_version == "2":
-            pdfv1 = _pdf_v2_to_v1(original_pdf)
+            package_entity_id = kwargs.get("package_entity_id", "")
+            app_entity_id = kwargs.get("app_entity_id", "")
+            pdfv1 = _pdf_v2_to_v1(original_pdf, package_entity_id, app_entity_id)
             get_cli_context_manager().set_project_definition(pdfv1)
         return func(*args, **kwargs)
 
-    return wrapper
+    return _options_decorator_factory(
```
sfc-gh-cgorrie, line 168: This is great -- I like how it allows us to keep all the v2 -> v1 stuff in one place.

However, I can see myself getting tripped up by needing to provide both. Can we infer the pkg from the app when only the app is provided (i.e. by introspecting the `from` clause)?

src/snowflake/cli/_plugins/nativeapp/v2_conversions/v2_to_v1_decorator.py
```diff
@@ -48,28 +51,37 @@ def _convert_v2_artifact_to_v1_dict(
     return v2_artifact
 
 
-def _pdf_v2_to_v1(v2_definition: DefinitionV20) -> DefinitionV11:
+def _pdf_v2_to_v1(
+    v2_definition: DefinitionV20,
+    package_entity_id: str = "",
+    app_entity_id: str = "",
```
sfc-gh-cgorrie, line 57: Should we enforce the expected type of these referenced entities?

src/snowflake/cli/_plugins/nativeapp/v2_conversions/v2_to_v1_decorator.py
```diff
@@ -48,28 +51,37 @@ def _convert_v2_artifact_to_v1_dict(
...
-                    "More than one application package entity exists in the project definition file."
-                )
-            app_package_definition = entity
+            if entity.entity_id == package_entity_id or not package_entity_id:
+                if app_package_definition:
+                    raise ClickException(
+                        "More than one application package entity exists in the project definition file, "
+                        "specify --package-entity-id to choose which one to operate on."
+                    )
+                app_package_definition = entity
```
sfc-gh-pjafari, line 72: out of curiosity - if pdf has two package entities with the same id, do we fail at a step before this? 
same question for any entity type really. where do we fail the case of two entities with the same type and id?

src/snowflake/cli/_plugins/nativeapp/v2_conversions/v2_to_v1_decorator.py
sfc-gh-fcampbell: yeah that's a good idea

src/snowflake/cli/_plugins/nativeapp/v2_conversions/v2_to_v1_decorator.py
sfc-gh-fcampbell: you mean fail if we find an entity with that ID but of a different type?

src/snowflake/cli/_plugins/nativeapp/v2_conversions/v2_to_v1_decorator.py
sfc-gh-fcampbell: you can't have two entities with the same ID since they're stored in a map/dict and the ID is the key

src/snowflake/cli/_plugins/nativeapp/v2_conversions/v2_to_v1_decorator.py
sfc-gh-fcampbell: this is done in a slightly different way; the new logic checks that the ID passed refers to the correct package  or app and fails if it doesn't find one, which would also happen if someone passed in an ID for an entity of a different type

src/snowflake/cli/_plugins/nativeapp/v2_conversions/v2_to_v1_decorator.py
sfc-gh-fcampbell: done

src/snowflake/cli/_plugins/nativeapp/v2_conversions/v2_to_v1_decorator.py
```diff
@@ -48,28 +51,77 @@ def _convert_v2_artifact_to_v1_dict(
...
+    app_entity_id: str = "",
+) -> DefinitionV11:
     pdfv1: Dict[str, Any] = {"definition_version": "1.1", "native_app": {}}
 
     app_package_definition: Optional[ApplicationPackageEntityModel] = None
     app_definition: Optional[ApplicationEntityModel] = None
 
+    # Enumerate all application package and application entities in the project definition
+    packages = {}
+    apps = {}
```
sfc-gh-gbloom, line 66: You can use `get_entities_by_type()` instead of the iteration below:
```suggestion
    packages: Dict[str, ApplicationPackageEntityModel] = v2_definition .get_entities_by_type("application package")
    apps: Dict[str, ApplicationEntityModel] = v2_definition .get_entities_by_type("application")
```

src/snowflake/cli/_plugins/nativeapp/v2_conversions/v2_to_v1_decorator.py
```diff
@@ -48,28 +51,77 @@ def _convert_v2_artifact_to_v1_dict(
...
+        app_definition = apps.get(app_entity_id)
+    elif len(apps) == 1:
+        # Otherwise, if there is only one app entity, fall back to that one
+        app_definition = next(iter(apps.values()))
+    elif len(apps) > 1:
+        # If there are multiple app entities, the user must specify which one to use
+        raise ClickException(
+            "More than one application entity exists in the project definition file, "
+            "specify --app-entity-id to choose which one to operate on."
+        )
```
sfc-gh-cgorrie, line 86: I think it's worth mentioning in comments that while we could support `snow app [bundle|deploy] --package-entity-id=pkg`, we're choosing not to in order to simplify the implementation.

sfc-gh-cgorrie (approved):
Approved, with one nitpick

RELEASE-NOTES.md
```diff
@@ -22,6 +22,7 @@
 
 ## New additions
 * Added templates expansion of arbitrary files for Native Apps through `templates` processor.
+* Added `--package-entity-id` and `--app-entity-id` options to `snow app` commands to allow targeting specific entities when the `definition_version` in `snowflake.yml` is `2` or higher and it contains multiple `application package` or `application` entities.
```
sfc-gh-gbloom, line 25: This should be under the 3.0.0 section

src/snowflake/cli/_plugins/nativeapp/v2_conversions/v2_to_v1_decorator.py
sfc-gh-gbloom: Aren't all commands decorated with this? Why won't bundle/deploy work out of the box?

sfc-gh-pjafari (dismissed):

src/snowflake/cli/_plugins/nativeapp/v2_conversions/v2_to_v1_decorator.py
sfc-gh-fcampbell: yeah it is, @sfc-gh-cgorrie I'm not sure I understand your comment

image

RELEASE-NOTES.md
sfc-gh-fcampbell: updated

src/snowflake/cli/_plugins/nativeapp/v2_conversions/v2_to_v1_decorator.py
sfc-gh-fcampbell: updated

sfc-gh-gbloom (dismissed):

sfc-gh-gbloom (approved):

src/snowflake/cli/_plugins/nativeapp/v2_conversions/v2_to_v1_decorator.py
sfc-gh-cgorrie: My point is that you need _app_ entity specified if you only provide one, even though the bundle and diff commands operate on the package only.

sfc-gh-fcampbell: hmm it shouldn't be necessary, the app entity remains optional as it was before

sfc-gh-fcampbell: ah i see what you mean, if there are multiple app entities you'd need to specify which even though the command doesn't use it. I'm open to following up this PR to make the app entity entirely optional per-command

sfc-gh-cgorrie: Sounds good, let's address in a follow-up.

sfc-gh-cgorrie (approved):
LGTM, follow-up to either document or address need for app entity ID even for `snow app` commands that only target the package

PR #1568, opened 2024-09-12
Add override tag support in mixins

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
Adds `!override` tag for mixins


sfc-gh-melnacouzi (commented):
I have an overall comment on this.
1) Have we considered having the flag on the mixin itself?
2) We don't have only 2 options. We could override, we could append, we could keep (!override). Do we still want to use "!" in the name?

src/snowflake/cli/api/project/definition.py
```diff
@@ -50,6 +52,10 @@ def _get_merged_definitions(paths: List[Path]) -> Optional[Definition]:
     loader.add_constructor(
         yaml.resolver.BaseResolver.DEFAULT_MAPPING_TAG, _no_duplicates_constructor
     )
+    loader.add_constructor(
+        yaml.resolver.BaseResolver.DEFAULT_MAPPING_TAG, _no_duplicates_constructor
+    )
```
sfc-gh-melnacouzi, line 57: is this duplicate line from the one before? Is it on purpose?

src/snowflake/cli/api/utils/definition_rendering.py
```diff
@@ -277,7 +277,9 @@ def _add_defaults_to_definition(original_definition: Definition) -> Definition:
     with context({"skip_validation_on_templates": True}):
         # pass a flag to Pydantic to skip validation for templated scalars
         # populate the defaults
-        project_definition = build_project_definition(**original_definition)
+        project_definition = build_project_definition(
+            **copy.deepcopy(original_definition)
```
sfc-gh-melnacouzi, line 281: I haven't see Pydantics modify the passed in object. Is the copy necessary here or is build_project_definition modifying it now?

sfc-gh-cgorrie (commented):
LGTM, but Michel's comments need addressing.

sfc-gh-turbaszek:
> 1. Have we considered having the flag on the mixin itself?

The override behavior is entity specific so we would have to change `use_mixins` to something like
```
use_mixins:
   - name: foo
     action: override
```

Possibly it's easier implementation. 

> 2. We don't have only 2 options. We could override, we could append, we could keep (!override). Do we still want to use "!" in the name?

The implementation was based on docker compose approach.

src/snowflake/cli/api/utils/definition_rendering.py
sfc-gh-turbaszek: It's necessary because applying mixin was changing YamlOverride to dictionary in original definition thus removing the necessary information.

src/snowflake/cli/api/project/definition.py
sfc-gh-turbaszek: Good catch! Fixed 👌

sfc-gh-astus (approved):

PR #1574, opened 2024-09-13
Add `snow ws version list` command

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Adds `snow ws version list` command to be invoked on an app package entity. Behaves the same as `snow app version list`.

### DIFF
--- a/src/snowflake/cli/_plugins/workspace/commands.py
+++ b/src/snowflake/cli/_plugins/workspace/commands.py
@@ -30,7 +30,7 @@ from snowflake.cli._plugins.nativeapp.common_flags import (
 from snowflake.cli._plugins.workspace.manager import WorkspaceManager
 from snowflake.cli.api.cli_global_context import get_cli_context
 from snowflake.cli.api.commands.decorators import with_project_definition
-from snowflake.cli.api.commands.snow_typer import SnowTyper
+from snowflake.cli.api.commands.snow_typer import SnowTyperFactory
 from snowflake.cli.api.entities.common import EntityActions
 from snowflake.cli.api.exceptions import IncompatibleParametersError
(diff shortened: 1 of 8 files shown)


sfc-gh-melnacouzi (commented):
Can we please revisit this?
I don't think we should use snow ws version-list. We should use snow ws version list instead.

sfc-gh-fcampbell:
Yeah I'm fine with it, that's what the auto-generation is gonna do anyways

sfc-gh-cgorrie (dismissed):

src/snowflake/cli/_plugins/workspace/commands.py
```diff
@@ -224,3 +224,31 @@ def validate(
         entity_id,
         EntityActions.VALIDATE,
     )
+
+
+version = SnowTyperFactory(
+    name="version",
+    help="Manages versions for project entities.",
```
sfc-gh-gbloom, line 231:
```suggestion
    help="Manages versions for application package entities.",
```
or something similar

sfc-gh-gbloom (commented):
Let's add integration test coverage as well?

src/snowflake/cli/_plugins/workspace/commands.py
sfc-gh-fcampbell: this command group isn't specific to just application packages though, we could have other entity types that would support version operations

sfc-gh-gbloom: Right, my bad. I'm not sure what "project entities" exactly means, but it probably doesn't really matter as we'll soon move the help messages to the concrete actions.

sfc-gh-melnacouzi (approved):

PR #1581, opened 2024-09-17
Move entities definitions to respective plugins

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
The change moves entities definitions and entities models from V2 implementation to respective plugins. The change is motivated by:
1. Avoiding coupling between CLI core and plugin definition
2. It's a step necessary to support external plugins adding new entities
3. Object from one domain (snowpark, streamlit) are encapsulated in a single place (plugin).

Additionally V1 models were moved under `v1` module to indicate that those should not longer be developed. It will also help with finding V2 dependencies on V1.


sfc-gh-astus (dismissed):

src/snowflake/cli/api/identifiers.py
```diff
@@ -142,8 +141,10 @@ def from_identifier_model_v1(cls, model: ObjectIdentifierBaseModel) -> "FQN":
         return fqn.set_database(model.database).set_schema(model.schema_name)
 
     @classmethod
-    def from_identifier_model_v2(cls, model: Identifier) -> "FQN":
+    def from_identifier_model_v2(cls, model) -> "FQN":
```
sfc-gh-fcampbell, line 144: how come this type annotation was removed?

sfc-gh-turbaszek: There's cyclic import. I will solve it in followup PR

sfc-gh-fcampbell (approved):

PR #1583, opened 2024-09-17
Fix negative execution time

sfc-gh-melnacouzi:
### Pre-review checklist

### Changes description
Fix negative execution time by capturing all exceptions.

Tested by adding print commands, and ctrl+C on the command line.
I got negative before my change.
I got proper value after my change.


src/snowflake/cli/api/commands/snow_typer.py
```diff
@@ -105,7 +105,7 @@ def command_callable_decorator(*args, **kw):
                     result = command_callable(*args, **kw)
                     self.process_result(result)
                     execution.complete(ExecutionStatus.SUCCESS)
-                except Exception as err:
+                except BaseException as err:
```
sfc-gh-turbaszek, line 108: Is it bulletproof? We were thinking about registering exit handlers.

Is it possible to add tests?

sfc-gh-melnacouzi: can always be forcefully killed - but in that case, we wouldn't get metrics at all.
I will add a test for this.

sfc-gh-melnacouzi: Added unit tests

sfc-gh-mchok, line 108: Wondering if it might be simpler to move the `execution.complete()` calls to the finally block instead? That way you don't need to worry about catching the wrong exceptions, and you guarantee calling the logic to set the exit time

sfc-gh-melnacouzi: There are other stuff that need to be handled during the exception, like setting complete with a failure status, and running the exception handler. This way, we ensure it goes in the right path during the exception (not only fix the time)

sfc-gh-turbaszek (approved):

PR #1589, opened 2024-09-18
Remove owner check in native app logic

sfc-gh-bdufour:
### Pre-review checklist
   * [] I've added or updated integration tests to verify correctness of my new code.

### Changes description

Remove ownership checks in snow app command logic. This logic is too restrictive for some advanced use cases. We should rely on RBAC for privileges rather than enforcing a stricter permission model. Fixes SNOW-1652272. 

### Testing

- Updated the current unit tests to reflect the new behaviour
- Manually verified correct and incorrect privileges by executing the CLI and forcing application upgrades as well as version drops with a different user.
- I tried writing integ tests to cover the behaviour change, but it turned out to be more complex than I expected to set them up correctly (and clean up). I decided to post the PR without integ tests at least for now, and try to backfill them in a subsequent change so that this bug fix could make the next release.


src/snowflake/cli/_plugins/nativeapp/application_entity.py
```diff
@@ -221,9 +220,6 @@ def drop(
             )
             return
 
-        # 2. Check for the right owner
-        ensure_correct_owner(row=show_obj_row, role=app_role, obj_name=app_name)
-
         # 3. Check if created by the Snowflake CLI
```
sfc-gh-cgorrie, line 223: Numbered comment is wrong now

src/snowflake/cli/_plugins/nativeapp/application_package_entity.py
```diff
@@ -449,11 +448,6 @@ def create_app_package(
         )
 
         if show_obj_row:
-            # 1. Check for the right owner role
-            ensure_correct_owner(
-                row=show_obj_row, role=package_role, obj_name=package_name
-            )
-
             # 2. Check distribution of the existing application package
```
sfc-gh-cgorrie, line 451: Numbered comment is wrong now

src/snowflake/cli/_plugins/nativeapp/application_entity.py
sfc-gh-bdufour: fixed

src/snowflake/cli/_plugins/nativeapp/application_package_entity.py
sfc-gh-bdufour: fixed

PR #1621, opened 2024-09-24
SNOW-1653357 Remove deploy_to_scratch_stage_fn() and call deploy() directly

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Now that `ApplicationPackageEntity` has a static `deploy()` method, we can just call that instead of having to pass a callback from the `NativeAppManager` to deploy the app files to the scratch stage when doing setup script validation. There's a bit of a an explosion in required params, but it's temporary and it'll be cleaned up when the v1 commands are made to operate on v2 entities directly.

### DIFF
--- a/src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
+++ b/src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
@@ -4,7 +4,7 @@ import json
 from contextlib import contextmanager
 from pathlib import Path
 from textwrap import dedent
-from typing import Callable, List, Literal, Optional, Union
+from typing import List, Literal, Optional, Union
 
 import typer
 from click import BadOptionUsage, ClickException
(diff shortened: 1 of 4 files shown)


src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
```diff
@@ -1175,20 +1214,54 @@ def validate_setup_script(
...
             stage_fqn = scratch_stage_fqn
-            deploy_to_scratch_stage_fn()
+            cls.deploy(
+                console=console,
+                project_root=project_root,
+                deploy_root=deploy_root,
+                bundle_root=bundle_root,
+                generated_root=generated_root,
+                artifacts=artifacts,
+                bundle_map=None,
```
sfc-gh-melnacouzi, line 1250: if we require bundle map here, can't we reduce the number parameters?

src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
```diff
@@ -439,12 +442,24 @@ def deploy(
...
+                recursive=recursive,
+                paths=paths,
                 stage_fqn=stage_fqn,
+                package_warehouse=package_warehouse,
+                post_deploy_hooks=post_deploy_hooks,
+                package_scripts=package_scripts,
+                policy=policy,
                 use_scratch_stage=False,
                 scratch_stage_fqn="",
-                deploy_to_scratch_stage_fn=lambda *args: None,
```
sfc-gh-melnacouzi, line 447: how is this translated? Do we skip because scratch_stage_fqn is empty? if so, should we use None instead to be more explicit?

sfc-gh-fcampbell: `use_scratch_stage=False`, so this doesn't run: https://github.com/snowflakedb/snowflake-cli/pull/1621/files/647e730c5df8cff0bd20b7ee849e045513041a18#diff-65efa774f47f1a181d9d65e29c111f229d7e6bd7eef48e70ec5fc81042bf3beeR1241-R1243, meaning that we run validation on the main app stage that was already deployed. I can change this to `None`, but I'm not a fan of `Optional[str]` when an empty string means the same thing (since in this case an empty stage name isn't valid)

src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
sfc-gh-fcampbell: the `deploy()` method can take a `bundle_map` if you have one or call `build_bundle()` to make one. in this case we don't have one, so we'd have to call `build_bundle()` with the same params anyways so it doesn't make a difference

sfc-gh-melnacouzi (approved):

PR #1631, opened 2024-09-26
Defer fetching role and warehouse until they're needed

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
We have some commands, like `snow app bundle`/`snow ws bundle` that don't even need a connection, so let's defer connecting to Snowflake until it's actually needed by the particular action being run.

For now, all `snow ws` commands are still marked as requiring a connection, that will be fixed separately.

### DIFF
--- a/src/snowflake/cli/_plugins/workspace/manager.py
+++ b/src/snowflake/cli/_plugins/workspace/manager.py
@@ -31,13 +31,6 @@ class WorkspaceManager:
         self._entities_cache: Dict[str, Entity] = {}
         self._project_definition: DefinitionV20 = project_definition
         self._project_root = project_root
-        self._default_role = default_role()
-        if self._default_role is None:
-            self._default_role = get_sql_executor().current_role()
-        self.default_warehouse = None
-        cli_context = get_cli_context()
-        if cli_context.connection.warehouse:
-            self.default_warehouse = to_identifier(cli_context.connection.warehouse)
 
     def get_entity(self, entity_id: str):
         """
@@ -62,8 +55,8 @@ class WorkspaceManager:
             action_ctx = ActionContext(
                 console=cc,
                 project_root=self.project_root(),
-                default_role=self._default_role,
-                default_warehouse=self.default_warehouse,
+                get_default_role=_get_default_role,
+                get_default_warehouse=_get_default_warehouse,
                 get_entity=self.get_entity,
             )
             return entity.perform(action, action_ctx, *args, **kwargs)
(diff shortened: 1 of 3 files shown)


src/snowflake/cli/_plugins/workspace/manager.py
```diff
@@ -31,13 +33,24 @@ def __init__(self, project_definition: ProjectDefinition, project_root: Path):
...
         self._project_root = project_root
-        self._default_role = default_role()
-        if self._default_role is None:
-            self._default_role = get_sql_executor().current_role()
-        self.default_warehouse = None
-        cli_context = get_cli_context()
-        if cli_context.connection.warehouse:
-            self.default_warehouse = to_identifier(cli_context.connection.warehouse)
+        self._default_role = self._UNSET
+        self._default_warehouse = self._UNSET
```
sfc-gh-fcampbell, line 37: `self.default_warehouse` wasn't used externally, so it's safe to change it to `self._default_warehouse`

src/snowflake/cli/_plugins/workspace/manager.py
```diff
@@ -31,13 +33,24 @@ def __init__(self, project_definition: ProjectDefinition, project_root: Path):
...
+
+    def _get_default_role(self):
```
sfc-gh-bdufour, line 39: nit: could you add a return type annotation? I think it's important here to label whether the return value is optional or not

src/snowflake/cli/_plugins/workspace/manager.py
```diff
@@ -31,13 +33,24 @@ def __init__(self, project_definition: ProjectDefinition, project_root: Path):
...
+        return self._default_role
+
+    def _get_default_warehouse(self):
+        if self._default_warehouse is self._UNSET:
+            cli_context = get_cli_context()
+            if cli_context.connection.warehouse:
+                self._default_warehouse = to_identifier(
+                    cli_context.connection.warehouse
+                )
+        return self._default_warehouse
```
sfc-gh-bdufour, line 53: there is no connection warehouse, isn't this going to return UNSET? That doesn't sound right. The role logic doesn't have this issue because if assigns `None` even if the fallback logic fails to produce a valid role.

src/snowflake/cli/_plugins/workspace/manager.py
sfc-gh-fcampbell: done

src/snowflake/cli/_plugins/workspace/manager.py
sfc-gh-fcampbell: @sfc-gh-bdufour good catch, I updated the implementation to use `@cached_property` in the `ActionContext` since we don't need to be caching them in `WorkspaceManager`

sfc-gh-bdufour (approved):

PR #1636, opened 2024-09-27
Disable snowflake.local.yml migration by default, add flag to enable it

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Fixes bug with `snow helpers v1-to-v2` if a `snowflake.local.yml` is present. Currently, the local file isn't renamed, so a v1 local can be overlaid onto a v2 main definition file after migration, which breaks subsequent commands. Since local overrides are no longer a desired feature in PDFv2, we don't convert them unless `--migrate-local-overrides` is used.

### DIFF
--- a/src/snowflake/cli/_plugins/helpers/commands.py
+++ b/src/snowflake/cli/_plugins/helpers/commands.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 import typer
 import yaml
 from snowflake.cli.api.commands.snow_typer import SnowTyperFactory
+from snowflake.cli.api.console import cli_console as cc
 from snowflake.cli.api.output.types import MessageResult
 from snowflake.cli.api.project.definition_conversion import (
     convert_project_definition_to_v2,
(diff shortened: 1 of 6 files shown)


src/snowflake/cli/_plugins/helpers/commands.py
```diff
@@ -35,10 +36,24 @@ def v1_to_v2(
     accept_templates: bool = typer.Option(
         False, "-t", "--accept-templates", help="Allows the migration of templates."
     ),
+    migrate_local_yml: bool = typer.Option(
+        False,
+        "-l",
+        "--migrate-local-overrides",
```
sfc-gh-bdufour, line 42: we might have to wordsmith this to explain that we'll include it in the conversion but not produce a separate file / a separate set of entities for it.

src/snowflake/cli/_plugins/helpers/commands.py
```diff
@@ -35,10 +36,24 @@ def v1_to_v2(
     accept_templates: bool = typer.Option(
         False, "-t", "--accept-templates", help="Allows the migration of templates."
     ),
+    migrate_local_yml: bool = typer.Option(
+        False,
```
sfc-gh-bdufour, line 40: As discussed offline, would prefer to make this abort the migration if a snowflake.local.yml is present, and make the user choose the behaviour explicitly.

sfc-gh-fcampbell: updated

src/snowflake/cli/_plugins/helpers/commands.py
sfc-gh-fcampbell: updated

sfc-gh-bdufour (approved):

sfc-gh-pjafari (approved):

sfc-gh-astus (approved):

PR #1637, opened 2024-09-27
Find deeply-nested manifest.yml when converting v1 to v2

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Fixes the manifest not being found if it isn't explicitly mentioned as an artifact or as part of a glob when running `snow helpers v1-to-v2`.

### DIFF
--- a/src/snowflake/cli/api/project/definition_conversion.py
+++ b/src/snowflake/cli/api/project/definition_conversion.py
@@ -6,7 +6,7 @@ from typing import Any, Dict, Literal, Optional
 
 from click import ClickException
 from snowflake.cli._plugins.nativeapp.artifacts import (
-    build_bundle,
+    BundleMap,
 )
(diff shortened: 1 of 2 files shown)


src/snowflake/cli/api/project/definition_conversion.py
```diff
@@ -216,17 +216,22 @@ def _find_manifest():
         # glob patterns. The simplest solution is to bundle the app and find the
         # manifest file from the resultant BundleMap, since the bundle process ensures
         # that only a single source path can map to the corresponding destination path
-        try:
-            bundle_map = build_bundle(
-                project_root, Path(native_app.deploy_root), native_app.artifacts
-            )
```
sfc-gh-fcampbell, line 222: switched from `build_bundle` to just `BundleMap` since we don't actually need to copy the artifacts to find the manifest, we just need to resolve the mapping between source and destination

sfc-gh-turbaszek (approved):

PR #1651, opened 2024-10-01
introduce subclasses and add more telemetry for ProgrammingErrors

sfc-gh-mchok:
### Pre-review checklist

### Changes description
1. Implemented custom errors derived from ClickException to replace places where we raise ProgrammingError
2. Added more telemetry (errno, sqlstate, error cause) to error events
3. Add shared code ownership to a couple of shared test files


src/snowflake/cli/_app/telemetry.py
```diff
@@ -81,6 +86,33 @@ class TelemetryEvent(Enum):
 TelemetryDict = Dict[Union[CLITelemetryField, TelemetryField], Any]
 
 
+def _is_cli_exception(exception: Exception) -> bool:
+    return issubclass(
+        exception.__class__,
+        (click.ClickException, typer.Exit, typer.Abort, BrokenPipeError),
+    )
```
sfc-gh-fcampbell, line 93:
```suggestion
    return isinstance(
        exception,
        (click.ClickException, typer.Exit, typer.Abort, BrokenPipeError),
    )
```

src/snowflake/cli/_app/telemetry.py
```diff
@@ -81,6 +86,33 @@ class TelemetryEvent(Enum):
...
+    """
+    additional_info = {}
+
+    if isinstance(exception, ProgrammingError):
+        additional_info[CLITelemetryField.ERROR_CODE] = exception.errno
+        additional_info[CLITelemetryField.SQL_STATE] = exception.sqlstate
+
+    if isinstance(exception.__cause__, ProgrammingError):
+        cause = exception.__cause__
+        additional_info[CLITelemetryField.ERROR_CAUSE] = cause.__class__.__name__
```
sfc-gh-fcampbell, line 109: how come we don't add this in the first `if`?

src/snowflake/cli/_app/telemetry.py
```diff
@@ -81,6 +86,33 @@ class TelemetryEvent(Enum):
...
+    Attach the errno and sqlstate if the exception or the
+    cause of the exception is a ProgrammingError
+    """
+    additional_info = {}
+
+    if isinstance(exception, ProgrammingError):
+        additional_info[CLITelemetryField.ERROR_CODE] = exception.errno
+        additional_info[CLITelemetryField.SQL_STATE] = exception.sqlstate
+
+    if isinstance(exception.__cause__, ProgrammingError):
```
sfc-gh-fcampbell, line 107: this will override the fields if the exception is a `ProgrammingError` and so is the cause, is that what we want?

tests/test_data/projects/napp_project_with_incorrect_pkg_warehouse/snowflake.yml
```diff
@@ -0,0 +1,15 @@
+definition_version: 1
```
sfc-gh-fcampbell, line 1: can we use `PdfV11Factory` instead?

tests_integration/nativeapp/test_telemetry_errors.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize(
+    "project_definition_files,command",
+    [("napp_post_deploy_missing_file", ["app", "deploy"])],
+    indirect=["project_definition_files"],
+)
+@mock.patch("snowflake.connector.telemetry.TelemetryClient.try_add_log_to_batch")
+def test_not_ProgrammingError_does_not_attach_any_info(
```
sfc-gh-fcampbell, line 32:
```suggestion
def test_not_programmingerror_does_not_attach_any_info(
```

tests_integration/nativeapp/test_telemetry_errors.py
```diff
@@ -0,0 +1,116 @@
...
+            )
+
+            assert CLITelemetryField.ERROR_CODE.value not in message
+            assert CLITelemetryField.SQL_STATE.value not in message
+            assert CLITelemetryField.ERROR_CAUSE.value not in message
+
+
+@pytest.mark.integration
+@mock.patch("snowflake.connector.telemetry.TelemetryClient.try_add_log_to_batch")
+def test_ProgrammingError_attaches_errno_and_sqlstate(
```
sfc-gh-fcampbell, line 59:
```suggestion
def test_programmingerror_attaches_errno_and_sqlstate(
```

tests_integration/nativeapp/test_telemetry_errors.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize(
+    "project_definition_files,command",
+    [("napp_project_with_incorrect_pkg_warehouse", ["app", "deploy"])],
+    indirect=["project_definition_files"],
+)
+@mock.patch("snowflake.connector.telemetry.TelemetryClient.try_add_log_to_batch")
+def test_ProgrammingError_cause_attaches_errno_and_sqlstate(
```
sfc-gh-fcampbell, line 88:
```suggestion
def test_programmingerror_cause_attaches_errno_and_sqlstate(
```

src/snowflake/cli/api/entities/utils.py
```diff
@@ -46,35 +50,13 @@ def generic_sql_error_handler(
...
-                Received error message '{err.msg}' while executing SQL statement.
-                '{role}' may not have access to warehouse '{warehouse}'.
-                Please grant usage privilege on warehouse to this role.
-                """
-            ),
-            errno=err.errno,
-        )
+        raise WarehouseDoesNotExistOrRoleUnauthorizedError(
+            warehouse=warehouse, role=role, msg=err.msg
+        ) from err
```
sfc-gh-pjafari, line 55: We gotta do something about this generic error handler :D 

Let's start by renaming it, because it's not a generic _sql error_ handler; we are only invoking it when we catch _programming errors_ (there are other exception types that connectors can raise in response to a sql call). So it's really a programming error handler.

Secondly, it seems that all three handled error codes here (warehouse does not exist, no warehouse selected, and does not exist or unauthorized) are classified as "not our fault", since we are raising them as ClickExceptions. Is that correct? If yes, can we extract these checks into a util: "is it our fault"; and call it at the top of this programming error handler? That way, we can add more error types that are "not our fault" to that util. That helps translate the mental model of exception classes (def our fault, def theirs, we don't know) into code. It's not a perfect translation, but it's a start!

Lastly, same error code returned from connector can be caused in different contexts. e.g. 
`use warehouse non_existant_warehouse` will return the same error code as `use role non_existant_role`. So we need context passed in to this handler from where the programming error is caught into this handler to be able to raise the proper error type here. 
  
@sfc-gh-bdufour thoughts?

tests/app/test_telemetry.py
```diff
@@ -147,3 +153,24 @@ def test_failing_executing_command_sends_telemetry_data(
         and result_command_event["message"]["command_execution_id"]
         == "8a2225b3800c4017a4a9eab941db58fa"
     )
+
+
+@pytest.mark.parametrize(
+    "error,expected",
```
sfc-gh-pjafari, line 159: super nit - "error, is_cli"

tests/test_data/projects/napp_project_with_incorrect_pkg_warehouse/snowflake.yml
sfc-gh-pjafari: Thank you! :D You'll want to use the `ProjectV11Factory`, I see you have other files for this project.

src/snowflake/cli/_app/telemetry.py
```diff
@@ -81,6 +86,33 @@ class TelemetryEvent(Enum):
 TelemetryDict = Dict[Union[CLITelemetryField, TelemetryField], Any]
 
 
+def _is_cli_exception(exception: Exception) -> bool:
+    return issubclass(
+        exception.__class__,
+        (click.ClickException, typer.Exit, typer.Abort, BrokenPipeError),
```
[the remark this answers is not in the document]
sfc-gh-bdufour: Correct, so "CLI" here means handled = not us.

src/snowflake/cli/_app/telemetry.py
sfc-gh-bdufour: Yeah, good catch. IMHO the logic should be:

```
if programming error:
    log errno/sqlstate
if cause:
    log cause
    if cause is programming error:
       log cause's errno/sqlstate if not already present
```

src/snowflake/cli/_app/telemetry.py
sfc-gh-bdufour: Marcus, see previous comment for the suggested flow.

src/snowflake/cli/_plugins/nativeapp/exceptions.py
```diff
@@ -111,3 +111,15 @@ class NoEventTableForAccount(ClickException):
 
     def __init__(self):
         super().__init__(f"{self.__doc__}\n\n{self.INSTRUCTIONS}")
+
+
+class DistributionAttributeNotFoundError(ClickException):
```
sfc-gh-bdufour, line 116: We can probably make that more generic. Something like "ObjectPropertyNotFound" or "UnexpectedApplicationPackageProperties" ?

src/snowflake/cli/api/entities/utils.py
sfc-gh-bdufour: Good thoughts, happy to discuss more, but I'd hold off on the suggested refactorings so that we can have a shot at including this at the very next opportunity (3.0.1 if possible).

src/snowflake/cli/api/exceptions.py
```diff
@@ -188,3 +189,54 @@ def __init__(self, options: list[str]):
         super().__init__(
             f"Parameters {comma_separated_options} and {options_with_quotes[-1]} are incompatible and cannot be used simultaneously."
         )
+
+
+class WarehouseDoesNotExistOrRoleUnauthorizedError(ClickException):
```
sfc-gh-bdufour, line 194: Why is warehouse special? We can probably just merge this with the more generic "DoesNotExist..." below.

src/snowflake/cli/api/entities/utils.py
sfc-gh-pjafari: Sure. For the sake of getting this PR in then, I'd say at minimum do the renaming. I'd also like to also see the separation of the log based on the context (for the same error code), but since we are calling this handler only in specific places I'm ok with marking that as TODO and fixing it in a separate PR.

sfc-gh-bdufour: Even the renaming doesn't buy us much. I'd fold that into the next work, since we'll pick it up right away.

sfc-gh-mchok: I added this comment thread to a new issue that I'll pick up right after, please feel free to add any other ideas: https://snowflakecomputing.atlassian.net/browse/SNOW-1722580

tests_integration/conftest.py
```diff
@@ -27,23 +27,22 @@
...
-from snowflake.cli._app.cli_app import app_factory
-from typer import Typer
-from typer.testing import CliRunner
-
 from snowflake.cli.api.project.util import TEST_RESOURCE_SUFFIX_VAR
 from tests.conftest import clean_logging_handlers_fixture  # noqa: F401
+from tests.testing_utils.files_and_dirs import merge_left
 from tests.testing_utils.fixtures import (
-    alter_snowflake_yml,  # noqa: F401
+    # noqa: F401
```
sfc-gh-fcampbell, line 43: nit: F401 is "unused import", but it's been removed so the `noqa` can be removed too

sfc-gh-fcampbell (approved):

sfc-gh-bdufour (approved):

tests_integration/conftest.py
sfc-gh-mchok: I'm not entirely sure why that import got removed, I'm guessing my IDE or pre-commit did it automatically, but all the tests are passing so I guess I'll remove the comment as well and hope for the best 👍

sfc-gh-fcampbell (dismissed):

sfc-gh-fcampbell (approved):

sfc-gh-fcampbell (approved):

src/snowflake/cli/_app/telemetry.py
```diff
@@ -81,6 +86,37 @@ class TelemetryEvent(Enum):
...
+
+def _get_additional_exception_information(exception: Exception) -> TelemetryDict:
+    """
+    Attach the errno and sqlstate if the exception or the
+    cause of the exception is a ProgrammingError
+    """
+    additional_info = {}
+
+    if isinstance(exception, ProgrammingError):
+        additional_info[CLITelemetryField.ERROR_CODE] = exception.errno
```
sfc-gh-turbaszek, line 104: Can we please cast it to integer?

sfc-gh-turbaszek, line 104:
```suggestion
        additional_info[CLITelemetryField.ERROR_CODE] = int(exception.errno)
```

src/snowflake/cli/_app/telemetry.py
```diff
@@ -81,6 +86,37 @@ class TelemetryEvent(Enum):
...
+        additional_info[CLITelemetryField.SQL_STATE] = exception.sqlstate
```
sfc-gh-turbaszek, line 105: What is `sqlstate`? Is it enum? Any chances of passing sensitive data under this key?

src/snowflake/cli/_plugins/nativeapp/exceptions.py
```diff
@@ -111,3 +111,15 @@ class NoEventTableForAccount(ClickException):
...
+class ObjectPropertyNotFoundError(ClickException):
+    def __init__(self, property_name: str, object_type: str, object_name: str):
+        super().__init__(
+            dedent(
+                f"""\
+                        Could not find the '{property_name}' attribute for {object_type} {object_name} in the output of SQL query:
+                        'describe {object_type} {object_name}'
+                        """
+            )
+        )
```
sfc-gh-turbaszek, line 125: What does exactly this error represents? What users should do if they see such an error?

sfc-gh-turbaszek: How different is it from `DoesNotExistOrUnauthorizedError`?

src/snowflake/cli/api/entities/utils.py
```diff
@@ -41,40 +44,15 @@
...
-                Received error message '{err.msg}' while executing SQL statement.
-                '{role}' may not have access to warehouse '{warehouse}'.
-                Please grant usage privilege on warehouse to this role.
-                """
-            ),
-            errno=err.errno,
-        )
+    if (
+        err.errno == DOES_NOT_EXIST_OR_CANNOT_BE_PERFORMED
+        or "does not exist or not authorized" in err.msg
```
sfc-gh-turbaszek, line 51: Why `errno` is not enough?

src/snowflake/cli/api/exceptions.py
```diff
@@ -188,3 +189,41 @@ def __init__(self, options: list[str]):
...
+        super().__init__(
+            f"Could not use {object_type} {name}. Object does not exist, or operation cannot be performed."
+        )
+
+
+class ShowSpecificObjectMultipleRowsError(ClickException):
+    def __init__(self, show_obj_query: str):
+        super().__init__(
+            f"Received multiple rows from result of SQL statement: {show_obj_query}. Usage of 'show_specific_object' may not be properly scoped."
+        )
```
sfc-gh-turbaszek, line 229: Is this user facing error? How should they act if they see "Usage of 'show_specific_object' may not be properly scoped." message?

sfc-gh-turbaszek:
Should we use `generic_sql_error_handler` for `SqlExecutionMixin_execute_string` or in `SnowTyper.post_execute` where we rewrite db errors into click exceptions?

src/snowflake/cli/_app/telemetry.py
sfc-gh-mchok: The type hints say it already should be an integer or None, in which case the casting would throw an error:
image

src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
```diff
@@ -702,7 +703,7 @@ def get_existing_version_info(
                 if err.msg.__contains__("does not exist or not authorized"):
                     raise ApplicationPackageDoesNotExistError(package_name)
                 else:
-                    generic_sql_error_handler(err=err, role=package_role)
```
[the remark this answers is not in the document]
sfc-gh-mchok: In refactoring `generic_sql_error_handler` we combined some similar errors to be more generic, so we didn't need to pass that as an argument anymore

sfc-gh-mchok:
> Should we use `generic_sql_error_handler` for `SqlExecutionMixin_execute_string` or in `SnowTyper.post_execute` where we rewrite db errors into click exceptions?

We are currently in the process of refactoring our error handling and one of our goals is actually to phase out the usage of `generic_sql_error_handler` since we want our error handling to be more specific and take into the context of when the error happened in order to provide more rich error messages/handling

src/snowflake/cli/api/entities/utils.py
sfc-gh-mchok: I did not write the original code and the person who did is on leave for a couple of weeks, so it's hard to say, but there may be some errors raised where the errno is not set or does not match what is expected. However, the two cases are similar enough that we decided to refactor the two branches into one and throw a more generic error

src/snowflake/cli/api/exceptions.py
sfc-gh-mchok: Just looking through the code, it seems like this is an error that is meant for us who is using the function, not the end users, and as the message says we need to make sure that our usage of this function is properly scoped (e.g. passing the right parameters)

src/snowflake/cli/_plugins/nativeapp/exceptions.py
sfc-gh-mchok: Sorry, both this (DoesNotExistOrUnauthorizedError) and ShowSpecificObjectMultipleRowsError should not be ClickExceptions since they are caused by us and not the user, just changed them to RuntimeErrors, thanks for the catch!

src/snowflake/cli/_app/telemetry.py
```diff
@@ -83,6 +88,43 @@ class TelemetryEvent(Enum):
...
+
+    if exception.__cause__:
+        cause = exception.__cause__
+        additional_info[CLITelemetryField.ERROR_CAUSE] = type(cause).__name__
```
sfc-gh-melnacouzi, line 117: what if we don't have a cause, don't we want to fallback to using the exception type itself?

src/snowflake/cli/_app/telemetry.py
```diff
@@ -224,14 +266,15 @@ def log_command_result(execution: ExecutionMetadata):
...
 def log_command_execution_error(exception: Exception, execution: ExecutionMetadata):
     exception_type: str = type(exception).__name__
-    is_cli_exception: bool = issubclass(exception.__class__, click.ClickException)
+    is_cli_exception: bool = _is_cli_exception(exception)
     _telemetry.send(
         {
             TelemetryField.KEY_TYPE: TelemetryEvent.CMD_EXECUTION_ERROR.value,
             CLITelemetryField.COMMAND_EXECUTION_ID: execution.execution_id,
             CLITelemetryField.ERROR_TYPE: exception_type,
             CLITelemetryField.IS_CLI_EXCEPTION: is_cli_exception,
```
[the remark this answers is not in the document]
sfc-gh-mchok: is_cli_exception = true -> handled/user error, and vice versa, I believe Snowcli introduced this naming themselves and are using it for their own dashboards

The line right above:
```
CLITelemetryField.ERROR_TYPE: exception_type,
```
contains the exception type which is what you're asking about I believe

src/snowflake/cli/_app/telemetry.py
sfc-gh-mchok: Is already being tracked, see https://github.com/snowflakedb/snowflake-cli/pull/1651#discussion_r1805238140

sfc-gh-melnacouzi (approved):

sfc-gh-turbaszek (approved):

PR #1652, opened 2024-10-01
Pass context in entity init instead of only to actions

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
We want to allow calling methods on entities but not making them actions exposed in the CLI. For some methods, it would be useful to have access to utilities like `ctx.default_role`, etc, so let's make context a property of the entity set when instantiated by the `WorkspaceManager`.


sfc-gh-bdufour:
Hmm, not sure about this one yet. It made sense to me to decouple the entity model from the action state, but this couples them again. Can you give me specifics about why we want to access things like `default_role` outside of an action method?

sfc-gh-fcampbell:
I'm trying to get rid of `NativeAppManager`, `NativeAppProjectModel`, etc since those classes can only handle the v1 data model, and make the `snow app` command operate directly on v2 entities. The problem is that these commands don't just call `action_` methods on entities, they need supporting methods like `get_existing_app_info`, `get_snowsight_url`, etc.

Right now, for an operation like `get_existing_app_info`, `NativeAppManager` calls a static method on the `ApplicationEntity`, passing in the app name and role to use:
```python
    def get_existing_app_info(self) -> Optional[dict]:
        return ApplicationEntity.get_existing_app_info(
            app_name=self.app_name,
            app_role=self.app_role,
        )
```
```python
    @staticmethod
    def get_existing_app_info(app_name: str, app_role: str) -> Optional[dict]:
        sql_executor = get_sql_executor()
        with sql_executor.use_role(app_role):
            return sql_executor.show_specific_object(
                "applications", app_name, name_col=NAME_COL
            )
```
In `NativeAppManager`/`NativeAppProjectModel`, `self.app_name` and `self.app_role` are properties with logic behind them, and if we're to get rid of `NativeAppManager`/`NativeAppProjectModel`, that logic needs to move somewhere. Since the same logic already exists for the v2 model in `WorkspaceManager`/`ActionContext`, I want to leverage it instead of duplicating it in `ApplicationEntity`. 

The problem I'm facing now is that `ActionContext` is only available to "implementations" of `EntityActions`, like `action_deploy`, `action_bundle`, etc. Since I don't consider `get_existing_app_info` to be a "top-level action", it doesn't merit its own entry in `EntityActions`, meaning it can't be invoked through `WorkspaceManager.perform()` (which is where the `ActionContext` is created). I also don't want to set a precedent that any instance methods of entities _need_ to be `EntityActions`, since that would require a bunch of boilerplate for actions that only one entity type would ever implement.

My solution to this is to set the `ActionContext` as an instance property of the entity so that any methods can access the current context data. With these changes, `get_existing_app_info` would become an instance method instead of a static:
```python
    def get_existing_app_info(self) -> Optional[dict]:
        model = self._entity_model
        ctx = self._action_ctx
        role = (model.meta and model.meta.role) or ctx.default_role  #  default_role used here
        sql_executor = get_sql_executor()
        with sql_executor.use_role(role):
            return sql_executor.show_specific_object(
                "applications", model.fqn.name, name_col=NAME_COL
            )
```
and could be called without needing the whole `WorkspaceManager.perform()`/`EntityActions` mechanism:
```python
    app = ws.get_entity(app_id)
    if app.get_existing_app_info():
        typer.launch(app.get_snowsight_url())
        return MessageResult(f"Snowflake Native App opened in browser.")
    else:
        return MessageResult(
            'Snowflake Native App not yet deployed! Please run "snow app run" first.'
        )
```

src/snowflake/cli/_plugins/workspace/manager.py
```diff
@@ -53,16 +59,13 @@ def perform_action(self, entity_id: str, action: EntityActions, *args, **kwargs)
...
-                project_root=self.project_root(),
-                get_default_role=_get_default_role,
-                get_default_warehouse=_get_default_warehouse,
                 get_entity=self.get_entity,
             )
             return entity.perform(action, action_ctx, *args, **kwargs)
         else:
             raise ValueError(f'This entity type does not support "{action.value}"')
 
+    @property
```
sfc-gh-fcampbell, line 68: drive-by improvement

sfc-gh-fcampbell:
Description of changes after that long post above. A previous version of this PR was passing the whole `ActionContext` to the entity's `__init__`, but now we've decided to split that class into `WorkspaceContext` to give access to workspace-wide config, like the console, default role, etc, and keep only `get_entity` in `ActionContext`.

sfc-gh-pjafari (approved):
I like this - the first time I read `ActionContext` I was wondering why things like `get_default_role` and `get_default_warehouse` are action context. This makes more sense :)

src/snowflake/cli/_plugins/workspace/context.py
```diff
@@ -7,16 +7,17 @@
 
 
 @dataclass
-class ActionContext:
+class WorkspaceContext:
     """
-    An object that is passed to each action when called by WorkspaceManager
+    An object that is passed to each entity when instantiated by WorkspaceManager
+    to allow access to the CLI context without required the entities to use
```
sfc-gh-bdufour, line 13: "without requiring"?

sfc-gh-fcampbell: fixed

sfc-gh-bdufour (approved):

PR #1678, opened 2024-10-07
Enforce that paths used for BundleMap and NativeAppCompiler are always absolute

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Subtle cwd-dependent bugs can creep in if we work with relative paths too much. Let's enforce that all paths used to initialize `BundleMap` and `NativeAppCompiler` should be absolute so there's no doubt about where files will be read from or written to. Paths to individual files within these roots are still allowed to be relative.

### DIFF
--- a/src/snowflake/cli/_plugins/nativeapp/artifacts.py
+++ b/src/snowflake/cli/_plugins/nativeapp/artifacts.py
@@ -240,9 +240,21 @@ class BundleMap:
     """
     Computes the mapping between project directory artifacts (aka source artifacts) to their deploy root location
     (aka destination artifact). This information is primarily used when bundling a native applications project.
+
+    :param project_root: The root directory of the project and base for all relative paths. Must be an absolute path.
+    :param deploy_root: The directory where artifacts should be copied to. Must be an absolute path.
     """
 
     def __init__(self, *, project_root: Path, deploy_root: Path):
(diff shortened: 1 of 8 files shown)


src/snowflake/cli/_plugins/nativeapp/artifacts.py
```diff
@@ -240,11 +240,21 @@ class BundleMap:
...
     def __init__(self, *, project_root: Path, deploy_root: Path):
+        if not project_root.is_absolute():
+            raise ValueError("Project root must be an absolute path")
         self._project_root: Path = resolve_without_follow(project_root)
-        self._deploy_root: Path = resolve_without_follow(deploy_root)
+        abs_deploy_root = (
+            deploy_root
+            if deploy_root.is_absolute()
+            else self._project_root / deploy_root
+        )
```
sfc-gh-bdufour, line 256: Is this necessary? I thought that `p1 / p2` already returns p2 if it's absolute.

sfc-gh-fcampbell: oh yeah that's true, I was just using that in an earlier PR 😅

sfc-gh-bdufour (approved):

sfc-gh-pjafari (approved):

sfc-gh-turbaszek (approved):

PR #1695, opened 2024-10-09
Execute user script refactor

sfc-gh-pjafari:
### Pre-review checklist

### Changes description
**NOTE - This is a PR against @sfc-gh-mchok's telemetry branch, since I use some error types he's introduced**
 
Started a SQLService abstraction layer. The goal is to extract  execute_query calls that are embedded in the logic out to their own unit. This PR uses package scripts to show an example of the restructuring.  
use role, warehouse, and database calls are moved to this layer. See `test_package_scripts_with_conn_warehouse` for and example of an updated test. 

NOTE - Since the start of this work, `apply_package_scripts` is no longer invoked by any commands (in favor of post-deploy hooks). That's why there are no integration tests associated with this change, even though some error handling has changed. 
Exhaustive test cases for `execute_user_script` as well as integ tests to be added when making similar change for running `post_deploy` hooks in an upcoming PR.   
...


src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
```diff
@@ -423,16 +424,16 @@ def deploy(
             console.warning(e.message)
             if not policy.should_proceed("Proceed with using this package?"):
                 raise typer.Abort() from e
-        with get_sql_executor().use_role(package_role):
```
sfc-gh-pjafari, line 426: use_role call removed from here, moved down.

src/snowflake/cli/api/sql_contract.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+class CannotUseRoleError(CouldNotUseObjectError):
+    def __init__(self, role):
+        self.role = role
+        super().__init__(ObjectType.ROLE, role)
+
+
+class SQLService:
+    _sql_executor = get_sql_executor()
```
sfc-gh-fcampbell, line 23: nit, since the SqlExecutor is stateful (it has a reference to a Snowflake connection), let's allow passing it in via constructor and default to calling `get_sql_executor()` if it's not passed

src/snowflake/cli/api/sql_contract.py
```diff
@@ -0,0 +1,134 @@
...
+        super().__init__(f"Unknown SQL error occurred. {msg}")
+
+
+class CannotUseRoleError(CouldNotUseObjectError):
+    def __init__(self, role):
+        self.role = role
+        super().__init__(ObjectType.ROLE, role)
+
+
+class SQLService:
```
sfc-gh-fcampbell, line 22:
```suggestion
class SqlService:
```
to match our other `Sql...` classes

src/snowflake/cli/api/sql_contract.py
```diff
@@ -0,0 +1,134 @@
...
+                self._sql_executor.execute_query(f"use warehouse {prev_wh}")
+
+    @contextmanager
+    def _use_role_optional(self, new_role: str | None):
+        """
+        Switches to a different role for a while, then switches back.
+        This is a no-op if the requested role is already active or if no role is passed in.
+        """
+        if new_role is None:
+            yield
```
sfc-gh-fcampbell, line 74:
```suggestion
            yield
            return
```

src/snowflake/cli/api/sql_contract.py
```diff
@@ -0,0 +1,134 @@
...
+        )
+        if is_different_role:
+            self._sql_executor.log_debug(f"Assuming different role: {new_role}")
+            try:
+                self._sql_executor.execute_query(f"use role {new_role}")
+            except ProgrammingError as err:
+                if err.errno == DOES_NOT_EXIST_OR_CANNOT_BE_PERFORMED:
+                    raise CannotUseRoleError(new_role) from err
+                else:
+                    raise UnknownSQLError(f"Failed to use role {new_role}") from err
```
sfc-gh-fcampbell, line 88:
```suggestion
                raise UnknownSQLError(f"Failed to use role {new_role}") from err
```
no need for `else` after `raise`

src/snowflake/cli/api/sql_contract.py
```diff
@@ -0,0 +1,134 @@
...
+            except:
```
sfc-gh-fcampbell, line 89:
```suggestion
            except Exception:
```
otherwise this will catch exceptions that inherit from `BaseException` too, like `KeyboardInterrupt` and `SystemExit`

src/snowflake/cli/api/sql_contract.py
```diff
@@ -0,0 +1,134 @@
...
+                raise UnknownSQLError(f"Failed to use role {new_role}")
+        try:
+            yield
+
```
sfc-gh-fcampbell, line 93:
```suggestion
```

src/snowflake/cli/api/sql_contract.py
```diff
@@ -0,0 +1,134 @@
...
+        """
+        Switches to a different role for a while, then switches back.
+        This is a no-op if the requested role is already active or if no role is passed in.
+        """
+        if new_role is None:
+            yield
+
+        prev_role = self._sql_executor.current_role()
+        is_different_role = (
+            new_role is not None and new_role.lower() != prev_role.lower()
```
sfc-gh-fcampbell, line 78:
```suggestion
            new_role.lower() != prev_role.lower()
```
we already abort if `new_role` is None above

src/snowflake/cli/api/sql_contract.py
```diff
@@ -0,0 +1,134 @@
...
+    def execute_user_script(
+        self,
+        queries: str,
+        role: str | None = None,
+        warehouse: str | None = None,
+        database: str | None = None,
+    ):
+        with self._use_role_optional(role):
+            with self._use_warehouse_optional(warehouse):
+                self._use_database_optional(database)
```
sfc-gh-fcampbell, line 120: how come this one doesn't restore the old DB?

sfc-gh-fcampbell (commented):
just some drive-by suggestions

src/snowflake/cli/api/sql_contract.py
sfc-gh-pjafari: done

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -0,0 +1,224 @@
...
+from snowflake.cli.api.project.util import to_identifier
+from snowflake.cli.api.sql_execution import SqlExecutor
+from snowflake.connector import ProgrammingError
+
+
+class UnknownSQLError(Exception):
+    """Exception raised when the root of the SQL error is unidentified by us."""
+
+    # PJ-question how do we ensure exit codes remain unique
+    exit_code = 3
```
sfc-gh-pjafari, line 36: any thoughts/insights on this? cc @sfc-gh-fcampbell

sfc-gh-fcampbell: i've done something like this before:
```python
class UniqueExitCode(type):
	_exit_code = 3
	
	def __new__(mcs, name, bases, attrs):
		c = mcs._exit_code
		if c > 255:
			raise ValueError(f"too many uses of {mcs.__name__}, max exit code reached")
		mcs._exit_code += 1
		attrs["exit_code"] = c
		return type(name, bases, attrs)

class MyError1(Exception, metaclass=UniqueExitCode):
	pass

class MyError2(Exception, metaclass=UniqueExitCode):
	pass
```

sfc-gh-mchok: Two questions:
1. Is there potentially a race condition where the exit codes for errors could change depending on when the code is called, refactors, and new errors that are introduced?
2. What is the purpose of making exit codes unique here? I'm assuming it's to identify the type of error, but it doesn't seem like there's any relationship between the exit code and the error type since they all fall under `UnknownSQLError`?

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -0,0 +1,223 @@
...
+            except Exception as err:
+                raise UnknownSQLError(
+                    f"Failed to use warehouse {valid_wh_name}"
+                ) from err
+        try:
+            yield
+        finally:
+            if is_different_wh and prev_wh is not None:
+                self._log.debug(f"Switching back to warehouse:{prev_wh}")
+                self._sql_executor.execute_query(f"use warehouse {prev_wh}")
```
sfc-gh-bdufour, line 110: ditto, what happens if this fails? Technically the user script could drop the warehouse or otherwise make it inaccessible.

tests/nativeapp/test_package_scripts.py
```diff
@@ -50,148 +54,170 @@ def _get_na_manager(working_dir):
...
+    temp_dir,
+):
+    scripts = use_project_with_package_scripts()
+    with mock.patch.object(
+        SnowflakeSQLFacade, "execute_user_script"
+    ) as mock_execute_user_script:
+        mock_execute_user_script.return_value = None
+        native_app_manager = _get_na_manager(str(temp_dir))
+        native_app_manager._apply_package_scripts()  # noqa: SLF001
+        assert mock_execute_user_script.call_count == 2
```
sfc-gh-bdufour, line 110: nit: when this fails, we don't get a lot of information. I prefer to either write the assertion against `mock_calls`, so equality means _only_ the expected calls were present, or at least moving the count check _after_ the other assertion so that failures carry good details.

tests/nativeapp/test_sf_sql_facade.py
```diff
@@ -0,0 +1,88 @@
...
+from snowflake.cli.api.entities.common import get_sql_executor
+
+from tests.nativeapp.utils import (
+    SQL_EXECUTOR_EXECUTE,
+    SQL_EXECUTOR_EXECUTE_QUERIES,
+    mock_execute_helper,
+)
+
+# TODO: Add more test coverage + define contract
+# TODO: Add test for non-identifier names
```
sfc-gh-bdufour, line 27: fast-follow PRs? What's the plan here?

tests/nativeapp/test_sf_sql_facade.py
```diff
@@ -0,0 +1,88 @@
...
+@mock.patch(SQL_EXECUTOR_EXECUTE)
+@mock.patch(SQL_EXECUTOR_EXECUTE_QUERIES)
+def test_execute_with_role_wh_db(mock_execute_queries, mock_execute_query, mock_cursor):
+    # Arrange
+    mock_script = "-- my comment\nselect 1;\nselect 2;"
+    mock_script_name = "test-user-sql-script.sql"
+    role = "mock_role"
+    wh = "mock_wh"
+    database = "mock_db"
+    side_effects, expected = mock_execute_helper(
```
sfc-gh-bdufour, line 40: if you have a better pattern, use it instead. This is pretty much NADE legacy at this point. We could also create a new abstraction / helper for SQL testing in this file, since it will be used here almost exclusively.

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
sfc-gh-pjafari: good point - no need for it to be unique. I had the wrong idea of what we might be using it for.

src/snowflake/cli/_plugins/nativeapp/sf_facade.py
```diff
@@ -0,0 +1,7 @@
+from snowflake.cli._plugins.nativeapp.sf_sql_facade import SnowflakeSQLFacade
+from snowflake.cli.api.sql_execution import SqlExecutor
+
+
+def get_snowflake_facade(sql_executor: SqlExecutor | None) -> SnowflakeSQLFacade:
+    """Returns a Snowflake Facade"""
+    return SnowflakeSQLFacade(sql_executor)
```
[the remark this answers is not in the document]
sfc-gh-pjafari: see @sfc-gh-fcampbell's comment here https://github.com/snowflakedb/snowflake-cli/pull/1695#discussion_r1794037846
Though we are currently always creating a new instance of sql_executor where we need it. imo the way we are using it rn either way ends up the same. But we can make a design decision as to if we want to pass the sql executor in or instantiate it in `SnowflakeSQLFacade`.

tests/nativeapp/test_sf_sql_facade.py
sfc-gh-pjafari: Yes, see note on my PR description. This class is not used in any code path (since Frank has removed apply_package_script in the meantime). So more tests and integ tests to be added in the upcoming PR for post deploy hooks.

src/snowflake/cli/_plugins/nativeapp/sf_facade.py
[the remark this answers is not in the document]
sfc-gh-bdufour: IMHO the executor is pretty much an implementation detail here, and I'd only expect it to be replaced for tests so we can inspect our interaction with a low-level executor. I do like the simplicity of just acquiring a facade object (could be a global or per-thread singleton) from anywhere in the code, as long as we can easily mock it out when we need to test code that interacts with the facade.

tests/nativeapp/test_package_scripts.py
sfc-gh-pjafari: good callout - asserting against mock_calls now :)

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
sfc-gh-pjafari: good point. and the same can be said about the `use database` case. See update. made a `_use_object` method that has the error handling it and using it for all `use object_type object_name` calls.

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -0,0 +1,223 @@
...
+        """
+        if new_wh is None:
+            yield
+            return
+
+        valid_wh_name = to_identifier(new_wh)
+
+        wh_result = self._sql_executor.execute_query(
+            f"select current_warehouse()"
+        ).fetchone()
```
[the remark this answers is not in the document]
sfc-gh-bdufour: It's a weird state for sure if something fails, but I'm not convinced that ProgrammingError is the right contract. We _know_ what failed here, and it's not wrong SQL.

tests/nativeapp/test_sf_sql_facade.py
sfc-gh-pjafari: I think using this helper is OK for when we actually want to test the sql. 
And its usage in other tests will be reduced and removed as I refactor more SQL calls into this class.
Of course there are small improvements I can do to it like switching the order of expected call and side_effect but not sure how much we'll gain from that.

tests/nativeapp/test_sf_sql_facade.py
sfc-gh-bdufour: We should be able to cover the facade in isolation though...

tests/nativeapp/test_package_scripts.py
```diff
@@ -50,148 +54,169 @@ def _get_na_manager(working_dir):
...
+    PdfV10Factory.with_filename("snowflake.local.yml")(
+        native_app__package__warehouse="myapp_pkg_warehouse"
+    )
+    with mock.patch.object(
+        SnowflakeSQLFacade, "execute_user_script"
+    ) as mock_execute_user_script:
+        mock_execute_user_script.return_value = None
+        native_app_manager = _get_na_manager(str(temp_dir))
+        native_app_manager._apply_package_scripts()  # noqa: SLF001
+        assert mock_execute_user_script.call_count == 2
```
sfc-gh-bdufour, line 132: sorry, I thought "and use the same pattern everywhere else" was implied

sfc-gh-pjafari: yes that was my bad I missed that there was more instances. fixed.

tests/nativeapp/test_sf_sql_facade.py
sfc-gh-pjafari: added tests

src/snowflake/cli/_plugins/nativeapp/sf_facade.py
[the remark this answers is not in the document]
sfc-gh-pjafari: updated

sfc-gh-pjafari:
closing this PR, new PR on main here #1732

PR #1699, opened 2024-10-09
Stop running some nativeapp tests on v1 projects and ws command

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Since we now convert v1 definitions to v2 in-memory when running `snow app` command, we don't need to run all integration tests for v1 example projects anymore. Also, since the status of the `snow ws` command is uncertain (`snow app` and `snow ws` now call the same code under the hood), we're removing the `snow ws` parametrization from most tests. All in all, this reduces the Native App integration test suite from 300 unique tests to 128, which greatly speeds it up (from about 35 minutes to 19 minutes).

For safety, we're keeping full parametrization on a simple test for each feature, to make sure that the v1 to v2 conversion and `snow ws` command don't regress.


sfc-gh-bdufour:
How comprehensive are those v1 -> v2 tests? I understand that running more integ tests is slower overall, and if we aren't getting anything for them, we should remove them. However, we need to make sure that we don't let v1 be impacted by new development, as the vast majority of our users are still on v1. What's our plan to ensure that we don't break v1 in the future?

tests_integration/nativeapp/test_version.py
```diff
@@ -48,48 +48,32 @@ def normalize_identifier(identifier: Union[str, int]) -> str:
...
-
-@pytest.fixture()
-def command_parametrization(
-    _create_command, _list_command, _drop_command, _test_project
-):
-    if "v1" in _test_project and (
-        "ws" in " ".join([_create_command, _list_command, _drop_command])
-    ):
-        pytest.skip("ws commands are not supported on v1 projects")
-    return _create_command, _list_command, _drop_command, _test_project
```
sfc-gh-fcampbell, line 79: https://github.com/snowflakedb/snowflake-cli/pull/1703#discussion_r1797015109

sfc-gh-fcampbell:
> How comprehensive are those v1 -> v2 tests? I understand that running more integ tests is slower overall, and if we aren't getting anything for them, we should remove them. However, we need to make sure that we don't let v1 be impacted by new development, as the vast majority of our users are still on v1. What's our plan to ensure that we don't break v1 in the future?

Reworked this PR and updated the description to describe what we're doing now.

tests_integration/nativeapp/test_version.py
```diff
@@ -355,15 +363,21 @@ def test_nativeapp_version_create_package_no_magic_comment(
 
 # Tests a simple flow of an existing project, executing snow app version create, drop and teardown, all with distribution=internal
 @pytest.mark.integration
+@pytest.mark.parametrize(
+    "create_command,list_command,drop_command,test_project",
+    [["app version create", "app version list", "app version drop", "napp_init_v2"]],
+)
```
sfc-gh-bdufour, line 369: I assume this is for consistency, since there doesn't seem to be a need to use a parametrized test here?

sfc-gh-bdufour (approved):

tests_integration/nativeapp/test_version.py
sfc-gh-fcampbell: yeah, I wanted to keep the parametrization for two reasons, 1. it makes this PR's diff smaller, and 2. it puts the relevant info in the test signature rather than in the body, keeping the test IDs consistent for stuff like snapshot names, debugging, etc

PR #1703, opened 2024-10-10
Fix `snow ws version drop` using the wrong action

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Fixes `snow ws version drop` accidentally performing the `VERSION_CREATE` action. Parametrized tests to use both `snow app version` and `snow ws version` commands, except for unsupported use-cases (`ws` command with a v1 project).

### DIFF
--- a/tests_integration/nativeapp/test_version.py
+++ b/tests_integration/nativeapp/test_version.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from shlex import split
-from typing import Any, Optional, Union
+from typing import Any, Union
 
 from yaml import safe_dump, safe_load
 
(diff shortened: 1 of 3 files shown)


tests_integration/nativeapp/test_version.py
```diff
@@ -48,29 +48,52 @@ def normalize_identifier(identifier: Union[str, int]) -> str:
...
+
+@pytest.fixture()
+def command_parametrization(
+    _create_command, _list_command, _drop_command, _test_project
+):
+    if "v1" in _test_project and (
+        "ws" in " ".join([_create_command, _list_command, _drop_command])
+    ):
+        pytest.skip("ws commands are not supported on v1 projects")
+    return _create_command, _list_command, _drop_command, _test_project
```
sfc-gh-bdufour, line 79: oof, maybe I'm just tired, but I find it hard to wrap my head around the combinations being assembled here. Does that mean we'll test all permutations of each of the 4 params?

sfc-gh-fcampbell: this fixture requests 4 fixtures, each of which is parametrized to return both the `snow app` and `snow ws` version of the command, for a total of 16 distinct test cases. Since `snow ws` can't operate on v1 projects, we skip the test in all the permutations where one of the commands is `snow ws`. In practice, this results in 9 test cases, 8 for v2 projects and 1 for a v1 project _and_ where the command sequence goes `snow app version create`, `snow app version list`, `snow app version drop`.

sfc-gh-pjafari (approved):

tests_integration/nativeapp/test_version.py
sfc-gh-bdufour: Not a fan of this pattern TBH, IMHO it makes it really tricky to grok what's being covered and what might be missing. I see this is merged already. I'd prefer not reusing this pattern elsewhere.

sfc-gh-fcampbell: Yeah sorry, forgot to disable the auto-merge. I'm gonna remove this when I do my test parametrization pruning.

PR #1722, opened 2024-10-15
SNOW-1706990 Convert test_manager.py to use v2 entities

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Updates tests in `test_manager.py` to use a v2 test project and call out to the v2 entities directly instead of going through the `NativeAppManager`. I've kept the diff as simple as possible and avoided introducing too many helpers or reorganizing the code, that can be done separately if necessary (the top priority is to just work towards removing `NativeAppManager` and all these static methods first).


src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
```diff
@@ -266,7 +272,7 @@ def action_validate(
             post_deploy_hooks=model.meta and model.meta.post_deploy,
             package_scripts=[],  # Package scripts are not supported in PDFv2
             policy=policy,
-            use_scratch_stage=True,
+            use_scratch_stage=use_scratch_stage,
```
sfc-gh-fcampbell, line 275: had to make this a param since we want `False` in tests

sfc-gh-bdufour: why is that? Didn't get it from the diff.

sfc-gh-fcampbell: in tests we don't want to call `deploy()` on the package, so we need to be able to pass `False` here. In the old `NativeAppManager.validate()`, the `use_scratch_stage` parameter defaults to `False`: https://github.com/snowflakedb/snowflake-cli/blob/b516dc29e59c8615d9e7497c9054e7de20b0aec2/src/snowflake/cli/_plugins/nativeapp/manager.py#L321

sfc-gh-fcampbell: this doesn't change prod code since the parameter now defaults to `True` in v2

sfc-gh-bdufour (approved):

PR #1730, opened 2024-10-16
Merge v2 entities into template context instead of overriding

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
...

### DIFF
--- a/src/snowflake/cli/_plugins/nativeapp/v2_conversions/compat.py
+++ b/src/snowflake/cli/_plugins/nativeapp/v2_conversions/compat.py
@@ -294,14 +294,15 @@ def force_project_definition_v2(
                 # Override the project definition so that the command operates on the new entities
                 cm.override_project_definition = pdfv2
 
-                # Override the template context so that templates refer to the new entities
-                # Reuse the old ctx.env and other top-level keys in the template context
-                # since they don't change between v1 and v2
-                pdfv2_dump = pdfv2.model_dump(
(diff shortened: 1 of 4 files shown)


src/snowflake/cli/_plugins/nativeapp/v2_conversions/compat.py
```diff
@@ -294,14 +294,15 @@ def wrapper(*args, **kwargs):
...
-                new_ctx = pdfv2_dump | dict(env=cm.template_context["ctx"]["env"])
-                cm.override_template_context = cm.template_context | dict(ctx=new_ctx)
+                # Add new entities to the template context so that templates in
+                # migrated package scripts can refer to the new entities
+                # Keep everything else in the template context since
+                # the user might still have other files with v1 templates in them
+                entities = {
+                    k: e.model_dump(exclude_none=True, warnings=False, by_alias=True)
+                    for k, e in pdfv2.entities.items()
+                }
```
sfc-gh-bdufour, line 304: It feels like that's not the right fix, since it can introduce a different behaviour once the contexts are merged (granted, I can't think of a way to trigger this with a _valid_ set of templates, but still...). I expected we'd instead preserve the entire v1 context, separately, and decide at expansion time which of the contexts should apply. I'm still thinking through what it means to merge the contexts though, so I might be off here.

sfc-gh-fcampbell:
Closing in favour of #1735

PR #1726, opened 2024-10-15
SNOW-1706990 Update/remove tests that use NativeAppRunProcessor

sfc-gh-fcampbell:
### Pre-review checklist
   * [x] I've confirmed that instructions included in README.md are still correct after my changes in the codebase.
   * [x] I've added or updated automated unit tests to verify correctness of my new code.
   * [ ] I've added or updated integration tests to verify correctness of my new code.
   * [ ] I've confirmed that my changes are working by executing CLI's commands manually on MacOS.
   * [ ] I've confirmed that my changes are working by executing CLI's commands manually on Windows.
   * [x] I've confirmed that my changes are up-to-date with the target branch.
   * [ ] I've described my changes in the release notes.
   * [x] I've described my changes in the section below.

### Changes description
Update/remove tests that use NativeAppRunProcessor. Removed `tests/nativeapp/test_package_scripts.py` because package scripts are now converted to post-deploy hooks for all commands, the package script code isn't used anymore.

PR #1732, opened 2024-10-16
SQL Facade

sfc-gh-pjafari:
### Pre-review checklist

### Changes description
 
Started a SQLService abstraction layer. The goal is to extract  execute_query calls that are embedded in the logic out to their own unit. This PR uses package scripts to show an example of the restructuring.  
use role, warehouse, and database calls are moved to this layer.

NOTE - Since the start of this work, `apply_package_scripts` is no longer invoked by any commands (in favor of post-deploy hooks). That's why there are no integration tests associated with this change, even though some error handling has changed.    
...


tests/nativeapp/test_package_scripts.py
```diff
@@ -17,19 +17,24 @@
 from unittest import mock
```
sfc-gh-pjafari, line 18: You can ignore all changes to this file. It's deleted in #1726

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -0,0 +1,199 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+from contextlib import contextmanager
+
+from cryptography.utils import cached_property
```
sfc-gh-fcampbell, line 18:
```suggestion
from functools import cached_property
```

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -0,0 +1,199 @@
...
+
+class SnowflakeSQLFacade:
+    def __init__(self, sql_executor: SqlExecutor | None = None):
+        self._sql_executor = (
+            sql_executor if sql_executor is not None else get_sql_executor()
+        )
+
+    @cached_property
+    def _log(self):
+        return logging.getLogger(__name__)
```
sfc-gh-fcampbell, line 43: this could just be an instance field set in init

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def _use_object(self, object_type: ObjectType, name: str):
```
sfc-gh-fcampbell, line 45: i don't think every `ObjectType` supports `use`, might be worthwhile to have a dedicated enum with just role/db/wh

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -0,0 +1,199 @@
...
+        Runs the user-provided sql script.
+        @param queries: Queries to run in this script
+        @param script_name: Name of the file containing the script. Used to show logs to the user.
+        @param [Optional] role: Role to switch to while running this script. Current role will be used if no role is passed in.
+        @param [Optional] warehouse: Warehouse to use while running this script.
+        @param [Optional] database: Database to use while running this script.
+        """
+        with self._use_role_optional(role):
+            with self._use_warehouse_optional(warehouse):
+                with self._use_database_optional(database):
```
sfc-gh-fcampbell, line 185: `with` can specify more than one CM to make this a bit less intented:
```python
with (
    self._use_role_optional(role),
    self._use_warehouse_optional(role),
    self._use_database_optional(role),
):
    try:
        ...
```

src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
```diff
@@ -1095,17 +1096,12 @@ def apply_package_scripts(
...
-                    console.step(f"Applying package script: {package_scripts[i]}")
-                    get_sql_executor().execute_queries(queries)
-            except ProgrammingError as err:
-                generic_sql_error_handler(
-                    err, role=package_role, warehouse=package_warehouse
-                )
+        for i, queries in enumerate(queued_queries):
+            script_name = package_scripts[i]
+            console.step(f"Applying package script: {script_name}")
+            get_snowflake_facade().execute_user_script(
```
sfc-gh-bdufour, line 1102: nit: for calls that have no obvious argument order, please use keyword args for extra readability

src/snowflake/cli/_plugins/nativeapp/sf_facade_exceptions.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    message = f"{context} {str(err)}"
+    if isinstance(err, DatabaseError):
+        raise UnknownSQLError(message) from err
+
+    if isinstance(err, Error):
+        raise UnknownConnectorError(message) from err
+
+    # Not a connector error
+    raise Exception(f"Unclassified exception occurred. {message}") from err
```
sfc-gh-bdufour, line 20: I'm not comfortable exposing this message to users. If these are Python exceptions, we should just let them bubble up. If it's something, it should be wrapped. We should consider just using the original exception here, provided that re-raising it maintains the original stack trace.

src/snowflake/cli/_plugins/nativeapp/sf_facade_exceptions.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class UserScriptError(ClickException):
+    """Exception raised when user-provided scripts fail."""
+
+    def __init__(self, script_name, msg):
+        super().__init__(f"Failed to run script {script_name}. {msg}")
+
+
+class SQLWithUserInputError(ClickException):
```
sfc-gh-bdufour, line 52: SQL is somewhat irrelevant here, it's just a `UserInputError` no?

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
sfc-gh-bdufour, line 18: odd, I think you want the one from functools

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -0,0 +1,199 @@
...
+        @param object_type: ObjectType, type of snowflake object to use
+        @param name: object name, has to be a valid snowflake identifier.
+        """
+        try:
+            self._sql_executor.execute_query(f"use {object_type} {name}")
+        except ProgrammingError as err:
+            if err.errno == DOES_NOT_EXIST_OR_CANNOT_BE_PERFORMED:
+                raise CouldNotUseObjectError(object_type, name) from err
+            else:
+                handle_unclassified_error(err, f"Failed to use {object_type} {name}.")
```
sfc-gh-bdufour, line 57: I think this implies that we'll raise an UnknownConnectorError in this case. That feels a little strange. I would have expected some kind of "CommandExecutionFailure", "InvalidSQLError" or something like that, otherwise we're discarding valuable information. It's a ProgrammingError, so we know the SQL was invalid. We just suspect that we're at fault, and not the user.

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -0,0 +1,199 @@
...
+            yield
+        finally:
+            if is_different_db and prev_db is not None:
+                self._log.debug(f"Switching back to database:{prev_db}")
+                self._use_object(ObjectType.DATABASE, prev_db)
+
+    def execute_user_script(
+        self,
+        queries: str,
+        script_name: str,
```
sfc-gh-bdufour, line 170: should it be a Path? What should it be relative to?

tests/nativeapp/test_sf_sql_facade.py
```diff
@@ -0,0 +1,841 @@
...
+    ]
+
+    # Act
+    sql_facade.execute_user_script(mock_script, mock_script_name, role, wh)
+
+    # Assert
+    assert mock_execute_query.mock_calls == expected
+    assert mock_execute_queries.mock_calls == [mock.call(mock_script)]
+    # Assert - order of calls
+    mock_parent.assert_has_calls(all_execute_calls)
```
sfc-gh-bdufour, line 153: these tests are mostly boilerplate. Would love to simplify, especially since we're checking the same thing two different ways.

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -0,0 +1,199 @@
...
+        @param [Optional] warehouse: Warehouse to use while running this script.
+        @param [Optional] database: Database to use while running this script.
+        """
+        with self._use_role_optional(role):
+            with self._use_warehouse_optional(warehouse):
+                with self._use_database_optional(database):
+                    try:
+                        self._sql_executor.execute_queries(queries)
+                    except ProgrammingError as err:
+                        if err.errno == NO_WAREHOUSE_SELECTED_IN_SESSION:
```
sfc-gh-mchok, line 189: Is this a special case of only some queries or was there a decision made to let it fail here? It feels a bit counterintuitive to me to say that `warehouse` parameter is optional and then we expect the call to fail if no warehouse was selected

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -0,0 +1,199 @@
...
+        except ProgrammingError as err:
+            if err.errno == DOES_NOT_EXIST_OR_CANNOT_BE_PERFORMED:
+                raise CouldNotUseObjectError(object_type, name) from err
+            else:
+                handle_unclassified_error(err, f"Failed to use {object_type} {name}.")
+        except Exception as err:
+            handle_unclassified_error(err, f"Failed to use {object_type} {name}.")
+
+    @contextmanager
+    def _use_warehouse_optional(self, new_wh: str | None):
```
sfc-gh-mchok, line 62: Was wondering what the design philosophy was making all of these optional? It seems like the original ones require non-optional arguments, it feels a bit weird to me to be able to call `use_warehouse_optional(None)` and have that be a noop, from an API standpoint; maybe it would be better to be explicit about our invocations?

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
sfc-gh-mchok: +1, feels like we might be falling into the same trap as the original `generic_sql_error_handler` but in a different form

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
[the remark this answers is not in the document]
sfc-gh-bdufour: I can comment on that one... That top-level "use role" is something we're specifically avoiding as it continues to cause issues. It should be essentially free in GS anyway, so I don't think it's work taking extra risk to optimize it away.

sfc-gh-bdufour: Notice these are internal. The general idea is that we're expressing high-level operations in the public facade API, and we let the caller specify the parameters. The only API we're introducing is for user scripts. It's a general API, so using the session role is a valid option, for example. Our NA use does prefer using the package role if one was specified. However, the abstractions here would let us move the fallback "use role" to the facade for cases where the role is optional. The helpers were written this way so that we can make the rest of the implementation clean. This is really just a "temporarily activate this WH/role/db if one was provided" helper.

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
sfc-gh-bdufour: WH is optional yes, but we don't know if the SQL commands will require one. So they can fail due to the user not having provided a WH. IMHO that's a fair failure mode, the user is escaping to SQL, they should know whether a WH is needed or not. We shouldn't force them to create/activate one just in case they need it.

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
sfc-gh-pjafari: I originally followed what SqlExecutor.use was doing. But I see that usage and the tests for that method only have role/db/wh too. 
I created a new enum.

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
sfc-gh-pjafari: this func uses the script_name for log/error purposes (not to read from disk, since template expansion etc happens before).
As for its type, the sql script paths are represented by `str` types right now, both in `apply_package_script` that I have used as an example here (though this function is never reached anymore and will be removed, beside the point) and in post_deploy_hooks, see `SqlScriptHookType`.

src/snowflake/cli/_plugins/nativeapp/sf_facade_exceptions.py
sfc-gh-pjafari: really good point. I'll just pass the message and re-raise it (the point of the message is to add context from wherever we called this `handle_unclassified_error`.

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
sfc-gh-pjafari: @sfc-gh-bdufour good callout to add handling for ProgrammingErrors specifically. As the handler is checking errors from higher specificity -> more general . ProgrammingError -> DatabaseError -> Error. 

@sfc-gh-mchok I'll have to disagree.  generic_sql_handler does somewhat the opposite of this. It handles known errors, whereas here the goal is to have a final catch-all filter after we've exhausted specific errors in each context the call is made.

tests/nativeapp/test_sf_sql_facade.py
sfc-gh-pjafari: I'm with you. Didn't like writing these tests either. I'll think on it and make a better solution/helper.

src/snowflake/cli/_plugins/nativeapp/sf_facade_exceptions.py
```diff
@@ -0,0 +1,90 @@
...
+        raise UnknownConnectorError(message) from err
+
+    # Not a connector error
+    raise Exception(message) from err
+
+
+class _BaseFacadeError(Exception):
+    """Base class for SnowflakeFacade Exceptions"""
+
+    msg = "Unknown Error"
```
sfc-gh-bdufour, line 43: why not simply make the message a required ctor param instead? This is odd, as the message should never really be used.

src/snowflake/cli/_plugins/nativeapp/sf_facade_exceptions.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def __str__(self):
+        return self.msg
+
+
+class InvalidSQLError(_BaseFacadeError):
+    "Exception raised when a ProgrammingError is caught but cause is unidentified" ""
```
sfc-gh-bdufour, line 50: Let's define these in a way that isn't so coupled with the specific way we talk to Snowflake underneath. I'd suggest something like:

`"Raised when Snowflake executed a SQL command but encountered an error, for example due to syntax or logical errors"`

src/snowflake/cli/_plugins/nativeapp/sf_facade_exceptions.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def __init__(self, msg):
+        self.msg = f"Invalid SQL error occurred. {msg}"
+        super().__init__(self.msg)
+
+
+class UnknownSQLError(_BaseFacadeError):
+    """Exception raised when the root of the DatabaseError is unidentified."""
```
sfc-gh-bdufour, line 58: `"Raised when Snowflake could not execute the SQL command"`

src/snowflake/cli/_plugins/nativeapp/sf_facade_exceptions.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def __init__(self, msg):
+        self.msg = f"Unknown SQL error occurred. {msg}"
+        super().__init__(self.msg)
+
+
+class UnknownConnectorError(_BaseFacadeError):
+    """Exception raised when the root of the error thrown by connector is unidentified."""
```
sfc-gh-bdufour, line 66: `"Raised when there was a problem reaching Snowflake to execute a SQL command"`

src/snowflake/cli/_plugins/nativeapp/sf_facade_exceptions.py
```diff
@@ -0,0 +1,90 @@
...
+
+class UnknownSQLError(_BaseFacadeError):
+    """Exception raised when the root of the DatabaseError is unidentified."""
+
+    def __init__(self, msg):
+        self.msg = f"Unknown SQL error occurred. {msg}"
+        super().__init__(self.msg)
+
+
+class UnknownConnectorError(_BaseFacadeError):
```
sfc-gh-bdufour, line 65: Here we can keep the Connector reference I guess, because it would help guide any kind of investigation into a spike of these.

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -0,0 +1,195 @@
...
+        # new_wh is not None, and should already be a valid identifier, no additional check is performed here.
+        is_different_wh = valid_wh_name != prev_wh
+        if is_different_wh:
+            self._log.debug(f"Using warehouse: {valid_wh_name}")
+            self._use_object(UseObjectType.WAREHOUSE, valid_wh_name)
+        try:
+            yield
+        finally:
+            if is_different_wh and prev_wh is not None:
+                self._log.debug(f"Switching back to warehouse:{prev_wh}")
```
sfc-gh-bdufour, line 91: nit: space after : for consistency

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -0,0 +1,195 @@
...
+        prev_role = prev_role_res[0]
+        is_different_role = valid_role_name.lower() != prev_role.lower()
+        if is_different_role:
+            self._log.debug(f"Assuming different role: {valid_role_name}")
+            self._use_object(UseObjectType.ROLE, valid_role_name)
+        try:
+            yield
+        finally:
+            if is_different_role:
+                self._log.debug(f"Switching back to role:{prev_role}")
```
sfc-gh-bdufour, line 122: nit: space after : for consistency

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -0,0 +1,195 @@
...
+        is_different_db = valid_name != prev_db
+        if is_different_db:
+            self._log.debug(f"Using database {valid_name}")
+            self._use_object(UseObjectType.DATABASE, valid_name)
+
+        try:
+            yield
+        finally:
+            if is_different_db and prev_db is not None:
+                self._log.debug(f"Switching back to database:{prev_db}")
```
sfc-gh-bdufour, line 160: nit: space after : for consistency

src/snowflake/cli/_plugins/nativeapp/sf_facade_exceptions.py
sfc-gh-mchok: Would it make sense to raise a custom error instead of `Exception` here? I don't think we will have access to the message on the dashboard, so I could see it being more helpful to see something like `UnknownFacadeError` so we know more specifically where a spike of errors could be coming from, instead of just `Exception`

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
sfc-gh-bdufour: We shouldn't, in general, represent paths as strings. Not a blocker here, but we should convert incoming strings to paths, and standardize how we refer to project artifacts across the board. IMHO this API should eventually accept something like a relative script path (relative to the project directory).

sfc-gh-bdufour (approved):

src/snowflake/cli/api/constants.py
```diff
@@ -33,7 +33,19 @@ def __str__(self):
...
+class BaseObjectType(Enum):
+    def __str__(self):
+        """This makes using this Enum easier in formatted string"""
+        return self.value.cli_name
+
+
+class UseObjectType(BaseObjectType):
+    DATABASE = ObjectNames("database", "database", "databases")
+    ROLE = ObjectNames("role", "role", "roles")
+    WAREHOUSE = ObjectNames("warehouse", "warehouse", "warehouses")
```
sfc-gh-bdufour, line 45: Sorry, somehow I had missed this. These changes don't make sense to me. If we wanted to group things by account-level, db-level and schema-level, maybe. But "UseObjectType" isn't really a thing IMHO. We can use a union + a type alias to achieve the same thing, and not modify this file at all.

src/snowflake/cli/_plugins/nativeapp/sf_facade_exceptions.py
```diff
@@ -0,0 +1,91 @@
...
+    def __str__(self):
+        return self.msg
+
+
+class InvalidSQLError(_BaseFacadeError):
+    """Raised when Snowflake executed a SQL command but encountered an error, for example due to syntax or logical errors"""
+
+    def __init__(self, msg):
+        self.msg = f"Invalid SQL error occurred. {msg}"
+        super().__init__(self.msg)
```
sfc-gh-bdufour, line 55: bad pattern, you're letting the parent ctor do it.

src/snowflake/cli/_plugins/nativeapp/sf_facade_exceptions.py
sfc-gh-bdufour: Not a custom error, but the original exception. These are basically python exceptions, we'd want to see what they are and where they come from (not in telemetry, but the user should be able to obtain a trace and submit a bug report)

src/snowflake/cli/api/constants.py
sfc-gh-pjafari: reverted the changes to this file and created a separate enum for facade for objects that are "usable".

sfc-gh-bdufour (approved):

PR #1735, opened 2024-10-17
Defer package script conversion until PDFv2 is loaded into Pydantic

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
When converting PDFv1 to PDFv2, we need to convert package scripts to post-deploy hooks, so we have to convert the `{{ package_name }}` template into the normal template syntax.

When converting using `snow helpers v1-to-v2`, we convert `{{ package_name }}` to `<% ctx.entities.pkg.identifier %>` since we know that the next time the CLI is run, the template context will be v2-based and that `ctx.entities.pkg.identifier` will be a valid reference.

When converting in-memory however, the template context is v1-based and since the conversion has to be transparent to the user (they could have other files with templates still using v1 references in it), we can't override the template context to v2, so `ctx.entities.pkg.identifier` is an invalid reference. 

Fortunately for in-memory conversions, package scripts are converted to post-deploy hooks using temp files (to avoid overwriting the user's files), so we can just insert the package name directly into the converted script instead of having to use a template reference. This requires us to defer conversion of package scripts until after the v2 definition is loaded into Pydantic since there are validators that can change the values of fields (like the test resource suffix being added to the package identifier).

### DIFF
--- a/src/snowflake/cli/_plugins/nativeapp/v2_conversions/compat.py
+++ b/src/snowflake/cli/_plugins/nativeapp/v2_conversions/compat.py
@@ -293,15 +293,6 @@ def force_project_definition_v2(
 
                 # Override the project definition so that the command operates on the new entities
                 cm.override_project_definition = pdfv2
-
-                # Override the template context so that templates refer to the new entities
-                # Reuse the old ctx.env and other top-level keys in the template context
-                # since they don't change between v1 and v2
-                pdfv2_dump = pdfv2.model_dump(
-                    exclude_none=True, warnings=False, by_alias=True
-                )
-                new_ctx = pdfv2_dump | dict(env=cm.template_context["ctx"]["env"])
-                cm.override_template_context = cm.template_context | dict(ctx=new_ctx)
             elif single_app_and_package:
                 package_entity_id = kwargs.get("package_entity_id", "")
                 app_entity_id = kwargs.get("app_entity_id", "")
--- a/src/snowflake/cli/api/project/definition_conversion.py
+++ b/src/snowflake/cli/api/project/definition_conversion.py
@@ -9,6 +9,9 @@ from click import ClickException
 from snowflake.cli._plugins.nativeapp.artifacts import (
     BundleMap,
 )
+from snowflake.cli._plugins.nativeapp.entities.application_package import (
+    ApplicationPackageEntityModel,
+)
 from snowflake.cli._plugins.snowpark.common import is_name_a_templated_one
 from snowflake.cli.api.constants import (
     DEFAULT_ENV_FILE,
@@ -19,6 +22,7 @@ from snowflake.cli.api.constants import (
 )
 from snowflake.cli.api.entities.utils import render_script_template
 from snowflake.cli.api.project.schemas.entities.common import (
+    MetaField,
     SqlScriptHookType,
 )
 from snowflake.cli.api.project.schemas.project_definition import (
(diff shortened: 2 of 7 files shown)


src/snowflake/cli/api/project/definition_conversion.py
```diff
@@ -83,23 +87,31 @@ def _is_field_defined(template_context: Optional[Dict[str, Any]], *path: str) ->
 
 def convert_project_definition_to_v2(
     project_root: Path,
-    pd: ProjectDefinition,
+    definition_v1: ProjectDefinition,
```
sfc-gh-fcampbell, line 90: renamed this since we now also have `definition_v2` below

src/snowflake/cli/api/project/definition_conversion.py
```diff
@@ -334,12 +324,11 @@ def _convert_package_script_files(package_scripts: list[str]):
...
-                native_app.package.scripts
-            )
-            package_meta["post_deploy"] = (
-                package_meta.get("post_deploy", []) + converted_post_deploy_hooks
-            )
+            # Package scripts are not supported in PDFv2 but we
+            # don't convert them here, conversion is deferred until
+            # the final v2 Pydantic model is available
+            # (see _convert_templates_in_files())
+            pass
```
sfc-gh-fcampbell, line 331: left this in instead of removing the `if` as a marker that this case is indeed being considered but consciously ignored

tests_integration/test_data/projects/napp_templates_processors_v1/app/setup_script.sql
```diff
@@ -8,3 +8,4 @@
 
 CREATE OR ALTER VERSIONED SCHEMA <% ctx.env.schema_name %>;
 EXECUTE IMMEDIATE from '/another_script.sql';
+select 'ctx.native_app.name: <% ctx.native_app.name %>';
```
sfc-gh-fcampbell, line 11: this tests that even if the PDF is implicitly converted to v2, the template context is still v1

src/snowflake/cli/_plugins/nativeapp/v2_conversions/compat.py
```diff
@@ -293,15 +293,6 @@ def wrapper(*args, **kwargs):
...
                 cm.override_project_definition = pdfv2
-
-                # Override the template context so that templates refer to the new entities
-                # Reuse the old ctx.env and other top-level keys in the template context
-                # since they don't change between v1 and v2
-                pdfv2_dump = pdfv2.model_dump(
-                    exclude_none=True, warnings=False, by_alias=True
-                )
-                new_ctx = pdfv2_dump | dict(env=cm.template_context["ctx"]["env"])
-                cm.override_template_context = cm.template_context | dict(ctx=new_ctx)
```
sfc-gh-fcampbell, line 304: stop overriding the template context, templates should be able to refer to fields as they exist in the PDF on-disk, since that's the one that the user sees

src/snowflake/cli/api/project/definition_conversion.py
```diff
@@ -269,34 +287,6 @@ def _find_manifest():
         # which use POSIX paths as default values
         return manifest_path.relative_to(project_root).as_posix()
 
-    def _make_template(template: str) -> str:
-        return f"{PROJECT_TEMPLATE_VARIABLE_OPENING} {template} {PROJECT_TEMPLATE_VARIABLE_CLOSING}"
-
-    def _convert_package_script_files(package_scripts: list[str]):
```
sfc-gh-fcampbell, line 275: moved to below nearly verbatim

src/snowflake/cli/api/project/definition_conversion.py
```diff
@@ -383,6 +372,73 @@ def convert_envs_to_v2(pd: ProjectDefinition):
...
 
 
+def _convert_templates_in_files(
+    project_root: Path,
+    definition_v1: ProjectDefinition,
+    definition_v2: ProjectDefinitionV2,
+    in_memory: bool,
+):
+    """Converts templates in other files to the new format"""
+    # TODO handle artifacts using the "templates" processor
```
sfc-gh-bdufour, line 382: what does this mean?

tests_integration/test_data/projects/napp_templates_processors_v1/app/setup_script.sql
sfc-gh-bdufour: but shouldn't we check the expansion somewhere then?

sfc-gh-fcampbell: We don't care about the expansion, we're just testing that arbitrary files in v1 projects can continue to use v1 template syntax. This test fails if the template context becomes converted implicitly (which is the original bug that Michel found)

src/snowflake/cli/api/project/definition_conversion.py
sfc-gh-fcampbell: Artifacts that use the templates processor to template arbitrary files don't have their templates converted to v2 when using `snow helpers v1-to-v2` (and if the snowflake.yml file doesn't have any templates in it, no warning is shown to the user, so they silently break).

We want to support automatic conversion of these templates eventually, but it's out of scope for this PR, this is just the function where we'd do it so left this comment as a placeholder/hint.

sfc-gh-bdufour (approved):

PR #1745, opened 2024-10-18
SNOW-1733913 Fix wrong script filename when running converted package scripts

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
When running a post-deploy hook that comes from an auto-converted package script, the filename seen is the absolute path to the tempfile, which is confusing to users. Let's store a `display_path` for these hooks so we can show users the original path that they had put in the PDF file.

Before:
image

After:
image


src/snowflake/cli/api/project/schemas/entities/common.py
```diff
@@ -28,6 +28,13 @@
 class SqlScriptHookType(UpdatableModel):
     sql_script: str = Field(title="SQL file path relative to the project root")
 
+    # Used to store a user-friendly path for this script, when the
+    # value of `sql_script` is a path to a different file
+    # This is used in the UI to display the path relative to the
+    # project root when `sql_script` is a actually path to a temp file
+    # generated by the in-memory PDF v1 to v2 conversion
+    display_path: str | None = Field(default=None, exclude=True)
```
sfc-gh-bdufour, line 36: Doesn't this feel wrong? A user shouldn't be able to override the display path in their snowflake.yml

sfc-gh-fcampbell: I didn't think users would know that this property exists (we wouldn't document it), but I went and changed it to a `PrivateAttr` just in case

sfc-gh-bdufour (approved):

sfc-gh-turbaszek (approved):

PR #1751, opened 2024-10-21
Make JWT command uses config connection details

### LINKED ISSUE #1744: SNOW-1752069: generate-jwt should use settings from config.toml
### Description

Currently `generate_jwt` command has its own set of flags and does not seem to take `config.toml` into account at all.

For example:

```
snow connection generate-jwt --private-key <p8-file-path>                             
Enter private key file password (Press enter if none) []: 

argument of type 'NoneType' is not iterable
```
The error
>argument of type 'NoneType' is not iterable
Probably happens because of the type [here](https://github.com/snowflakedb/snowflake-cli/blob/main/src/snowflake/cli/_plugins/connection/commands.py#L360):

```python
def generate_jwt(
    account: str = typer.Option(
        None,
        "--account",
        "-a",
        "--accountname",
        help="Account name to use when authenticating with Snowflake.",
        show_default=False,
    ),
```

This also applies to the `user` and `private_key_file` values.

* Other commands use values from the configuration file
* `argument of type 'NoneType' is not iterable` is not immediately apparent on how to fix. `--debug` does show the location, but the error message could be better

### Context

_No response_

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
Closes #1744

### DIFF
--- a/src/snowflake/cli/_plugins/connection/commands.py
+++ b/src/snowflake/cli/_plugins/connection/commands.py
@@ -16,10 +16,9 @@ from __future__ import annotations
 
 import logging
 import os.path
-from pathlib import Path
 
 import typer
-from click import ClickException, Context, Parameter  # type: ignore
+from click import ClickException, Context, Parameter, UsageError  # type: ignore
 from click.core import ParameterSource  # type: ignore
 from click.types import StringParamType
 from snowflake import connector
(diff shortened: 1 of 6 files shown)


src/snowflake/cli/_plugins/connection/commands.py
```diff
@@ -355,36 +354,23 @@ def set_default(
...
+
+    msq_template = (
+        "{} is not set in the connection context, but required for JWT generation."
+    )
+    if not connection_details.user:
+        raise UsageError(msq_template.format("User"))
+    if not connection_details.account:
+        raise UsageError(msq_template.format("Account"))
+    if not connection_details.private_key_file:
+        raise UsageError(msq_template.format("Private key file"))
```
sfc-gh-astus, line 372: No password?

sfc-gh-turbaszek: What do you mean?

sfc-gh-astus: We are not able to provide password? We have to use private key?

sfc-gh-astus (approved):

src/snowflake/cli/_plugins/connection/commands.py
sfc-gh-turbaszek: Yes, only private key is neccessary

PR #1764, opened 2024-10-22
Add package scripts warning back

sfc-gh-melnacouzi:
### Pre-review checklist

### Changes description
Add package scripts warning back.
It disappeared due to the conversion of package scripts to post deploy scripts.

### DIFF
--- a/tests_integration/nativeapp/test_feature_metrics.py
+++ b/tests_integration/nativeapp/test_feature_metrics.py
@@ -103,7 +103,7 @@ def test_sql_templating_emits_counter(
                 CLICounterField.TEMPLATES_PROCESSOR: 0,
                 CLICounterField.PDF_TEMPLATES: 1,
                 CLICounterField.POST_DEPLOY_SCRIPTS: 1,
-                CLICounterField.PACKAGE_SCRIPTS: 0,
+                CLICounterField.PACKAGE_SCRIPTS: 1,
             },
         ),
         # ensure post deploy scripts are picked up for v2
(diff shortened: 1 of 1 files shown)


tests_integration/nativeapp/test_feature_metrics.py
```diff
@@ -103,7 +103,7 @@ def test_sql_templating_emits_counter(
                 CLICounterField.TEMPLATES_PROCESSOR: 0,
                 CLICounterField.PDF_TEMPLATES: 1,
                 CLICounterField.POST_DEPLOY_SCRIPTS: 1,
-                CLICounterField.PACKAGE_SCRIPTS: 0,
+                CLICounterField.PACKAGE_SCRIPTS: 1,
```
sfc-gh-melnacouzi, line 106: test files have package scripts, so this should be 1.

tests_integration/nativeapp/test_feature_metrics.py
```diff
@@ -115,7 +115,6 @@ def test_sql_templating_emits_counter(
                 CLICounterField.TEMPLATES_PROCESSOR: 0,
                 CLICounterField.PDF_TEMPLATES: 1,
                 CLICounterField.POST_DEPLOY_SCRIPTS: 1,
-                CLICounterField.PACKAGE_SCRIPTS: 0,
```
sfc-gh-melnacouzi, line 118: this is PDF v2, so no need to emit metric for this one.

sfc-gh-fcampbell (dismissed):

tests_integration/nativeapp/test_feature_metrics.py
sfc-gh-mchok: Do we want to make the distinction that since the pdf is v1, then POST_DEPLOY_SCRIPTS should count as 0 then?

sfc-gh-melnacouzi: i think it's something we can distinguish in the graphs themselves (like exclude package_scripts = 1). I do not want to touch the post_deploy_scripts part because the goal is to have the bulk part of the code operate in pdf v2 mode, and would make sense to keep the counters for this one with the code itself.

sfc-gh-turbaszek (dismissed):

sfc-gh-turbaszek (approved):

PR #1781, opened 2024-10-23
Extract BaseSqlExecutor

sfc-gh-pjafari:
### Pre-review checklist

### Changes description
Extract base methods of SqlExecutor to a parent class. Allows better encapsulation while SnowflakeFacade is underway.   
Also clean up the redundant interface (removed private `_execute_quer(y|ies)` methods)


src/snowflake/cli/api/sql_execution.py
```diff
@@ -42,9 +42,8 @@
 from snowflake.connector.errors import ProgrammingError
 
 
-class SqlExecutor:
-    def __init__(self, connection: SnowflakeConnection | None = None):
-        self._snowpark_session = None
+class BaseSqlExecutor:
+    def __init__(self, connection: SnowflakeConnection | None = None) -> None:
```
sfc-gh-fcampbell, line 46: IMO the connection should be a param of the concrete `SqlExecutor`, since other implementations might not need one (like a `TestSqlExecutor` for example)

src/snowflake/cli/api/sql_execution.py
```diff
@@ -42,9 +42,8 @@
 from snowflake.connector.errors import ProgrammingError
 
 
-class SqlExecutor:
-    def __init__(self, connection: SnowflakeConnection | None = None):
-        self._snowpark_session = None
+class BaseSqlExecutor:
```
sfc-gh-bdufour, line 45: How about we take the opportunity to simply the various "execute" variants in here? I'm not sure why we need all these private helpers. From the POV of mocking the contract, it would be a lot easier if we just had a single point to mock (I know that everything eventually rolls up to _execute_string, but there's a rather complex set of dependencies between the calls).

src/snowflake/cli/api/sql_execution.py
```diff
@@ -90,6 +89,12 @@ def execute_query(self, query: str, **kwargs):
     def execute_queries(self, queries: str, **kwargs):
```
sfc-gh-bdufour, line 89: can we backfill the doc comments too while we're here for all public methods in the contract?

src/snowflake/cli/api/sql_execution.py
sfc-gh-bdufour: I think the point is this _is_ a concrete executor. Just a restricted one that doesn't overlap with the SQL facade in its contract.

sfc-gh-mchok: Just wanted to clarify the desired end state of this project

In my head based on previous conversations, we want the following abstractions:
1. One to execute all the raw SQL which we can mock or switch out the implementation for if the API for calling SQL changes, e.g. aforementioned REST APIs, and 
2. Another abstraction on top of that that only executes business logic, which doesn't need to change depending on how we execute queries and we can test/reason about easily

Am I correct in the classification that:
- BaseSqlExecutor = 1. (Data Access/Repository Layer)?
- SnowflakeSQLFacade = 2. (Business Logic/Service Layer)?

However I see that SnowflakeSQLFacade is still putting together raw SQL queries with strings, just passing them to the executor, and I'm a bit confused on the separation between these classes, since it seems now that SnowflakeSQLFacade and SQLExecutor are doing similar jobs in that they both extend BaseSQLExecutor's functionality, but with individual flavours of API definitions and error handling?

sfc-gh-pjafari: the main difference between SqlExecutor and SnowflakeSQLFacade is in those API definitions and error handling. 
SnowflakeFacade defines the interactions from cli (or nativeapps for now) with snowflake on a higher level than `execute_query` and `use object`.
Currently we are managing sql calls (their logic, formulation, and their context e.g. role, wh,db) embedded within our logic whereas with SnowflakeFacade it's all encapsulated in its own layer. For an example, see how `execute_user_script` from SQLFacade is used VS how `execute_post_deploy_hooks` are written and used.

sfc-gh-bdufour: Marcus, 1 & 2 are the same to me, they're both encapsulated in the facade. BaseSQLExecutor is very SQL-specific, and is there mostly so that we can add convenience methods to manipulate raw query results.

sfc-gh-bdufour (dismissed):

sfc-gh-mchok (approved):

sfc-gh-astus (approved):

PR #1786, opened 2024-10-23
SNOW-1758715 Convert execute_post_deploy_hooks to instance method

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Converts `execute_post_deploy_hooks` for both entities to instance methods. Removes `test_invalid_hook_type` since it's now impossible to pass an invalid hook since Pydantic will fail to validate the model object (and will also fail if trying to set `model.meta.post_deploy` to an invalid hook afterwards).

### DIFF
--- a/tests/nativeapp/test_post_deploy_for_app.py
+++ b/tests/nativeapp/test_post_deploy_for_app.py
@@ -58,6 +58,7 @@ def test_sql_scripts(
     mock_execute_queries,
     mock_execute_query,
     temp_dir,
+    workspace_context,
 ):
     mock_conn.return_value = MockConnectionCtx()
     post_deploy_1 = dedent(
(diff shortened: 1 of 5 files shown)


tests/nativeapp/test_post_deploy_for_app.py
```diff
@@ -172,53 +168,20 @@ def test_sql_scripts_with_no_warehouse_no_database(
...
-                app_warehouse=app_model.meta.warehouse or "MockWarehouse",
-            )
+            app.execute_post_deploy_hooks()
 
         assert err.value.message == 'Script "scripts/missing.sql" does not exist'
 
 
-@mock.patch(SQL_EXECUTOR_EXECUTE)
-@mock_connection()
-def test_invalid_hook_type(
```
sfc-gh-fcampbell, line 198: it was previously possible to call `ApplicationEntity.execute_post_deploy_hooks` with an invalid hook type, but now that it's an instance method, it requires a valid `ApplicationEntityModel` instance, which doesn't allow loading or setting an invalid hook type, so this test is no longer required (or possible)

sfc-gh-bdufour: awesome!

sfc-gh-bdufour (approved):

PR #1788, opened 2024-10-23
implement metric steps with tests

sfc-gh-mchok:
### Pre-review checklist
   * [X] I've confirmed that my changes are working by executing CLI's commands manually on Windows.

### Changes description
To support instrumenting commands, we're implementing tracking steps of a command
We support both a context manager and manually calling start/end_step functions for flexible use cases
See the implemented test plan for the general contract


tests/api/test_metrics_counters.py
sfc-gh-mchok, line 1: tests were getting long, so I separated the two different use cases in their own test files

src/snowflake/cli/api/metrics.py
```diff
@@ -11,8 +11,21 @@
...
+from typing import Dict, List, Optional, Union
+
+
+class CLIMetricsInvalidUsageError(RuntimeError):
+    """
+    Indicative of bug in the code where a call to CLIMetrics was made erroneously
+    """
+
+    def __init__(self, message: str):
+        super().__init__(message)
```
sfc-gh-fcampbell, line 28:
```suggestion
    pass
```

src/snowflake/cli/api/metrics.py
```diff
@@ -54,17 +67,125 @@ class CLICounterField:
...
+    def to_dict(self) -> Dict[str, Union[str, int, float, None]]:
+        return {
+            self.ID_KEY: self.step_id,
+            self.NAME_KEY: self.name,
+            self.PARENT_KEY: self.parent,
+            self.PARENT_ID_KEY: self.parent_id,
+            self.START_TIME_KEY: self.start_time,
+            self.EXECUTION_TIME_KEY: self.execution_time,
+            self.ERROR_KEY: self.error,
+        }
```
sfc-gh-fcampbell, line 168: can this class be an `@dataclass`? then we wouldn't need this method or all the getters

src/snowflake/cli/api/metrics.py
```diff
@@ -86,7 +207,109 @@ def increment_counter(self, name: str, value: int = 1) -> None:
...
+
+        :returns: id of the started step which can be passed to end_step
+
+        :raises CliMetricsInvalidUsageError: if the step name is empty
+        """
+        parent_step = self._current_step
+        new_step = _CLIMetricsStep(name, parent_step)
+        self._executing_steps.append(new_step)
+        new_step.start()
+        return new_step.step_id
```
sfc-gh-fcampbell, line 247: what's the use-case for only returning the ID? can we return the whole step?

src/snowflake/cli/api/metrics.py
```diff
@@ -86,7 +207,109 @@ def increment_counter(self, name: str, value: int = 1) -> None:
...
+        Recommended general use API for tracking a step throughout the execution of a command
+        Assumes that parent steps contain the entirety of their child steps
+
+        start_step and end_step are exposed in case the caller needs more flexibility
+
+        :raises CliMetricsInvalidUsageError: if the step name is empty
+        """
+        step_id = self.start_step(name)
+        try:
+            yield
```
sfc-gh-fcampbell, line 226: could we yield the whole step here?

tests/api/test_metrics_steps.py
```diff
@@ -0,0 +1,407 @@
...
+    with metrics.track_step("step1"):
+        pass
+
+    # then
+    assert len(metrics.steps) == 1
+    step1 = metrics.steps[0]
+    assert step1[_CLIMetricsStep.NAME_KEY] == "step1"
+
+
+def test_metrics_steps_name_is_valid():
```
sfc-gh-fcampbell, line 48: i think the names of these two tests are reversed

tests/api/test_metrics_steps.py
```diff
@@ -0,0 +1,407 @@
...
+    with metrics.track_step("step1"):
+        try:
+            raise RuntimeError()
+        except RuntimeError:
+            pass
+
+    # then
+    assert len(metrics.steps) == 1
+    step1 = metrics.steps[0]
+    assert step1[_CLIMetricsStep.ERROR_KEY] is None
```
sfc-gh-fcampbell, line 78: this test is a bit redundant, it's obvious that the exception is caught and not propagated

src/snowflake/cli/api/metrics.py
```diff
@@ -54,17 +67,125 @@ class CLICounterField:
...
+    def __init__(self, name: str, parent: _CLIMetricsStep | None = None):
+        if not name:
+            raise CLIMetricsInvalidUsageError("step name must not be empty")
+
+        # to disambiguate steps with same name
+        self._step_id: int = next(self._id_counter)
+        self._name: str = name
+
+        self._parent: Optional[str] = parent.name if parent is not None else None
+        self._parent_id: Optional[int] = parent.step_id if parent is not None else None
```
sfc-gh-bdufour, line 95: nit: not a huge fan of expanding things this way. Why can't we just keep a ref to the parent and read the data on the fly as needed?

src/snowflake/cli/api/metrics.py
```diff
@@ -54,17 +67,125 @@ class CLICounterField:
...
+        self._monotonic_start = time.monotonic()
+
+    def end(self, error: Optional[BaseException] = None) -> None:
+        if self._execution_time:
+            raise CLIMetricsInvalidUsageError("step has already ended")
+        if not self._monotonic_start:
+            raise CLIMetricsInvalidUsageError("step has not started")
+
+        if error:
+            self._error = type(error).__name__
```
sfc-gh-bdufour, line 118: why not preserve the entire object?

src/snowflake/cli/api/metrics.py
```diff
@@ -54,17 +67,125 @@ class CLICounterField:
...
+        return self._execution_time
+
+    @property
+    def error(self) -> Optional[str]:
+        """
+        name of the error class that was raised during the execution of this step
+        """
+        return self._error
+
+    def to_dict(self) -> Dict[str, Union[str, int, float, None]]:
```
sfc-gh-bdufour, line 159: this is for reporting? I'd add a comment to that effect, since it makes this method a contract that should be stable.

src/snowflake/cli/api/metrics.py
```diff
@@ -86,7 +207,109 @@ def increment_counter(self, name: str, value: int = 1) -> None:
         else:
             self._counters[name] += value
 
+    @property
+    def _current_step(self) -> Optional[_CLIMetricsStep]:
+        return self._executing_steps[-1] if len(self._executing_steps) > 0 else None
+
+    @contextmanager
+    def track_step(self, name: str):
```
sfc-gh-bdufour, line 215: let's workshop the name

src/snowflake/cli/api/metrics.py
```diff
@@ -54,17 +67,125 @@ class CLICounterField:
...
+
 class CLIMetrics:
     """
     Class to track various metrics across the execution of a command
     """
 
     def __init__(self):
         self._counters: Dict[str, int] = {}
+        # stack of current steps as command is executing
+        self._executing_steps: List[_CLIMetricsStep] = []
```
sfc-gh-bdufour, line 179: executing -> in progress maybe?

sfc-gh-bdufour (commented):
(partial review)

src/snowflake/cli/api/metrics.py
sfc-gh-bdufour: esp since, AFAICT, step IDs aren't globally unique. Ideally for a start/stop scenario, one should be able to obtain some kind of unique token / structure from the start method, and pass that unmodified to the stop method.

src/snowflake/cli/api/metrics.py
```diff
@@ -86,7 +207,109 @@ def increment_counter(self, name: str, value: int = 1) -> None:
...
+                None,
+            )
+        else:
+            found_step = self._current_step
+
+        if not found_step:
+            if step_id or step_name:
+                raise CLIMetricsInvalidUsageError(
+                    f"step with {'id' if step_id else 'name'} '{step_id or step_name}' could not be ended because it could not be found"
+                )
```
sfc-gh-bdufour, line 284: this hurts my brain. if step_id elif step_name else would be a lot clearer.

src/snowflake/cli/api/metrics.py
```diff
@@ -54,17 +67,125 @@ class CLICounterField:
     PACKAGE_SCRIPTS = f"{_TypePrefix.FEATURES}.{_DomainPrefix.APP}.package_scripts"
 
 
+class _CLIMetricsStep:
+    """
+    class for holding metrics step data and encapsulating related operations
+    """
+
+    _id_counter = count(start=1, step=1)
```
sfc-gh-fcampbell, line 75: since steps end up in a single snowflake table, the ID should probably be a UUID

src/snowflake/cli/api/metrics.py
sfc-gh-fcampbell: AFAIK the common nomenclature is "trace", which are composed of "spans". Both datadog and opentelemetry use this: https://opentelemetry.io/docs/concepts/signals/traces/

src/snowflake/cli/api/metrics.py
```diff
@@ -86,7 +207,109 @@ def increment_counter(self, name: str, value: int = 1) -> None:
...
+                )
+
+        found_step.end(error)
+        self._executing_steps.remove(found_step)
+        self._finished_steps.append(found_step)
+
+    def flush_steps(self, error: Optional[BaseException] = None) -> None:
+        """
+        Useful if you are using the manual start/end steps and an error
+        propagated up, requiring you to clear out all the executing steps
```
sfc-gh-bdufour, line 297: doesn't say what this actually does. Please document the contract of the method.

src/snowflake/cli/api/metrics.py
```diff
@@ -86,7 +207,109 @@ def increment_counter(self, name: str, value: int = 1) -> None:
...
+            else:
+                raise CLIMetricsInvalidUsageError(
+                    "current step could not be ended because no steps are executing"
+                )
+
+        found_step.end(error)
+        self._executing_steps.remove(found_step)
+        self._finished_steps.append(found_step)
+
+    def flush_steps(self, error: Optional[BaseException] = None) -> None:
```
sfc-gh-bdufour, line 294: I can't see a use case for calling this without an error. Can you?

src/snowflake/cli/api/metrics.py
```diff
@@ -86,7 +207,109 @@ def increment_counter(self, name: str, value: int = 1) -> None:
...
     @property
     def counters(self) -> Dict[str, int]:
         # return a copy of the original dict to avoid mutating the original
         return self._counters.copy()
+
+    @property
+    def steps(self) -> List[Dict[str, Union[str, int, float, None]]]:
+        """
+        returns the finished steps tracked throughout a command, sorted by start time
+        """
```
sfc-gh-bdufour, line 311: I have no idea what the return type represents here, other than it's a list of dictionaries. Would `List[dict]` suffice? Basically you're returning a list of JSON objects.

src/snowflake/cli/api/metrics.py
```diff
@@ -86,7 +207,109 @@ def increment_counter(self, name: str, value: int = 1) -> None:
...
+
+        :raises CliMetricsInvalidUsageError: if the step name is empty
+        """
+        parent_step = self._current_step
+        new_step = _CLIMetricsStep(name, parent_step)
+        self._executing_steps.append(new_step)
+        new_step.start()
+        return new_step.step_id
+
+    def end_step(
```
sfc-gh-bdufour, line 249: have we considered modelling this as a self-contained Step structure, such that start/stop would be instance methods of a step? It would fit nicely with the context manager abstraction as well. I think it would help simplify the implementation here quite a bit too.

tests/api/test_metrics_steps.py
```diff
@@ -0,0 +1,407 @@
...
+    metrics = CLIMetrics()
+
+    # when
+    with metrics.track_step("step1"):
+        pass
+
+    # then
+    assert len(metrics.steps) == 1
+    step1 = metrics.steps[0]
+    assert step1[_CLIMetricsStep.NAME_KEY] == "step1"
```
sfc-gh-bdufour, line 45: please just merge a bunch of assertions into the same test. There is no need to write a single test case per field in the output.

tests/api/test_metrics_steps.py
```diff
@@ -0,0 +1,407 @@
...
+def test_metrics_steps_error_caught():
+    # given
+    metrics = CLIMetrics()
+
+    # when
+    with metrics.track_step("step1"):
+        try:
+            raise RuntimeError()
+        except RuntimeError:
+            pass
```
sfc-gh-bdufour, line 73: you're testing the python interpreter here. No value added.

tests/api/test_metrics_steps.py
```diff
@@ -0,0 +1,407 @@
...
+def test_metrics_steps_error_uncaught():
+    # given
+    metrics = CLIMetrics()
+
+    # when
+    try:
+        with metrics.track_step("step1"):
+            raise RuntimeError()
+    except RuntimeError:
+        pass
```
sfc-gh-bdufour, line 90: never do this in tests. assert that the exception is indeed propagated instead. Pytest makes that trivial.

tests/api/test_metrics_steps.py
```diff
@@ -0,0 +1,407 @@
...
+        with metrics.track_step("child"):
+            pass
+
+    # then
+    assert len(metrics.steps) == 2
+    parent, child = metrics.steps
+
+    assert child[_CLIMetricsStep.NAME_KEY] == "child"
+    assert child[_CLIMetricsStep.PARENT_KEY] == "parent"
+    assert parent[_CLIMetricsStep.NAME_KEY] == "parent"
```
sfc-gh-bdufour, line 127: weak checks. I'd make sure the returned info looks right more generally. Did both steps get terminated properly, etc

tests/api/test_metrics_steps.py
```diff
@@ -0,0 +1,407 @@
...
+
+
+def test_metrics_steps_with_duplicate_names_proper_timing():
+    # given
+    metrics = CLIMetrics()
+
+    # when
+    with metrics.track_step("duplicate"):
+        with metrics.track_step("duplicate"):
+            pass
```
sfc-gh-bdufour, line 154: this example worries me a little. We could blow up the metrics if we try to time a recursive method. Should this be allowed?

tests/api/test_metrics_steps.py
```diff
@@ -0,0 +1,407 @@
...
+
+    # then
+    assert len(metrics.steps) == 2
+    parent, child = metrics.steps
+
+    assert (
+        child[_CLIMetricsStep.START_TIME_KEY] > parent[_CLIMetricsStep.START_TIME_KEY]
```
sfc-gh-bdufour, line 161: likely flaky, would use `>=` at least since there is no guarantee that execution will keep being slow enough to trigger a different start time for both steps.

tests/api/test_metrics_steps.py
```diff
@@ -0,0 +1,407 @@
...
+    )
+    assert (
+        parent[_CLIMetricsStep.EXECUTION_TIME_KEY]
+        > child[_CLIMetricsStep.EXECUTION_TIME_KEY]
+    )
+
+
```
sfc-gh-bdufour, line 168: missing a test for multiple children under the same parent, esp multiple with the same name but diff ids?

tests/api/test_metrics_steps.py
```diff
@@ -0,0 +1,407 @@
...
+def test_metrics_steps_manual_start_and_end_overlapping_proper_parent():
+    # given
+    metrics = CLIMetrics()
+
+    # when
+    step1_id = metrics.start_step("step1")
+    step2_id = metrics.start_step("step2")
+
+    metrics.end_step(step_id=step2_id)
+    metrics.end_step(step_id=step1_id)
```
sfc-gh-bdufour, line 197: I don't get the reference to a parent in the test case name, it looks like there's no nesting at play here.

tests/api/test_metrics_steps.py
```diff
@@ -0,0 +1,407 @@
...
+
+    # then
+    assert len(metrics.steps) == 2
+    step1, step2 = metrics.steps
+
+    assert step2[_CLIMetricsStep.PARENT_KEY] == "step1"
```
sfc-gh-bdufour, line 203: yikes, that's surprising. Steps are implicitly nested? Let's discuss offline.

tests/api/test_metrics_steps.py
```diff
@@ -0,0 +1,407 @@
...
+    step1, step2 = metrics.steps
+    step1_end_time = (
+        step1[_CLIMetricsStep.START_TIME_KEY]
+        + step1[_CLIMetricsStep.EXECUTION_TIME_KEY]
+    )
+    step2_end_time = (
+        step2[_CLIMetricsStep.START_TIME_KEY]
+        + step2[_CLIMetricsStep.EXECUTION_TIME_KEY]
+    )
+    assert step1_end_time > step2_end_time
```
sfc-gh-bdufour, line 229: ditto, could be flaky

src/snowflake/cli/api/metrics.py
sfc-gh-mchok: Converted to dataclass

Regarding your suggestion to implement `to_dict` as `dict((field.name, getattr(obj, field.name)) for field in fields(obj) if field.name in FIELDS_TO_SERIALIZE)`, I think that would add additional complexity and make it harder to reason about, since we also add additional keys to the dict that are extracted (i.e. parent name, parent id, error name) from the dataclass fields. I think it's simpler to be explicit here since there aren't too many keys

Let me know if you have other thoughts!

src/snowflake/cli/api/metrics.py
sfc-gh-mchok: I chose "start_span" as the new name based on my research below:

Every library seems to yield some kind of `Span` object, but in terms of function names, datadog uses `.trace()`, otel uses `.start_as_current_span()`, and sentry uses `.start_span()`, so I went with the simplest common denominator which to me was `start_span()`, since I guess we would count our instance of `CLIMetrics` as a trace in our context. I'm also open to renaming that class if we feel the nomenclature does not fit the concept

src/snowflake/cli/api/metrics.py
sfc-gh-mchok: Since we are no longer allowing manual start/stop in the initial version of this API, this is no longer needed

src/snowflake/cli/api/metrics.py
sfc-gh-mchok: Since we are no longer allowing manual start/stop in the initial version of this API, this function is no longer needed

src/snowflake/cli/api/metrics.py
sfc-gh-mchok: Since we are no longer allowing manual start/stop in the initial version of this API, this function is no longer needed

src/snowflake/cli/api/metrics.py
```diff
@@ -86,7 +207,109 @@ def increment_counter(self, name: str, value: int = 1) -> None:
...
         return self._counters.copy()
+
+    @property
+    def steps(self) -> List[Dict[str, Union[str, int, float, None]]]:
+        """
+        returns the finished steps tracked throughout a command, sorted by start time
+        """
+        return [
+            step.to_dict()
+            for step in sorted(self._finished_steps, key=lambda step: step.start_time)
```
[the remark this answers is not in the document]
sfc-gh-mchok: renamed to `completed_spans`

src/snowflake/cli/api/metrics.py
sfc-gh-mchok: Discussed offline: we yield the span here in case we want to perform operations with it in the future, however for now we're not supporting manual start/stop and only supporting through context managers

tests/api/test_metrics_steps.py
sfc-gh-mchok: 407 lines -> 283 😅

tests/api/test_metrics_steps.py
sfc-gh-mchok: Discussed offline: adding a depth/total limit for number of spans, added tests for those as well

tests/api/test_metrics_steps.py
sfc-gh-mchok: Added: `test_metrics_spans_parent_with_two_children_same_name`

tests/api/test_metrics_steps.py
sfc-gh-mchok: Discussed offline: no more manual start/stop for spans in initial version of API, only supported through context managers, so spans are explicitly nested now

tests/api/test_metrics_steps.py
sfc-gh-mchok: Discussed offline: no more manual start/stop for spans in initial version of API, only supported through context managers, so spans are explicitly nested now

src/snowflake/cli/api/metrics.py
```diff
@@ -86,7 +207,109 @@ def increment_counter(self, name: str, value: int = 1) -> None:
...
+        self._executing_steps.remove(found_step)
+        self._finished_steps.append(found_step)
+
+    def flush_steps(self, error: Optional[BaseException] = None) -> None:
+        """
+        Useful if you are using the manual start/end steps and an error
+        propagated up, requiring you to clear out all the executing steps
+        """
+        while self._current_step:
+            self.end_step(error=error)
```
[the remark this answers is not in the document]
sfc-gh-mchok: No more manual start/stop, so this method is no longer needed

.github/CODEOWNERS
```diff
@@ -11,6 +11,8 @@ tests_common/__init__.py @snowflakedb/snowcli @snowflakedb/nade
 tests_common/conftest.py @snowflakedb/snowcli @snowflakedb/nade
 tests_integration/conftest.py @snowflakedb/snowcli @snowflakedb/nade
 tests_integration/test_utils.py @snowflakedb/snowcli @snowflakedb/nade
+src/snowflake/cli/api/metrics.py @snowflakedb/snowcli @snowflakedb/nade
+tests/api/metrics/ @snowflakedb/snowcli @snowflakedb/nade
```
sfc-gh-mchok, line 15: I anticipate some friction in the future with us modifying these files a lot and then requiring cli for review when we are the sole contributors, so I added them to shared ownership to prevent that

tests/api/test_global_cli_context.py
```diff
@@ -84,3 +85,22 @@ def test_forked_context():
...
+                pass
+
+    assert outer_metrics != inner_metrics
+    assert outer_metrics.completed_spans != inner_metrics.completed_spans
+
+    assert len(outer_metrics.completed_spans) == len(inner_metrics.completed_spans) == 1
+
+    assert outer_metrics.completed_spans[0][CLIMetricsSpan.NAME_KEY] == "outer_span"
+
+    assert inner_metrics.completed_spans[0][CLIMetricsSpan.NAME_KEY] == "inner_span"
```
sfc-gh-mchok, line 106: To me it makes sense that we should reset the spans when we fork the context, since otherwise it could eventually hit the limit if we keep forking the same context and starting spans, but would like to hear if you think otherwise

src/snowflake/cli/api/metrics.py
```diff
@@ -54,18 +70,96 @@ class CLICounterField:
...
+            self.PARENT_KEY: self.parent.name if self.parent is not None else None,
+            self.PARENT_ID_KEY: self.parent.step_id
+            if self.parent is not None
+            else None,
+            self.EXECUTION_TIME_KEY: self.execution_time,
+            self.ERROR_KEY: type(self.error).__name__ if self.error else None,
+        }
+
+
+@dataclass
```
sfc-gh-mchok, line 136: Converted to dataclass so that we get equality comparison and cloning for free to help with forking the cli context

tests/api/metrics/test_metrics_spans.py
```diff
@@ -0,0 +1,288 @@
...
+def test_metrics_spans_parent_with_one_child():
+    # given
+    metrics = CLIMetrics()
+
+    # when
+    with metrics.start_span("parent") as parent:
+        assert metrics.current_span is parent
+        # on windows, these sleeps are required to avoid the steps ending up on the same
+        # timestamps/monotonic starts and ensure a deterministic order for the completed steps
+        time.sleep(0.1)
```
sfc-gh-mchok, line 112: I ran into an issue with testing on windows where `time.monotonic()` only has 7 significant digits of precision, causing parent/child to be sorted in a non-deterministic and flaky order since they could have the same start time. I tried to remedy this by introducing `time.time()` as another tiebreaker for sorting, but even that ended up being the same for both spans during my tests

To remedy this, I added some sleeps in the tests which I believe simulates more realistic scenarios (since we probably won't be starting spans immediately after each other) and allows us to do a couple of things:
- Keep implementation simple
- Test whether we have the proper execution/start time inequalities (> instead of >=) between spans

In total, we have 0.7s of sleep time in these tests, which I believe is a negligible amount and shouldn't make our tests significantly longer

Let me know if you have any thoughts on alternative approaches!

src/snowflake/cli/api/metrics.py
```diff
@@ -86,7 +181,66 @@ def increment_counter(self, name: str, value: int = 1) -> None:
...
+        Only one span can be active at a time; once one span finishes its parent becomes the active span
+
+        Spans are not emitted in telemetry if depth/total limits are exceeded
+
+        :raises CliMetricsInvalidUsageError: if the step name is empty
+        """
+        new_step = CLIMetricsSpan(
+            name=name,
+            start_time=time.monotonic() - self._monotonic_start,
+            parent=self.current_span,
```
sfc-gh-fcampbell, line 205: do we want to allow passing in a custom parent? that would allow us to trace parallel operations:
```python
@metrics.start_span(name="put_file", parent=metrics.current_span)
def _put_file(path: Path):
    ...

with concurrent.futures.ThreadPoolExecutor() as executor:
    for file in files_to_upload:
        executor.submit(_put_file, file)
```

which would give us a span structure like
```
+------------------------------------+
| root                               |
+------------------------------------+
  +--------------------------+
  | parent                   |
  +--------------------------+
    +-----------------+
    | put_file        |
    +-----------------+
    +-------------+
    | put_file    |
    +-------------+
    +---------------+
    | put_file      |
    +---------------+  
```

src/snowflake/cli/api/metrics.py
```diff
@@ -86,7 +181,66 @@ def increment_counter(self, name: str, value: int = 1) -> None:
...
+        )
+
+        if len(self._in_progress_spans) >= self.IN_PROGRESS_SPANS_DEPTH_LIMIT:
+            self.num_spans_past_depth_limit += 1
+        else:
+            self._in_progress_spans.append(new_step)
+
+        try:
+            yield new_step
+        except BaseException as err:
```
sfc-gh-fcampbell, line 215:
```suggestion
        except Exception as err:
```
we don't want to catch `KeyboardInterrupt`, `SystemExit`, etc since they're not errors

src/snowflake/cli/api/metrics.py
```diff
@@ -86,7 +181,66 @@ def increment_counter(self, name: str, value: int = 1) -> None:
...
+        new_step = CLIMetricsSpan(
+            name=name,
+            start_time=time.monotonic() - self._monotonic_start,
+            parent=self.current_span,
+        )
+
+        if len(self._in_progress_spans) >= self.IN_PROGRESS_SPANS_DEPTH_LIMIT:
+            self.num_spans_past_depth_limit += 1
+        else:
+            self._in_progress_spans.append(new_step)
```
sfc-gh-fcampbell, line 211: this will break `current_span` once we're past the depth limit (all future spans won't have the correct parent)

tests/api/metrics/test_metrics_spans.py
sfc-gh-fcampbell: we could also use https://github.com/spulec/freezegun

tests/api/test_global_cli_context.py
sfc-gh-fcampbell: if we start a span before forking, that context will be gone after the fork. it's probably ok, since we most likely don't want to start any spans pre-fork (otherwise it would be a very long span since the interpreter could be very long-lived)

src/snowflake/cli/api/metrics.py
sfc-gh-mchok: I think it *might* be helpful for something like `KeyboardInterrupt` if our users are stuck on a super long step and they consistently exit out, wdyt?

src/snowflake/cli/api/metrics.py
sfc-gh-mchok: I added support for this, had to add a contextvar in the class to make sure that `metrics.current_span` is thread local, but I couldn't find a way to make it so that it is automatically propagated between threads

E.g. for the below code, assertion 1 will fail but assertion 2 will fail:
```python
    with metrics.start_span("parent") as parent_span:

        def parallel_operation(seq_num: int):
            assert metrics.current_span is parent_span # ASSERTION 1
            with metrics.start_span("child", parent=parent_span) as local_span:
                assert metrics.current_span is local_span # ASSERTION 2

        executor = concurrent.futures.ThreadPoolExecutor()
        futures = [executor.submit(parallel_operation, seq_num=i) for i in range(3)]
```

Do you know if there is a way around this or are we ok with declaring this as undefined behaviour?

tests/api/metrics/test_metrics_spans.py
sfc-gh-mchok: From my understanding that library freezes the values returned by the `time` functions? I'm not sure if that solves the issue that two calls to `time.monotonic()` on windows in short succession could return the same value, which breaks the sort key, but maybe I'm misunderstanding the library?

src/snowflake/cli/api/metrics.py
sfc-gh-mchok: Fixed and added a test for it here: https://github.com/snowflakedb/snowflake-cli/pull/1788/files#diff-839165dddafe1dd908ef843269b12e1cd4a7158ea95e5e7a2ee210ea7ec56de0R350

src/snowflake/cli/api/metrics.py
sfc-gh-fcampbell: discussed offline, let's hold off on this to avoid complexity for now. we can revisit later if necessary

tests/api/metrics/test_metrics_spans.py
sfc-gh-fcampbell: it also allows you to fake the time by setting it to a fixed value or advancing/rewinding it at will

sfc-gh-fcampbell: for this fast of a sleep it's ok, just wanted to make you aware of the lib since it can come in handy

sfc-gh-bdufour: Sorry, unresolving this. Relying on actual time + sleep in tests is a red flag, and almost always ends up having negative side effects down the line. We don't need to include a special library, we should be able to get away with mocking directly here. I've also used dep injection in the past for testing stuff like this, but it's probably overkill here.

sfc-gh-fcampbell (dismissed):

tests/api/metrics/test_metrics_spans.py
sfc-gh-mchok: Good callout, added some [mocks for time.monotonic in the tests](https://github.com/snowflakedb/snowflake-cli/pull/1788/files#diff-839165dddafe1dd908ef843269b12e1cd4a7158ea95e5e7a2ee210ea7ec56de0R163-R177), had to convert the default_factory of the dataclass fields into lambdas instead of direct references so that they could be patched properly

sfc-gh-fcampbell (approved):

sfc-gh-turbaszek (approved):

PR #1791, opened 2024-10-24
[SNOW-1739737] BUGFIX: Setting a default connection copies the connection from connections.toml to config.toml

sfc-gh-pjob:
### Pre-review checklist

### Changes description
  * Changed scope of unit tests' `test_snowcli_config` fixture from `session` to `function` to avoid side effects between tests.
  * Added an unit test confirming the bug
  * Bugfix in `config.py`. We were saving whole content of "conf file cache" to `config.toml`. Now we're saving connections to `connections.toml` if that file exists.

### Bug description
  * Conditions:
    * No connections in `config.toml`
    * Only one connection in `connections.toml` (“dev” connection)
    * Invoked command: `snow connection set-default "dev"`
  * Incorrect behaviour:
    * Default connection was changed correctly but “dev” connection details were copied from `connections.toml` to `config.toml`. User may don’t want to keep connection secrets in config.toml so IMO it looks like security issue.
  * Reason:
    * We were saving whole content of "conf file cache" to `config.toml` forgetting about existence of connections.toml as a source of connections.

### DIFF
--- a/src/snowflake/cli/api/config.py
+++ b/src/snowflake/cli/api/config.py
@@ -345,6 +345,9 @@ def _dump_config(config_and_connections: Dict):
 
     if CONNECTIONS_FILE.exists():
         # update connections in connections.toml
+        # it will add only connections (maybe updated) which were originally read from connections.toml
+        # it won't add connections from config.toml
+        # because config manager doesn't have connections from config.toml if connections.toml exists
         _update_connections_toml(config_and_connections.get("connections") or {})
         # to config.toml save only connections from config.toml
         connections_to_save_in_config_toml = _read_config_file_toml().get("connections")
--- a/src/snowflake/cli/api/config.py
+++ b/src/snowflake/cli/api/config.py
@@ -340,12 +340,12 @@ def _get_envs_for_path(*path) -> dict:
     }
 
 
-def _dump_config(whole_conf_cache: Dict):
-    config_toml_dict = whole_conf_cache.copy()
+def _dump_config(config_and_connections: Dict):
+    config_toml_dict = config_and_connections.copy()
 
     if CONNECTIONS_FILE.exists():
         # update connections in connections.toml
-        _update_connections_toml(whole_conf_cache.get("connections") or {})
+        _update_connections_toml(config_and_connections.get("connections") or {})
         # to config.toml save only connections from config.toml
         connections_to_save_in_config_toml = _read_config_file_toml().get("connections")
         config_toml_dict["connections"] = connections_to_save_in_config_toml
@@ -383,13 +383,11 @@ def get_feature_flags_section() -> Dict[str, bool | Literal["UNKNOWN"]]:
 
 
 def _read_config_file_toml() -> dict:
-    with open(CONFIG_MANAGER.file_path, "r") as f:
-        return tomlkit.loads(f.read()).unwrap()
+    return tomlkit.loads(CONFIG_MANAGER.file_path.read_text())
 
 
 def _read_connections_toml() -> dict:
-    with open(CONNECTIONS_FILE, "r") as f:
-        return tomlkit.loads(f.read()).unwrap()
+    return tomlkit.loads(CONNECTIONS_FILE.read_text())
 
 
 def _update_connections_toml(connections: dict):
--- a/src/snowflake/cli/api/config.py
+++ b/src/snowflake/cli/api/config.py
@@ -340,9 +340,18 @@ def _get_envs_for_path(*path) -> dict:
     }
 
 
-def _dump_config(conf_file_cache: Dict):
+def _dump_config(whole_conf_cache: Dict):
+    config_toml_dict = whole_conf_cache.copy()
+
+    if CONNECTIONS_FILE.exists():
+        # update connections in connections.toml
+        _update_connections_toml(whole_conf_cache.get("connections") or {})
+        # to config.toml save only connections from config.toml
+        connections_to_save_in_config_toml = _read_config_file_toml().get("connections")
+        config_toml_dict["connections"] = connections_to_save_in_config_toml
+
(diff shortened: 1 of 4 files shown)


tests/testing_utils/fixtures.py
```diff
@@ -268,7 +268,7 @@ def temp_directory_for_app_zip(temp_dir) -> Generator:
     yield temp_dir.name
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="function")
```
sfc-gh-pjob, line 271: Changed scope of unit tests' `test_snowcli_config` fixture from `session` to `function` to avoid side effects between tests.

tests/test_config.py
```diff
@@ -223,6 +224,78 @@ def test_not_found_default_connection_from_evn_variable(test_root_path):
...
+            "asdf_b": {
+                "database": "asdf_b_database",
+                "user": "asdf_b",
+                "account": "asdf_b",
+            },
+        }
+
+    assert_correct_connections_loaded()
+
+    with open(connections_toml) as f:
```
sfc-gh-pjob, line 265: A lot of assertions to be sure that everything lands where it belongs.

tests/test_config.py
```diff
@@ -223,6 +224,78 @@ def test_not_found_default_connection_from_evn_variable(test_root_path):
...
+    [asdf_b]
+    database = "asdf_b_database"
+    user = "asdf_b"
+    account = "asdf_b"
+    """
+    )
+    config_init(test_snowcli_config)
+    set_config_value(section=None, key="default_connection_name", value="asdf_b")
+
+    def assert_correct_connections_loaded():
```
sfc-gh-pjob, line 248: Used 2 times:
* after setting `default_connection_name` to be sure that everything is ok in the memory at that moment
* after reinit of the config from the file to be sure that loading from file still is ok

src/snowflake/cli/api/config.py
```diff
@@ -340,9 +340,18 @@ def _get_envs_for_path(*path) -> dict:
     }
 
 
-def _dump_config(conf_file_cache: Dict):
+def _dump_config(whole_conf_cache: Dict):
```
sfc-gh-pjob, line 343: The whole fix is here

src/snowflake/cli/api/config.py
```diff
@@ -373,6 +382,11 @@ def _bool_or_unknown(value):
     return {k: _bool_or_unknown(v) for k, v in flags.items()}
 
 
+def _read_config_file_toml() -> dict:
+    with open(CONFIG_MANAGER.file_path, "r") as f:
+        return tomlkit.loads(f.read()).unwrap()
```
sfc-gh-mraba, line 387:
```suggestion
    return tomlkit.loads(CONFIG_MANAGER.file_path.read_text())
```

tests/test_config.py
sfc-gh-mraba: What do you think about reflecting this comments in test code by splitting in functions?
This may help in future read the code without searching for PR.

src/snowflake/cli/api/config.py
```diff
@@ -340,9 +340,18 @@ def _get_envs_for_path(*path) -> dict:
     }
 
 
-def _dump_config(conf_file_cache: Dict):
+def _dump_config(whole_conf_cache: Dict):
+    config_toml_dict = whole_conf_cache.copy()
+
+    if CONNECTIONS_FILE.exists():
+        # update connections in connections.toml
+        _update_connections_toml(whole_conf_cache.get("connections") or {})
```
sfc-gh-turbaszek, line 348: Won't it add all connections from config.toml to connections.toml?

src/snowflake/cli/api/config.py
```diff
@@ -340,9 +340,18 @@ def _get_envs_for_path(*path) -> dict:
     }
 
 
-def _dump_config(conf_file_cache: Dict):
+def _dump_config(whole_conf_cache: Dict):
+    config_toml_dict = whole_conf_cache.copy()
```
sfc-gh-turbaszek, line 344: Can we replace `whole` with something like `config_and_connections`?

sfc-gh-pjob: Done :)

src/snowflake/cli/api/config.py
sfc-gh-pjob: No, it won't. It will add only connections loaded to config manager and if CONNECTIONS_FILE exists then there will be no config.toml connections there (in the manager).

There is a test assertion for it (that "jwt" connection from config.toml isn't copied to connections.toml): https://github.com/snowflakedb/snowflake-cli/pull/1791/files#diff-da24b355349b53e4c8047034767d008e4c2ce9ea6c8933a3acd50d0828f00528R276

src/snowflake/cli/api/config.py
sfc-gh-pjob: Done - for this file and few lines later for connections.toml :)

tests/test_config.py
sfc-gh-pjob: IMO it won't help in the case of this test. IMO it only increase complexity of reading.

I added comments to PR only to simplify your reviews but IMO they are not required to read the code because these are not complicated changes.

Anyway I added one more comment line in the test case to make it more explanatory.

src/snowflake/cli/api/config.py
sfc-gh-turbaszek: Let's add this comment. Imho it's not obvious that only connection.toml connections are loaded when both files are present and ahve connection details

sfc-gh-pjob: Done :)

sfc-gh-turbaszek (approved):

sfc-gh-mraba (approved):

PR #1796, opened 2024-10-25
Move usage telemetry to post_execute callback

sfc-gh-turbaszek:
### Pre-review checklist

### Changes description
Move telemetry sending to post execute callback to improve behavior in cases like `snow connection set-default`

### DIFF
--- a/src/snowflake/cli/api/commands/snow_typer.py
+++ b/src/snowflake/cli/api/commands/snow_typer.py
@@ -125,10 +125,7 @@ class SnowTyper(typer.Typer):
         Pay attention to make this method safe to use if performed operations are not necessary
         for executing the command in proper way.
         """
-        from snowflake.cli._app.telemetry import log_command_usage
-
         log.debug("Executing command pre execution callback")
-        log_command_usage(execution)
         if require_warehouse and not SqlExecutionMixin().session_has_warehouse():
             raise ClickException(
                 "The command requires warehouse. No warehouse found in current connection."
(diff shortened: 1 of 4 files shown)


tests/app/test_telemetry.py
```diff
@@ -174,7 +174,7 @@ def test_failing_executing_command_sends_telemetry_data(
     # The method is called with a TelemetryData type, so we cast it to dict for simpler comparison
     result_command_event = (
         mock_conn.return_value._telemetry.try_add_log_to_batch.call_args_list[  # noqa: SLF001
-            1
+            0
```
sfc-gh-pjob, line 177: Why this change in the test was required?
Does it mean that failing command won't send telemetry data?

sfc-gh-turbaszek: It will, but it will first send error then usage info. The change is caused by the order of callbacks (pre, error, post). By moving usage from pre to post we changed the events order.

src/snowflake/cli/api/connections.py
```diff
@@ -80,6 +80,7 @@ def update(self, **updates):
             setattr(self, key, value)
 
     def update_from_config(self) -> ConnectionContext:
+        self.validate_and_complete()
```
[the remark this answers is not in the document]
sfc-gh-turbaszek: Sending telemetry data before execution was causing creation of connection, including connection name resolution (for example falling back to default conn). This method is called first here because it resolves `self.connection_name`. Without calling it the value is `None`

sfc-gh-pjob (approved):

sfc-gh-fcampbell (approved):

src/snowflake/cli/api/commands/snow_typer.py
```diff
@@ -166,9 +163,14 @@ def post_execute(execution: ExecutionMetadata):
...
         """
-        from snowflake.cli._app.telemetry import flush_telemetry, log_command_result
+        from snowflake.cli._app.telemetry import (
+            flush_telemetry,
+            log_command_result,
+            log_command_usage,
+        )
 
         log.debug("Executing command post execution callback")
+        log_command_usage(execution)
```
sfc-gh-melnacouzi, line 173: does this mean that commands that do not require connection will not be logged anymore?

PR #1800, opened 2024-10-25
SNOW-1758715 Move small helper methods out of ApplicationEntity class

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Moves `application_objects_to_str()` out from the `ApplicationEntity` class to be a module-level function and `get_account_event_table()` to the SQL facade since they're not part of the public API of the entity

### DIFF
--- a/src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
+++ b/src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
@@ -194,7 +194,16 @@ class SnowflakeSQLFacade:
             except Exception as err:
                 handle_unclassified_error(err, f"Failed to run script {script_name}.")
 
-    def get_account_event_table(self) -> str:
+    def get_account_event_table(self, role: str | None = None) -> str | None:
+        """
+        Returns the name of the event table for the account.
+        If the account has no event table set up or the event table is set to NONE, returns None.
+        @param [Optional] role: Role to switch to while running this script. Current role will be used if no role is passed in.
+        """
         query = "show parameters like 'event_table' in account"
-        results = self._sql_executor.execute_query(query, cursor_class=DictCursor)
-        return next((r["value"] for r in results if r["key"] == "EVENT_TABLE"), "")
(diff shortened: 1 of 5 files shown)


src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -193,3 +193,8 @@ def execute_user_script(
                     raise UserScriptError(script_name, err.msg) from err
             except Exception as err:
                 handle_unclassified_error(err, f"Failed to run script {script_name}.")
+
+    def get_account_event_table(self) -> str:
```
sfc-gh-bdufour, line 197: 1. Is the current role always fine here?
2. Shouldn't the RV be optional, since the table may not exist, instead of returning "" ?
3. Should we handle "NONE" here and return no event table?

sfc-gh-fcampbell: yeah those are fair points. 
1. We could at least try the package role (and warehouse) since this would be fetching events in the provider's account
2. I'm not a fan of returning null values where zero values could be used, but in this case since `""` is a valid identifier, i'd be ok
3. yeah makes sense. Right now the check is made where this function is called, but I can move it into here

Can I do these as a followup to keep the change separate from the rest of this PR (since this PR is meant just as a straight refactor)?

sfc-gh-bdufour: 1. That could be a follow up, but I'd suggest making it possible to specify an optional role in the facade API. That's consistent with Parya's work.
2. Yeah, the issue is that empty is ambiguous. For identifiers we should return an `Optional[Identifier]` instance now I guess, or `Optional[str]` if we really don't need an FQN.
3. Ok, agreed.

I think 2 & 3 can probably be done here since it's a small PR anyway. As a reviewer, it's so much easier if we close the discussions in a PR.

sfc-gh-bdufour (approved):

PR #1803, opened 2024-10-25
SNOW-1789074: SPCS SnowCLI Log Streaming

sfc-gh-ashen:
### Pre-review checklist

### Changes description
The updated snow spcs service logs command now supports additional parameters for enhanced log retrieval and monitoring:
--since: Start log retrieval from a specified UTC timestamp.
--include-timestamps: Add timestamps to log entries for log streaming.
--follow (-f): Stream logs in real-time.
--follow-interval: Customize polling intervals during log streaming.


src/snowflake/cli/_plugins/spcs/services/commands.py
```diff
@@ -221,20 +225,68 @@ def logs(
     num_lines: int = typer.Option(
         500, "--num-lines", help="Number of lines to retrieve."
     ),
+    previous_logs: bool = typer.Option(
```
sfc-gh-tkommineni, line 228: why are we exposing previous_logs

src/snowflake/cli/_plugins/spcs/common.py
```diff
@@ -95,5 +95,41 @@ def handle_object_already_exists(
...
 
+def extract_log(log):
+    return log[0] if isinstance(log, tuple) else log
+
+
+def filter_log_timestamp(log: str, include_timestamps: bool) -> str:
+    if include_timestamps:
+        return log
+    else:
+        return log.split(" ", 1)[1] if " " in log else log
```
sfc-gh-tkommineni, line 106: why can we not write simply
log.split(" ")[1]

src/snowflake/cli/_plugins/spcs/services/manager.py
```diff
@@ -133,12 +135,65 @@ def status(self, service_name: str) -> SnowflakeCursor:
...
+                    new_log_records = log_block.split("\n")
+                    new_log_records = [line for line in new_log_records if line.strip()]
+                    if new_log_records:
+                        new_log_lines = _new_logs_only(
+                            prev_log_records, new_log_records
+                        )
+                        yield new_log_lines
+
+                        prev_log_records = new_log_lines
+                        prev_timestamp = prev_log_records[-1].split(" ", 1)[0]
```
sfc-gh-tkommineni, line 190: if new_log_records:
                        dedup_log_records = _new_logs_only(
                            prev_log_records, new_log_records
                        )
                        yield dedup_log_records

      
                        prev_timestamp = dedup_log_records[-1].split(" ", 1)[0]
                        prev_log_records = dedup_log_records

src/snowflake/cli/_plugins/spcs/services/commands.py
sfc-gh-ashen: But as you mentioned, it's okay to keep the code this way. The previous_logs parameter is included in the system function, and if we remove it, it would be confusing to pass a false in every call. This is because we should only pass a value for since when previous_logs has a value. When calling the system function, we can also demonstrate the functionality of previous_logs using SnowCLI during the presentation.

sfc-gh-turbaszek: From end user perspective "Retrieve previous logs" is very ambiguous. How many previous logs?

src/snowflake/cli/_plugins/spcs/services/commands.py
```diff
@@ -221,20 +225,68 @@ def logs(
...
+        False, "--include-timestamps", help="Include timestamps in logs", is_flag=True
+    ),
+    follow: bool = typer.Option(
+        False, "--follow", "-f", help="Continue polling for logs.", is_flag=True
+    ),
+    follow_interval: int = typer.Option(
+        2,
+        "--follow-interval",
+        help="Polling interval in seconds when using the --follow flag",
+    ),
```
sfc-gh-turbaszek, line 244: Why not recommend using `watch -n snow spcs service logs`?

sfc-gh-turbaszek, line 244: If we decide to have this functionality I would suggest combining this in into `--refresh-interval` that if provided will cause logs to be refreshed every few seconds. This also requires users to provide a number.

src/snowflake/cli/_plugins/spcs/services/commands.py
```diff
@@ -221,20 +225,68 @@ def logs(
...
     ),
+    previous_logs: bool = typer.Option(
+        False, "--previous-logs", "-p", help="Retrieve previous logs", is_flag=True
+    ),
+    since_timestamp: Optional[str] = typer.Option(
+        "", "--since", help="Timestamp to retrieve logs from"
+    ),
+    include_timestamps: bool = typer.Option(
+        False, "--include-timestamps", help="Include timestamps in logs", is_flag=True
+    ),
```
sfc-gh-turbaszek, line 236: What's the purpose of logs if there's no timestamp? Is this a necessary feature? If we don;'t have strong need for it I would suggest to add it later on once we see business/user need.

src/snowflake/cli/_plugins/spcs/services/commands.py
sfc-gh-ashen: This is another feature not related to log streaming, it is the log from last terminated instance.

src/snowflake/cli/_plugins/spcs/services/commands.py
sfc-gh-ashen: It is necessary, we need timestamps to track logs accurately. Since not all logs have timestamps, tracking becomes difficult. To address this, we introduced the include-timestamps, which adds a timestamp to each log, allowing us to reliably track the last log for the next retrieval cycle.

src/snowflake/cli/_plugins/spcs/services/commands.py
sfc-gh-tkommineni: There are several reasons for this

1. `watch -n 2 snow spcs service logs` has some drawbacks, this call will get last 1000 log lines every time invoked.
2.  The functionality that we want to support is streaming (To append new logs to a current set of logs).
      - We offer customers to stream from a particular timestamp using --since flag
      - We offer deduplication logic
      - We are also going to perform load testing to understand backend api limits anc come up with a follow interval
      - We plan to implement exponential backoff
3.  Here is the design doc - https://docs.google.com/document/d/1hvGC97jfvhfcg6gyPnFUl5j2-Xc1MKWy6C7ftc5JZ1c/edit?tab=t.0#heading=h.an8exbgprj8y
4.  This is also similar to what the events does here currently - 
https://github.com/snowflakedb/snowflake-cli/blob/dfefe77e68ee88a4f529d9eef6a6f6c1bf71b09c/src/snowflake/cli/_plugins/nativeapp/entities/application.py#L816

src/snowflake/cli/_plugins/spcs/services/commands.py
sfc-gh-tkommineni: The intention for this flag is as follows, may be we can chose better naming:
1. For every log record K8s adds a timestamp
2. Customers services which emit logs can also have timestamps in some cases.
3. If customer services already emit timestamps k8s timestamps might be redundant.
4. Hence, this flag is to turn on/off the k8s appended timestamps to each log record

PR #1805, opened 2024-10-25
Create shortcut properties for entity model fields

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
Adds properties to the native app entities to make usage of model fields a bit simpler and safer (especially the derived ones like paths).

For example, to get the package name, we can call `self.name` instead of `self._entity_model.fqn.name`.

### DIFF
--- a/src/snowflake/cli/_plugins/nativeapp/entities/application.py
+++ b/src/snowflake/cli/_plugins/nativeapp/entities/application.py
@@ -119,24 +119,24 @@ class ApplicationEntity(EntityBase[ApplicationEntityModel]):
     """
 
     @property
-    def project_root(self):
+    def project_root(self) -> Path:
         return self._workspace_ctx.project_root
 
     @property
-    def package_entity_id(self):
+    def package_entity_id(self) -> str:
         return self._entity_model.from_.target
 
     @property
-    def name(self):
+    def name(self) -> str:
         return self._entity_model.fqn.name
 
     @property
-    def role(self):
+    def role(self) -> str:
         model = self._entity_model
         return (model.meta and model.meta.role) or self._workspace_ctx.default_role
 
     @property
-    def warehouse(self):
+    def warehouse(self) -> str:
         model = self._entity_model
         return (
             model.meta and model.meta.warehouse and to_identifier(model.meta.warehouse)
--- a/src/snowflake/cli/_plugins/nativeapp/entities/application.py
+++ b/src/snowflake/cli/_plugins/nativeapp/entities/application.py
@@ -64,6 +64,7 @@ from snowflake.cli.api.metrics import CLICounterField
 from snowflake.cli.api.project.schemas.entities.common import (
     EntityModelBase,
     Identifier,
+    PostDeployHook,
     TargetField,
 )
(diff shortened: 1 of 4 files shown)


src/snowflake/cli/_plugins/nativeapp/entities/application.py
```diff
@@ -118,6 +118,35 @@ class ApplicationEntity(EntityBase[ApplicationEntityModel]):
     A Native App application object, created from an application package.
     """
 
+    @property
+    def project_root(self):
```
sfc-gh-bdufour, line 122: can we add type annotations to these properties, for clarity?

src/snowflake/cli/_plugins/nativeapp/entities/application.py
```diff
@@ -118,6 +118,35 @@ class ApplicationEntity(EntityBase[ApplicationEntityModel]):
...
+        return self._workspace_ctx.project_root
+
+    @property
+    def package_entity_id(self):
```
sfc-gh-bdufour, line 126: is this guaranteed to be a valid SF identifier? If not, should it be? Same question for all identifier properties here.

src/snowflake/cli/_plugins/nativeapp/entities/application.py
```diff
@@ -118,6 +118,35 @@ class ApplicationEntity(EntityBase[ApplicationEntityModel]):
...
+
+    @property
+    def warehouse(self):
+        model = self._entity_model
+        return (
+            model.meta and model.meta.warehouse and to_identifier(model.meta.warehouse)
+        ) or to_identifier(self._workspace_ctx.default_warehouse)
+
+    @property
+    def post_deploy_hooks(self):
```
sfc-gh-bdufour, line 146: looks like this can return a None value. Should it always return a valid list, even if empty, instead?

src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
```diff
@@ -147,6 +147,51 @@ class ApplicationPackageEntity(EntityBase[ApplicationPackageEntityModel]):
     A Native App application package.
     """
 
+    @property
```
sfc-gh-bdufour, line 150: ditto for package entity

src/snowflake/cli/_plugins/nativeapp/entities/application.py
sfc-gh-fcampbell: no, this is the entity ID from snowflake.yml

src/snowflake/cli/_plugins/nativeapp/entities/application.py
sfc-gh-fcampbell: added

src/snowflake/cli/_plugins/nativeapp/entities/application.py
sfc-gh-mchok: Does it make sense to return Optional[List[...]] here then or did we skip annotating this on purpose?

sfc-gh-fcampbell: oops, updated

sfc-gh-mchok (approved):

PR #1806, opened 2024-10-28
remove old error and use new version from sql facade

sfc-gh-mchok:
### Pre-review checklist

### Changes description
- Two versions of `CouldNotUseObjectError` made it into the codebase separately. This PR consolidates both versions and chooses the one from the sql facade to be used in place of the old one

### DIFF
--- a/src/snowflake/cli/_plugins/connection/commands.py
+++ b/src/snowflake/cli/_plugins/connection/commands.py
@@ -26,6 +26,7 @@ from snowflake.cli._plugins.connection.util import (
     strip_and_check_if_exists,
     strip_if_value_present,
 )
+from snowflake.cli._plugins.nativeapp.sf_facade_constants import UseObjectType
 from snowflake.cli._plugins.object.manager import ObjectManager
 from snowflake.cli.api.cli_global_context import get_cli_context
 from snowflake.cli.api.commands.flags import (
@@ -42,7 +43,6 @@ from snowflake.cli.api.config import (
     set_config_value,
 )
 from snowflake.cli.api.console import cli_console
-from snowflake.cli.api.constants import ObjectType
 from snowflake.cli.api.output.types import (
     CollectionResult,
     CommandResult,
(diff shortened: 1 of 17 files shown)


src/snowflake/cli/api/sql_execution.py
```diff
@@ -21,11 +21,11 @@
 from textwrap import dedent
 from typing import Iterable, Optional, Tuple
 
+from snowflake.cli._plugins.nativeapp.sf_facade_constants import UseObjectType
+from snowflake.cli._plugins.nativeapp.sf_facade_exceptions import CouldNotUseObjectError
```
sfc-gh-mchok, line 25: Unsure if we want to move these into the `api` folder if we're going this route or leave it like this

src/snowflake/cli/_plugins/connection/commands.py
```diff
@@ -309,13 +309,13 @@ def test(
         schema = conn.schema
 
         if conn.role:
-            om.use(object_type=ObjectType.ROLE, name=f'"{conn.role}"')
+            om.use(object_type=UseObjectType.ROLE, name=f'"{conn.role}"')
```
sfc-gh-melnacouzi, line 312: any reason we switched to UseObjectType? 
We don't need to link the object type itself to how it is used.

src/snowflake/cli/api/sql_execution.py
sfc-gh-fcampbell: the facede is nativeapp-only for now, we'd move the whole thing later if we wanted to

sfc-gh-fcampbell: we probably shouldn't be importing from plugins from core code though, does this file need to change?

sfc-gh-mchok: If we're using the sql facade version of the exception that has the proper exception hierarchy that we want then we would need to import it from `snowflake.cli._plugins.nativeapp.sf_facade_exceptions`, unless we want to use a different error

sfc-gh-melnacouzi (approved):

src/snowflake/cli/api/sql_execution.py
sfc-gh-fcampbell: i was under the impression that we wanted to keep developing the SQL facade stuff separate from the rest of the CLI so we don't have to constantly get approval from them. i'd vote for just leaving any non-nativeapp file as-is. If that's not doable (or it would interfere with our metrics), we should move the enum and error class to the core CLI so that core files don't import from plugins (at some point we had a pre-commit check for it, not sure how this passed in the first place)

src/snowflake/cli/_plugins/connection/commands.py
sfc-gh-bdufour: Why are we changing ObjectManager calls though? We shouldn't do this right now.

src/snowflake/cli/api/sql_execution.py
```diff
@@ -90,9 +90,9 @@ def execute_query(self, query: str, **kwargs):
     def execute_queries(self, queries: str, **kwargs):
         return self._execute_queries(queries, **kwargs)
 
-    def use(self, object_type: ObjectType, name: str):
+    def use(self, object_type: UseObjectType, name: str):
```
sfc-gh-bdufour, line 93: let's not do this now, there is very little benefit in doing this.

src/snowflake/cli/api/sql_execution.py
sfc-gh-bdufour: yeah, there are way too many changes here. Can we take a step back and see what's required? Our strategy so far is that the facade is NA-specific, since it's under active development. We're not changing the rest of the CLI, that introduces more risk that we're not ready for ATM.

PR #1818, opened 2024-10-29
SNOW-1772200 Convert templates in artifacts when converting PDF from v1 to v2

sfc-gh-fcampbell:
### Pre-review checklist

### Changes description
When running `snow helpers v1-to-v2`, the structure of the project definition changes, and so does the structure of the template context. If artifacts are using the `templates` processor, the files could have references to v1 fields, like `ctx.native_app.stage`, for example. These references would then be invalid after the v2 conversion, since the template context now only has data about v2 entities. We therefore need to convert these template references to the equivalent v2 references using information from the conversion process. We don't convert `ctx.env.` or `fn.` references since those don't change from v1 to v2.

### DIFF
--- a/src/snowflake/cli/api/project/definition_conversion.py
+++ b/src/snowflake/cli/api/project/definition_conversion.py
@@ -490,6 +490,7 @@ def _convert_templates_in_files(
         # and if we're doing a permanent conversion. If we're doing an in-memory conversion,
         # the CLI global template context is already populated with the v1 definition, so
         # we don't want to convert the v1 template references in artifact files
+        metrics.set_counter_default(CLICounterField.TEMPLATES_PROCESSOR, 0)
         if not in_memory and any(
             processor.name == TEMPLATES_PROCESSOR
             for artifact in pkg_model.artifacts
--- a/src/snowflake/cli/api/project/definition_conversion.py
+++ b/src/snowflake/cli/api/project/definition_conversion.py
@@ -1,6 +1,9 @@
 from __future__ import annotations
 
 import logging
+import shutil
+import tempfile
+from copy import deepcopy
 from pathlib import Path
 from tempfile import TemporaryDirectory, mkstemp
 from typing import Any, Dict, Literal, Optional
@@ -8,6 +11,12 @@ from typing import Any, Dict, Literal, Optional
 from click import ClickException
 from snowflake.cli._plugins.nativeapp.artifacts import (
     BundleMap,
+    symlink_or_copy,
+)
+from snowflake.cli._plugins.nativeapp.bundle_context import BundleContext
+from snowflake.cli._plugins.nativeapp.codegen.compiler import TEMPLATES_PROCESSOR
+from snowflake.cli._plugins.nativeapp.codegen.templates.templates_processor import (
+    TemplatesProcessor,
 )
 from snowflake.cli._plugins.nativeapp.entities.application_package import (
     ApplicationPackageEntityModel,
@@ -46,6 +55,7 @@ from snowflake.cli.api.project.schemas.v1.snowpark.snowpark import Snowpark
 from snowflake.cli.api.project.schemas.v1.streamlit.streamlit import Streamlit
 from snowflake.cli.api.rendering.jinja import get_basic_jinja_env
 from snowflake.cli.api.utils.definition_rendering import render_definition_template
+from snowflake.cli.api.utils.dict_utils import deep_merge_dicts
 
 log = logging.getLogger(__name__)
 
@@ -107,7 +117,7 @@ def convert_project_definition_to_v2(
         if definition_v1.streamlit
         else {}
     )
-    native_app_data = (
+    native_app_data, native_app_template_context = (
         convert_native_app_to_v2_data(
             project_root, definition_v1.native_app, template_context
         )
@@ -137,7 +147,15 @@ def convert_project_definition_to_v2(
 
     # If the user's files have any template tags in them, they
     # also need to be migrated to point to the v2 entities
-    _convert_templates_in_files(project_root, definition_v1, definition_v2, in_memory)
+    replacement_template_context = deepcopy(template_context) or {}
+    deep_merge_dicts(replacement_template_context, native_app_template_context)
+    _convert_templates_in_files(
+        project_root,
+        definition_v1,
+        definition_v2,
+        in_memory,
+        replacement_template_context,
+    )
 
     return definition_v2
 
@@ -246,7 +264,7 @@ def convert_native_app_to_v2_data(
     project_root: Path,
     native_app: NativeApp,
     template_context: Optional[Dict[str, Any]] = None,
-) -> Dict[str, Any]:
+) -> tuple[dict[str, Any], dict[str, Any]]:
     def _make_meta(obj: Application | Package):
         meta = {}
         if obj.role:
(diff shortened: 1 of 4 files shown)


src/snowflake/cli/api/project/definition_conversion.py
```diff
@@ -380,33 +449,118 @@ def _convert_templates_in_files(
...
+            )
+
+            converted_post_deploy_hooks = _convert_package_script_files(
+                project_root, pkg.scripts, pkg_model, in_memory
+            )
+            if pkg_model.meta is None:
+                pkg_model.meta = MetaField()
+            if pkg_model.meta.post_deploy is None:
+                pkg_model.meta.post_deploy = []
+            pkg_model.meta.post_deploy += converted_post_deploy_hooks
```
sfc-gh-fcampbell, line 563: this hasn't changed, it's just an indentation change

src/snowflake/cli/api/project/definition_conversion.py
```diff
@@ -380,33 +450,118 @@ def _convert_templates_in_files(
...
+                    bundle_root=project_root / pkg_model.bundle_root,
+                    deploy_root=deploy_root,
+                    generated_root=(
+                        project_root / deploy_root / pkg_model.generated_root
+                    ),
+                )
+                template_processor = TemplatesProcessor(bundle_ctx)
+                for artifact in na.artifacts:
+                    for processor in artifact.processors:
+                        if processor.name == TEMPLATES_PROCESSOR:
```
sfc-gh-mchok, line 518: If we're inside the outer if block that checks if the templates processor is present, why do we need this loop and the check again?

src/snowflake/cli/api/project/definition_conversion.py
```diff
@@ -380,33 +450,118 @@ def _convert_templates_in_files(
...
+        # We only convert files on-disk if the "templates" processor is used in the artifacts
+        # and if we're doing a permanent conversion. If we're doing an in-memory conversion,
+        # the CLI global template context is already populated with the v1 definition, so
+        # we don't want to convert the v1 template references in artifact files
+        if not in_memory and any(
+            processor.name == TEMPLATES_PROCESSOR
+            for artifact in pkg_model.artifacts
+            for processor in artifact.processors
+        ):
+            metrics.set_counter(CLICounterField.TEMPLATES_PROCESSOR, 1)
```
sfc-gh-mchok, line 498: I don't think I see anything that would set this to 0 in the case where we don't invoke the template processor in the conversion? In that case maybe I would turn the expression in the if statement into a boolean and pass that as 0/1 to set_counter

sfc-gh-mchok:
I like the idea of re-using the template processor to convert templates in files 🙂 

The only thing I would say is that the logic seems brittle in the sense that it's reconstructing some of our implementation details/logic e.g. with bundling, so if that ever changes then this might have to be updated as well. If we're fine with that then I think LGTM

src/snowflake/cli/api/project/definition_conversion.py
sfc-gh-fcampbell: yeah that's correct, the first check bypasses everything if none of the artifacts use the templates processor, this check makes us apply the processor to only the artifacts that have opted-in to it (since other artifact entries may have files with tags that they don't want templated)

src/snowflake/cli/api/project/definition_conversion.py
sfc-gh-fcampbell: yeah i could do that. i think i just copied this from the prod code

sfc-gh-mchok: In the other commands we have the set_counter_defaults in `compile_artifacts` which is called when bundling, but that doesn't happen here I don't think since we're manually building the bundle map etc.

src/snowflake/cli/api/project/definition_conversion.py
sfc-gh-mchok: small nit then is that maybe we can build the list of artifacts that opted-in to the processor, then in the check we can make sure that's not empty, and loop over it for the conversion

(e.g. something like `[artifact for artifact in pkg_model.artifacts for processor in artifact.processors if processor.name == TEMPLATES_PROCESSOR]`)

to reduce the nesting a bit and make it easier to reason about, wdyt?

sfc-gh-fcampbell: yeah i can do that

sfc-gh-fcampbell:
Yeah that's a fair point. I originally had a more bare-bones approach of manually templating each file, but I decided to re-use the TemplatesProcessor since it encapsulates most of what I was doing anyways, making the implementation simpler. The rest of the glue code is indeed lifted from the main CLI implementation since the processor makes some assumptions about the state of files on disk (since it's normally called after bundling artifacts).

sfc-gh-mchok:
That makes sense, so it's just a side effect of how we've coupled our business logic and processing

I'm good with the approach then, but it might be helpful to get a second pair of eyes 🙂

sfc-gh-mchok:
> Yeah that's a fair point. I originally had a more bare-bones approach of manually templating each file, but I decided to re-use the TemplatesProcessor since it encapsulates most of what I was doing anyways, making the implementation simpler. The rest of the glue code is indeed lifted from the main CLI implementation since the processor makes some assumptions about the state of files on disk (since it's normally called after bundling artifacts).

That makes sense, so it's just a side effect of how we've coupled our business logic and processing

I'm good with the approach then, but it might be helpful to get a second pair of eyes 🙂

sfc-gh-mchok (dismissed):

src/snowflake/cli/api/project/definition_conversion.py
```diff
@@ -380,33 +450,119 @@ def _convert_templates_in_files(
...
+                        if src.is_dir():
+                            continue
+                        # We call the implementation directly instead of calling process()
+                        # since we need access to the BundleMap to copy files anyways
+                        template_processor.expand_templates_in_file(
+                            src, dest, replacement_template_context
+                        )
+                        # Copy the expanded file back to its original source location if it was modified
+                        if not dest.is_symlink():
+                            shutil.copyfile(dest, src)
```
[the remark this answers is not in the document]
sfc-gh-fcampbell: we always want to copy the destination back to the source, but if it's a symlink we can't because it's the same file, so we only need to do it if it's not a symlink. If we stop using symlinks, the `if` check will pass and the copy will still happen

src/snowflake/cli/api/project/definition_conversion.py
```diff
@@ -380,33 +450,119 @@ def _convert_templates_in_files(
...
+                    bundle_map.add(artifact)
+
+                    for src, dest in bundle_map.all_mappings(
+                        absolute=True, expand_directories=False
+                    ):
+                        # Copy the mapping from the source root to the deploy root,
+                        # since the processor expects the artifacts to have
+                        # already been bundled (we can't call build_bundle() since it
+                        # checks that the deploy_root is a child of the project_root,
+                        # which isn't the case here)
```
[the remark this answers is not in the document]
sfc-gh-fcampbell: I didn't want to modify any app code, but if that's preferable, I can do it

src/snowflake/cli/api/project/definition_conversion.py
```diff
@@ -380,33 +450,119 @@ def _convert_templates_in_files(
...
+                        symlink_or_copy(src, dest, deploy_root=deploy_root)
+
+                    for src, dest in bundle_map.all_mappings(
+                        absolute=True, expand_directories=True
+                    ):
+                        if src.is_dir():
+                            continue
+                        # We call the implementation directly instead of calling process()
+                        # since we need access to the BundleMap to copy files anyways
```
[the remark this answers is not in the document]
sfc-gh-fcampbell: it's not really because of the directory, it's because the implementation of `TemplatesProcessor.process()` is just
```python
        get_cli_context().metrics.set_counter(CLICounterField.TEMPLATES_PROCESSOR, 1)

        bundle_map = BundleMap(
            project_root=self._bundle_ctx.project_root,
            deploy_root=self._bundle_ctx.deploy_root,
        )
        bundle_map.add(artifact_to_process)

        for src, dest in bundle_map.all_mappings(
            absolute=True,
            expand_directories=True,
        ):
            self.expand_templates_in_file(src, dest)
```
and since we need access to the `BundleMap` to copy the files back to the source directory, it's simpler to just copy the implementation and add the `copytree` call in the same `for` loop. If we call `.process()`, we'd still have to instantiate the same `BundleMap` and iterate over it again to copy the edited files back

src/snowflake/cli/api/project/definition_conversion.py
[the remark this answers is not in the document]
sfc-gh-fcampbell: updated to re-use parts of `build_bundle`

sfc-gh-turbaszek (approved):

PR #1823, opened 2024-10-31
Enable Event Sharing V2 in SnowCLI

sfc-gh-melnacouzi:
### Pre-review checklist
   * [x] I've confirmed that instructions included in README.md are still correct after my changes in the codebase.
   * [x] I've added or updated automated unit tests to verify correctness of my new code.
   * [ ] I've added or updated integration tests to verify correctness of my new code.
   * [x] I've confirmed that my changes are working by executing CLI's commands manually on MacOS.
   * [x] I've confirmed that my changes are working by executing CLI's commands manually on Windows.
   * [x] I've confirmed that my changes are up-to-date with the target branch.
   * [x] I've described my changes in the release notes.
   * [x] I've described my changes in the section below.

### Changes description
Enable Event Sharing V2 in SnowCLI.
Example:
```
  app:
    type: application
    from:
      target: pkg
    telemetry:
      authorize_event_sharing: true
      shared_events:
        - DEBUG_LOGS
        - ERRORS_AND_WARNINGS
 ```
 
Above example authorize telemetry to be shared to provider. In addition to required telemetry by the provider, consumer can specify a list of the events to share with the provider. If events are specified, all mandatory events from your manifest will need to be included as well.
There are 2 flags introduced to control the BCR in the backend. One to enable telemetry on consumer for same-account applications, and another to enforce that the consumer authorize event sharing when the provider has mandatory events.

- Added unit tests.
- Tested manually - optional events not recognized - reached out to corresponding team.
- Integ tests will be added as a follow up PR.

PR #1824, opened 2024-10-31
Add utils for event sharing PR

sfc-gh-melnacouzi:
### Pre-review checklist

### Changes description
Add utils for event sharing PR.
This is to simplify the merging of https://github.com/snowflakedb/snowflake-cli/pull/1823/files, by splitting the files by ownership.

### DIFF
--- a/src/snowflake/cli/_plugins/connection/util.py
+++ b/src/snowflake/cli/_plugins/connection/util.py
@@ -71,7 +71,7 @@ def get_ui_parameter(
     return ui_parameters.get(parameter, default)
 
 
-@lru_cache(maxsize=None)
+@lru_cache()
 def get_ui_parameters(conn: SnowflakeConnection) -> Dict[UIParameter, Any]:
     """
     Returns the UI parameters from the SYSTEM$BOOTSTRAP_DATA_REQUEST function
--- a/src/snowflake/cli/_plugins/connection/util.py
+++ b/src/snowflake/cli/_plugins/connection/util.py
@@ -19,6 +19,7 @@ import logging
 import os
 from enum import Enum
 from functools import lru_cache
+from textwrap import dedent
 from typing import Any, Dict, Optional
 
 from click.exceptions import ClickException
@@ -51,9 +52,9 @@ class MissingConnectionRegionError(ClickException):
 
 
 class UIParameter(Enum):
-    ENABLE_REGIONLESS_REDIRECT = "UI_SNOWSIGHT_ENABLE_REGIONLESS_REDIRECT"
-    EVENT_SHARING_V2 = "ENABLE_EVENT_SHARING_V2_IN_THE_SAME_ACCOUNT"
-    ENFORCE_MANDATORY_FILTERS = (
+    NA_ENABLE_REGIONLESS_REDIRECT = "UI_SNOWSIGHT_ENABLE_REGIONLESS_REDIRECT"
+    NA_EVENT_SHARING_V2 = "ENABLE_EVENT_SHARING_V2_IN_THE_SAME_ACCOUNT"
+    NA_ENFORCE_MANDATORY_FILTERS = (
         "ENFORCE_MANDATORY_FILTERS_FOR_SAME_ACCOUNT_INSTALLATION"
     )
 
@@ -66,9 +67,6 @@ def get_ui_parameter(
     If the parameter is not found, the default value is returned.
     """
 
-    if not isinstance(parameter, UIParameter):
-        raise ValueError("Parameter must be a UIParameters enum")
-
     ui_parameters = get_ui_parameters(conn)
     return ui_parameters.get(parameter, default)
 
@@ -79,16 +77,16 @@ def get_ui_parameters(conn: SnowflakeConnection) -> Dict[UIParameter, Any]:
     Returns the UI parameters from the SYSTEM$BOOTSTRAP_DATA_REQUEST function
     """
 
-    parameters_to_fetch = sorted(
-        [param.value for param in UIParameter.__members__.values()]
-    )
+    parameters_to_fetch = sorted([param.value for param in UIParameter])
 
-    query = f"""
-    select value['value']::string as PARAM_VALUE, value['name']::string as PARAM_NAME from table(flatten(
-        input => parse_json(SYSTEM$BOOTSTRAP_DATA_REQUEST()),
-        path => 'clientParamsInfo'
-    )) where value['name'] in ('{"', '".join(parameters_to_fetch)}');
-    """
+    query = dedent(
+        f"""
+        select value['value']::string as PARAM_VALUE, value['name']::string as PARAM_NAME from table(flatten(
+            input => parse_json(SYSTEM$BOOTSTRAP_DATA_REQUEST()),
+            path => 'clientParamsInfo'
+        )) where value['name'] in ('{"', '".join(parameters_to_fetch)}');
+        """
+    )
 
     *_, cursor = conn.execute_string(query, cursor_class=DictCursor)
 
@@ -107,7 +105,7 @@ def is_regionless_redirect(conn: SnowflakeConnection) -> bool:
     try:
         return (
             get_ui_parameter(
-                conn, UIParameter.ENABLE_REGIONLESS_REDIRECT, "true"
+                conn, UIParameter.NA_ENABLE_REGIONLESS_REDIRECT, "true"
             ).lower()
             == "true"
         )
--- a/src/snowflake/cli/_plugins/connection/util.py
+++ b/src/snowflake/cli/_plugins/connection/util.py
@@ -17,7 +17,9 @@ from __future__ import annotations
 import json
 import logging
 import os
-from typing import Optional
+from enum import Enum
+from functools import lru_cache
+from typing import Any, Dict, Optional
 
 from click.exceptions import ClickException
 from snowflake.connector import SnowflakeConnection
@@ -25,12 +27,6 @@ from snowflake.connector.cursor import DictCursor
 
 log = logging.getLogger(__name__)
 
-REGIONLESS_QUERY = """
-    select value['value'] as REGIONLESS from table(flatten(
-        input => parse_json(SYSTEM$BOOTSTRAP_DATA_REQUEST()),
-        path => 'clientParamsInfo'
-    )) where value['name'] = 'UI_SNOWSIGHT_ENABLE_REGIONLESS_REDIRECT';
-"""
 
 ALLOWLIST_QUERY = "SELECT SYSTEM$ALLOWLIST()"
 SNOWFLAKE_DEPLOYMENT = "SNOWFLAKE_DEPLOYMENT"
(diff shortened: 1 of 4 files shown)


src/snowflake/cli/_plugins/connection/util.py
```diff
@@ -54,6 +50,53 @@ def __init__(self, host: str | None):
...
+def get_ui_parameter(
+    conn: SnowflakeConnection, parameter: UIParameter, default: Any
+) -> str:
+    """
+    Returns the value of a single UI parameter.
+    If the parameter is not found, the default value is returned.
+    """
+
+    if not isinstance(parameter, UIParameter):
+        raise ValueError("Parameter must be a UIParameters enum")
```
sfc-gh-fcampbell, line 70: mypy should take care of this, no?

src/snowflake/cli/_plugins/connection/util.py
```diff
@@ -54,6 +50,53 @@ def __init__(self, host: str | None):
...
+
+def get_ui_parameter(
+    conn: SnowflakeConnection, parameter: UIParameter, default: Any
+) -> str:
+    """
+    Returns the value of a single UI parameter.
+    If the parameter is not found, the default value is returned.
+    """
+
+    if not isinstance(parameter, UIParameter):
```
sfc-gh-mchok, line 69: Is the reason for having this check explicitly along with the type hint just to be extra safe?

sfc-gh-mchok:
I think it could be valuable to add an integ test that actually calls the real function to see if the parameters we get are what we expect, just to make sure we don't get spooked by a BCR or something, would like to know how the others feel about this

src/snowflake/cli/_plugins/connection/util.py
```diff
@@ -54,6 +50,53 @@ def __init__(self, host: str | None):
...
+def get_ui_parameters(conn: SnowflakeConnection) -> Dict[UIParameter, Any]:
+    """
+    Returns the UI parameters from the SYSTEM$BOOTSTRAP_DATA_REQUEST function
+    """
+
+    parameters_to_fetch = sorted(
+        [param.value for param in UIParameter.__members__.values()]
+    )
+
+    query = f"""
```
sfc-gh-fcampbell, line 86: can we throw a dedent on this?

src/snowflake/cli/_plugins/connection/util.py
```diff
@@ -62,8 +105,12 @@ def is_regionless_redirect(conn: SnowflakeConnection) -> bool:
     assume it's regionless, as this is true for most production deployments.
     """
     try:
-        *_, cursor = conn.execute_string(REGIONLESS_QUERY, cursor_class=DictCursor)
-        return cursor.fetchone()["REGIONLESS"].lower() == "true"
+        return (
+            get_ui_parameter(
+                conn, UIParameter.ENABLE_REGIONLESS_REDIRECT, "true"
```
sfc-gh-fcampbell, line 110: the previous version didn't assume that `true` was the default, is it safe to change?

tests/test_utils.py
```diff
@@ -286,3 +290,119 @@ def test_get_context_local_non_regionless_gets_local_region(
...
+
+
+def test_get_ui_parameters_no_params():
+    connection = MagicMock()
+    cursor = MagicMock()
+    connection.execute_string.return_value = (None, cursor)
+    cursor.fetchall.return_value = []
+    assert get_ui_parameters(connection) == {}
+
+    assert connection.execute_string.has_calls([])
```
sfc-gh-fcampbell, line 302: i don't think this assert can ever fail, it checks that an empty list is part of the set of calls that `execute_string`, which is always true

tests/testing_utils/files_and_dirs.py
```diff
@@ -31,6 +31,7 @@ def create_temp_file(suffix: str, dir_name: str, contents: List[str]) -> str:
 
 def create_named_file(file_name: str, dir_name: str, contents: List[str]):
     file_path = os.path.join(dir_name, file_name)
+    os.makedirs(os.path.dirname(file_path), exist_ok=True)
```
sfc-gh-fcampbell, line 34: why is this in this PR? it doesn't seem necessary for tests you're touching

src/snowflake/cli/_plugins/connection/util.py
```diff
@@ -54,6 +50,53 @@ def __init__(self, host: str | None):
...
+    select value['value']::string as PARAM_VALUE, value['name']::string as PARAM_NAME from table(flatten(
+        input => parse_json(SYSTEM$BOOTSTRAP_DATA_REQUEST()),
+        path => 'clientParamsInfo'
+    )) where value['name'] in ('{"', '".join(parameters_to_fetch)}');
+    """
+
+    *_, cursor = conn.execute_string(query, cursor_class=DictCursor)
+
+    return {
+        UIParameter(row["PARAM_NAME"]): row["PARAM_VALUE"] for row in cursor.fetchall()
```
sfc-gh-fcampbell, line 96: could casing break this? maybe we throw a `.upper()` just _in case_? (😂)

src/snowflake/cli/_plugins/connection/util.py
```diff
@@ -54,6 +50,53 @@ def __init__(self, host: str | None):
...
+    )
+
+    query = f"""
+    select value['value']::string as PARAM_VALUE, value['name']::string as PARAM_NAME from table(flatten(
+        input => parse_json(SYSTEM$BOOTSTRAP_DATA_REQUEST()),
+        path => 'clientParamsInfo'
+    )) where value['name'] in ('{"', '".join(parameters_to_fetch)}');
+    """
+
+    *_, cursor = conn.execute_string(query, cursor_class=DictCursor)
```
sfc-gh-fcampbell, line 93: why are we using the raw connection object? can we use SqlExecutor instead?

src/snowflake/cli/_plugins/connection/util.py
```diff
@@ -54,6 +50,53 @@ def __init__(self, host: str | None):
         )
 
 
+class UIParameter(Enum):
+    ENABLE_REGIONLESS_REDIRECT = "UI_SNOWSIGHT_ENABLE_REGIONLESS_REDIRECT"
+    EVENT_SHARING_V2 = "ENABLE_EVENT_SHARING_V2_IN_THE_SAME_ACCOUNT"
+    ENFORCE_MANDATORY_FILTERS = (
```
sfc-gh-fcampbell, line 56: can we give this one a slightly more descriptive name? it's not obvious that this has to do with nativeapp installation

src/snowflake/cli/_plugins/connection/util.py
sfc-gh-melnacouzi: yes

src/snowflake/cli/_plugins/connection/util.py
sfc-gh-melnacouzi: I'll remove it

tests/test_utils.py
sfc-gh-melnacouzi: sorry this test was left here by mistake - I removed it - there is a proper test under it.

tests/testing_utils/files_and_dirs.py
sfc-gh-melnacouzi: I split the main PR into 2 parts. This one involves all the files where snowCLI approval is required. This will be needed for the event sharing PR: please see https://github.com/snowflakedb/snowflake-cli/pull/1823/files for more info

src/snowflake/cli/_plugins/connection/util.py
sfc-gh-melnacouzi: it shouldn't. I tested it and it's working as expected.

src/snowflake/cli/_plugins/connection/util.py
sfc-gh-fcampbell: yeah good point. if this parameter is ever removed from the response, is `true` the default we want to assume?

src/snowflake/cli/_plugins/connection/util.py
```diff
@@ -54,6 +50,53 @@ def __init__(self, host: str | None):
...
+
+
+@lru_cache(maxsize=None)
+def get_ui_parameters(conn: SnowflakeConnection) -> Dict[UIParameter, Any]:
+    """
+    Returns the UI parameters from the SYSTEM$BOOTSTRAP_DATA_REQUEST function
+    """
+
+    parameters_to_fetch = sorted(
+        [param.value for param in UIParameter.__members__.values()]
```
sfc-gh-mchok, line 83: nit: I think you can simplify this to `[param.value for param in UIParameter]`

src/snowflake/cli/_plugins/connection/util.py
sfc-gh-melnacouzi: I would like to minimize the risk for this change, as it was using it earlier.

src/snowflake/cli/_plugins/connection/util.py
sfc-gh-fcampbell: we could make that argument about any function. we need to decide if we trust our tools or not (in this case mypy). i think this is safe enough without, especially since this function is in the code path for many integ tests

src/snowflake/cli/_plugins/connection/util.py
sfc-gh-melnacouzi: added NA prefix

src/snowflake/cli/_plugins/connection/util.py
sfc-gh-melnacouzi: I am not aware of the logic behind it, but I am relying on the default that Cam used. My understanding, this is a configuration rather than a parameter for migration.

src/snowflake/cli/_plugins/connection/util.py
sfc-gh-melnacouzi: thanks

sfc-gh-fcampbell (approved):

sfc-gh-mchok (dismissed):

src/snowflake/cli/_plugins/connection/util.py
```diff
@@ -54,6 +51,50 @@ def __init__(self, host: str | None):
...
+    """
+    Returns the value of a single UI parameter.
+    If the parameter is not found, the default value is returned.
+    """
+
+    ui_parameters = get_ui_parameters(conn)
+    return ui_parameters.get(parameter, default)
+
+
+@lru_cache(maxsize=None)
```
sfc-gh-bdufour, line 74: LRU cache with no max size? Makes me a bit twitchy. In the context of a language server, for example, that could be a real issue. In any case this will be cached by connection, so it should be safe to use a rather small number here for the the LRU cache size.

tests/testing_utils/files_and_dirs.py
sfc-gh-bdufour, line 34: doesn't need to be here. We're co-owners, so no CLI approval needed. I'd remove it from this PR.

src/snowflake/cli/_plugins/connection/util.py
sfc-gh-melnacouzi: I removed maxsize, so I rely on the default, which is currently 128.

sfc-gh-bdufour (approved):

sfc-gh-turbaszek (approved):

PR #1835, opened 2024-11-06
implement depth & total span limits with tests

sfc-gh-mchok:
### Pre-review checklist
   * [X] I've confirmed that my changes are working by executing CLI's commands manually on Windows.

### Changes description
- implement depth and total limits for spans so that the column housing the telemetry doesn't blow up
- added `count_spans_in_subtree`, `span_depth`, `trimmed` fields to reporting dict to make investigation easier if we do reach the limits
- chose to do metadata augmentation on the span tree as we're making it to make the postprocessing easier (e.g. no need to build an adjacency matrix and bfs at the end, we can use simple python builtins, which makes the implementation cleaner/more declarative, and easier to read and maintain imo

### DIFF
--- a/src/snowflake/cli/api/metrics.py
+++ b/src/snowflake/cli/api/metrics.py
@@ -263,13 +263,7 @@ class CLIMetrics:
 
     @property
     def num_spans_past_total_limit(self) -> int:
-        return max(
-            0,
-            # we don't count the spans past the depth limit since they would not be reported
-            len(self._completed_spans)
-            - self.num_spans_past_depth_limit
-            - self.SPAN_TOTAL_LIMIT,
-        )
+        return max(0, len(self._completed_spans) - self.SPAN_TOTAL_LIMIT)
 
     @property
     def completed_spans(self) -> List[Dict]:
--- a/src/snowflake/cli/api/metrics.py
+++ b/src/snowflake/cli/api/metrics.py
@@ -17,7 +17,8 @@ import time
 import uuid
 from contextlib import contextmanager
 from dataclasses import dataclass, field, replace
-from typing import ClassVar, Dict, Iterator, List, Optional
+from heapq import nsmallest
+from typing import ClassVar, Dict, Iterator, List, Optional, Set
 
 
 class CLIMetricsInvalidUsageError(RuntimeError):
@@ -84,24 +85,52 @@ class CLIMetricsSpan:
     START_TIME_KEY: ClassVar[str] = "start_time"
     EXECUTION_TIME_KEY: ClassVar[str] = "execution_time"
     ERROR_KEY: ClassVar[str] = "error"
+    COUNT_SPANS_IN_SUBTREE_KEY: ClassVar[str] = "count_spans_in_subtree"
+    SPAN_DEPTH_KEY: ClassVar[str] = "span_depth"
+    # denotes whether direct children were pruned from telemetry payload
+    TRIMMED_KEY: ClassVar[str] = "trimmed"
 
+    # constructor vars
     name: str
     start_time: float  # relative to when the command first started executing
     parent: Optional[CLIMetricsSpan] = None
 
-    # ensure we get unique ids for each step for the parent-child link in case of steps with the same name
-    step_id: str = field(init=False, default_factory=lambda: uuid.uuid4().hex)
+    # vars for reporting
+    span_id: str = field(init=False, default_factory=lambda: uuid.uuid4().hex)
     execution_time: Optional[float] = field(init=False, default=None)
     error: Optional[BaseException] = field(init=False, default=None)
+    span_depth: int = field(init=False, default=1)
+    count_spans_in_subtree: int = field(init=False, default=1)
 
+    # vars for postprocessing
+    children: Set[CLIMetricsSpan] = field(init=False, default_factory=set)
+
+    # private state
     # start time of the step from the monotonic clock in order to calculate execution time
     _monotonic_start: float = field(
         init=False, default_factory=lambda: time.monotonic()
     )
 
+    def __hash__(self) -> int:
+        return hash(self.span_id)
+
     def __post_init__(self):
         if not self.name:
-            raise CLIMetricsInvalidUsageError("step name must not be empty")
+            raise CLIMetricsInvalidUsageError("span name must not be empty")
+
+        if self.parent:
+            self.parent.add_child(self)
+            self.span_depth = self.parent.span_depth + 1
+
+    def increment_subtree_node_count(self) -> None:
+        self.count_spans_in_subtree += 1
+
+        if self.parent:
+            self.parent.increment_subtree_node_count()
+
+    def add_child(self, child: CLIMetricsSpan) -> None:
+        self.children.add(child)
+        self.increment_subtree_node_count()
 
     def finish(self, error: Optional[BaseException] = None) -> None:
         """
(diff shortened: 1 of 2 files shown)


tests/api/metrics/test_metrics_spans.py
```diff
@@ -265,4 +298,99 @@ def test_metrics_spans_empty_name_raises_error():
...
+        110542.781,
+        110543.39,
+        110544.359,
+        110545.296,
+        110545.812,
+        110546.265,
+        110546.5,
+        110547.078,
+        110547.984,
+        110548.078,
```
sfc-gh-mchok, line 331: I recognize this is a bit of a red flag but I'm unsure how to reconcile this, since otherwise the start_times end up being all 0 when running the tests on windows

tests/api/metrics/test_metrics_spans.py
```diff
@@ -22,25 +23,19 @@
...
 
-# helper for testing span depth limit edge case
-def create_nested_spans_recursively(metrics: CLIMetrics, num_spans: int = 1) -> None:
-    @metrics.start_span("nested_span")
-    def create_span():
-        nonlocal num_spans
-        num_spans -= 1
+# helper for testing span limits
+def create_spans(metrics: CLIMetrics, width: int, depth: int):
+    counter = count()
```
sfc-gh-mchok, line 28: This is here in case the test fails then we will have something to work with in the logs

src/snowflake/cli/api/metrics.py
```diff
@@ -213,15 +248,57 @@ def counters(self) -> Dict[str, int]:
...
+
+        # sort by start time to make reading the payload easier
+        sorted_spans_to_report = sorted(
+            spans_to_report, key=lambda span: span.start_time
+        )
+
+        return [
+            {
+                **span.to_dict(),
+                CLIMetricsSpan.TRIMMED_KEY: not span.children <= spans_to_report,
```
sfc-gh-mchok, line 301: I chose to put this here and not in the span class since it feels like this additional information we're tagging on in post-processing rather than having to do with the state of the span class itself, but I kept the key in the span class to centralize them all. Not sure if there is a better approach, let me know if you have any thoughts

src/snowflake/cli/api/metrics.py
```diff
@@ -84,24 +85,52 @@ class for holding metrics span data and encapsulating related operations
...
+    children: Set[CLIMetricsSpan] = field(init=False, default_factory=set)
+
+    # private state
     # start time of the step from the monotonic clock in order to calculate execution time
     _monotonic_start: float = field(
         init=False, default_factory=lambda: time.monotonic()
     )
 
+    def __hash__(self) -> int:
+        return hash(self.span_id)
```
sfc-gh-mchok, line 115: for post processing, since we need to be able to put the spans in a set

src/snowflake/cli/api/metrics.py
```diff
@@ -213,15 +248,57 @@ def counters(self) -> Dict[str, int]:
...
+        # since we care more about the big picture than granularity
+        spans_to_report = set(
+            nsmallest(
+                n=self.SPAN_TOTAL_LIMIT,
+                iterable=(
+                    span
+                    for span in self._completed_spans
+                    if span.span_depth <= self.SPAN_DEPTH_LIMIT
+                ),
+                key=lambda span: (span.span_depth, span.start_time),
```
sfc-gh-fcampbell, line 289: smart, i like how this avoids a tree traversal

src/snowflake/cli/api/metrics.py
sfc-gh-fcampbell, line 301: we can use the inverse operator to avoid the `not`. (there's `.issuperset()` but unfortunately it would return `Ture` if they're equal, and there's no named method for strict superset checking)
```suggestion
                CLIMetricsSpan.TRIMMED_KEY: span.children > spans_to_report,
```

tests/api/metrics/test_metrics_spans.py
sfc-gh-fcampbell: how about making `side_effect` a function that returns an increasing number? (maybe `side_effect=lambda: *_, **__: next(GLOBAL_COUNT)` would be enough?

src/snowflake/cli/api/metrics.py
```diff
@@ -213,15 +248,57 @@ def counters(self) -> Dict[str, int]:
...
+                for span in self._completed_spans
+                if span.span_depth > self.SPAN_DEPTH_LIMIT
+            ]
+        )
+
+    @property
+    def num_spans_past_total_limit(self) -> int:
+        return max(
+            0,
+            # we don't count the spans past the depth limit since they would not be reported
```
sfc-gh-fcampbell, line 265: why is this the case? i thought we wanted to report on the full number of spans opened by the app?

src/snowflake/cli/api/metrics.py
```diff
@@ -84,24 +85,52 @@ class for holding metrics span data and encapsulating related operations
     START_TIME_KEY: ClassVar[str] = "start_time"
     EXECUTION_TIME_KEY: ClassVar[str] = "execution_time"
     ERROR_KEY: ClassVar[str] = "error"
+    COUNT_SPANS_IN_SUBTREE_KEY: ClassVar[str] = "count_spans_in_subtree"
+    SPAN_DEPTH_KEY: ClassVar[str] = "span_depth"
+    # denotes whether direct children were pruned from telemetry payload
+    TRIMMED_KEY: ClassVar[str] = "trimmed"
```
sfc-gh-bdufour, line 91: nit: pruned or trimmed? You're using both for the same thing.

src/snowflake/cli/api/metrics.py
sfc-gh-bdufour: +1, these limits are just guardrails.

src/snowflake/cli/api/metrics.py
```diff
@@ -84,24 +85,52 @@ class for holding metrics span data and encapsulating related operations
     START_TIME_KEY: ClassVar[str] = "start_time"
     EXECUTION_TIME_KEY: ClassVar[str] = "execution_time"
     ERROR_KEY: ClassVar[str] = "error"
+    COUNT_SPANS_IN_SUBTREE_KEY: ClassVar[str] = "count_spans_in_subtree"
+    SPAN_DEPTH_KEY: ClassVar[str] = "span_depth"
```
sfc-gh-bdufour, line 89: can you please document these? Span depth seems to represent the depth of the current span (is it 0-based though? 1-based?). COUNT_SPANS sees to be the total size of the subtree rooted at this span. I would have expected it to be named "SPAN_COUNT" and not "COUNT_SPANS", if so.

tests/api/metrics/test_metrics_spans.py
sfc-gh-bdufour: please, let's not do this, it's not maintainable and generally not best practice. You can definitely just make the clock "tick" as Frank suggested, that should be sufficient.

sfc-gh-mchok: Good call, added a local fixture for it here: https://github.com/snowflakedb/snowflake-cli/pull/1835/files#diff-839165dddafe1dd908ef843269b12e1cd4a7158ea95e5e7a2ee210ea7ec56de0R26-R31

src/snowflake/cli/api/metrics.py
sfc-gh-mchok: Thank you :'D

src/snowflake/cli/api/metrics.py
sfc-gh-mchok: In my head these two are measuring which of the two guardrails we hit: total for horizontal and depth for vertical, and it would help investigation if they were isolated

We can infer the full number by summing the subtree counts (or introduce an explicit variable), and the semantics of having the spans exceeding depth limit also contribute to the total limit seems weird to me:
e.g. we have a path of 6 spans nested within each other, we then would say that 1 exceeded depth limit and 1 exceeded total limit, when we only report 5 of them and our total limit is 100 (which was not hit)

Happy to discuss further since I might be missing something though

src/snowflake/cli/api/metrics.py
sfc-gh-mchok: I thought they were synonyms, checked Google and seems trimmed works better so I kept that one 😓

src/snowflake/cli/api/metrics.py
sfc-gh-mchok: Updated, tried to make the description as agnostic to the names (e.g. not using the words depth and count) as possible: https://github.com/snowflakedb/snowflake-cli/pull/1835/files#diff-db39ee1dd59a874df561d1420daafd90b948a34c614128415110ee101aa55cc7R88-R92

src/snowflake/cli/api/metrics.py
sfc-gh-fcampbell: > we then would say that 1 exceeded depth limit and 1 exceeded total limit
woudn't it be 0 exceeded total limit, since 6 < 100?

sfc-gh-mchok: Discussed offline, changes made here: https://github.com/snowflakedb/snowflake-cli/pull/1835/commits/a2982b84a7b3d99baf3a70c8d334b84516859270

sfc-gh-bdufour (approved):

PR #1843, opened 2024-11-08
remove limitation on nested phases/steps

sfc-gh-mchok:
### Pre-review checklist

### Changes description
- Allow nesting phases/steps and also their usages in indents, since we found that we've had bugs multiple times from this and we don't want the console message mechanism to be breaking commands in prod


sfc-gh-fcampbell:
I'm OK with this, might wanna run it y @sfc-gh-bdufour since he added the limitation initially

tests/api/console/test_cli_console_output.py
```diff
@@ -74,11 +73,19 @@ def test_error_messages(cli_console, capsys):
     assert_output_matches("42\n  73\n  ops\nOPS\n", capsys)
 
 
-def test_phase_nesting_not_allowed(cli_console):
+def test_phase_nesting(cli_console, capsys):
```
sfc-gh-bdufour, line 76: step nesting isn't allowed still, right? We should capture that in a test since it's part of the contract.

tests/api/console/test_cli_console_output.py
```diff
@@ -91,18 +98,35 @@ def test_phase_is_cleaned_up_on_exception(cli_console):
...
+        Outside of Indent
+          Enter
+        """
+    )
 
-def test_step_cannot_be_indented(cli_console):
+    assert_output_matches(expected_output, capsys)
+
+
+def test_step_with_indent(cli_console, capsys):
```
sfc-gh-bdufour, line 117: do we actually need this? It's a bit surprising, that wasn't the intention behind `indented`. I would expect phases can be nested, as you've done here, but steps cannot. Indented should only be used as a leaf construct, not in between phases and steps, unless I'm missing something.

src/snowflake/cli/api/console/console.py
```diff
@@ -70,28 +66,21 @@ def _format_message(self, message: str, output: Output) -> Text:
     @contextmanager
     def phase(self, enter_message: str, exit_message: Optional[str] = None):
```
sfc-gh-bdufour, line 67: Looks like we never checked that, but shouldn't we disallow phase when within a step? If so, could you please fix the implementation?

sfc-gh-bdufour: right, steps don't have a duration. Ignore me :)

tests/api/console/test_cli_console_output.py
sfc-gh-mchok: Similar question to https://github.com/snowflakedb/snowflake-cli/pull/1843#discussion_r1838759925, I'm not entirely sure what you mean by step nesting, since a step is just a message and cannot contain things within it

tests/api/console/test_cli_console_output.py
sfc-gh-mchok: The intent was to allow everything, or else we might end up in another situation where an unexpected flow ends up raising an error, when it probably doesn't make sense for our commands to not work if we mess up the formatting a little, similar to how we decided that our metrics shouldn't blow up our commands if something goes wrong there either

tests/api/console/test_cli_console_output.py
sfc-gh-bdufour: as you pointed out, steps are just messages, not blocks like phases, so this comment doesn't make sense. My bad.

sfc-gh-fcampbell (approved):

tests/api/console/test_cli_console_output.py
```diff
@@ -91,18 +100,35 @@ def test_phase_is_cleaned_up_on_exception(cli_console):
...
 
 
-def test_phase_cannot_be_indented(cli_console):
+def test_phase_inside_indented(cli_console, capsys):
+    cli_console.step("Outside of Indent")
     with cli_console.indented():
-        with pytest.raises(CliConsoleNestingProhibitedError):
-            with cli_console.phase("Enter"):
-                pass
+        with cli_console.phase("Enter"):
```
sfc-gh-bdufour, line 106: the check is too weak IMHO. Add a step above to capture that phase + step are indented to the same level, and something inside the phase to capture that it indents _on top of_ the existing indented block

sfc-gh-astus (approved):

PR #1844, opened 2024-11-08
instrument NADE commands using metric spans

sfc-gh-mchok:
### Pre-review checklist

### Changes description
- Added metric spans to telemetry payload being sent
- Instrumented native app commands with metric spans API
- Added integ tests to verify the proper spans are being sent in telemetry for example commands (see [tests_integration/nativeapp/test_metrics.py](https://github.com/snowflakedb/snowflake-cli/pull/1844/files#diff-67c0486726129d525081aab2dd54797c3dcbd46345b2a4cb8152c9c1e1674743) for examples of traces being sent)
- Refactored test files to fit with convention, combining metrics counters + spans tests

An example trace of `snow app run` looks like the following (truncated for brevity):

```python
{
        'message': {
            'command_execution_time': 30.307388332905248,
            'counters': {
                'features.global.pdf_templates': 1,
                'features.global.templates_processor': 1,
                'features.app.snowpark_processor': 0,
                'features.app.post_deploy_scripts': 1,
                'features.app.event_sharing': 0,
                'features.app.event_sharing_warning': 0,
                'features.app.event_sharing_error': 0
            },
            'spans': {
                'completed_spans': [{
                    'id': 'a45d527defc44719a0b0d8e8ab9d02b5',
                    'name': 'action.app.deploy',
                    'start_time': 0.6444029589183629,
                    'parent': None,
                    'parent_id': None,
                    'execution_time': 27.891165500041097,
                    'error': None,
                    'span_count_in_subtree': 10,
                    'span_depth': 1,
                    'trimmed': False
                }, {
                    'id': '8dd033f84f4f481b8542fff24a759ccf',
                    'name': 'action.app_pkg.deploy',
                    'start_time': 0.644967291969806,
                    'parent': 'action.app.deploy',
                    'parent_id': 'a45d527defc44719a0b0d8e8ab9d02b5',
                    'execution_time': 11.048065917100757,
                    'error': None,
                    'span_count_in_subtree': 7,
                    'span_depth': 2,
                    'trimmed': False
                }, {
                    'id': '088ae6c6ac0947e9ac8796fe83e60dd9',
                    'name': 'bundle',
                    'start_time': 0.6450087090488523,
                    'parent': 'action.app_pkg.deploy',
                    'parent_id': '8dd033f84f4f481b8542fff24a759ccf',
                    'execution_time': 0.0043289999011904,
                    'error': None,
                    'span_count_in_subtree': 1,
                    'span_depth': 3,
                    'trimmed': False
                }, {
                    'id': 'ae49cdc6fe944579a057fa4850d92c1d',
                    'name': 'artifact_processors',
                    'start_time': 0.6495900419540703,
                    'parent': 'action.app_pkg.deploy',
                    'parent_id': '8dd033f84f4f481b8542fff24a759ccf',
                    'execution_time': 0.00230620801448822,
                    'error': None,
                    'span_count_in_subtree': 2,
                    'span_depth': 3,
                    'trimmed': False
                }, {
                    'id': 'b83c2257cedd41a38a89bc79b60f0214',
                    'name': 'templates_processor',
                    'start_time': 0.6496683340519667,
                    'parent': 'artifact_processors',
                    'parent_id': 'ae49cdc6fe944579a057fa4850d92c1d',
                    'execution_time': 0.0022297080140560865,
                    'error': None,
                    'span_count_in_subtree': 1,
                    'span_depth': 4,
                    'trimmed': False
                }, {
                    'id': '399ba61144824d25861401573f9447c4',
                    'name': 'sync_deploy_root_with_stage',
                    'start_time': 2.6214112089946866,
                    'parent': 'action.app_pkg.deploy',
                    'parent_id': '8dd033f84f4f481b8542fff24a759ccf',
                    'execution_time': 5.49098225007765,
                    'error': None,
                    'span_count_in_subtree': 1,
                    'span_depth': 3,
                    'trimmed': False
                }, {
                    'id': '483b5fe50c1c4e6ea4e81598a0bf926d',
                    'name': 'post_deploy_hooks',
                    'start_time': 8.112657208926976,
                    'parent': 'action.app_pkg.deploy',
                    'parent_id': '8dd033f84f4f481b8542fff24a759ccf',
                    'execution_time': 3.1565516660921276,
                    'error': None,
                    'span_count_in_subtree': 1,
                    'span_depth': 3,
                    'trimmed': False
                }, {
                    'id': 'cd065a3c3148457d82d45f9b62d3ec62',
                    'name': 'validate_setup_script',
                    'start_time': 11.27043924992904,
                    'parent': 'action.app_pkg.deploy',
                    'parent_id': '8dd033f84f4f481b8542fff24a759ccf',
                    'execution_time': 0.422550291987136,
                    'error': None,
                    'span_count_in_subtree': 1,
                    'span_depth': 3,
                    'trimmed': False
                }, {
                    'id': '99829e4b459b48fb92e111c951f7f58c',
                    'name': 'update_app_object',
                    'start_time': 11.693080459022895,
                    'parent': 'action.app.deploy',
                    'parent_id': 'a45d527defc44719a0b0d8e8ab9d02b5',
                    'execution_time': 16.84247687482275,
                    'error': None,
                    'span_count_in_subtree': 2,
                    'span_depth': 2,
                    'trimmed': False
                }, {
                    'id': '7f79381f00f442bc8d499e7950ca9ac0',
                    'name': 'post_deploy_hooks',
                    'start_time': 22.447485291864723,
                    'parent': 'update_app_object',
                    'parent_id': '99829e4b459b48fb92e111c951f7f58c',
                    'execution_time': 6.088065165793523,
                    'error': None,
                    'span_count_in_subtree': 1,
                    'span_depth': 3,
                    'trimmed': False
                }, {
                    'id': '71938e6500ac427db621280c229bdf4e',
                    'name': 'get_snowsight_url_for_app',
                    'start_time': 28.535641624825075,
                    'parent': None,
                    'parent_id': None,
                    'execution_time': 1.8380501249339432,
                    'error': None,
                    'span_count_in_subtree': 1,
                    'span_depth': 1,
                    'trimmed': False
                }],
                'num_spans_past_depth_limit': 0,
                'num_spans_past_total_limit': 0
            }
        }
```


src/snowflake/cli/_plugins/nativeapp/entities/application.py
```diff
@@ -420,6 +421,7 @@ def get_objects_owned_by_application(self) -> List[ApplicationOwnedObject]:
             ).fetchall()
             return [{"name": row[1], "type": row[2]} for row in results]
 
+    @get_cli_context().metrics.start_span("create_or_upgrade_app")
```
sfc-gh-fcampbell, line 424: this will get the CLI context at module load time, which won't reflect forking. I think we need to always use the context manager form to make sure that we get the latest global context

tests_integration/nativeapp/test_feature_metrics.py
```diff
@@ -143,4 +146,4 @@ def test_nativeapp_feature_counter_has_expected_value(
             mock_telemetry, TelemetryEvent.CMD_EXECUTION_RESULT.value
         )
 
-        assert message["counters"] == expected_counters
+        assert message[CLITelemetryField.COUNTERS.value] == expected_counters
```
sfc-gh-fcampbell, line 149: can we also add a test for spans?

src/snowflake/cli/_plugins/nativeapp/entities/application.py
sfc-gh-mchok: What about writing a custom decorator? I think that should grab the global context at runtime, and it avoids nesting the whole function, maybe something like:
```
def start_cli_metrics_span(span_name: str):
    def decorator(func):
        @wraps(func)
        def wrapper(*args, **kwargs):
            with get_cli_context().metrics.start_span(span_name):
                return func(*args, **kwargs)
        return wrapper
    return decorator
```

sfc-gh-fcampbell: yeah that would work

tests_integration/nativeapp/test_command_metric_counters.py
sfc-gh-mchok, line 1: file renamed from `test_feature_metrics` since it was going to be too large with the spans, and also converted the tests to use the factories

tests_integration/nativeapp/test_command_metric_counters.py
```diff
@@ -0,0 +1,248 @@
...
+    )
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize(
+    "command,project_factory,expected_counters",
+    [
+        (
+            "app deploy",
+            lambda: ProjectV11Factory(
```
sfc-gh-mchok, line 75: This is the best I could come up with to have parametrization with factories, if anyone has a better approach I'd be interested

sfc-gh-bdufour:
description + pre-review checklist aren't optional, please fill them in

tests_integration/nativeapp/test_command_metric_counters.py
sfc-gh-bdufour: Generally the test directory mirrors the layout of the source directory so that tests are easy to locate. Breaking down the files like this because of some arbitrary size constraint makes things easier to manage, and makes it easier for us to review the changes because we can no longer see the diff. Please follow the convention instead.

src/snowflake/cli/_plugins/nativeapp/codegen/compiler.py
```diff
@@ -88,7 +88,7 @@ def compile_artifacts(self):
         if not self._should_invoke_processors():
             return
 
-        with cc.phase("Invoking artifact processors"):
+        with cc.phase("Invoking artifact processors", span_name="compile_artifacts"):
```
sfc-gh-bdufour, line 91: there's no compilation happening. How about simply "artifact_processors"? We should probably also have individual processors as sub-spans.

src/snowflake/cli/api/cli_global_context.py
```diff
@@ -213,6 +214,26 @@ def get_cli_context() -> _CliGlobalContextAccess:
     return _CliGlobalContextAccess(get_cli_context_manager())
 
 
+def start_cli_metrics_span(span_name: str):
```
sfc-gh-bdufour, line 217: I don't think "start" makes sense here, since this both starts and stops the span at the appropriate time. Let's workshop the name of this + the `start_span` method it calls.

src/snowflake/cli/_plugins/nativeapp/entities/application.py
```diff
@@ -420,6 +421,7 @@ def get_objects_owned_by_application(self) -> List[ApplicationOwnedObject]:
             ).fetchall()
             return [{"name": row[1], "type": row[2]} for row in results]
 
+    @start_cli_metrics_span("create_or_upgrade_app")
```
sfc-gh-bdufour, line 424: spans shouldn't be implementation-specific, but rather represent a logic step in the overall operation being performed.

src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
```diff
@@ -915,6 +916,7 @@ def validate_setup_script(
             if validation_result["status"] == "FAIL":
                 raise SetupScriptFailedValidation()
 
+    @start_cli_metrics_span("get_validation_result")
```
sfc-gh-bdufour, line 919: ditto, let's not name things according to the implementation. That can change over time.

src/snowflake/cli/api/console/abc.py
```diff
@@ -68,6 +68,7 @@ def phase(
         self,
         enter_message: str,
         exit_message: Optional[str] = None,
+        span_name: Optional[str] = None,
```
sfc-gh-bdufour, line 71: not sure if this is truly needed, since we can use multiple context managers on the same line. Thoughts?

src/snowflake/cli/api/entities/utils.py
```diff
@@ -304,6 +308,7 @@ def validation_item_to_str(item: dict[str, str | int]):
     return s
 
 
+@start_cli_metrics_span("drop_generic_object")
```
sfc-gh-bdufour, line 311: same here

tests_integration/nativeapp/test_command_metric_counters.py
sfc-gh-bdufour: what do we need parametrized tests at all? I don't think I can read this anymore. I think overall we make way, way too frequent use of parametrized tests in our codebase, but that's a different story.

src/snowflake/cli/api/cli_global_context.py
sfc-gh-melnacouzi: +1, I feel @span would suffice

tests_integration/nativeapp/test_command_metric_counters.py
sfc-gh-fcampbell: if the arguments to the factories are the same, you can parametrize just on the class and instantiate it in the test
```python
@pytest.mark.parametrize("factory_class", [FooFactory, BarFactory])
def test_factory(factory_class):
    obj = factory_class(arg=123)
```

tests_integration/nativeapp/test_command_metric_counters.py
sfc-gh-fcampbell: i don't follow this comment, did you mean it makes it harder if we split the test file?

tests_integration/nativeapp/test_command_metric_counters.py
sfc-gh-mchok: I ended up getting rid of all the parametrization and separating the tests out since the factories didn't have the same arguments, I think it's more clear and separated this way, let me know what you think

sfc-gh-mchok:
> description + pre-review checklist aren't optional, please fill them in

My mistake, I usually do it at the same time as when I make the PR but I started it as a draft this time and forgot to do it when I set it as ready for review, should be populated now!

src/snowflake/cli/api/cli_global_context.py
sfc-gh-mchok: Changed both the method name and the decorator name to just `span`, let me know if this works for you

src/snowflake/cli/_plugins/nativeapp/entities/application.py
sfc-gh-mchok: Adjusted the names and locations of all the spans to better encapsulate this, let me know if they work for you!

src/snowflake/cli/api/console/abc.py
sfc-gh-mchok: I forgot about that feature, I adjusted these instances, thanks for the reminder

src/snowflake/cli/api/metrics.py
```diff
@@ -109,10 +109,8 @@ class for holding metrics span data and encapsulating related operations
     children: Set[CLIMetricsSpan] = field(init=False, default_factory=set)
 
     # private state
-    # start time of the step from the monotonic clock in order to calculate execution time
-    _monotonic_start: float = field(
-        init=False, default_factory=lambda: time.monotonic()
-    )
+    # start time of the step from a performance counter in order to calculate execution time
+    _start_time: float = field(init=False, default_factory=time.perf_counter)
```
sfc-gh-mchok, line 113: Ran into failing integ tests where the ordering of the spans was not deterministic, did some more research and found that `time.perf_counter` should just be a better version of `time.monotonic` with appropriate granularity on Windows, so replaced those calls here

tests/api/test_metrics.py
```diff
@@ -23,12 +23,90 @@
 )
 
 
-# we need to mock time.monotonic because on windows it does not
-# capture enough precision for these tests to not be flaky
-@pytest.fixture
-def mock_time_monotonic():
-    with mock.patch("time.monotonic", side_effect=count()) as mocked_time_monotonic:
-        yield mocked_time_monotonic
+def test_metrics_no_counters():
```
sfc-gh-mchok, line 26: All counter tests moved from `test_metrics_counters.py` to consolidate the files and keep to the convention

tests_integration/nativeapp/test_metrics.py
```diff
@@ -0,0 +1,544 @@
...
+            CLICounterField.TEMPLATES_PROCESSOR: 0,
+            CLICounterField.PDF_TEMPLATES: 0,
+            CLICounterField.POST_DEPLOY_SCRIPTS: 1,
+            CLICounterField.PACKAGE_SCRIPTS: 0,
+        }
+
+
+@pytest.mark.integration
+@mock.patch("snowflake.connector.telemetry.TelemetryClient.try_add_log_to_batch")
+def test_feature_counters_v2_post_deploy_not_available_in_bundle(
```
sfc-gh-mchok, line 108: counter tests moved from `test_feature_metrics.py` to consolidate them in one file and keep to the convention, and converted to use factories

src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
```diff
@@ -220,6 +221,7 @@ def action_deploy(
             force=force,
         )
 
+    @span("drop_app_package")
```
sfc-gh-fcampbell, line 224: i thought we said we didn't want spans to be implementation-specific? shouldn't this be just "drop"? same for the others here

src/snowflake/cli/_plugins/nativeapp/entities/application.py
```diff
@@ -769,6 +772,7 @@ def stream_events(
         except KeyboardInterrupt:
             return
 
+    @span("get_snowsight_url_for_app")
```
sfc-gh-fcampbell, line 775: IMO this doesn't need to be recorded

sfc-gh-mchok: If you take a look at the example payload in the PR description it's actually a sizeable chunk, around 3s~ which is bigger than some of the other spans, and when it's not there it seems like the trace is missing something since the total execution time isn't roughly equal to the sum of the spans' execution times. I'm not against removing it if we feel it's unnecessary and we agree we can infer what's happening in the blind spot, just wanted to bring attention to it and see if you have any other thoughts

src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
sfc-gh-fcampbell: i guess it depends what we want to see when we're looking at the metrics. i was under the impression that we'd have generic span names like "bundle", "deploy", "drop", etc so that we could aggregate metrics from different entity types together, but if we want to keep them separate then this name is ok

src/snowflake/cli/api/entities/utils.py
```diff
@@ -76,6 +76,7 @@ def _get_stage_paths_to_sync(
     return stage_paths
 
 
+@span("sync_remote_and_local_files")
```
sfc-gh-bdufour, line 79: this is one of the rare cases where I'd use the actual function name, since it's exactly the logical concept we're timing.

src/snowflake/cli/api/metrics.py
sfc-gh-melnacouzi: I don't get that part - Why is time determining the order? The most accurate way is first call goes first. Or are we dealing with multithreads  here?

sfc-gh-mchok: On Windows, `time.monotonic()` does not have as much precision, so multiple spans would end up having the same start time, so when we sort them by start time (to make it so that the first call goes first) the order ends up not being the same as the order as when we start the spans. This is fixed with `time.perf_counter()`

sfc-gh-bdufour: weird that the sorting function wasn't stable though.

sfc-gh-mchok: When we take spans breadth-first (to make sure they conform to limits) we sort by depth first, which breaks the ordering of when the spans get inserted: https://github.com/snowflakedb/snowflake-cli/blob/main/src/snowflake/cli/api/metrics.py#L293

sfc-gh-bdufour: Not sure I follow. This should all be deterministic, even if two spans have the same start time. IMHO that's a hard requirement. I've been burnt by stuff like this in the past, so I'm extra careful now. Happy to discuss possible solutions if you want.

sfc-gh-mchok: Took a stab at it, let me know if this works for you: https://github.com/snowflakedb/snowflake-cli/pull/1844/commits/ae43f25d8835785a4887339f688dfa953f86ca22

src/snowflake/cli/api/metrics.py
```diff
@@ -99,6 +100,9 @@ class for holding metrics span data and encapsulating related operations
     # denotes whether direct children were trimmed from telemetry payload
     TRIMMED_KEY: ClassVar[str] = "trimmed"
 
+    # counter for sorting by creation order
+    _CREATION_COUNTER: ClassVar[count] = count()
```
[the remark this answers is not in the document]
sfc-gh-mchok: Commented [here](https://github.com/snowflakedb/snowflake-cli/pull/1844#discussion_r1848875513)

tests_integration/nativeapp/test_metrics.py
```diff
@@ -0,0 +1,552 @@
...
+
+        message = extract_first_telemetry_message_of_type(
+            mock_telemetry, TelemetryEvent.CMD_EXECUTION_RESULT.value
+        )
+
+        spans = message[CLITelemetryField.SPANS.value]
+
+        assert_spans_within_limits(spans)
+
+        assert extract_span_keys_to_check(spans) == [
```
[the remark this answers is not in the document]
sfc-gh-bdufour: ok

src/snowflake/cli/api/metrics.py
```diff
@@ -294,9 +299,11 @@ def completed_spans(self) -> List[Dict]:
             )
         )
 
-        # sort by start time to make reading the payload easier
         sorted_spans_to_report = sorted(
-            spans_to_report, key=lambda span: span.start_time
+            spans_to_report,
+            # start_time can be the same, so we want to sort by something more
+            # deterministic, so we use a counter to see which spans are started first
+            key=lambda span: span.creation_key,
```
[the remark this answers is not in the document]
sfc-gh-mchok: Discussed offline: requirement to have spans sorted by start_time during reporting was waived, the new requirement is to only do one sort, and keep the spans sorted by (depth, start_time, end_time), in that order for tiebreaking

src/snowflake/cli/api/metrics.py
```diff
@@ -275,34 +274,29 @@ def num_spans_past_total_limit(self) -> int:
...
-                key=lambda span: (span.span_depth, span.start_time),
-            )
+        spans_to_report = nsmallest(
+            n=self.SPAN_TOTAL_LIMIT,
+            iterable=(
+                span
+                for span in self._completed_spans
+                if span.span_depth <= self.SPAN_DEPTH_LIMIT
+            ),
+            key=lambda span: (span.span_depth, span.start_time, span.execution_time),
```
sfc-gh-mchok, line 291: As discussed offline, only one sort, however the last tiebreaker is `execution_time` here instead of end_time since they are logically equivalent and we track execution time explicitly, not end time

sfc-gh-bdufour (approved):

sfc-gh-melnacouzi (approved):

sfc-gh-astus (approved):

PR #1845, opened 2024-11-08
Add --label to version create

sfc-gh-pjafari:
### Pre-review checklist

### Changes description
Added `--label` option to `snow app version create`. version label in `manifest.yml` will be ignored if this option is provided. 
    
Other changes:
- `--patch` can be provided to the cli even if there is no version name specified in the cli. `manifest.yml will be checked for the version name. if there is a conflicting patch in the manifest, the user will have to confirm to use the cli-provided patch.   

- If a version name is provided through the cli, all version info from manifest (version name, patch, and label) are ignored.

### DIFF
--- a/src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
+++ b/src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
@@ -667,7 +667,7 @@ class ApplicationPackageEntity(EntityBase[ApplicationPackageEntityModel]):
             f"Defining a new version {version}{with_label_prompt} in application package {self.name}"
         )
         get_snowflake_facade().create_version_in_package(
-            package_role=self.role,
+            role=self.role,
             package_name=self.name,
             stage_fqn=self.stage_fqn,
             version=version,
@@ -691,7 +691,7 @@ class ApplicationPackageEntity(EntityBase[ApplicationPackageEntityModel]):
             f"Adding new patch to version {version}{with_label_prompt} defined in application package {self.name}"
         )
         new_patch = get_snowflake_facade().add_patch_to_package_version(
-            package_role=self.role,
+            role=self.role,
             package_name=self.name,
             stage_fqn=self.stage_fqn,
             version=version,
--- a/src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
+++ b/src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
@@ -363,7 +363,7 @@ class ApplicationPackageEntity(EntityBase[ApplicationPackageEntityModel]):
             git_policy = AllowAlwaysPolicy()
 
         bundle_map = self._bundle()
-        version_info = self.resolve_version_info(
+        resolved_version, resolved_patch, resolved_label = self.resolve_version_info(
             version=version,
             patch=patch,
             label=label,
@@ -371,7 +371,6 @@ class ApplicationPackageEntity(EntityBase[ApplicationPackageEntityModel]):
             policy=policy,
             interactive=interactive,
         )
-        resolved_version, resolved_patch, resolved_label = version_info
 
         if git_policy.should_proceed():
             self.check_index_changes_in_git_repo(policy=policy, interactive=interactive)
@@ -665,7 +664,7 @@ class ApplicationPackageEntity(EntityBase[ApplicationPackageEntityModel]):
         with_label_prompt = f" labeled {label}" if label else ""
 
         console.step(
-            f"Defining a new version {version} in application package {self.name}"
+            f"Defining a new version {version}{with_label_prompt} in application package {self.name}"
         )
         get_snowflake_facade().create_version_in_package(
             package_role=self.role,
@@ -973,18 +972,16 @@ class ApplicationPackageEntity(EntityBase[ApplicationPackageEntityModel]):
                 dedent(
                     f"""\
                         Version was not provided through the Snowflake CLI. Checking version in the manifest.yml instead.
-                        This step will bundle your app artifacts to determine the location of the manifest.yml file.
                     """
                 )
             )
             if bundle_map is None:
                 self._bundle()
-            version_info = find_version_info_in_manifest_file(self.deploy_root)
-            resolved_version, patch_manifest, label_manifest = (
-                version_info.version_name,
-                version_info.patch_number,
-                version_info.label,
-            )
+            (
+                resolved_version,
+                patch_manifest,
+                label_manifest,
+            ) = find_version_info_in_manifest_file(self.deploy_root)
             if resolved_version is None:
                 raise ClickException(
                     "Manifest.yml file does not contain a value for the version field."
--- a/src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
+++ b/src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
@@ -330,7 +330,6 @@ class ApplicationPackageEntity(EntityBase[ApplicationPackageEntityModel]):
 
             return show_obj_cursor
 
-    # PJ-TODO: check what label=None does
     def action_version_create(
         self,
         action_ctx: ActionContext,
@@ -736,7 +735,7 @@ class ApplicationPackageEntity(EntityBase[ApplicationPackageEntityModel]):
 
         # Label must be a string literal
         with_label_query = (
-            f"label={to_string_literal(label)}" if label is not None else ""
+            f"\nlabel={to_string_literal(label)}" if label is not None else ""
         )
         # Use raw string in prompt
         with_label_prompt = f" labeled {label}" if label else ""
@@ -749,8 +748,7 @@ class ApplicationPackageEntity(EntityBase[ApplicationPackageEntityModel]):
                 f"""\
                     alter application package {self.name}
                         add version {version}
-                        using @{self.stage_fqn}
-                        {with_label_query}
+                        using @{self.stage_fqn}{with_label_query}
                 """
             )
             sql_executor.execute_query(add_version_query, cursor_class=DictCursor)
@@ -773,7 +771,7 @@ class ApplicationPackageEntity(EntityBase[ApplicationPackageEntityModel]):
 
         # Label must be a string literal
         with_label_query = (
-            f"label={to_string_literal(label)}" if label is not None else ""
+            f"\nlabel={to_string_literal(label)}" if label is not None else ""
         )
         # Use raw string in prompt
         with_label_prompt = f" labeled {label}" if label else ""
@@ -786,8 +784,7 @@ class ApplicationPackageEntity(EntityBase[ApplicationPackageEntityModel]):
                 f"""\
                     alter application package {self.name}
                         add patch {patch if patch else ""} for version {version}
-                        using @{self.stage_fqn}
-                        {with_label_query}
+                        using @{self.stage_fqn}{with_label_query}
                 """
             )
             result_cursor = sql_executor.execute_query(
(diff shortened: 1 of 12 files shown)


src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
```diff
@@ -357,10 +361,22 @@ def action_version_create(
         else:
             git_policy = AllowAlwaysPolicy()
 
-        # Make sure version is not None before proceeding any further.
-        # This will raise an exception if version information is not found. Patch can be None.
         bundle_map = None
-        if not version:
+        version_resolved = None
```
sfc-gh-bdufour, line 365: super nit: my brain wants this to be `resolved_version` (same for the other 2)

src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
```diff
@@ -357,10 +361,22 @@ def action_version_create(
...
+        patch_resolved = None
+        label_resolved = ""
+
+        # If version is specified in CLI, no version information from manifest.yml is used (except for comment, we can't control comment as of now).
```
sfc-gh-bdufour, line 369: did you try if we could set the comment to NULL explicitly? We should file an internal ticket if there's no workaround.

src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
```diff
@@ -357,10 +361,22 @@ def action_version_create(
...
+        if version is not None:
+            console.message(
+                "Version information in manifest.yml is ignored since you provided version in your command."
+            )
+            patch_resolved = patch
+            label_resolved = label if label is not None else ""
+            version_resolved = version
+
+        # When version is not set by CLI, version name is read from manifest.yml. patch and label from CLI will be used, if provided.
+        else:
```
sfc-gh-bdufour, line 379: this method is getting really massive. How about we split the version/patch/label resolution into a helper, and return your new named tuple directly? It would greatly simplify the version creation logic proper.

src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
```diff
@@ -370,24 +386,57 @@ def action_version_create(
...
-        # Check if --patch needs to throw a bad option error, either if application package does not exist or if version does not exist
-        if patch is not None:
+            # If patch is set in CLI and is also present in manifest.yml with different value, confirmation from user is required to ignore patch from manifest.yml and proceed with CLI value.
+            if (
+                patch is not None
+                and patch_manifest is not None
+                and patch_manifest != patch
+            ):
+                console.warning(
+                    f"Cannot resolve version. Found patch: {patch_manifest} in manifest.yml which is different from provided patch {patch}. Re-run command with --force to ignore patch in manifest.yml"
```
sfc-gh-bdufour, line 407: Not sure we should instruct the user to re-run unless we're in non-interactive mode.

src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
```diff
@@ -370,24 +386,57 @@ def action_version_create(
...
+
+        # Check if patch needs to throw a bad option error, either if application package does not exist or if version does not exist
+        if patch_resolved is not None:
             try:
-                if not self.get_existing_version_info(version):
+                if not self.get_existing_version_info(version_resolved):
                     raise BadOptionUsage(
                         option_name="patch",
-                        message=f"Cannot create a custom patch when version {version} is not defined in the application package {self.name}. Try again without using --patch.",
+                        message=f"Cannot create patch {patch_resolved} when version {version_resolved} is not defined in the application package {self.name}. Try again without specifying patch.",
```
sfc-gh-bdufour, line 434: nit: specifying _a_ patch?

src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
```diff
@@ -370,24 +386,57 @@ def action_version_create(
...
                     )
             except ApplicationPackageDoesNotExistError as app_err:
                 raise BadOptionUsage(
                     option_name="patch",
-                    message=f"Cannot create a custom patch when application package {self.name} does not exist. Try again without using --patch.",
+                    message=f"Cannot create patch {patch_resolved} when application package {self.name} does not exist. Try again without specifying patch.",
```
sfc-gh-bdufour, line 439: ditto

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -181,6 +189,102 @@ def get_account_event_table(self, role: str | None = None) -> str | None:
             return None
         return table
 
+    def create_version_in_package(
```
sfc-gh-bdufour, line 192: Nice :)

sfc-gh-bdufour (approved):

src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
sfc-gh-pjafari: great idea. done

src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
sfc-gh-pjafari: the SQL syntaxt does not accept a `comment` parameter. see https://docs.snowflake.com/en/sql-reference/sql/alter-application-package-version#parameters
Will open a ticket internally for it.

src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
```diff
@@ -357,38 +362,16 @@ def action_version_create(
...
-                    raise BadOptionUsage(
-                        option_name="patch",
-                        message=f"Cannot create a custom patch when version {version} is not defined in the application package {self.name}. Try again without using --patch.",
-                    )
-            except ApplicationPackageDoesNotExistError as app_err:
-                raise BadOptionUsage(
-                    option_name="patch",
-                    message=f"Cannot create a custom patch when application package {self.name} does not exist. Try again without using --patch.",
-                )
+        bundle_map = self._bundle()
```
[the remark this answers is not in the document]
sfc-gh-pjafari: We always deploy during version create which means we always bundle so that message didn't make sense before these changes anyway. I've removed it.

src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
```diff
@@ -671,60 +657,51 @@ def get_existing_release_directive_info_for_version(
...
-                        add version {version}
-                        using @{self.stage_fqn}
-                """
-            )
-            sql_executor.execute_query(add_version_query, cursor_class=DictCursor)
-            console.message(
-                f"Version {version} created for application package {self.name}."
-            )
+        console.step(
+            f"Defining a new version {version} in application package {self.name}"
```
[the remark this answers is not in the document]
sfc-gh-pjafari: I had it here too but removed it. I felt like the label is not necessary in the step, but I'll re-add it :)

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -181,6 +189,102 @@ def get_account_event_table(self, role: str | None = None) -> str | None:
...
+            except Exception as err:
+                handle_unclassified_error(
+                    err,
+                    f"Failed to create patch {patch_query} for version {version} in application package {package_name}.",
+                )
+            try:
+                show_row = result_cursor[0]
+            except IndexError as err:
+                raise err
+            new_patch = show_row["patch"]
```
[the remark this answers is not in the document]
sfc-gh-pjafari: good point. I augmented the error.

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -128,6 +131,11 @@ def _use_schema_optional(self, schema_name: str | None):
         """
         return self._use_object_optional(UseObjectType.SCHEMA, schema_name)
 
+    def _use_role(self, role: str):
+        if role is None:
+            raise ExpectedObjectNotProvidedError(UseObjectType.ROLE)
```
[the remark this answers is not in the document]
sfc-gh-pjafari: oh I wasn't sure if the type-checker will error so that was my safeguard. Removed the check and the error type.

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -181,6 +189,102 @@ def get_account_event_table(self, role: str | None = None) -> str | None:
...
+        self,
+        package_role: str,
+        package_name: str,
+        stage_fqn: str,
+        version: str,
+        label: str | None = None,
+    ):
+        """
+        Creates a new version in an existing application package.
+        @param package_role: Switch to this role while executing create version.
```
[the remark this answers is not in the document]
sfc-gh-pjafari: The package role should always be necessary since we are altering application package. No db/schema/wh is needed I don't think. The current implementation of creating new version or adding a patch does not specify a warehouse.

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -181,6 +189,102 @@ def get_account_event_table(self, role: str | None = None) -> str | None:
...
+
+        # Label must be a string literal
+        with_label_query = (
+            f"\nlabel={to_string_literal(label)}" if label is not None else ""
+        )
+        add_version_query = dedent(
+            f"""\
+                alter application package {package_name}
+                    add version {version}
+                    using @{stage_fqn}{with_label_query}
```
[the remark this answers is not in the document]
sfc-gh-pjafari: Thank you for the wording I thought about it couldn't come up with the word `clause` :D.  I had it in its own line but the extra blank line broke tests, which is why I added "\n" in the clause.

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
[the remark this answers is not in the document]
sfc-gh-bdufour: Frank has a fair point though, being consistent with argument order across the APIs exposed by the facade makes things easier for developers overall. As for the role being required, I'm not sure we should enforce it here. The caller has full context. If they know the current role is fine, then so be it.

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -181,6 +187,102 @@ def get_account_event_table(self, role: str | None = None) -> str | None:
...
+                self._sql_executor.execute_query(add_version_query)
+            except Exception as err:
+                handle_unclassified_error(
+                    err,
+                    f"Failed to add version {version} to application package {package_name}.",
+                )
+
+    def add_patch_to_package_version(
+        self,
+        package_role: str,
```
sfc-gh-bdufour, line 232: ditto here then I guess

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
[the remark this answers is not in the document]
sfc-gh-pjafari: hmm I can see both sides of the argument for and against this method enforcing the role. But to stay consistent, I updated the order and made role optional :)

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
sfc-gh-pjafari: done

sfc-gh-bdufour (approved):

PR #1870, opened 2024-11-19
move create/alter app upgrade calls to sql facade to reclassify errors cause by setup script execution

sfc-gh-mchok:
### Pre-review checklist

### Changes description
- Move sql calls that execute setup scripts (CREATE APPLICATION and ALTER APPLICATION UPGRADE) to the SQL Facade, and reclassify all ProgrammingErrors coming out of these calls as UserInputErrors
- Update tests that use these calls to mock the sql facade


src/snowflake/cli/_plugins/nativeapp/same_account_install_method.py
```diff
@@ -10,23 +11,12 @@
 from snowflake.cli._plugins.stage.manager import StageManager
 
 
+@dataclass
```
sfc-gh-mchok, line 14: Turning this to a dataclass for automatically implemented equality comparison, which is useful for tests which compare mock call args

src/snowflake/cli/_plugins/nativeapp/entities/application.py
```diff
@@ -719,19 +720,28 @@ def create_or_upgrade_app(
                         self.execute_post_deploy_hooks()
                         return
 
-                    except ProgrammingError as err:
-                        if err.errno == CANNOT_DISABLE_MANDATORY_TELEMETRY:
+                    except (UserInputError, ProgrammingError) as err:
```
sfc-gh-mchok, line 723: Not in love with this style of error handling, but unsure what would be a better approach without further refactoring

tests/nativeapp/test_event_sharing.py
```diff
@@ -487,11 +519,22 @@ def test_event_sharing_disabled_no_change_to_current_behavior(
...
 
-    assert mock_execute_query.mock_calls == mock_execute_query_expected
-    mock_console.warning.assert_not_called()
+    assert [
+        *mock_execute_query.mock_calls,
+        *mock_sql_facade_upgrade_application.mock_calls,
+        *mock_sql_facade_create_application.mock_calls,
+    ] == expected
+
+    mock_console.warning.assert_called_once_with(
```
sfc-gh-melnacouzi, line 528: why is application creation a warning? or has it always been the case?

tests/nativeapp/test_event_sharing.py
```diff
@@ -550,13 +597,27 @@ def test_event_sharing_disabled_but_we_add_event_sharing_flag_in_project_definit
...
+    ] == expected
+
+    assert mock_console.warning.mock_calls == [
+        mock.call(
+            "WARNING: Same-account event sharing is not enabled in your account, therefore, application telemetry section will be ignored."
+        ),
+        mock.call(
+            DEFAULT_UPGRADE_SUCCESS_MESSAGE
+            if is_upgrade
+            else DEFAULT_CREATE_SUCCESS_MESSAGE
```
sfc-gh-melnacouzi, line 613: for simplicity, you can have the same message since this is within the same test. We already check the mocks for upgrade/create, so it's hard to mess this up.

tests/nativeapp/test_event_sharing.py
sfc-gh-mchok: I believe it's always been the case if we look [here](https://github.com/snowflakedb/snowflake-cli/pull/1870/files#diff-e9438874d872da0c1e6711d40f24dd6162ad6cb31684b46eb61688a8503237bbL336-R333), but we've always had the mock return None, which I don't think reflects reality, so I updated the tests for that

tests/nativeapp/test_event_sharing.py
sfc-gh-mchok: For sure, updated here: https://github.com/snowflakedb/snowflake-cli/pull/1870/commits/17f761b0ef35025d5e2f07f9f0e6ceb47d3068f6

sfc-gh-pjafari (commented):
stil reading :D

src/snowflake/cli/_plugins/nativeapp/same_account_install_method.py
```diff
@@ -53,8 +44,9 @@ def using_clause(
             return ""
 
         if self.version:
+            version_clause = f"version {to_identifier(self.version)}"
```
sfc-gh-mchok, line 47: Bug where numbers like `2` are passed in directly without being converted to identifier, leading to syntax errors

src/snowflake/cli/_plugins/nativeapp/entities/application.py
```diff
@@ -634,6 +591,75 @@ def get_objects_owned_by_application(self) -> List[ApplicationOwnedObject]:
...
+        install_method: SameAccountInstallMethod,
+        event_sharing: EventSharingHandler,
+        current_app_row: dict,
+        policy: PolicyBase,
+        interactive: bool,
+    ) -> list[tuple[str]] | None:
+        console = self._workspace_ctx.console
+        self._workspace_ctx.console.step(
+            f"Upgrading existing application object {self.name}."
+        )
```
sfc-gh-pjafari, line 606: `console.step` teehee

src/snowflake/cli/_plugins/nativeapp/entities/application.py
```diff
@@ -634,6 +591,75 @@ def get_objects_owned_by_application(self) -> List[ApplicationOwnedObject]:
...
+
+        try:
+            return get_snowflake_facade().upgrade_application(
+                name=self.name,
+                current_app_row=current_app_row,
+                install_method=install_method,
+                stage_fqn=stage_fqn,
+                debug_mode=self._entity_model.debug,
```
sfc-gh-pjafari, line 614: kinda nit: 
create a `@property` getter for `self._entity_model.debug`, same for `self._entity_model.telemetry`

`self._workspace_ctx.console` is used in various places too.

src/snowflake/cli/_plugins/nativeapp/entities/application.py
```diff
@@ -634,6 +591,75 @@ def get_objects_owned_by_application(self) -> List[ApplicationOwnedObject]:
...
+                    )
+                    stage_schema = extract_schema(stage_fqn)
+                    sql_executor.execute_query(
+                        f"grant usage on schema {package.name}.{stage_schema} to role {self.role}"
+                    )
+                    sql_executor.execute_query(
+                        f"grant read on stage {stage_fqn} to role {self.role}"
+                    )
+        except ProgrammingError as err:
+            generic_sql_error_handler(err)
```
sfc-gh-pjafari, line 650: Move this to facade too while we're here :) something like: `grant_create_app` or such. 
Will upgrade our error handling too

src/snowflake/cli/_plugins/nativeapp/entities/application.py
```diff
@@ -643,20 +669,14 @@ def create_or_upgrade_app(
...
+
         with sql_executor.use_role(self.role):
             event_sharing = EventSharingHandler(
-                telemetry_definition=model.telemetry,
+                telemetry_definition=self._entity_model.telemetry,
                 deploy_root=package.deploy_root,
                 install_method=install_method,
-                console=console,
+                console=self._workspace_ctx.console,
             )
```
sfc-gh-pjafari, line 680: So about this use_role block :D
instantiating an EventSharingHandler shouldn't need to be in a use_role block. If it needs to be done with a specific role, pass the role in. 

Looking at implementation of `get_existing_app_info`, it seems that it could be moved as a whole to the facade. Just make sure you pass an optional role in, similar to other methods in the facade. 

For `_upgrade_app` and `_create_app`, you're already doing role handling in them.

It seems that this `use_role` is trivial now upon changes above and can be removed.

src/snowflake/cli/_plugins/nativeapp/same_account_install_method.py
```diff
@@ -8,25 +9,15 @@
...
-        patch: Optional[int] = None,
-        from_release_directive: bool = False,
-    ):
-        self._requires_created_by_cli = requires_created_by_cli
-        self.version = version
-        self.patch = patch
-        self._from_release_directive = from_release_directive
+    version: Optional[str] = None
+    patch: Optional[int] = None
+    _from_release_directive: bool = False
```
sfc-gh-pjafari, line 21: Thank you :D

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -42,6 +60,15 @@
...
 
+# Reasons why an `alter application ... upgrade` might fail
+UPGRADE_RESTRICTION_CODES = {
+    CANNOT_UPGRADE_FROM_LOOSE_FILES_TO_VERSION,
+    CANNOT_UPGRADE_FROM_VERSION_TO_LOOSE_FILES,
+    ONLY_SUPPORTED_ON_DEV_MODE_APPLICATIONS,
+    NOT_SUPPORTED_ON_DEV_MODE_APPLICATIONS,
+    APPLICATION_NO_LONGER_AVAILABLE,
+}
+
```
sfc-gh-pjafari, line 71: You can move these to the `sf_facade_exceptions` :)

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -503,6 +530,138 @@ def show_release_directives(
...
+                    raise UserInputError(
+                        "Could not disable telemetry event sharing for the application because it contains mandatory events. Please set 'share_mandatory_events' to true in the application telemetry section of the project definition file."
+                    ) from err
+
+                raise UserInputError(
+                    f"Failed to upgrade application {name} with the following error message:\n"
+                    f"{err.msg}"
+                ) from err
+            except Exception as err:
+                handle_unclassified_error(err, f"Failed to upgrade application {name}.")
```
sfc-gh-pjafari, line 603: In create case, it makes sense that you have all the error handling in one block since you send one query with the clause. 
Here however, you can separate your sql calls in try blocks of their own to make the error handling more readable. 
You can also move the telemetry upgrade block into its own private method here, like `_update_telemetry_authorization` if you want

src/snowflake/cli/_plugins/nativeapp/entities/application.py
```diff
@@ -643,173 +657,64 @@ def create_or_upgrade_app(
...
             # 1. Need to use a warehouse to create an application object
             with sql_executor.use_warehouse(self.warehouse):
 
                 # 2. Check for an existing application by the same name
                 show_app_row = self.get_existing_app_info()
 
+                stage_fqn = stage_fqn or package.stage_fqn
+
                 # 3. If existing application is found, perform a few validations and upgrade the application object.
+                create_or_upgrade_result = None
```
sfc-gh-melnacouzi, line 680: if we remove the self._upgrade_app() and _create_app() functions (since they do not do much), would this logic become clearer?

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -503,6 +587,148 @@ def show_release_directives(
...
+        debug_mode: bool | None,
+        should_authorize_event_sharing: bool | None,
+    ) -> list[tuple[str]]:
+        """
+        Creates a new application object using an application package,
+        running the setup script of the application package
+        """
+
+        if package_role != role:
+            self._grant_privileges_for_create_application(
```
sfc-gh-melnacouzi, line 681: i don't think this belongs here tbh - I feel this should be called outside of this function, since it's supposed to "create_application" with specific parameters (not grant stuff)

src/snowflake/cli/_plugins/nativeapp/entities/application.py
sfc-gh-pjafari: I think it's ok as it is now - since create and upgrade do different things, e.g. create needs permission grants and upgrade has specific error handling. Either way I don't have strong feelings about it :D

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
sfc-gh-mchok: I extracted the logic without creating a helper, I think it looks alright as it is right now, but might be weird with the error raising duplication, but i made the error messages more specialized. Let me know what you think

cc @sfc-gh-melnacouzi

sfc-gh-melnacouzi (dismissed):

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
sfc-gh-melnacouzi: - separate error handling makes sense if there are no shared errors -might look ugly though if there is shared error handling that we need to repeat.
- separating telemetry authorization is optional given that it is not used somewhere else yet.
I think you can make the call on this @sfc-gh-mchok based on which one is more readable.

src/snowflake/cli/_plugins/nativeapp/entities/application.py
```diff
@@ -208,35 +193,6 @@ def should_authorize_event_sharing_during_create(
...
-            return None
-
-        current_app_authorization = (
-            upgraded_app_properties.get(AUTHORIZE_TELEMETRY_COL, "false").lower()
-            == "true"
-        )
-
-        # Skip the update if the current value is the same as the one we want to set
-        if current_app_authorization == self._share_mandatory_events:
-            return None
```
sfc-gh-mchok, line 236: this part moved to sql facade upgrade_application at suggestion

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -136,6 +155,31 @@ def _use_schema_optional(self, schema_name: str | None):
...
+        @param object_type: Type of snowflake object to grant to a role
+        @param object_identifier: Valid identifier of the snowflake object to grant to a role
+        @param role_to_grant: Name of the role to grant privileges to
+        @param [Optional] role_to_use: Name of the role to use to grant privileges
+        """
+        comma_separated_privileges = ", ".join(privileges)
+        object_type_and_name = f"{object_type.value.sf_name} {object_identifier}"
+
+        with self._use_role_optional(role_to_use):
+            self._sql_executor.execute_query(
```
sfc-gh-mchok, line 179: No error handling on this to defer to the caller to categorize the errors (e.g. whether or not we can safely say it's user vs. our errors), let me know any thoughts

sfc-gh-melnacouzi: what if we just plug in the generic_error_handling function here?

sfc-gh-melnacouzi (approved):

src/snowflake/cli/_plugins/nativeapp/entities/application.py
```diff
@@ -765,50 +765,53 @@ def drop_application_before_upgrade(
         interactive: bool,
         cascade: bool = False,
     ):
-        if cascade:
-            try:
-                if application_objects := self.get_objects_owned_by_application():
-                    application_objects_str = _application_objects_to_str(
-                        application_objects
+        sql_executor = get_sql_executor()
+        with sql_executor.use_role(self.role):
```
sfc-gh-mchok, line 769: Just adding this line at the top and indenting the rest of the function. I believe this doesn't need a warehouse from my tests so I did not add it, but please let me know if I am mistaken

src/snowflake/cli/api/constants.py
```diff
@@ -62,14 +62,22 @@ class ObjectType(Enum):
...
         return self.value.cli_name
 
 
 OBJECT_TO_NAMES = {o.value.cli_name: o.value for o in ObjectType}
-SUPPORTED_OBJECTS = sorted(OBJECT_TO_NAMES.keys())
+UNSUPPORTED_OBJECTS = {
+    ObjectType.APPLICATION.value.cli_name,
+    ObjectType.APPLICATION_PACKAGE.value.cli_name,
+}
+SUPPORTED_OBJECTS = sorted(OBJECT_TO_NAMES.keys() - UNSUPPORTED_OBJECTS)
```
sfc-gh-mchok, line 80: I wanted to add application and application package as constants to support the new sql facade operations with a cleaner API but I did not want to have them be supported in commands like object list/drop/etc., unsure if there's a better way of handling this

src/snowflake/cli/_plugins/nativeapp/entities/application.py
```diff
@@ -838,64 +759,61 @@ def get_existing_app_info(self) -> Optional[dict]:
         Check for an existing application object by the same name as in project definition, in account.
         It executes a 'show applications like' query and returns the result as single row, if one exists.
         """
-        sql_executor = get_sql_executor()
-        with sql_executor.use_role(self.role):
-            return sql_executor.show_specific_object(
-                "applications", self.name, name_col=NAME_COL
-            )
+        return get_snowflake_facade().get_existing_app_info(self.name, self.role)
```
sfc-gh-melnacouzi, line 762: i am confused, why do we use facade version sometimes, and sometimes use the application one (self.get_existing_app_info)?

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -503,6 +553,198 @@ def show_release_directives(
                 )
             return cursor.fetchall()
 
+    def get_existing_app_info(self, name: str, role: str) -> dict | None:
```
sfc-gh-melnacouzi, line 556: why is this different than show_specific_object?

src/snowflake/cli/_plugins/nativeapp/entities/application.py
sfc-gh-mchok: My mistake, I forgot to refactor the other places where we call that method, I just replaced them all now

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
sfc-gh-mchok: The goal was to also move `show_specific_object` calls to the facade eventually, this is one of those instances, see https://github.com/snowflakedb/snowflake-cli/pull/1870#discussion_r1858668505 for initial context

sfc-gh-melnacouzi (dismissed):

sfc-gh-melnacouzi (approved):

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -503,6 +553,198 @@ def show_release_directives(
...
+                object_type_plural = ObjectType.APPLICATION.value.sf_plural_name
+                show_obj_query = f"show {object_type_plural} like {identifier_to_show_like_pattern(name)}".strip()
+
+                show_obj_cursor = self._sql_executor.execute_query(
+                    show_obj_query, cursor_class=DictCursor
+                )
+
+                show_obj_row = find_first_row(
+                    show_obj_cursor,
+                    lambda row: row[NAME_COL] == unquote_identifier(name),
```
sfc-gh-pjafari, line 572: I think you'll have a more robust comparison if you use `_same_identifier` method for equality check here.

sfc-gh-melnacouzi (approved):

sfc-gh-jsikorski (approved):

PR #1909, opened 2024-12-02
Moved Snowpark and Streamlit artifacts to separate directory in outpu…

sfc-gh-astus:
…t/bundle

### Pre-review checklist

### Changes description
Moved Snowpark and Streamlit artifacts to separate directory in output/bundle

### DIFF
--- a/src/snowflake/cli/_plugins/snowpark/snowpark_project_paths.py
+++ b/src/snowflake/cli/_plugins/snowpark/snowpark_project_paths.py
@@ -25,7 +25,7 @@ from snowflake.cli.api.console import cli_console
 from snowflake.cli.api.constants import DEPLOYMENT_STAGE
 from snowflake.cli.api.feature_flags import FeatureFlag
 from snowflake.cli.api.identifiers import FQN
-from snowflake.cli.api.project.project_paths import ProjectPaths
+from snowflake.cli.api.project.project_paths import ProjectPaths, bundle_root
 from snowflake.cli.api.project.schemas.entities.common import PathMapping
 from snowflake.cli.api.secure_path import SecurePath
 
@@ -45,6 +45,7 @@ class SnowparkProjectPaths(ProjectPaths):
         if FeatureFlag.ENABLE_SNOWPARK_GLOB_SUPPORT.is_enabled():
             return Artefact(
                 project_root=self.project_root,
+                bundle_root=self.bundle_root,
                 dest=artifact_path.dest,
                 path=Path(artifact_path.src),
             )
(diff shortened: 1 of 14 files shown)


src/snowflake/cli/_plugins/snowpark/snowpark_project_paths.py
```diff
@@ -74,19 +78,29 @@ def snowflake_requirements(self) -> SecurePath:
     def requirements(self) -> SecurePath:
         return SecurePath(self.path_relative_to_root(Path("requirements.txt")))
 
+    @property
+    def deploy_root(self) -> Path:
+        return self.project_root / "output" / "deploy" / "snowpark"
+
```
sfc-gh-mraba, line 84: This path is build in multiple places in repeated manner.
How about creating a common tool for a deploy root path

```python
from pathlib import Path


def deploy_root(root: Path, app_type: str | None = None) -> Path:
    if app_type:
        return root / "output" / "deploy" / app_type
    return root / "output" / "deploy"
```
---

```python
@property
def deploy_root(self) -> Path:
   return deploy_root(self.project_root, "snowpark")
```
---

```python
@property
def deploy_root(self) -> Path:
   return deploy_root(self.project_root, "streamlit")
```
---
```python
@property
def deploy_root(self) -> Path:
   return deploy_root(self.project_root)
```
---

sfc-gh-astus: Changed

src/snowflake/cli/api/project/project_paths.py
```diff
@@ -9,8 +9,14 @@ class ProjectPaths:
...
 
     def remove_up_deploy_root(self) -> None:
         if self.deploy_root.exists():
             rmtree(self.deploy_root)
+
+
+def deploy_root(root: Path, app_type: str | None = None) -> Path:
+    if app_type:
+        return root / "output" / "deploy" / app_type
+    return root / "output" / "deploy"
```
sfc-gh-bdufour, line 22: 1. Native apps have made this a configurable property in snowflake.yml since day 1. I think all domains should support that as well.
2. This isn't resilient to having multiple entities of the same type, which is a big deal if we start deploying multiple entities in a dependency chain. I think we should either use `output/deploy`, as native apps do, for all domains now (it's consistent and simple for now), or make `output/deploy/<entity id>` work across the board. That last one is probably significantly bigger in terms of implementation however.

sfc-gh-astus: 1. I will add.
2. You mean to mix artifacts from all entities in `output/deploy`? I added `output/deploy/snowpark` and `output/deploy/streamlit` to separate artifacts for each entity type, to keep similar solution what we have right now.

sfc-gh-bdufour: there won't be any mixing, each bundle/deploy step should wipe out `output/deploy`. Segregating by type doesn't buy us anything since multiple entities for the same type could have conflicting artifacts.

sfc-gh-bdufour: About #1, I factored out the model portion on our feature branch already.

PR #1910, opened 2024-12-02
Add feature flag for enable_release_channels

sfc-gh-melnacouzi:
### Pre-review checklist

### Changes description
- Add feature flag (in config.toml) for enable_release_channels
- Feature flag has to be explicitly enabled or disabled for it to take effect. Otherwise, we will rely on the default in the backend.
- For the feature flag to be respected, release channels need to be enabled in the account. That's why we check the parameter first before attempting to update the flag on the application package.


src/snowflake/cli/_plugins/nativeapp/codegen/compiler.py
```diff
@@ -155,5 +155,5 @@ def _should_invoke_processors(self):
 
     def _is_enabled(self, processor: ProcessorMapping) -> bool:
         if processor.name.lower() == NA_SETUP_PROCESSOR:
-            return FeatureFlag.ENABLE_NATIVE_APP_PYTHON_SETUP.is_enabled()
+            return FeatureFlag.ENABLE_NATIVE_APP_PYTHON_SETUP.get_flag_value() is True
```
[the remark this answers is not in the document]
sfc-gh-melnacouzi: How do you suggest doing this?
- is_enabled: True if set to true, False if set to False, None otherwise?
- is_disabled: True if set to false, False if set to True, None otherwise?

OR:
- is_enabled: True if set to true, False otherwise (including None).
- is_disabled: False if set to false, False otherwise (including None).
Optional: - is_set: True if explicitly set.

WDYT?

sfc-gh-astus (dismissed):

src/snowflake/cli/_plugins/connection/util.py
```diff
@@ -76,26 +72,22 @@ def get_ui_parameter(
...
 
-    query = dedent(
-        f"""
-        select value['value']::string as PARAM_VALUE, value['name']::string as PARAM_NAME from table(flatten(
-            input => parse_json(SYSTEM$BOOTSTRAP_DATA_REQUEST()),
-            path => 'clientParamsInfo'
-        )) where value['name'] in ('{"', '".join(parameters_to_fetch)}');
-        """
-    )
+    query = "call system$bootstrap_data_request('CLIENT_PARAMS_INFO')"
```
sfc-gh-bdufour, line 82: I'm afraid the knowledge will get lost quickly here, please add a comment regarding how this logic avoids requiring a warehouse.

src/snowflake/cli/api/config.py
```diff
@@ -286,8 +286,12 @@ def get_config_value(*path, key: str, default: Optional[Any] = Empty) -> Any:
         raise
 
 
-def get_config_bool_value(*path, key: str, default: Optional[Any] = Empty) -> bool:
-    value = get_config_value(*path, key=key, default=default)
+def get_config_bool_value(*path, key: str, default: Any = Empty) -> bool | None:
```
sfc-gh-bdufour, line 289: that signature doesn't make sense to me, looks like we can return `Optional[Any]` with the current logic. That wasn't the case before because we tried to cast any value to a bool.

Do we really need `Any` anyway? `Optional[bool]` is probably a lot more intuitive.

src/snowflake/cli/_plugins/connection/util.py
sfc-gh-melnacouzi: added comment.

src/snowflake/cli/api/config.py
```diff
@@ -286,8 +286,12 @@ def get_config_value(*path, key: str, default: Optional[Any] = Empty) -> Any:
         raise
 
 
-def get_config_bool_value(*path, key: str, default: Optional[Any] = Empty) -> bool:
-    value = get_config_value(*path, key=key, default=default)
+def get_config_bool_value(*path, key: str, default: Optional[bool]) -> Optional[bool]:
+    value = get_config_value(*path, key=key, default=None)
+
+    if value is None:
+        return default
```
sfc-gh-astus, line 293: What is the difference between default get_config_value and `if`?

sfc-gh-astus (approved):

src/snowflake/cli/api/config.py
sfc-gh-melnacouzi: With if statement, I skip the boolean parsing after

PR #1911, opened 2024-12-02
Changed snapshot assert to field check

sfc-gh-astus:
### Pre-review checklist

### Changes description
Changed snapshot assert to field check.


tests_integration/nativeapp/test_version.py
```diff
@@ -377,6 +377,8 @@ def test_nativeapp_version_create_package_no_magic_comment(
             row.pop("created_on", None)
             # Remove metric_level field
             row.pop("metric_level", None)
+            # Remove auditlog_level field
+            row.pop("auditlog_level", "OFF")
```
sfc-gh-mraba, line 381: How about creating a minimal set of required fields for this snapshot to work and discard all others.
Another option is to create a tuple of fields to discard.

With first case we won't be notice if any new field is added - only removed ones will trigger test to fail
With second we will need to update each time a new field is added.

Finally, do we really need to test for payload in audit log? Do we control the content?
Maybe we could only check if response exists and/or response status?

sfc-gh-astus: Changed

sfc-gh-mraba (approved):

tests_integration/nativeapp/test_version.py
```diff
@@ -372,12 +372,18 @@ def test_nativeapp_version_create_package_no_magic_comment(
...
         # app package contains version v1 with 2 patches
         actual = runner.invoke_with_connection_json(split(list_command))
-        for row in actual.json:
-            # Remove date field
-            row.pop("created_on", None)
-            # Remove metric_level field
-            row.pop("metric_level", None)
-        assert actual.json == snapshot
+        assert len(actual.json) == 2
+        _assert_list_of_dicts_contains_key_with_value(actual.json, "patch", 0)
```
sfc-gh-bdufour, line 376: This is significantly weakening the assertion. Since the JSON output is part of the contract, I'd prefer checking that the actual dict is a superset of the expected one instead. Do you want to make the change or would you prefer that we do it?

sfc-gh-astus: I didn't know which fields should I include, treat this as a proposition. Would be great if you could do this 🙏 .

sfc-gh-bdufour: ok feel free to close this, I'll submit a PR soon.

sfc-gh-bdufour:
Replaced with https://github.com/snowflakedb/snowflake-cli/pull/1918

PR #1928, opened 2024-12-06
Downgrade snowpark lib until a fix is identified

sfc-gh-bdufour:
### Pre-review checklist

### Changes description

Snowpark 1.26.0 breaks native app's snowpark integration. It looks like a new piece of logic was introduced that incorrectly assumes that Snowpark has a valid session, which is not the case when we execute Snowpark in the CLI sandbox. NADE will need to work with the Snowpark team on the fix or workaround, until then we shouldn't use the buggy library.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ dependencies = [
   "setuptools==75.6.0",
   'snowflake.core==1.0.2; python_version < "3.12"',
   "snowflake-connector-python[secure-local-storage]==3.12.3",
-  'snowflake-snowpark-python>=1.15.0;python_version < "3.12"',
+  'snowflake-snowpark-python>=1.15.0,<1.26.0;python_version < "3.12"',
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -35,7 +35,7 @@ dependencies = [
   "setuptools==75.6.0",
   'snowflake.core==1.0.2; python_version < "3.12"',
   "snowflake-connector-python[secure-local-storage]==3.12.3",
-  'snowflake-snowpark-python>=1.15.0;python_version < "3.12"',
+  'snowflake-snowpark-python>=1.15.0,<1.26.0;python_version < "3.12"',
```
sfc-gh-bdufour, line 38: not entirely sure why the context.py alone wasn't sufficient, but I had to do this too for E2E to pass.

sfc-gh-astus (approved):

PR #1856, opened 2024-11-13
POC: Add child entities to application package

sfc-gh-gbloom:
### Pre-review checklist
   * [x] I've confirmed that instructions included in README.md are still correct after my changes in the codebase.
   * [x] I've added or updated automated unit tests to verify correctness of my new code.
   * I've added or updated integration tests to verify correctness of my new code.
   * [x] I've confirmed that my changes are working by executing CLI's commands manually on MacOS.
   * I've confirmed that my changes are working by executing CLI's commands manually on Windows.
   * [x] I've confirmed that my changes are up-to-date with the target branch.
   *  I've described my changes in the release notes.
   * [x] I've described my changes in the section below.

### Changes description

This PR adds the foundation to enable composability in native application packages. Eventually, users will be able to specify existing Streamlit/Snowpark/SPCS entities as children of `ApplicationPackageEntity` to have them bundled in the app package.

For demonstration purposes, a basic functionality was added to `StreamlitEntity`, showing the two required methods entities will need to implement: `bundle()` and `get_deploy_sql()`.

1. Adding a feature flag `enable_native_app_children`
2. Adding `children` field to `ApplicationPackageEntityModel`:
    a. `target` is a reference to the entity to be included as a child, validating that only the supported children types are specified.
    b. `identifier.schema` can be specified to create the child in a specific schema.
    c. `application_role` can be specified to auto-grant `USAGE` on the child object.
3. A new `_entities` directory will be created under the deploy root to hold the artifacts of each child entity.
  a. The directory name can be configured with the new `entities_dir` field.
  b. A sub-directory will be created for each child (as the entity name), and then `ChildEntity.bundle(dir)` will be called for the child to bundle itself there.
4. Each child is called with `ChildEntity.get_deploy_sql()`, which returns the SQL without executing it. This SQL is appended to the setup script alongside `GRANT USAGE` statements.

### Manual test

Starting with the following `snowflake.yml` configuration:
```yml
definition_version: 2
entities:
  pkg:
    type: application package
    identifier: my_pkg
    artifacts:
      - src: app/*
        dest: ./
      - src: add.py
    children_artifacts_dir: _entities
    children:
      - target: my_streamlit
        identifier:
          schema: my_v_schema
        grant_usage:
          application_roles: ["my_app_role"]

  app:
    type: application
    identifier: my_app
    debug: true
    from:
      target: pkg

  my_streamlit:
    type: streamlit
    main_file: streamlit_app.py
    artifacts:
      - streamlit_app.py
```

[...]

PR #1934, opened 2024-12-09
Add streamlit entities

sfc-gh-jsikorski:
### Pre-review checklist

### Changes description
Added streamlit entity outline for introducing composability in PRDF v2


src/snowflake/cli/_plugins/streamlit/streamlit_entity.py
```diff
@@ -1,12 +1,97 @@
...
 from snowflake.cli._plugins.streamlit.streamlit_entity_model import (
     StreamlitEntityModel,
 )
-from snowflake.cli.api.entities.common import EntityBase
+from snowflake.cli._plugins.workspace.context import ActionContext
+from snowflake.cli.api.entities.common import EntityBase, get_sql_executor
+from snowflake.connector.cursor import SnowflakeCursor
 
 
 class StreamlitEntity(EntityBase[StreamlitEntityModel]):
```
sfc-gh-pczajka, line 13: This class is currently used only in tests - commands still use Manager

src/snowflake/cli/_plugins/streamlit/streamlit_entity.py
```diff
@@ -1,12 +1,97 @@
...
+
+    def action_get_url(
+        self, action_ctx: ActionContext, *args, **kwargs
+    ):  # maybe this should be a property
+        name = self._entity_model.fqn.using_connection(self._conn)
+        return make_snowsight_url(
+            self._conn, f"/#/streamlit-apps/{name.url_identifier}"
+        )
+
+    def action_get_deploy_sql(self, action_ctx: ActionContext, *args, **kwargs):
```
sfc-gh-pczajka, line 80: Could you add a comment why this is left empty?

src/snowflake/cli/_plugins/streamlit/streamlit_entity.py
```diff
@@ -1,12 +1,97 @@
...
+    def action_get_drop_sql(self, action_ctx: ActionContext, *args, **kwargs):
+        return f"DROP STREAMLIT {self._entity_model.fqn}"
+
+    def action_get_execute_sql(self, action_ctx: ActionContext, *args, **kwargs):
+        return f"EXECUTE STREAMLIT {self._entity_model.fqn}()"
+
+    def get_share_sql(
+        self, action_ctx: ActionContext, to_role: str, *args, **kwargs
+    ) -> str:
+        return f"GRANT USAGE ON STREAMLIT {{self._entity_model.fqn}} to role {to_role}"
```
sfc-gh-pczajka, line 97:
```suggestion
        return f"GRANT USAGE ON STREAMLIT {{self._entity_model.fqn}} TO ROLE {to_role}"
```
to be consistent :)

tests/streamlit/__snapshots__/test_commands.ambr
```diff
@@ -11,38 +11,6 @@
   
   '''
 # ---
-# name: test_deploy_put_files_on_stage[example_streamlit-merge_definition1]
```
sfc-gh-pczajka, line 14: Was this removed by your change?

src/snowflake/cli/_plugins/streamlit/streamlit_entity.py
sfc-gh-gbloom, line 13: I added a temporary mock implementation of `StreamlitEntity` for composability with Native Apps in #1856.
The fake logic will be replaced by your PR, but please take a look at the proposed `ApplicationPackageChildInterface` contract.

tests/test_data/projects/example_streamlit_v2/utils/utils.py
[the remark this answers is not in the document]
sfc-gh-jsikorski: Nope, it's listed in `snowflake.yml`

src/snowflake/cli/_plugins/streamlit/streamlit_entity.py
```diff
@@ -1,12 +1,97 @@
...
+        output_folder.mkdir(parents=True, exist_ok=True)
+
+        output_files = []
+
+        # This is far from , but will be replaced by bundlemap mappings.
+        for file in artifacts:
+            output_file = output_folder / file.name
+
+            if file.is_file():
+                shutil.copy(file, output_file)
```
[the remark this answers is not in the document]
sfc-gh-jsikorski: Done

src/snowflake/cli/_plugins/streamlit/streamlit_entity.py
sfc-gh-jsikorski: Fixed

tests/streamlit/__snapshots__/test_commands.ambr
sfc-gh-jsikorski: It was a leftover snapshot- not used by any test

src/snowflake/cli/_plugins/streamlit/streamlit_entity.py
sfc-gh-jsikorski: Renamed some methods to allign it with your interface.
Bundle and deploy actions are left intact, as they will have to be heavily remodeled, after introducing project-wide bundle map and other changes to output files management

src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
```diff
@@ -833,6 +833,7 @@ def _bundle_children(self, action_ctx: ActionContext) -> List[str]:
                 child_entity.get_deploy_sql(
                     artifacts_dir=child_artifacts_dir.relative_to(self.deploy_root),
                     schema=child_schema,
+                    replace=True,
```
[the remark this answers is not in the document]
sfc-gh-jsikorski: For snowpark entities- for shure. We want user to be able to ensure, if the deploy operation is allowed to overwrite any data.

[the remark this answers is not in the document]
sfc-gh-gbloom: We should probably pass the `replace` value as it exists on the base entity, and then allow to override it for a specific child when used in an app. How will users control `replace` in a standalone Streamlit entity?

It doesn't have to block this PR though. Just please add a TODO comment here to address this issue, since it will now be `True` for all app children without being able to control it.

sfc-gh-gbloom (approved):

src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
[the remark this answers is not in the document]
sfc-gh-jsikorski: You got me a bit lost here. I would assume, that users will control `replace` just like they do now- by passing `replace` flag to command/action.
Assuming by the expected query in your test, it was done implicitly, without user control, as `CREATE OR REPLACE` was a default value.

src/snowflake/cli/_plugins/streamlit/streamlit_entity.py
```diff
@@ -28,43 +25,141 @@ def __init__(self, *args, **kwargs):
...
-            [
-                PathMapping(src=str(artifact))
-                for artifact in self._entity_model.artifacts
-            ],
+    def action_deploy(self, action_ctx: ActionContext, *args, **kwargs):
+        # After adding bundle map- we should use it's mapping here
+
+        query = self.get_deploy_sql()
+        result = self._sql_executor.execute_query(query)
+        return result
```
sfc-gh-pczajka, line 55:
```suggestion
        return self._sql_executor.execute_query(self.get_deploy_sql())
```
To be consistent with the rest of the functions :)

sfc-gh-pczajka, line 55:
```suggestion
    def action_deploy(self, action_ctx: ActionContext, *args, **kwargs):
        # After adding bundle map- we should use it's mapping here
        return self._sql_executor.execute_query(self.get_deploy_sql())
```

To be consistent with the rest of the functions :)

src/snowflake/cli/_plugins/streamlit/streamlit_entity.py
```diff
@@ -28,43 +25,141 @@ def __init__(self, *args, **kwargs):
...
 
-    def bundle(self, bundle_root=None):
-        return build_bundle(
-            self.project_root,
-            bundle_root or self.deploy_root,
-            [
-                PathMapping(src=str(artifact))
-                for artifact in self._entity_model.artifacts
-            ],
+    def action_deploy(self, action_ctx: ActionContext, *args, **kwargs):
```
sfc-gh-pczajka, line 50: Should deploy action also copy the artifacts to the stage (after resolving `--replace` conflicts)?

tests/streamlit/test_streamlit_entity.py
```diff
@@ -1,53 +1,183 @@
...
+                )
+
+                workspace_context = WorkspaceContext(
+                    console=mock.MagicMock(),
+                    project_root=pdir,
+                    get_default_role=lambda: "test_role",
+                    get_default_warehouse=lambda: "test_warehouse",
+                )
+
+                return (
```
sfc-gh-pczajka, line 43:
```suggestion
                yield (
```

src/snowflake/cli/_plugins/streamlit/streamlit_entity.py
sfc-gh-jsikorski: Done

src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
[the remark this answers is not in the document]
sfc-gh-gbloom: A `--replace` flag works when we're deploying a single entity. But here we will have multiple child entities deployed as part of the native app, and the user might want to `replace` only some of them.
For now we can keep `replace=True` to be the implied behavior, but eventually we need to enable a way to override it.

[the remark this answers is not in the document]
sfc-gh-jsikorski: Oh right, so, we should think of- in your PoC, or while implementing `depends_on` of a way, in which user can specify which mode should be used for each dependency/child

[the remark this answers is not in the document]
sfc-gh-gbloom: Yep. We will figure out the exact syntax later on. I was thinking about something like this:
```yml
entities:
  pkg:
    type: application package
    children:
      - target: my_sproc
        replace: false
```

sfc-gh-pczajka (approved):

src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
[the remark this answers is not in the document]
sfc-gh-gbloom:
```suggestion
                    # TODO Allow users to override the hard-coded value for specific children
                    replace=True,
```

tests/nativeapp/test_children.py
```diff
@@ -140,13 +139,13 @@ def test_children_bundle_with_custom_dir(project_directory):
...
-                )
+                f"""
+-- AUTO GENERATED CHILDREN SECTION
+CREATE OR REPLACE STREAMLIT IDENTIFIER('v_schema.my_streamlit')
+FROM = '{custom_dir_path}'
+MAIN_FILE = 'streamlit_app.py';
+CREATE APPLICATION ROLE IF NOT EXISTS my_app_role;
+GRANT USAGE ON SCHEMA v_schema TO APPLICATION ROLE my_app_role;
+GRANT USAGE ON STREAMLIT v_schema.my_streamlit TO APPLICATION ROLE my_app_role;
+"""
```
sfc-gh-gbloom, line 150: Why no `dedent`?

src/snowflake/cli/_plugins/streamlit/streamlit_entity.py
```diff
@@ -28,43 +25,140 @@ def __init__(self, *args, **kwargs):
...
-            return f"CREATE OR REPLACE STREAMLIT {entity_id} MAIN_FILE='{self._entity_model.main_file}';"
+            query = "CREATE "
+
+        schema_to_use = schema or self._entity_model.fqn.schema
+        query += f"STREAMLIT {self._entity_model.fqn.set_schema(schema_to_use).sql_identifier}"
+
+        if from_stage_name:
+            query += f"\nROOT_LOCATION = '{from_stage_name}'"
+        elif artifacts_dir:
+            query += f"\nFROM = '{artifacts_dir}'"
```
sfc-gh-gbloom, line 128: I think having the `=` here is invalid syntax ([docs](https://docs.snowflake.com/en/developer-guide/native-apps/adding-streamlit#create-the-streamlit-object-in-the-setup-script))
```suggestion
            query += f"\nFROM '{artifacts_dir}'"
```

tests/nativeapp/test_children.py
sfc-gh-jsikorski: I left it out while fixing this test, but it seems it was not the problem. Fixed.

sfc-gh-gbloom (approved):

sfc-gh-pczajka (approved):

PR #1953, opened 2024-12-13
Add release channels list support

sfc-gh-melnacouzi:
### Pre-review checklist

### Changes description
Add support for `snow app release-channel list [CHANNEL]` command.

Integ tests will be added once UIParam is in production

Screenshot 2024-12-16 at 11 36 30 AM

Screenshot 2024-12-16 at 11 40 49 AM

### DIFF
--- a/src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
+++ b/src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import json
 import os
 import re
-from datetime import datetime
 from pathlib import Path
 from textwrap import dedent
 from typing import Any, List, Literal, Optional, Set, Union
@@ -54,6 +53,7 @@ from snowflake.cli._plugins.nativeapp.sf_facade import get_snowflake_facade
 from snowflake.cli._plugins.nativeapp.sf_facade_exceptions import (
     InsufficientPrivilegesError,
 )
+from snowflake.cli._plugins.nativeapp.sf_sql_facade import ReleaseChannel
 from snowflake.cli._plugins.nativeapp.utils import needs_confirmation, sanitize_dir_name
 from snowflake.cli._plugins.snowpark.snowpark_entity_model import (
     FunctionEntityModel,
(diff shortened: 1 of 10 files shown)


src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
```diff
@@ -134,6 +136,99 @@ def ensure_app_roles_is_a_set(
         return application_roles
 
 
+class ReleaseChannel(dict[str, Any]):
```
sfc-gh-gbloom, line 139: Maybe specify the exact dict structure we're expecting with a `TypedDict`?
It's the same type that `show_release_channels()` returns, right? We can reuse it there too.
That way we would also be able to remove the `type: ignore` comments below.

src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
```diff
@@ -134,6 +136,99 @@ def ensure_app_roles_is_a_set(
...
+
+    @property
+    def targets(self) -> dict[str, Any]:
+        return self.get("targets")  # type: ignore
+
+    @property
+    def versions(self) -> list[str]:
+        return self.get("versions")  # type: ignore
+
+    def print_to_console(self, console: AbstractConsole) -> None:
```
sfc-gh-gbloom, line 191: Can you add a screenshot with an output example?

src/snowflake/cli/_plugins/nativeapp/entities/application_package.py
```diff
@@ -134,6 +136,99 @@ def ensure_app_roles_is_a_set(
...
+        console.message(f"""[bold]{self.name}[/bold]""")
+        accounts_list: Optional[list[str]] = self.targets.get("accounts")
+        target_accounts = (
+            f"({', '.join(accounts_list)})"
+            if accounts_list is not None
+            else "ALL ACCOUNTS"
+        )
+
+        formatted_created_on = (
+            self.created_on.astimezone().strftime("%Y-%m-%d %H:%M:%S.%f %Z")
```
sfc-gh-gbloom, line 204: nit:
Why this specific format? Do we already print some datetimes elsewhere?

sfc-gh-melnacouzi: From the design doc, Bruno wanted something more human readable for console output - in this case, we are outputting the current timezone. 
Also, this is not set in stone, as it's only a console output. `--format=json` part is printed using the default.

sfc-gh-gbloom (approved):

PR #1956, opened 2024-12-18
[SNOW-1833500] New command: snow helpers import-snowsql-connections

sfc-gh-pjob:
### Pre-review checklist

### Changes description
* Added new command: `snow helpers import-snowsql-connections`
  * It allows to import connections from SnowSQL config files.


src/snowflake/cli/_plugins/helpers/commands.py
```diff
@@ -88,3 +100,185 @@ def v1_to_v2(
...
+            section_name
+            for section_name in snowsql_config.sections()
+            if section_name.startswith("connections.")
+        ]
+        for snowsql_connection_section_name in other_snowsql_connection_section_names:
+            log.debug(
+                "Reading SnowSQL's connection configuration [%s] from [%s]",
+                snowsql_connection_section_name,
+                str(file.path),
+            )
```
sfc-gh-turbaszek, line 186: I would consider `cli_console.step` instead of debug. Imo showing users what is being read what is happening may be helpful in case of this command.

src/snowflake/cli/_plugins/helpers/commands.py
```diff
@@ -88,3 +100,185 @@ def v1_to_v2(
...
+        if imported_default_connection
+        else {}
+    )
+    if (
+        named_default_connection
+        and default_cli_connection_name in imported_named_connections
+    ):
+        raise ClickException(
+            f"Default connection name [{default_cli_connection_name}] conflicts with the name of one of connections from SnowSQL. Please specify a different name for your default connection."
+        )
```
sfc-gh-turbaszek, line 222: How about making this interactive?
```
Connection 'foo' exists in both configuration files, do you want to use SnowSQL definition? [y/N]
```
WDYT?

src/snowflake/cli/_plugins/helpers/commands.py
```diff
@@ -88,3 +100,185 @@ def v1_to_v2(
...
+
+    key_names_replacements = {
+        "accountname": "account",
+        "username": "user",
+        "databasename": "database",
+        "dbname": "database",
+        "schemaname": "schema",
+        "warehousename": "warehouse",
+        "rolename": "role",
+        "private_key_path": "private_key_file",
```
sfc-gh-turbaszek, line 240: Is this full list of options?

src/snowflake/cli/_plugins/helpers/commands.py
```diff
@@ -88,3 +100,185 @@ def v1_to_v2(
...
+    }
+
+    def parse_value(value: Any):
+        try:
+            parsed_value = literal_eval(value)
```
sfc-gh-turbaszek, line 245: Why we need eval?

src/snowflake/cli/_plugins/helpers/commands.py
sfc-gh-pczajka: I like the idea. I'd also consider `--no-interactive` or `--skip-conflicting-connections` so user can choose not to be prompted, but this might wait until someone actually requests it.

src/snowflake/cli/_plugins/helpers/commands.py
```diff
@@ -88,3 +100,185 @@ def v1_to_v2(
...
+                )
+                continue
+
+            connection_name = snowsql_connection_section_name.removeprefix(
+                "connections."
+            )
+            imported_named_conenction = _convert_connection_from_snowsql_config_section(
+                snowsql_named_connection
+            )
+            if connection_name in imported_named_connections:
```
sfc-gh-pczajka, line 204: Shouldn't we raise / prompt if connection is defined twice instead of overriding (or at least log `console.step(connection redefined in X, overriding)`)

src/snowflake/cli/_plugins/helpers/commands.py
sfc-gh-pjob: It's only `literal_eval` and it's used as parser of values (to correct types) because using raw values would add them to CLI always as strings and CLI wouldn't be able to use such config. Reading everything as strings is the specific behaviour of configparser. At the beginning I wanted to use TOML parser to read SQL's config but SQL's config can contain strings without quotes (even example connection added after installation contains such strings) and they are not valid TOML values. So as the result I read everything as strings and try to parse literals to correct types before saving them in CLI.

src/snowflake/cli/_plugins/helpers/commands.py
sfc-gh-pjob: To be honest, I don't know.

src/snowflake/cli/_plugins/helpers/commands.py
```diff
@@ -88,3 +100,185 @@ def v1_to_v2(
...
         )
     return MessageResult("Project definition migrated to version 2.")
+
+
+@app.command(name="import-snowsql-connections", requires_connection=False)
+def import_snowsql_connections(
+    custom_snowsql_config_files: Optional[List[Path]] = typer.Option(
+        None,
+        "--snowsql-config-file",
+        help="Specifies file paths to SnowSQL configuration. The option can be used multiple times to specify more than 1 file.",
```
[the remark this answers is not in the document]
sfc-gh-pjob: I applied your suggestion :)

src/snowflake/cli/_plugins/helpers/commands.py
sfc-gh-pjob: I applied your suggestion and changed debug to console.step.

src/snowflake/cli/_plugins/helpers/commands.py
sfc-gh-pjob: I added prompts. At the moment I'm not adding `--no-interactive` and `--skip-conflicting-connections` flags.

src/snowflake/cli/_plugins/helpers/commands.py
sfc-gh-pjob: I suggest to use this list and extend it later if needed.

sfc-gh-pczajka (approved):

PR #1982, opened 2025-01-09
Add support for --create-version in snow app publish

sfc-gh-melnacouzi:
### Pre-review checklist

### Changes description
Allow --create-version in `snow app publish` command.

If user wants us to create the version and patch that they provided, they need to pass `--create-version` flag. This will delegate to the version create command to handle the version creation and then continue with the regular publish workflow.

The same logic of falling back to the manifest will also apply in that case, which makes version and patch optional.

### DIFF
--- a/src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
+++ b/src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
@@ -51,6 +51,7 @@ from snowflake.cli.api.errno import (
     ACCOUNT_DOES_NOT_EXIST,
     ACCOUNT_HAS_TOO_MANY_QUALIFIERS,
     APPLICATION_PACKAGE_MAX_VERSIONS_HIT,
+    APPLICATION_PACKAGE_PATCH_ALREADY_EXISTS,
     APPLICATION_REQUIRES_TELEMETRY_SHARING,
     CANNOT_DEREGISTER_VERSION_ASSOCIATED_WITH_CHANNEL,
     CANNOT_DISABLE_MANDATORY_TELEMETRY,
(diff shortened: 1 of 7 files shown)


src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -407,11 +412,12 @@ def add_patch_to_package_version(
...
             f"\nlabel={to_string_literal(label)}" if label is not None else ""
         )
-        patch_query = f"{patch}" if patch else ""
+
+        patch_query = f" {patch}" if patch is not None else ""
         add_patch_query = dedent(
             f"""\
                  alter application package {package_name}
-                     add patch {patch_query} for version {version}
+                     add patch{patch_query} for version {version}
```
sfc-gh-xke, line 420: Apparently I didn't pay attention to L416 and miss the prefix space in the new statement. And it took me a while to figure out whether the space is required in this line. Could you add a short comment here?

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
```diff
@@ -421,9 +427,14 @@ def add_patch_to_package_version(
                     add_patch_query, cursor_class=DictCursor
                 ).fetchall()
             except Exception as err:
+                if isinstance(err, ProgrammingError):
+                    if err.errno == APPLICATION_PACKAGE_PATCH_ALREADY_EXISTS:
+                        raise UserInputError(
+                            f"Patch {patch} already exists for version {version} in application package {package_name}."
```
sfc-gh-xke, line 433: to be consistent, can we use `Patch{patch_query}` here?

sfc-gh-xke (approved):
LGTM! Just a minor comment.

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
sfc-gh-melnacouzi: patch_query is only needed to adjust the spacing in case we're using auto-patch. In this case, error implies there is actually a patch, and we won't get it when auto-creating a patch.

src/snowflake/cli/_plugins/nativeapp/sf_sql_facade.py
sfc-gh-melnacouzi: Auto-merge was on. I'll add that in the following PR.

PR #1996, opened 2025-01-16
Fix notebook create on Windows

sfc-gh-pczajka:
### Pre-review checklist

### Changes description
Fix notebook stage path handling on Windows


tests/notebook/test_notebook_commands.py
```diff
@@ -65,3 +65,30 @@ def test_create(mock_create, runner):
         notebook_name=FQN.from_string("my_notebook"),
         notebook_file=notebook_file,
     )
+
+
+@mock.patch("snowflake.connector.connect")
```
sfc-gh-jwilkowski, line 70: I've seen we have this exact patch repeated around 100 times in the codebase - wdyt about extracting it to a fixture? Requesting that fixture would still be repetitive, but at least we would not have to worry about proper ordering of mocks/args. If you think it makes sense, I'd be happy to take it as a warmup :)

tests/notebook/test_notebook_commands.py
```diff
@@ -65,3 +66,33 @@ def test_create(mock_create, runner):
...
+    )
+    assert result.exit_code == 0, result.output
+    assert ctx.get_query() == (
+        "\n"
+        "CREATE OR REPLACE NOTEBOOK "
+        f"IDENTIFIER('MockDatabase.MockSchema.{notebook_name}')\n"
+        f"FROM '{stage_path}'\n"
+        "QUERY_WAREHOUSE = 'MockWarehouse'\n"
+        "MAIN_FILE = 'notebook.ipynb';\n"
+        "// Cannot use IDENTIFIER(...)\n"
```
sfc-gh-astus, line 96: This comment in query is intentional?

tests/notebook/test_notebook_commands.py
sfc-gh-jsikorski: In testing entites we use constants to define those paths, however they repeat in every test file.
So +1 for any effort to go through tests, and create conts in test utils dir, to call them in such situation

sfc-gh-mraba (approved):

PR #2027, opened 2025-01-27
[wip] Jw/snow 1769565 add stage entity model

sfc-gh-jwilkowski:
### Pre-review checklist

### Changes description
https://snowflakecomputing.atlassian.net/browse/SNOW-1769565


src/snowflake/cli/_plugins/stage/commands.py
```diff
@@ -272,3 +273,12 @@ def _put(
...
         return QueryResult(cursor)
+
+
+@app.command("createv2", requires_connection=True)
+def stage_createv2(stage_name: FQN = StageNameArgument, **options) -> dict:
+    """
+    needed only for testing purposes
+    """
+    stage = StageEntity.create(stage_name.name)
+    return stage.to_dict()
```
sfc-gh-jwilkowski, line 284: returning a dict is not supported, but that's what was printed by `stage.to_dict()`
```
hatch run snow stage createv2 jwilkowski_test2
{'created_on': datetime.datetime(2025, 1, 27, 14, 13, 24, 718000, tzinfo=TzInfo(UTC)),
 'directory_table': {'auto_refresh': False,
                     'enable': False,
                     'refresh_on_create': True},
 'has_credentials': False,
 'has_encryption_key': False,
 'kind': 'PERMANENT',
 'name': 'JWILKOWSKI_TEST2',
 'owner': 'ACCOUNTADMIN',
 'owner_role_type': 'ROLE',
 'type': 'stage',
 'url': ''}
```

Stage gets created on integration env:
image

src/snowflake/cli/_plugins/stage/stage_entity.py
```diff
@@ -0,0 +1,76 @@
...
+        stage = Stage(
+            name=name,
+            kind=KindType.TEMPORARY.value if temporary else KindType.PERMANENT.value,
+            comment=comment,
+        )
+        stage_resource = stage_collection.create(stage, mode=CreateMode.if_not_exists)
+        entity_model_cls = cls.get_entity_model_type()
+        temp = stage_resource.fetch()
+
+        entity_model = entity_model_cls(type="stage", **temp.to_dict())
```
sfc-gh-jsikorski, line 60: I don't think we need adding discriminator field here. Also, we can call `StageEntityModel` directly- `get_entity_model_type` is rathe meant for use in situations, when we have many entities of different types

src/snowflake/cli/_plugins/stage/commands.py
```diff
@@ -272,3 +273,12 @@ def _put(
             auto_compress=auto_compress,
         )
         return QueryResult(cursor)
+
+
+@app.command("createv2", requires_connection=True)
```
sfc-gh-jsikorski, line 278: Just to be safe for now, lets add `hidden=True` here

src/snowflake/cli/_plugins/stage/stage_entity.py
```diff
@@ -0,0 +1,76 @@
...
+    #             yield Root(conn)
+
+    # TODO: discuss: those 2 methods could go to a mixing class for resources that support using snowapis
+    #       figure out typing as we might often deal with <T>Collection, <T>Resource and <T>
+    @classmethod
+    def get_collection(cls, root: Optional[Root] = None) -> StageCollection:
+        conn: SnowflakeConnection = get_cli_context().connection
+        if root is None:
+            root = cls.get_root()
+        return root.databases[conn.database].schemas[conn.schema].stages
```
sfc-gh-jsikorski, line 39: It doesn't have to be `conn.database/schema` - we should derive it from `FQN` if possible

src/snowflake/cli/_plugins/stage/stage_entity.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def get_resource(self, resource_name: str, root: Optional[Root]) -> StageResource:
+        return self.get_collection(root)[resource_name]
+
+    # TODO: discuss: this code looks mostly generic so might go as well to the mixin class. Don't do it too fast to
+    #       avoid issues of premature abstraction
+    @classmethod
```
sfc-gh-jsikorski, line 46: Maybe we could make it an instance method- so it would instantiate a model, as user defines a stage in the project, and then we would use instance method to create said stage server-side. WDYT?

src/snowflake/cli/_plugins/stage/commands.py
sfc-gh-jsikorski: `StageCollection.create` returns `return StageResource(stage.name, self)`- so just a `MessageResult` with `stage.name` should be ok

src/snowflake/cli/_plugins/stage/stage_entity.py
```diff
@@ -0,0 +1,76 @@
...
+from snowflake.core import CreateMode, Root
+from snowflake.core.stage import Stage, StageCollection, StageResource
+
+
+class StageEntity(EntityBase[StageEntityModel]):
+    # TODO: discuss: twose 2 methods might go to parent class
+    @staticmethod
+    def get_root() -> Root:
+        conn: SnowflakeConnection = get_cli_context().connection
+        return Root(conn)
```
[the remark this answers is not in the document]
sfc-gh-jwilkowski: Sure, that's why I extracted that method for easier refactor once it's available upstream :)

src/snowflake/cli/_plugins/stage/stage_entity_model.py
```diff
@@ -0,0 +1,31 @@
...
+class KindType(Enum):
+    PERMANENT = "PERMANENT"
+    TEMPORARY = "TEMPORARY"
+
+
+class StageEntityModel(EntityModelBase, Stage):
+    type: Literal["stage"] = DiscriminatorField()  # noqa: A003
+    # TODO: discuss: inherit or compose? composition would require either a lot of magic or double keying of fields,
+    #       while inheritance does that for free + offers autocompletion etc
+    # api_resource: Stage
```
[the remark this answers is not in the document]
sfc-gh-jwilkowski: Ok, I'll give it a try

src/snowflake/cli/_plugins/stage/commands.py
sfc-gh-jsikorski: Yes, but this will hide the command from any help messages, so the tests will pass

src/snowflake/cli/_plugins/stage/stage_entity_model.py
```diff
@@ -0,0 +1,80 @@
...
+            _model = dict()
+        if isinstance(_model, Stage):
+            self._snowapi_model = _model
+        else:
+            api_model_kwargs = {
+                key: value
+                for key, value in _model.items()
+                if key in self.Meta.supported_fields
+            }
+            # api_model_kwargs["has_encryption_key"] = 123
```
sfc-gh-jwilkowski, line 49: we can pass in data that's violating validation rules and it'll raise an exception only once `revalidate` gets called.

PR #2035, opened 2025-01-30
Pjob/poc/pyapp binary with plugins

sfc-gh-pjob:
### Pre-review checklist

### Changes description
PoC: PyApp + plugins installed and loaded to/from custom directory


src/snowflake/cli/__about__.py
```diff
@@ -14,4 +14,4 @@
 
 from __future__ import annotations
 
-VERSION = "3.4.0.dev0"
+VERSION = "3.4.0.dev8"
```
sfc-gh-pjob, line 17: I have issues with caching of my builds, that's why I bump version in my PoC. This should be resolved.

build_pyapp_binary.sh
```diff
@@ -0,0 +1,5 @@
+pip install -U hatch
+hatch clean
+hatch build
+pip install git+https://github.com/hobbsd/hatch-build-isolated-binary.git@v1.0.0
```
sfc-gh-pjob, line 4: 1. It's external script. It should be reviewed.
2. The script uses tomllib, so it requires python 3.11.

pyproject.toml
```diff
@@ -188,3 +188,6 @@ markers = [
 
 [tool.codespell]
 skip = 'tests/*,snow.spec'
+
+[tool.hatch.build.targets.binary]
+python-version = "3.11"
```
sfc-gh-pjob, line 193: I use python 3.11 but we had issues with pyinstaller binaries when using python 3.11 - we should check it.

src/snowflake/cli/_app/commands_registration/__init__.py
```diff
@@ -12,10 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import sys
 from dataclasses import dataclass
 
 from snowflake.cli.api.plugins.command import CommandSpec
 
+sys.path.append("/tmp/cli_plugins/lib/python3.11/site-packages")
```
sfc-gh-pjob, line 20: We should consider if we can do it better. For sure we need to load plugins from configurable directory.

src/snowflake/cli/_plugins/plugin/commands.py
```diff
@@ -26,10 +28,35 @@
...
+@app.command(name="install", requires_connection=False)
+def install(
+    package: str = typer.Argument(
+        None, help="Pip compatible package (PyPI name / URL / local path)"
+    ),
+    **options,
+) -> CommandResult:
+    """Installs a plugin from a package"""
+    target_dir: str = (
+        "/tmp/cli_plugins"  # TODO make it configurable in config.toml with some default
```
sfc-gh-pjob, line 43: It should be configurable.

build_pyapp_binary.sh
sfc-gh-pjob, line 1: 1. This script is just PoC - probably not all commands are needed.
2. I don't know how it works on different operating systems and CPUs.
3. I don't how will work signing of the binaries.
4. We should probably move the script to `scripts` directory.

src/snowflake/cli/_plugins/plugin/commands.py
```diff
@@ -26,10 +28,35 @@
...
+    )
+    subprocess.check_call(
+        [
+            sys.executable,
+            "-m",
+            "pip",
+            "install",
+            "--prefix",
```
sfc-gh-pjob, line 51: `--prefix`, not `--target` to install only the plugin and new dependencies, without installation of existing CLI's dependencies.

PR #2078, opened 2025-02-18
Jw/SNOW-1890085 snow dbt commands

sfc-gh-jwilkowski:
### Pre-review checklist

### Changes description
...
https://snowflakecomputing.atlassian.net/browse/SNOW-1890085


src/snowflake/cli/_plugins/dbt/commands.py
```diff
@@ -0,0 +1,86 @@
...
+from snowflake.cli._plugins.dbt.constants import DBT_COMMANDS
+from snowflake.cli._plugins.dbt.manager import DBTManager
+from snowflake.cli.api.commands.decorators import global_options_with_connection
+from snowflake.cli.api.commands.snow_typer import SnowTyperFactory
+from snowflake.cli.api.output.types import CommandResult, QueryResult
+
+app = SnowTyperFactory(
+    name="dbt",
+    help="Manages dbt on Snowflake projects",
+    is_hidden=lambda: True,
```
sfc-gh-jwilkowski, line 30: Currently I've only set is_hidden - do you think I should also use feature flags or is this enough?

src/snowflake/cli/_plugins/dbt/commands.py
```diff
@@ -0,0 +1,86 @@
...
+    "list",
+    requires_connection=True,
+)
+def list_dbts(
+    **options,
+) -> CommandResult:
+    """
+    List all dbt on Snowflake projects.
+    """
+    return QueryResult(DBTManager().list())
```
sfc-gh-jwilkowski, line 45: nice to have: we could add `--like` option if you see value/consistency in it

src/snowflake/cli/_plugins/dbt/commands.py
```diff
@@ -0,0 +1,86 @@
...
+dbt_execute_app = SnowTyperFactory(
+    name="execute",
+    help="Execute a dbt command",
+)
+app.add_typer(dbt_execute_app)
+
+
+@dbt_execute_app.callback()
+@global_options_with_connection
+def before_callback(
```
sfc-gh-jwilkowski, line 59:
```
hatch run snow dbt execute --help

 Usage: snow dbt execute [OPTIONS] NAME COMMAND [ARGS]...

 Execute a dbt command

╭─ Arguments ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ *    name      TEXT  [default: None] [required]                                                                                                                                                                                                     │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Options ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --help  -h        Show this message and exit.                                                                                                                                                                                                       │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Connection configuration ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --connection,--environment             -c      TEXT     Name of the connection, as defined in your config.toml file. Default: default.                                                                                                              │

...
│ --diag-allowlist-path                          TEXT     Diagnostic report path to optional allowlist                                                                                                                                                │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Global configuration ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --format           [TABLE|JSON]  Specifies the output format.                                                                                                                                                                                       │
│                                  [default: TABLE]                                                                                                                                                                                                   │
│ --verbose  -v                    Displays log entries for log levels info and higher.                                                                                                                                                               │
│ --debug                          Displays log entries for log levels debug and higher; debug logs contain additional information.                                                                                                                   │
│ --silent                         Turns off intermediate output to console.                                                                                                                                                                          │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Commands ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ build      Execute build command on dbt on Snowflake project.                                                                                                                                                                                       │
│ clean      Execute clean command on dbt on Snowflake project.                                                                                                                                                                                       │
│ clone      Execute clone command on dbt on Snowflake project.                                                                                                                                                                                       │
│ compile    Execute compile command on dbt on Snowflake project.                                                                                                                                                                                     │
│ debug      Execute debug command on dbt on Snowflake project.                                                                                                                                                                                       │
│ deps       Execute deps command on dbt on Snowflake project.                                                                                                                                                                                        │
│ docs       Execute docs command on dbt on Snowflake project.                                                                                                                                                                                        │
│ init       Execute init command on dbt on Snowflake project.                                                                                                                                                                                        │
│ list       Execute list command on dbt on Snowflake project.                                                                                                                                                                                        │
│ parse      Execute parse command on dbt on Snowflake project.                                                                                                                                                                                       │
│ retry      Execute retry command on dbt on Snowflake project.                                                                                                                                                                                       │
│ run        Execute run command on dbt on Snowflake project.                                                                                                                                                                                         │
│ seed       Execute seed command on dbt on Snowflake project.                                                                                                                                                                                        │
│ show       Execute show command on dbt on Snowflake project.                                                                                                                                                                                        │
│ snapshot   Execute snapshot command on dbt on Snowflake project.                                                                                                                                                                                    │
│ source     Execute source command on dbt on Snowflake project.                                                                                                                                                                                      │
│ test       Execute test command on dbt on Snowflake project.                                                                                                                                                                                        │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

src/snowflake/cli/_plugins/dbt/constants.py
```diff
@@ -0,0 +1,20 @@
...
+    "parse",
+    "retry",
+    "run",
+    "run",
+    "seed",
+    "show",
+    "snapshot",
+    "source",
+    "test",
+]
```
sfc-gh-turbaszek, line 20: Should we consider some feature flag for enabling new commands. Something like `SNOWFLAKE_FEATURES_ENABLE_DBT_COMMANDS=foo,bar,baz`?

src/snowflake/cli/_plugins/dbt/commands.py
```diff
@@ -0,0 +1,129 @@
...
+        False,
+        help="Overwrites conflicting files in the project, if any.",
+    ),
+    dbt_version: Optional[str] = typer.Option(
+        None,
+        help="Version of dbt tool to be used. Taken from dbt_project.yml if not provided.",
+    ),
+    dbt_adapter_version: str = typer.Option(
+        help="dbt-snowflake adapter version to be used",
+    ),
```
sfc-gh-jwilkowski, line 75: I don't like this too much - I believe that we should give an option to specify adapter the version, but making it required can be inconvenient. Imho it would be best if we could skip this from sql (if version is not provided) and the server could default to newest supported version.
Unlike dbt version, adapter version is not stored in dbt_project.yml nor profiles.yml. `dbt-core` reads it from imported lib, but that would indicate that we need to have it installed, which imho is not always granted https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/version.py#L200

src/snowflake/cli/api/feature_flags.py
```diff
@@ -68,3 +68,4 @@ class FeatureFlag(FeatureFlagMixin):
     )
     ENABLE_SNOWPARK_GLOB_SUPPORT = BooleanFlag("ENABLE_SNOWPARK_GLOB_SUPPORT", False)
     ENABLE_SPCS_SERVICE_EVENTS = BooleanFlag("ENABLE_SPCS_SERVICE_EVENTS", False)
+    ENABLE_DBT_POC = BooleanFlag("ENABLE_DBT_POC", False)
```
sfc-gh-turbaszek, line 71: Do we plan to rename this flag later on? POC sounds bit unstable even for PrPr

src/snowflake/cli/_plugins/dbt/commands.py
```diff
@@ -0,0 +1,145 @@
...
+        dbt_cli_args = ctx.args
+        dbt_command = ctx.command.name
+        name = ctx.parent.params["name"]
+        run_async = ctx.parent.params["run_async"]
+        execute_args = (dbt_command, name, run_async, *dbt_cli_args)
+
+        if run_async is True:
+            result = DBTManager().execute(*execute_args)
+            return MessageResult(
+                f"Command submitted. You can check the result with `snow sql -q \"select execution_status from table(information_schema.query_history_by_user()) where query_id in ('{result.sfqid}');\"`"
```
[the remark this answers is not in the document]
sfc-gh-jwilkowski: This will be replaced with `snow logs`, but that's still not merged https://github.com/snowflakedb/snowflake-cli/pull/2130

src/snowflake/cli/api/feature_flags.py
sfc-gh-jwilkowski: I didn't get any feedback about it before, but I agree that we could change it to just `ENABLE_DBT` (or if you have better ideas LMK). This is the right time to do it, as we still don't have many people using it

sfc-gh-turbaszek: I would opt for `ENABLE_DBT_PROJECT`

src/snowflake/cli/_plugins/dbt/commands.py
```diff
@@ -0,0 +1,145 @@
...
+            )
+
+        with Progress(
+            SpinnerColumn(),
+            TextColumn("[progress.description]{task.description}"),
+            transient=True,
+        ) as progress:
```
sfc-gh-turbaszek, line 143: Would it make sense to extract this ctx manager to a common helper?

src/snowflake/cli/_plugins/dbt/commands.py
```diff
@@ -0,0 +1,145 @@
...
+        ctx: typer.Context,
+    ) -> CommandResult:
+        dbt_cli_args = ctx.args
+        dbt_command = ctx.command.name
+        name = ctx.parent.params["name"]
+        run_async = ctx.parent.params["run_async"]
+        execute_args = (dbt_command, name, run_async, *dbt_cli_args)
+
+        if run_async is True:
+            result = DBTManager().execute(*execute_args)
```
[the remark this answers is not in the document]
sfc-gh-turbaszek: In fact wdyt about this structure
```py
result = dbt.execute(*execute_args)
if run_async:
   return MessageResult(...)

with Progress(...):
   return QueryResult(result)
```

In this way we have "execute" logic invoked only once.

src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def deploy(
+        self,
+        path: SecurePath,
+        name: FQN,
+        force: bool,
+    ) -> SnowflakeCursor:
+        dbt_project_path = path / "dbt_project.yml"
+        if not dbt_project_path.exists():
+            raise ClickException(f"dbt_project.yml does not exist in provided path.")
```
sfc-gh-turbaszek, line 47: Personally I would put full path in the message, it helps with debugging

src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -0,0 +1,74 @@
...
+
+        if self.exists(name=name) and force is not True:
+            raise ClickException(
+                f"DBT project {name} already exists. Use --force flag to overwrite"
```
sfc-gh-turbaszek, line 51: Will this also clean up related stage?

src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -0,0 +1,74 @@
...
+
+        with cli_console.phase("Creating DBT project"):
+            query = f"""{'CREATE OR REPLACE' if force is True else 'CREATE'} DBT PROJECT {name}
+FROM {stage_name}"""
+
+            return self.execute_query(query)
+
+    def execute(self, dbt_command: str, name: str, run_async: bool, *dbt_cli_args):
+        if dbt_cli_args:
+            dbt_command = dbt_command + " " + " ".join([arg for arg in dbt_cli_args])
```
sfc-gh-turbaszek, line 72:
```suggestion
            dbt_command = " ".join([dbt_command, *dbt_cli_args])
```
WDYT?

src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -0,0 +1,74 @@
...
+        query = f"EXECUTE DBT PROJECT {name} args='{dbt_command.strip()}'"
```
sfc-gh-turbaszek, line 73: I would move strip to previous line where we define dbt_command

src/snowflake/cli/api/feature_flags.py
sfc-gh-jwilkowski: renamed to `ENABLE_DBT` in https://github.com/snowflakedb/snowflake-cli/pull/2078/commits/d9075b03bc5a79b3f357c4c4b20be5d1fd0c6c28

src/snowflake/cli/_plugins/dbt/commands.py
sfc-gh-jwilkowski: I usually extract code to shared after it's copied a couple of times, but if you consider it as a good candidate, then I can add it 👌

src/snowflake/cli/_plugins/dbt/manager.py
sfc-gh-jwilkowski: shlex.join calls shelx.quote behind the scenes, and that's trying too hard to escape parameters such as https://github.com/snowflakedb/snowflake-cli/pull/2078/files#diff-7ab4937785ac1c58eaca5020325c5e75aaf5ce82e0b2bb42f1eafe36df302168R226. In any case the format you suggested improves it a lot and replacing shlex with " " does the job

src/snowflake/cli/_plugins/dbt/manager.py
sfc-gh-jwilkowski: good idea. btw, do we have any good practices around asserting that the right directory is printed? I've had issues where lines was broken, so it couldn't be done with `assert path in output.response`

src/snowflake/cli/_plugins/dbt/commands.py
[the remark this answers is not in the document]
sfc-gh-jwilkowski: This will not work as underneath `dbt.execute` calls `BaseSqlExecutor.execute_query` which waits for the result from server. So the whole waiting would be done before we start the spinner.

An alternative would be to use BaseSqlExecutor.execute_string, which returns a generator, but then we'd have to leak the logic for fetching last result outside of BaseSqlExecutor. In the end, I'd prefer to repeat `result = dbt.execute(*execute_args)` as those 2 calls are close to each other

PR #2113, opened 2025-03-10
Added auth keypair status command and extended connection add

sfc-gh-astus:
### Pre-review checklist

### Changes description
Added `snow auth keypair status` command to verify the key pair configuration and test the connection.
The `snow connection add` command has been enhanced to include key pair authentication when a password is provided.
Added support for ~ in SecurePath

### DIFF
--- a/src/snowflake/cli/_app/snow_connector.py
+++ b/src/snowflake/cli/_app/snow_connector.py
@@ -21,7 +21,7 @@ from typing import Dict, Optional
 
 import snowflake.connector
 from click.exceptions import ClickException
-from snowflake.cli.__about__ import VERSION
+from snowflake.cli import __about__
 from snowflake.cli._app.constants import (
     INTERNAL_APPLICATION_NAME,
     PARAM_APPLICATION_NAME,
@@ -247,7 +247,7 @@ def _update_internal_application_info(connection_parameters: Dict):
     """Update internal application data if ENABLE_SEPARATE_AUTHENTICATION_POLICY_ID is enabled."""
     if FeatureFlag.ENABLE_SEPARATE_AUTHENTICATION_POLICY_ID.is_enabled():
         connection_parameters["internal_application_name"] = INTERNAL_APPLICATION_NAME
-        connection_parameters["internal_application_version"] = VERSION
+        connection_parameters["internal_application_version"] = __about__.VERSION
 
 
 def _load_pem_from_file(private_key_file: str) -> SecretType:
--- a/src/snowflake/cli/_plugins/auth/keypair/commands.py
+++ b/src/snowflake/cli/_plugins/auth/keypair/commands.py
@@ -31,12 +31,6 @@ def _show_connection_name_prompt(ctx: typer.Context, value: str):
     return value
 
 
-def _resolve_default_and_convert(value: Path) -> Path:
-    if value:
-        return value
-    return Path.home()
-
-
 _new_connection_option = typer.Option(
     True,
     help="Create a new connection.",
@@ -69,7 +63,6 @@ _output_path_option = typer.Option(
     "--output-path",
     help="The output path for private and public keys",
     prompt="Enter output path",
-    callback=_resolve_default_and_convert,
 )
 
 
--- a/src/snowflake/cli/_plugins/auth/keypair/commands.py
+++ b/src/snowflake/cli/_plugins/auth/keypair/commands.py
@@ -19,6 +19,9 @@ app = SnowTyperFactory(
 )
 
 
+KEY_PAIR_DEFAULT_PATH = "~/.ssh"
+
+
 def _show_connection_name_prompt(ctx: typer.Context, value: str):
     for param in ctx.command.params:
         if param.name == "connection_name":
@@ -38,7 +41,6 @@ _new_connection_option = typer.Option(
     True,
     help="Create a new connection.",
     prompt="Create a new connection?",
-    confirmation_prompt=True,
     callback=_show_connection_name_prompt,
     show_default=False,
     hidden=True,
@@ -63,7 +65,7 @@ _key_length_option = typer.Option(
 
 
 _output_path_option = typer.Option(
-    "~/.ssh",
+    KEY_PAIR_DEFAULT_PATH,
     "--output-path",
     help="The output path for private and public keys",
     prompt="Enter output path",
@@ -148,3 +150,12 @@ def remove(
     return SingleQueryResult(
         AuthManager().remove_public_key(public_key_property=public_key_property)
     )
+
+
+@app.command("status", requires_connection=True)
+def status(**options):
+    """
+    Verifies the key pair configuration and tests the connection.
+    """
+    AuthManager().status()
+    return MessageResult("Status check completed.")
--- a/src/snowflake/cli/_plugins/auth/keypair/manager.py
+++ b/src/snowflake/cli/_plugins/auth/keypair/manager.py
@@ -82,7 +82,7 @@ class AuthManager(SqlExecutionMixin):
         if not connection_name:
             connection_name = cli_context.connection_context.connection_name
 
-        self._check_if_connection_has_private_key(
+        self._ensure_connection_has_private_key(
             cli_context.connection_context.connection_name
         )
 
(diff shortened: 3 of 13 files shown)


src/snowflake/cli/_plugins/auth/keypair/manager.py
```diff
@@ -159,14 +175,56 @@ def remove_public_key(
...
+            ).path
+        )
+        if connection_options.get("password"):
+            del connection_options["password"]
+
+        return connection_options
+
     @staticmethod
-    def _check_if_connection_has_private_key(connection_name: str):
+    def _check_if_connection_has_private_key(connection_name: str) -> None:
```
sfc-gh-pczajka, line 215: How about naming this "_assert_conntion_has..." or "_ensure_connection_has..."? I'd expect that "_check..." returns True/False, this raises an error if condition is not met

src/snowflake/cli/_plugins/auth/keypair/manager.py
```diff
@@ -159,14 +175,56 @@ def remove_public_key(
             f"ALTER USER {cli_context.connection.user} UNSET {public_key_property.value}"
         )
 
+    def status(self):
+        cli_context = get_cli_context()
+        self._check_if_connection_has_private_key(
+            cli_context.connection_context.connection_name
+        )
+        cli_console.step("Private key set for connection - OK")
+        self._check_connection(cli_context)
```
sfc-gh-pczajka, line 184: If I'm correct this only gets connection configuration, not testing whether connection actually works

tests/auth/test_auth.py
```diff
@@ -659,6 +755,18 @@ def test_remove(
     mock_execute_query.assert_called_once_with(f"ALTER USER test_user UNSET {key}")
 
 
+@pytest.fixture
```
sfc-gh-pczajka, line 758: IT might be worth to add it to "conftest" / "fixtures" - it would be useful in other tests

tests/test_connection.py
```diff
@@ -1397,3 +1398,170 @@ def test_connection_add_no_interactive(mock_add, runner):
...
+            "--connection-name",
+            "conn",
+            "--user",
+            "user",
+            "--account",
+            "account",
+            "--no-interactive",
+        ],
+        input="conn\n"  # connection name: zz
+        "test\n"  # account:
```
sfc-gh-pczajka, line 1550: "user" and "account" are already defined via flags. Do we need "input" if test is "--no-interactive"?

src/snowflake/cli/_plugins/auth/keypair/commands.py
```diff
@@ -19,6 +19,9 @@
 )
 
 
+KEY_PAIR_DEFAULT_PATH = "~/.ssh"
```
sfc-gh-pczajka, line 22: Is this a good location for all OS we support? Not sure about Windows

src/snowflake/cli/_plugins/connection/commands.py
```diff
@@ -402,3 +410,50 @@ def _decrypt(passphrase: str | None):
...
+        )
+        if configure_key_pair:
+            key_length = typer.prompt(
+                "Key length",
+                default=2048,
+                show_default=True,
+            )
+
+            def output_path_parser(value: str) -> SecurePath:
+                if value == KEY_PAIR_DEFAULT_PATH:
```
sfc-gh-pczajka, line 433: Why not just `SecurePath(value)`?

src/snowflake/cli/_plugins/auth/keypair/manager.py
sfc-gh-astus: I was surprised too, but I checked and it open connection.

src/snowflake/cli/_plugins/auth/keypair/manager.py
sfc-gh-astus: Changed to ensure

src/snowflake/cli/_plugins/auth/keypair/manager.py
sfc-gh-pczajka: I found it, thx

tests/auth/test_auth.py
sfc-gh-astus: Moved

tests/test_connection.py
sfc-gh-astus: Whole input is not needed 😱 , removed

src/snowflake/cli/_plugins/auth/keypair/commands.py
sfc-gh-astus: I had some problems with Path, but checked again and changed to Path.home() / .ssh

tests/test_connection.py
```diff
@@ -1397,3 +1398,156 @@ def test_connection_add_no_interactive(mock_add, runner):
...
+        [
+            "connection",
+            "add",
+            "--connection-name",
+            "conn",
+            "--user",
+            "user",
+            "--account",
+            "account",
+            "--no-interactive",
```
sfc-gh-pczajka, line 1547:
```suggestion
            "--no-interactive",
            "--private-key-path",
            str(key.resolve()),
```

src/snowflake/cli/_plugins/auth/keypair/commands.py
sfc-gh-pczajka: Right, I've just found `_resolve_default_and_convert`.

Could you rename it to `_output_path_option_callback` if it's used only there?
Also if I'm correct it should look like this:
```
    if value != KEY_PAIR_DEFAULT_PATH:
        return Path(value)
    return Path.home() / '.ssh'
```
as the callback will be given default value if not provided, not None

tests/test_connection.py
sfc-gh-astus: I don't want to add the key, removed.

src/snowflake/cli/_app/snow_connector.py
```diff
@@ -21,7 +21,7 @@
 
 import snowflake.connector
 from click.exceptions import ClickException
-from snowflake.cli.__about__ import VERSION
+from snowflake.cli import __about__
```
sfc-gh-mraba, line 24: Why need to import whole module?
There are other possibilities to consider:
1.
`from snowflake.cli.__about__ import VERSION as CLI_VERSION`

2.
in `snowflake.cli.__init__.py`
```from snowflake.cli.__about__ import VERSION as CLI_VERSION

# extend or add
__all__ = ('CLI_VERSION',)
```
 and use `from snowflake.cli import CLI_VERSION`

I guess the issue here has something with using `VERSION` as const/variable name

src/snowflake/cli/_plugins/auth/keypair/commands.py
sfc-gh-mraba: what about 

`Path('~/.ssh').expanduser()`

src/snowflake/cli/_plugins/auth/keypair/manager.py
```diff
@@ -250,14 +308,9 @@ def _create_or_update_connection(
         private_key_path: SecurePath,
     ):
         connection = get_connection_dict(current_connection)
-        del connection["password"]
+        if connection.get("password"):
+            del connection["password"]
```
sfc-gh-mraba, line 312:
```suggestion
        connection.pop('password', None)
```

sfc-gh-mraba (approved):

src/snowflake/cli/_plugins/auth/keypair/manager.py
sfc-gh-astus: 1. `Path('~/.ssh').expanduser()` - It didn't work, we have help tests where this path is printed and we compare it with snapshot, on linux/mac and widnows path is different.
```
FAILED tests/test_help_messages.py::test_help_messages[auth.keypair.setup] - AssertionError: assert [+ received] == [- snapshot]
    '''
    ...
    |                                            public keys                       |
  - |                                            [default: ~/.ssh]                 |
  + |                                            [default:                         |
  + |                                            C:\Users\runneradmin\~\.ssh]      |
```

2. changed

sfc-gh-pczajka (approved):

PR #2122, opened 2025-03-11
Fix for symlink problem in tests

sfc-gh-jsikorski:
### Pre-review checklist

### Changes description
Fix for temporary directory returning symlink in tests. This should prevent any problems with paths starting with `/private` in the future.

### DIFF
--- a/tests_common/conftest.py
+++ b/tests_common/conftest.py
@@ -30,7 +30,7 @@ def temp_dir():
     with tempfile.TemporaryDirectory() as tmp_dir:
         try:
             os.chdir(tmp_dir)
-            yield tmp_dir
+            yield str(os.path.realpath(tmp_dir))
         finally:
             # this has to happen before tmp_dir is cleaned up
             # so that we don't try to remove the cwd of the process


tests_common/conftest.py
```diff
@@ -30,7 +30,7 @@ def temp_dir():
     with tempfile.TemporaryDirectory() as tmp_dir:
         try:
             os.chdir(tmp_dir)
-            yield tmp_dir
+            yield str(os.path.realpath(tmp_dir))
```
sfc-gh-astus, line 33: Why not tmp_dir.resolve()?

sfc-gh-jsikorski: `tmp_dir` is a string

sfc-gh-astus (approved):

tests_common/conftest.py
sfc-gh-jwilkowski: Are our needs specific enough to discard using [tmp_path](https://docs.pytest.org/en/stable/how-to/tmp_path.html) fixture? I'm not sure I understand what this PR solves, but maybe using something more widely supported would be beneficial in the long term?

PR #2149, opened 2025-03-25
feat: [SNOW-2002186] save password based connection if subsequent key…

sfc-gh-jwilkowski:
… pair setup fails

### Pre-review checklist

### Changes description
https://snowflakecomputing.atlassian.net/browse/SNOW-2002186

Given a user that already has key-pair configured
When user adds new connection with just password and decides to actually setup key pair auth
Then the command should save connection based on password and inform about underlying failure and how to fix it

### DIFF
--- a/src/snowflake/cli/_plugins/auth/keypair/manager.py
+++ b/src/snowflake/cli/_plugins/auth/keypair/manager.py
@@ -50,18 +50,13 @@ class AuthManager(SqlExecutionMixin):
             connection_name = cli_context.connection_context.connection_name
 
         key_name = AuthManager._get_free_key_name(output_path, connection_name)  # type: ignore[arg-type]
-        try:
-            self._generate_key_pair_and_set_public_key(
-                user=cli_context.connection.user,
-                key_length=key_length,
-                output_path=output_path,
-                key_name=key_name,  # type: ignore[arg-type]
-                private_key_passphrase=private_key_passphrase,
-            )
-        except exceptions.CouldNotSetKeyPairError:
-            raise ClickException(
-                "The public key is set already. Use the rotate command instead."
-            )
+        self._generate_key_pair_and_set_public_key(
+            user=cli_context.connection.user,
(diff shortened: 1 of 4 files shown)


tests/test_connection.py
```diff
@@ -1546,3 +1546,83 @@ def test_connection_add_no_key_pair_setup_if_no_interactive(
...
+        Enter region: 
+        Enter authenticator: 
+        Enter private key file: 
+        Enter token file path: 
+        Do you want to configure key pair authentication? [y/N]: y
+        Key length [2048]: 
+        Output path [~/.ssh]: {tmp_path}
+        Private key passphrase: 
+        Wrote new password-based connection conn to {test_snowcli_config}, however there were some issues during key pair setup:
+        The public key is set already. Use the rotate command instead.
```
sfc-gh-astus, line 1621: But the rotate won't work for that connection :/ , it required private key in connection.

src/snowflake/cli/_plugins/auth/keypair/manager.py
```diff
@@ -49,13 +50,18 @@ def setup(
...
-        )
+        try:
+            self._generate_key_pair_and_set_public_key(
+                user=cli_context.connection.user,
+                key_length=key_length,
+                output_path=output_path,
+                key_name=key_name,  # type: ignore[arg-type]
+                private_key_passphrase=private_key_passphrase,
+            )
+        except exceptions.CouldNotSetKeyPairError:
```
sfc-gh-astus, line 61: `CouldNotSetKeyPairError` could be

```python
class CouldNotSetKeyPairError(ClickException):
    pass
```

and you can omit reraise.

sfc-gh-jwilkowski: I didn't like the fact that I'm changing the error message when handling it in connection add, but maybe you're right - after all we're handling that error in a significantly different way.

PR #2152, opened 2025-03-25
SNOW-1994541: Change cortex calls to use REST

sfc-gh-asravi:
### Pre-review checklist

### Changes description
The Snowflake CLI has a snowflake cortex complete command which calls into Cortex complete, but uses the SQL backend. This means that latency will be more variable for users using this tool.
This PR:
1. Changes the default model to `llama3.1-70b`.
2. Adds a `--backend` flag to switch between SQL and REST backends and implements support for calling into both.
3. Enables REST backend by default.


src/snowflake/cli/_plugins/cortex/commands.py
```diff
@@ -142,17 +152,36 @@ def complete(
...
 
     manager = CortexManager()
+    backend = Backend(backend)
 
     if text:
-        result_text = manager.complete_for_prompt(
-            text=Text(text),
-            model=Model(model),
-        )
+        if backend == Backend("SQL") or backend == Backend("sql"):
```
sfc-gh-bprosnitz, line 159: You can do

`backend.lower() == "sql"` (or similar with the Backend type)

src/snowflake/cli/_plugins/cortex/commands.py
```diff
@@ -142,17 +152,36 @@ def complete(
...
+            result_text = manager.complete_for_conversation(
+                conversation_json_file=SecurePath(file),
+                model=Model(model),
+            )
+        elif backend == Backend("REST") or backend == Backend("rest"):
+            result_text = manager.rest_complete_for_conversation(
+                conversation_json_file=SecurePath(file),
```
sfc-gh-bprosnitz, line 180: To avoid needing 4 cases (REST/SQL x file/text), I think we want to separate out the file/text specific parts from the REST/SQL specific parts

So then the flow is:
- Process input
- Call REST/SQL
- Process output

src/snowflake/cli/_plugins/cortex/manager.py
```diff
@@ -187,3 +188,46 @@ def _query_cortex_result_str(self, query: str) -> str:
...
+    def make_rest_request_body(
+        self,
+        model: Model,
+        prompt: Text,
+    ) -> Dict[str, Any]:
+        data = {
+            "model": str(model),
+            "stream": False,
+        }
+        data["messages"] = [{"content": str(prompt)}]
```
sfc-gh-bprosnitz, line 202: You can add this to data inline.
```
{
...
"messages": [{"content"...
}
```

pyproject.toml
```diff
@@ -39,6 +39,7 @@ dependencies = [
   "snowflake-connector-python[secure-local-storage]==3.14.0",
```
sfc-gh-bprosnitz, line 39: Could you add some tests? https://github.com/snowflakedb/snowflake-cli/blob/5094954f79681fe8b2db61dde7f49e40a804784c/tests/cortex/test_cortex_commands.py

src/snowflake/cli/_plugins/cortex/commands.py
sfc-gh-asravi: But there are 2 different functions for file/test: `complete_for_prompt()` and `complete_for_conversation()`.
So we'd still end up with 4 cases right?

src/snowflake/cli/_plugins/cortex/commands.py
sfc-gh-asravi: updated

src/snowflake/cli/_plugins/cortex/manager.py
sfc-gh-asravi: updated

sfc-gh-astus (changes requested):
Snyk found issue in snowflake-ml-python lib.

src/snowflake/cli/_plugins/cortex/commands.py
```diff
@@ -142,17 +153,36 @@ def complete(
...
+                model=Model(model),
+            )
+        elif backend.lower() == "rest":
+            result_text = manager.rest_complete_for_prompt(
+                text=Text(text),
+                model=Model(model),
+                url=REST_COMPLETE_URL,
+            )
+        else:
+            raise UsageError("--backend option should be either REST or SQL.")
```
sfc-gh-astus, line 171: Backend can be enum type and this will be handled by typer.

src/snowflake/cli/_plugins/cortex/commands.py
sfc-gh-bprosnitz: Any more thoughts on this? As a refresher, we discussed splitting this into two phases so there is less repeat. Happy to talk more about it

src/snowflake/cli/_plugins/cortex/manager.py
```diff
@@ -69,6 +75,62 @@ def complete_for_conversation(
...
+        cortex_inference_service = CortexInferenceService(root=root)
+        try:
+            raw_resp = cortex_inference_service.complete(
+                complete_request=complete_request
+            )
+        except Exception as e:
+            raise
+        result = ""
+        for event in raw_resp.events():
+            data = json.loads(event.data)["choices"][0]["delta"]
```
sfc-gh-bprosnitz, line 105: This needs to not crash if we send messages that don't have these fields. In that case, we should just skip the message

src/snowflake/cli/_plugins/cortex/manager.py
```diff
@@ -69,6 +75,62 @@ def complete_for_conversation(
...
```
sfc-gh-bprosnitz, line 130: This duplicates logic above

src/snowflake/cli/_plugins/cortex/manager.py
```diff
@@ -69,6 +75,62 @@ def complete_for_conversation(
...
+    ) -> str:
+        complete_request = self.make_rest_complete_request(model=model, prompt=text)
+        cortex_inference_service = CortexInferenceService(root=root)
+        try:
+            raw_resp = cortex_inference_service.complete(
+                complete_request=complete_request
+            )
+        except Exception as e:
+            raise
+        result = ""
```
sfc-gh-bprosnitz, line 103: What about reporting midstream errors that have the 'error' SSE type

sfc-gh-asravi: I can raise a midstream exception here. Although I might not be able to get request id here

src/snowflake/cli/_plugins/cortex/manager.py
sfc-gh-asravi: updated it

sfc-gh-asravi:
splitting this into two phases is a bigger change, I can push a separate PR for that. wdyt?

sfc-gh-bprosnitz (dismissed):

src/snowflake/cli/_plugins/cortex/commands.py
```diff
@@ -143,18 +157,30 @@ def complete(
...
-        result_text = manager.complete_for_prompt(
-            text=Text(text),
+        prompt = text
+    elif file:
+        prompt = SecurePath(file).read_text(file_size_limit_mb=DEFAULT_SIZE_LIMIT_MB)
+        is_file_input = True
+    else:
+        raise UsageError("Either --file option or TEXT argument has to be provided.")
+
+    if backend == Backend.SQL:
```
sfc-gh-turbaszek, line 169: Do we need to have sql and rest backends, is there any difference in behaviour?

src/snowflake/cli/_plugins/cortex/constants.py
```diff
@@ -14,4 +14,5 @@
 
 from snowflake.cli._plugins.cortex.types import Model
 
-DEFAULT_MODEL: Model = Model("snowflake-arctic")
+DEFAULT_MODEL: Model = Model("mistral-large2")
```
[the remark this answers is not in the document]
sfc-gh-asravi: Thanks, I've updated the default model to `llama3.1-70b`!

sfc-gh-melnacouzi (dismissed):
test_docs_generation_output.ambr looks good from NADE side

sfc-gh-astus (dismissed):

sfc-gh-turbaszek (dismissed):

src/snowflake/cli/_plugins/cortex/commands.py
```diff
@@ -130,6 +139,11 @@ def complete(
         "--model",
         help="String specifying the model to be used.",
     ),
+    backend: Optional[Backend] = typer.Option(
+        DEFAULT_BACKEND,
+        "--backend",
+        help="String specifying whether to use sql or rest backend. Default is rest.",
```
sfc-gh-pczajka, line 145:
```suggestion
        help="String specifying whether to use sql or rest backend.",
```

sfc-gh-asravi: Thanks, updated!

sfc-gh-asravi:
duplicate of https://github.com/snowflakedb/snowflake-cli/pull/2260

PR #2172, opened 2025-04-02
Snow 1994008 change project version flow

sfc-gh-pczajka:
### Pre-review checklist

### Changes description
Change flow of project versions according to suggestions from https://snowflakecomputing.atlassian.net/browse/SNOW-1994008
Bugfixes:
* `ENTITY_ID` parameter of `snow project list-versions` is now required
* `alias` is passed in "double quotes" to sql queries (otherwise `--alias v1.1` caused SQL compilation error)


src/snowflake/cli/_plugins/project/commands.py
```diff
@@ -117,65 +167,60 @@ def create_version(
...
+        ProjectManager(),
+        project=project,
+        prune=prune,
         from_stage=_from,
         alias=_alias,
         comment=comment,
     )
-    return MessageResult("Version added.")
+    alias_str = "" if _alias is None else _alias + " "
+    return MessageResult(f"Project version {alias_str}added to project {project.fqn}")
```
sfc-gh-astus, line 214: `{alias_str}added` missing space

sfc-gh-mraba, line 214:
```suggestion
    return MessageResult(f"Project version '{alias_str}' added to project {project.fqn}")
```

src/snowflake/cli/_plugins/project/commands.py
```diff
@@ -117,65 +167,60 @@ def create_version(
...
-
-    # Create project and version
-    with cli_console.phase("Creating project and version"):
-        stage_name = FQN.from_stage(project.stage)
-
-        pm = ProjectManager()
-        cli_console.step(f"Creating project {project.fqn}")
-        pm.create(project_name=project.fqn)
+    if no_version:
+        return QueryResult(result)
```
sfc-gh-jwilkowski, line 175: WDYT about returning the same message and type of result regardless of whether you create version or not? I would find it confusing that sometimes I get back a table and sometimes a string

src/snowflake/cli/_plugins/project/commands.py
```diff
@@ -101,14 +114,51 @@ def dry_run(
     return SingleQueryResult(result)
 
 
+def _add_version_to_project(
```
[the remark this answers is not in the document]
sfc-gh-jwilkowski: I agree, creating new versions feels like a domain operation, so should be in the manager layer

src/snowflake/cli/_plugins/project/commands.py
```diff
@@ -101,14 +114,51 @@ def dry_run(
...
     entity_id: str = entity_argument("project"),
+    no_version: bool = no_version_flag(
+        mutually_exclusive=["prune"],
+    ),
+    prune: bool = PruneOption(mutually_exclusive=["no_version"]),
     **options,
 ):
     """
-    Upload local files and create a new version of a project using those files. If the stage does not exist, it will be created.
+    Creates a project in snowflake and initializes it with a new version created from local files.
```
sfc-gh-jwilkowski, line 161:
```suggestion
    Uploads local files and creates a project.
```
I'm also fine with the original wording as long as we skip the part about creating versions since it's configurable

src/snowflake/cli/_plugins/project/commands.py
```diff
@@ -117,65 +167,60 @@ def create_version(
...
     )
-    return MessageResult("Version added.")
+    alias_str = "" if _alias is None else _alias + " "
+    return MessageResult(f"Project version {alias_str}added to project {project.fqn}")
 
 
 @app.command(requires_connection=True)
-def list_versions(entity_id: str = entity_argument("project"), **options):
+def list_versions(
+    entity_id: str = entity_argument("project", required=True), **options
```
sfc-gh-jwilkowski, line 219: oh, thank you :D

src/snowflake/cli/_plugins/project/commands.py
```diff
@@ -101,14 +114,51 @@ def dry_run(
...
+
 @app.command(requires_connection=True)
 @with_project_definition()
-def create_version(
+def create(
     entity_id: str = entity_argument("project"),
+    no_version: bool = no_version_flag(
+        mutually_exclusive=["prune"],
+    ),
+    prune: bool = PruneOption(mutually_exclusive=["no_version"]),
```
sfc-gh-jwilkowski, line 157: Why do we need this? At least in case of create, the stage does not exist yet, so instead of adding this as option we could set it always to False in call to _add_version_to_project

tests/dcm_project/test_dcm_project.py
```diff
@@ -12,19 +13,39 @@
...
+        if no_version:
+            command.append("--no-version")
+        result = runner.invoke(command)
         assert result.exit_code == 0, result.output
 
+    mock_pm().create.assert_called_once_with(expected_project_fqn)
+    if no_version:
+        mock_create.assert_not_called()
+        mock_pm().add_version.assert_not_called()
+        mock_create.assert_not_called()
```
sfc-gh-jwilkowski, line 40: duplicated in L38?

src/snowflake/cli/_plugins/project/commands.py
sfc-gh-pczajka: It might be possible, in my case it was:
- create project
- drop project (which doesn't remove the stage)
- create project again

I think it might be worth to throw an error if stage exists if we want to ensure starting from the clean state

src/snowflake/cli/_plugins/project/commands.py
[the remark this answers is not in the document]
sfc-gh-pczajka: Done

src/snowflake/cli/_plugins/project/commands.py
sfc-gh-pczajka: Changed

src/snowflake/cli/_plugins/project/commands.py
sfc-gh-pczajka: After short discussion:
* `snow project create` will fail if the project or the stage already exists
* `--prune` flag is removed

src/snowflake/cli/_plugins/project/commands.py
sfc-gh-pczajka: Changed to
```
    Creates a project in snowflake.
    By default, the project is initialised with a new version created from local files.
```

src/snowflake/cli/_plugins/project/commands.py
```diff
@@ -117,65 +131,68 @@ def create_version(
...
+    pm = ProjectManager()
+    with cli_console.phase(f"Creating project '{project.fqn}'"):
+        pm.create(project.fqn)
+    if no_version:
+        return MessageResult(f"Project '{project.fqn}' successfully created.")
+
+    pm.add_version(project=project)
+    return MessageResult(
+        f"Project '{project.fqn}' successfully created and initial version is added."
+    )
```
sfc-gh-jwilkowski, line 151: Food for thought: personally I'd put most of this logic into manager, because we're trying here to perform a domain operation of creating a project based on some parameters. Commands files (I'll refer to that as interface) should not be responsible for orchestrating different domain actions, but rather for collecting input from users and presenting them output.

This is just my opinion though and I'm not asking to change anything. Feels like such problems and use cases should be addressed by your upcoming doc :)

src/snowflake/cli/_plugins/project/commands.py
```diff
@@ -103,12 +111,18 @@ def dry_run(
...
+    no_version: bool = typer.Option(
+        False,
+        "--no-version",
+        help="Do not initialize project with a new version, only create the snowflake object.",
+    ),
     **options,
 ):
     """
-    Upload local files and create a new version of a project using those files. If the stage does not exist, it will be created.
+    Creates a project in snowflake.
```
sfc-gh-jwilkowski, line 124: nit: Snowflake

sfc-gh-jwilkowski (approved):
LGTM, but uhmmm... windows :D

src/snowflake/cli/_plugins/project/commands.py
```diff
@@ -103,12 +111,18 @@ def dry_run(
...
+    By default, the project is initialised with a new version created from local files.
```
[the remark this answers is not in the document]
sfc-gh-pczajka: Fixed

src/snowflake/cli/_plugins/project/commands.py
sfc-gh-pczajka: :+1:

src/snowflake/cli/_plugins/project/commands.py
sfc-gh-pczajka: I like the idea ;) changed

sfc-gh-jwilkowski (approved):

PR #2190, opened 2025-04-09
Change table behaviour in snow logs

sfc-gh-jsikorski:
### Pre-review checklist

### Changes description
Small change to `snow logs` behaviour.
Instead of querying for logs table, default table name will be used - as described here https://docs.snowflake.com/en/developer-guide/logging-tracing/event-table-setting-up

User can provide custom event table name, using `--table`

### DIFF
--- a/src/snowflake/cli/_plugins/logs/commands.py
+++ b/src/snowflake/cli/_plugins/logs/commands.py
@@ -36,6 +36,11 @@ def get_logs(
         "--refresh",
         help="If set, the logs will be streamed with the given refresh time in seconds",
     ),
+    event_table: str = typer.Option(
+        None,
+        "--table",
+        help="The table to query for logs. If not provided, the default table will be used",
+    ),
     **options,
 ):
     """
@@ -55,6 +60,7 @@ def get_logs(
             object_name=object_name,
             from_time=from_time,
             refresh_time=refresh_time,
+            event_table=event_table,
         )
         logs = itertools.chain(
             (MessageResult(log.log_message) for logs in logs_stream for log in logs)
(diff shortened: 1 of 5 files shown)


src/snowflake/cli/_plugins/logs/commands.py
```diff
@@ -36,6 +36,11 @@ def get_logs(
         "--refresh",
         help="If set, the logs will be streamed with the given refresh time in seconds",
     ),
+    event_table: str = typer.Option(
```
sfc-gh-jwilkowski, line 39: won't that be?
```suggestion
    event_table: Optional[str] = typer.Option(
```

src/snowflake/cli/_plugins/logs/manager.py
```diff
@@ -82,7 +81,11 @@ def get_raw_logs(
         object_name: FQN = NameArgument,
         from_time: Optional[datetime] = None,
         to_time: Optional[datetime] = None,
+        event_table: Optional[str] = None,
     ) -> SnowflakeCursor:
+
+        table = event_table if event_table else "SNOWFLAKE.TELEMETRY.EVENTS"
```
sfc-gh-jwilkowski, line 87: hmmm, that feels like a step back in user friendliness. I don't know how often clients use different tables for logs than the default one, but asking the user to provide the table name (while we can do it) doesn't feel like great UX.

Is this change coming from different column numbers while fetching parameters? How about querying parameters as before and then using `result_scan` with last query's query_id/request_id to transform it into table and fetch exactly wha you need?

sfc-gh-jsikorski: But what do we get from the `result_scan`? It's either the default snowflake name, or the query will result in multiple results, as user can have different event tables for different objects. We can query for default value, but, according to the docs - it's set to this value.
But if user defines multiple event tables, how do we get the one he needs?

sfc-gh-jwilkowski (approved):

PR #2237, opened 2025-04-30
Add support for queries, result and abort commands

sfc-gh-pczajka:
### Pre-review checklist

### Changes description
Add support for `!queries`, `!result` and `!abort` commands for SnowSQL. They're meant to be used in `--output=TABLE` format and might malfunction when mixed with other commands in `--json` output format.

### DIFF
--- a/src/snowflake/cli/_plugins/sql/commands.py
+++ b/src/snowflake/cli/_plugins/sql/commands.py
@@ -29,7 +29,6 @@ from snowflake.cli.api.commands.snow_typer import SnowTyperFactory
 from snowflake.cli.api.commands.utils import parse_key_value_variables
 from snowflake.cli.api.output.types import (
     CommandResult,
-    MessageResult,
     MultipleResults,
     QueryResult,
 )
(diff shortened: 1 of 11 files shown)


src/snowflake/cli/_plugins/sql/commands.py
```diff
@@ -115,13 +114,13 @@ def execute_sql(
...
         # ends gracefully with no message for consistency with snowsql.
-        return MessageResult("")
+        sys.exit(0)
 
     if expected_results_cnt == 1:
         # evaluate the result to schedule async queries
         results = list(cursors)
         if not results:
-            return MessageResult("")
+            return sys.exit(0)
```
sfc-gh-astus, line 123: Why you changed to exit?

sfc-gh-pczajka: Otherwise `--format=json` works even worse:
```
➜ snow sql -q '!result 01bc0934-090f-d660-0001-c1be146444ae' --format=json
[
    {
        "42*6": 252
    }
]
{
    "message": ""
}
```

sfc-gh-pczajka: This helps with integration tests, but still doesn't work when you mix "normal" queries with commands.
Further solution would be to integrate `SnowSQLCommands` with cursors, but that would require too much work for now as commands are mostly designed to be used in inline / interactive mode. We can fix json format if there is a demand.

sfc-gh-astus (approved):

PR #2266, opened 2025-05-08
Snow 2059612 single transaction support

sfc-gh-mraba:
### Pre-review checklist

### Changes description
`--single-transaction` flag was added to `snow sql`. It can be used with `-q, -f` and `-i` flags to disable autocommit and wrap all the queries in `begin/commit` block

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -19,6 +19,7 @@
 ## Deprecations
 
 ## New additions
+* Added `--single-transaction` flag to `snow sql` Wraps BEGIN/COMMIT around statements to execute them as a single transaction, ensuring all commands complete successfully or no change is applied.
 
 ## Fixes and improvements
 
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -19,7 +19,7 @@
 ## Deprecations
 
 ## New additions
-* Added `--single-transaction` flag to `snow sql` Wraps BEGIN/COMMIT around statements to execute them as a single transaction, ensuring all commands complete successfully or no change is applied.
+* Added `--single-transaction` flag to `snow sql`. Wraps BEGIN/COMMIT around statements to execute them as a single transaction, ensuring all commands complete successfully or no change is applied.
 
 ## Fixes and improvements
 
--- a/src/snowflake/cli/_plugins/sql/commands.py
+++ b/src/snowflake/cli/_plugins/sql/commands.py
@@ -15,6 +15,7 @@
 from __future__ import annotations
 
 import sys
+from logging import getLogger
 from pathlib import Path
 from typing import List, Optional
 
@@ -27,12 +28,14 @@ from snowflake.cli.api.commands.flags import (
 from snowflake.cli.api.commands.overrideable_parameter import OverrideableOption
 from snowflake.cli.api.commands.snow_typer import SnowTyperFactory
 from snowflake.cli.api.commands.utils import parse_key_value_variables
+from snowflake.cli.api.exceptions import CliArgumentError
 from snowflake.cli.api.output.types import (
     CommandResult,
     MultipleResults,
     QueryResult,
 )
 
+logger = getLogger(__name__)
 # simple Typer with defaults because it won't become a command group as it contains only one command
 app = SnowTyperFactory()
 
@@ -74,6 +77,12 @@ def execute_sql(
         "--retain-comments",
         help="Retains comments in queries passed to Snowflake",
     ),
+    single_transaction: Optional[bool] = typer.Option(
+        False,
+        help="Connects with autocommit disabled. Wraps BEGIN/COMMIT around statements to execute them as a single transaction, ensuring all commands complete successfully or no change is applied.",
+        flag_value=False,
+        is_flag=True,
+    ),
     **options,
 ) -> CommandResult:
     """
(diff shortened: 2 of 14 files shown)


RELEASE-NOTES.md
```diff
@@ -19,6 +19,7 @@
 ## Deprecations
 
 ## New additions
+* Added `--single-transaction` flag to `snow sql` Wraps BEGIN/COMMIT around statements to execute them as a single transaction, ensuring all commands complete successfully or no change is applied.
```
sfc-gh-turbaszek, line 22:
```suggestion
* Added `--single-transaction` flag to `snow sql`. Wraps BEGIN/COMMIT around statements to execute them as a single transaction, ensuring all commands complete successfully or no change is applied.
```

src/snowflake/cli/_plugins/sql/commands.py
```diff
@@ -102,13 +112,30 @@ def execute_sql(
...
+    manager = SqlManager()
+
+    if single_transaction:
+        logger.debug("disabling AUTOCOMMIT")
+        manager.autocommit(False)
+        query = f"BEGIN; {query}"
+        if query.rstrip().endswith(";"):
+            query = f"{query} COMMIT;"
+        else:
+            query = f"{query}; COMMIT;"
```
sfc-gh-turbaszek, line 131:
```suggestion
        query = query.rstrip(";") + " COMMIT;"
```
WDYT?

src/snowflake/cli/_plugins/sql/commands.py
```diff
@@ -102,13 +112,30 @@ def execute_sql(
...
         sys.exit(0)
 
-    expected_results_cnt, cursors = SqlManager().execute(
-        query, files, std_in, data=data, retain_comments=retain_comments
+    manager = SqlManager()
+
+    if single_transaction:
+        logger.debug("disabling AUTOCOMMIT")
+        manager.autocommit(False)
+        query = f"BEGIN; {query}"
```
sfc-gh-turbaszek, line 127: Will this work with `BEGIN; BEGIN ... `?

src/snowflake/cli/api/sql_execution.py
```diff
@@ -279,6 +284,9 @@ def show_specific_object(
         )
         return show_obj_row
 
+    def autocommit(self, value: bool):
```
sfc-gh-turbaszek, line 287:
```suggestion
    def set_autocommit(self, value: bool):
```
WDYT? Eventually, maybe we just need `disable_autocommit`?

tests_integration/test_single_transaction_autocommit.py
```diff
@@ -0,0 +1,108 @@
...
+            "-q",
+            sql,
+        ]
+    )
+    assert result.exit_code == 0, result.output
+    assert (
+        result.json[1][0]["status"] == f"Table {tbl_name} successfully created."
+    ), result.json[1]
+    assert result.json[2][0]["COUNT(*)"] == 0, result.json[1]
+    assert result.json[5][0]["COUNT(*)"] == 2, result.json[1]
```
sfc-gh-pczajka, line 98:
```suggestion
    assert result.json[2][0]["COUNT(*)"] == 0, result.json[2]
    assert result.json[5][0]["COUNT(*)"] == 2, result.json[5]
```

tests_integration/test_single_transaction_autocommit.py
```diff
@@ -0,0 +1,108 @@
...
+        [
+            "sql",
+            "-q",
+            "show parameters like 'autocommit'",
+        ]
+    )
+    assert result.exit_code == 0, result.output
+    assert result.json[0]["key"] == "AUTOCOMMIT"
+    assert result.json[0]["default"] == "true"
+    assert result.json[0]["value"] == "true"
```
sfc-gh-pczajka, line 18: Those two tests could be merged using `parametrize`, wdyth?

```suggestion
@pytest.mark.integration
@pytest.mark.parametrize("autocommit", [True, False])
def test_autocommit(runner, autocommit):
    cmd = [
        "sql", "-q", "show parameters like 'autocommit'"
    ]
    if autocommit:
        cmd.append("--single-transaction")

    result = runner.invoke_with_connection_json(cmd)
    assert result.exit_code == 0, result.output
    assert result.json[0]["key"] == "AUTOCOMMIT"
    assert result.json[0]["default"] == "true"
    assert result.json[0]["value"] == str(autocommit).lower()
```

src/snowflake/cli/_plugins/sql/commands.py
sfc-gh-mraba: Yes.
If two BEGIN statements in a row are executed (within the same [scope](https://docs.snowflake.com/en/sql-reference/transactions.html#label-scoped-transactions)), the second one is ignored.

src/snowflake/cli/_plugins/sql/commands.py
sfc-gh-mraba: We need to keep in mind that query might end with white characters after `;`
```python
query = f"BEGIN; {query.rstrip().rstrip(';')}; COMMIT;"

tests_integration/test_single_transaction_autocommit.py
sfc-gh-mraba: I prefer to avoid conditional logic in tests.
Also with autocommit disabled we wrap the query in BEGIN/COMMIT so additional logic will be required to handle different statement count in result testing

sfc-gh-pczajka (approved):

PR #2300, opened 2025-05-15
Snow 2097403 add encryption to stage create

sfc-gh-pczajka:
### Pre-review checklist

### Changes description
Added `--encryption` flag to `snow stage create` command defining the type of encryption for all files on the stage.

### DIFF
--- a/src/snowflake/cli/_plugins/stage/commands.py
+++ b/src/snowflake/cli/_plugins/stage/commands.py
@@ -29,7 +29,7 @@ from snowflake.cli._plugins.stage.diff import (
     DiffResult,
     compute_stage_diff,
 )
-from snowflake.cli._plugins.stage.manager import StageManager
+from snowflake.cli._plugins.stage.manager import StageEncryption, StageManager
 from snowflake.cli._plugins.stage.utils import print_diff_to_console
 from snowflake.cli.api.cli_global_context import get_cli_context
 from snowflake.cli.api.commands.common import OnErrorType
@@ -150,11 +150,19 @@ def copy(
 
 
 @app.command("create", requires_connection=True)
-def stage_create(stage_name: FQN = StageNameArgument, **options) -> CommandResult:
+def stage_create(
+    stage_name: FQN = StageNameArgument,
+    encryption: StageEncryption = typer.Option(
+        StageEncryption.SNOWFLAKE_FULL.value,
+        "--encryption",
+        help="Type of encryption supported for all files stored on the stage.",
+    ),
+    **options,
+) -> CommandResult:
     """
     Creates a named stage if it does not already exist.
     """
-    cursor = StageManager().create(fqn=stage_name)
+    cursor = StageManager().create(fqn=stage_name, encryption=encryption)
     return SingleQueryResult(cursor)
 
 
--- a/src/snowflake/cli/_plugins/stage/manager.py
+++ b/src/snowflake/cli/_plugins/stage/manager.py
@@ -25,6 +25,7 @@ import time
 from collections import deque
 from contextlib import nullcontext
 from dataclasses import dataclass
+from enum import Enum
 from os import path
 from pathlib import Path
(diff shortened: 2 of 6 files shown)


src/snowflake/cli/_plugins/stage/commands.py
```diff
@@ -150,11 +150,19 @@ def copy(
 
 
 @app.command("create", requires_connection=True)
-def stage_create(stage_name: FQN = StageNameArgument, **options) -> CommandResult:
+def stage_create(
+    stage_name: FQN = StageNameArgument,
+    encryption: StageEncryption = typer.Option(
```
sfc-gh-jsikorski, line 155: Just for the sake of readability, i would suggest moving this to `StageEncryptionArgument`

src/snowflake/cli/_plugins/stage/manager.py
```diff
@@ -68,6 +69,11 @@
 STAGE_PATH_REGEX = rf"(?P<prefix>(@|{re.escape('snow://')}))?(?:(?P<first_qualifier>{VALID_IDENTIFIER_REGEX})\.)?(?:(?P<second_qualifier>{VALID_IDENTIFIER_REGEX})\.)?(?P<name>{VALID_IDENTIFIER_REGEX})/?(?P<directory>([^/]*/?)*)?"
 
 
+class StageEncryption(Enum):
```
sfc-gh-jsikorski, line 72: Maybe we could use StageEncryption from `snowflake.core` - this would make it easier to maintain. WDYT?

tests_integration/test_stage.py
```diff
@@ -18,6 +18,7 @@
 import tempfile
 import time
 from pathlib import Path
+from unittest import result
```
sfc-gh-jwilkowski, line 21: looks like wrong import :)

src/snowflake/cli/_plugins/stage/manager.py
sfc-gh-jwilkowski: or https://github.com/snowflakedb/snowflake-cli/blob/d6ecae383955c96bc458a6b66d3c3e5618e6b904/src/snowflake/cli/api/utils/python_api_utils.py#L1-L12

src/snowflake/cli/_plugins/stage/manager.py
```diff
@@ -515,10 +521,16 @@ def remove(
             return self.execute_query(f"remove {stage_path.path_for_sql()}")
 
     def create(
-        self, fqn: FQN, comment: Optional[str] = None, temporary: bool = False
+        self,
+        fqn: FQN,
+        comment: Optional[str] = None,
+        temporary: bool = False,
+        encryption: StageEncryption | None = None,
```
sfc-gh-jwilkowski, line 528: wouldn't the default be `SNOWFLAKE_FULL`?

sfc-gh-pczajka: This is what Snowflake defaults to when encryption is not provided, and adding not-None here would break a lot of unit mocks :D

tests_integration/test_stage.py
sfc-gh-pczajka: Whoops
Thanks for noticing :D

src/snowflake/cli/_plugins/stage/manager.py
sfc-gh-pczajka: That contains encryption types from both [internal and external](https://docs.snowflake.com/en/sql-reference/sql/create-stage) stage creation parameters - we would have to add parameters for aws/gcs/azure and validation as SQL does not return helpful message:
```
| 001062 (42601): 01bc6280-080f-fb12-0001-c1be156479e2: SQL compilation error:                                                                                                                                                                                                                                                                 │
│ Cannot set URL, credentials, or encryption key of an internal or temporary stage. 
```
which is far behind the feature scope

sfc-gh-mraba (approved):

PR #2341, opened 2025-05-27
Check brew version instead of pip

sfc-gh-pczajka:
### Pre-review checklist

### Changes description
As brew index takes longer to propagate the change, check brew version instead of pypi, revert to pypi if brew is not available.

### DIFF
--- a/src/snowflake/cli/_app/version_check.py
+++ b/src/snowflake/cli/_app/version_check.py
@@ -9,7 +9,8 @@ from snowflake.cli.api.cli_global_context import get_cli_context
 from snowflake.cli.api.secure_path import SecurePath
 from snowflake.connector.config_manager import CONFIG_MANAGER
 
-REPOSITORY_URL = "https://pypi.org/pypi/snowflake-cli/json"
+REPOSITORY_URL_PIP = "https://pypi.org/pypi/snowflake-cli/json"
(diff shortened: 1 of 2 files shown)


src/snowflake/cli/_app/version_check.py
```diff
@@ -48,12 +49,19 @@ def _save_latest_version(self, version: str):
...
+
+    @staticmethod
+    def _get_version_from_brew() -> str | None:
+        response = requests.get(REPOSITORY_URL_BREW, timeout=3)
+        response.raise_for_status()
+        return response.json().get("versions", {}).get("stable", None)
 
     def _update_latest_version(self) -> Version | None:
-        version = self._get_version_from_pypi()
+        # use brew version, revert to pypi if it is not available
```
sfc-gh-daniszewski, line 63: two small remarks:
- `fallback` sounds IMO better than `revert`, WDYT?
- let's add a short explanation why brew has priority

sfc-gh-daniszewski (commented):
almost LGTM :)

src/snowflake/cli/_app/version_check.py
sfc-gh-pczajka: Reworded :)

sfc-gh-daniszewski (approved):
now LGTM, thx!

PR #2352, opened 2025-06-02
Jw/snow 2131951 less freqent update messages

sfc-gh-jwilkowski:
### Pre-review checklist

### Changes description
...


src/snowflake/cli/_app/version_check.py
```diff
@@ -6,17 +6,37 @@
...
 from snowflake.cli.api.secure_path import SecurePath
 from snowflake.connector.config_manager import CONFIG_MANAGER
 
 REPOSITORY_URL_PIP = "https://pypi.org/pypi/snowflake-cli/json"
 REPOSITORY_URL_BREW = "https://formulae.brew.sh/api/formula/snowflake-cli.json"
 
+# How often to refresh the version cache (seconds)
+VERSION_CACHE_REFRESH_INTERVAL = 60 * 60  # 1 hour
+# How often to show the new version message (seconds)
+NEW_VERSION_MSG_INTERVAL = 60 * 60 * 24 * 7  # 1 week
```
sfc-gh-astus: `NEW_VERSION_MESSAGE_INTERVAL` WDYT?

src/snowflake/cli/_app/version_check.py
```diff
@@ -73,13 +116,29 @@ def _read_latest_version(self) -> Version | None:
...
-
         return self._update_latest_version()
 
     def get_last_version(self) -> Version | None:
         try:
             return self._read_latest_version()
         except:  # anything, this it not crucial feature
             return None
+
+    def was_warning_shown_recently(self) -> bool:
```
sfc-gh-pczajka: I'd separate this logic from `_VersionCache` - I imagine this class as a layer for reading/writing the cache file

src/snowflake/cli/_app/version_check.py
```diff
@@ -44,6 +65,28 @@ def _save_latest_version(self, version: str):
             _VersionCache._last_time: time.time(),
             _VersionCache._version: str(version),
         }
+        if self._cache_file.exists():
+            try:
+                old_data = json.loads(self._cache_file.read_text(file_size_limit_mb=1))
+                if _VersionCache._last_time_shown in old_data:
```
sfc-gh-pczajka, line 71: Will this ever be a case? Lifecycle of this class is read_file_if_exists -> update_data -> write_new_data_to_the_file

src/snowflake/cli/_app/version_check.py
sfc-gh-jwilkowski: What would you suggest instead? I agree that the original intent of _VersionCache was probably to not hit pypi too often, but our requirements have changed. If we were very strict, the info when upgrade message was last displayed could go to another file, but that seems to be adding more complexity if you ask me. After all _last_time_shown is also related to the functionality of displaying version upgrade message, so I'd rather have that in a single file instead of worrying about management of another file.

src/snowflake/cli/_app/version_check.py
sfc-gh-jwilkowski: done

sfc-gh-astus (approved):

src/snowflake/cli/_app/version_check.py
sfc-gh-pczajka: How about returning `last_time_shown: Optional[int]` and doing a comparison with a constant in `get_new_version_msg`?

sfc-gh-jwilkowski: ok, that sounds like a good idea 👌

sfc-gh-pczajka (approved):

PR #2363, opened 2025-06-04
Fix integration flaky test report mechanism

sfc-gh-astus:
### Pre-review checklist

### Changes description
Fix integration flaky test report mechanism

### DIFF
--- a/.github/workflows/test_e2e.yaml
+++ b/.github/workflows/test_e2e.yaml
@@ -21,6 +21,7 @@ concurrency:
 
 permissions:
   contents: none
+  issues: write
 
 jobs:
   define-matrix:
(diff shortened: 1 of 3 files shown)


tests_common/deflake.py
```diff
@@ -270,11 +270,6 @@ def update_issue_body(self, issue: dict, root: Path, test: TestResult) -> None:
 
     def comment_on_issue(self, issue: dict, test: TestResult) -> dict:
         body = "+1"
-        for phase in PHASES:
-            if getattr(test, phase).outcome in (FAILED, FLAKY):
-                body += f"\n```python\n{getattr(test, phase).crash.message}\n```\n"
-                break
```
sfc-gh-daniszewski: can you elaborate a bit on why we shrunken body here?

sfc-gh-astus: We can print tokens and passwords here, it is not safe.

sfc-gh-astus: My bad, checked what we have in integration workflow and it looks the same like this, so reverted the code.

PR #2372, opened 2025-06-06
Fixed failing snow sql command when snowflake.yml is invalid

sfc-gh-astus:
### Pre-review checklist

### Changes description
Fixed failing snow sql command when snowflake.yml is invalid


sfc-gh-jwilkowski (approved):

src/snowflake/cli/api/rendering/sql_templates.py
```diff
@@ -95,7 +95,11 @@ def snowflake_sql_jinja_render(content: str, data: Dict | None = None) -> str:
                 f"{reserved_key} in user defined data. The `{reserved_key}` variable is reserved for CLI usage."
             )
 
-    context_data = get_cli_context().template_context
+    try:
+        context_data = get_cli_context().template_context
+    except Exception:
+        cli_console.warning("Invalid snowflake.yml file.")
+        context_data = {}
```
sfc-gh-daniszewski: Umm we're doing this change in `snowflake_sql_jinja_render` which makes me wonder if the current behaviour of failing the `snow sql` command isn't actually good. If we'd have jinja templating in the SQL being executed and there source of template values is broken, then I'd expect it to fail.

sfc-gh-daniszewski (changes requested):
I think we might have omitted some aspects of this JIRA ticket, let's discuss on Monday

src/snowflake/cli/api/rendering/sql_templates.py
sfc-gh-astus: Current behaviour not allow to execute `snow sql -q "select 1"`, query doesn't use snowflake.yml, but we fail because of invalid snowflake.yml.

sfc-gh-daniszewski: But the code doesn't know that query has no params - I don't see check if query has templating directives. Yet if query has templating, but yaml is broken then we will render query with empty context and it can result in either failed query, or - worse - unpredicted query.

sfc-gh-astus: Right. I added new proposition.

RELEASE-NOTES.md
```diff
@@ -22,6 +22,7 @@
 * Add support for passing OAuth token via `--token` flag.
 
 ## Fixes and improvements
+* Fixed failing snow sql command when snowflake.yml is invalid
```
sfc-gh-daniszewski: How about
> Fixed failing snow sql command when snowflake.yml is invalid and query has no templating.

?

src/snowflake/cli/api/rendering/sql_templates.py
```diff
@@ -94,13 +94,19 @@ def snowflake_sql_jinja_render(content: str, data: Dict | None = None) -> str:
...
+        CLICounterField.SQL_TEMPLATES, int(has_templates)
     )
 
-    return env.from_string(content).render(context_data)
+    if has_templates:
+        try:
+            context_data = get_cli_context().template_context
+        except Exception as e:
+            raise CliArgumentError(f"Failed to read snowflake.yml file: {e}")
+        context_data.update(data)
```
sfc-gh-daniszewski, line 109: is `context_data` reachable here? it was defined in try block.

tests/sql/test_sql.py
```diff
@@ -523,3 +523,39 @@ def test_comments_are_handled_correctly_from_query(
...
+    snowflake_yml_path.write_text(
+        """
+        invalid
+        snowflake
+        yml
+        """
+    )
+    result = runner.invoke(["sql", "-q", "select 1"])
+    assert result.exit_code == 0, result.output
+    assert "\n" == result.output
```
sfc-gh-daniszewski: shouldn't we expect `1` as an output?

src/snowflake/cli/api/rendering/sql_templates.py
sfc-gh-astus: It is.

tests/sql/test_sql.py
sfc-gh-astus: No, the query doesn't contains markups so I skip the snowflake.yml load.

sfc-gh-daniszewski: right, but you query `select 1` so why result.output is empty?

src/snowflake/cli/api/rendering/sql_templates.py
sfc-gh-daniszewski: 🤦 my brain neglected that in case of an exception we will raise CliArgError and not go any further. All good! :)

RELEASE-NOTES.md
sfc-gh-astus: Changed

PR #2375, opened 2025-06-09
Snow 2122462 add flags to configure templating syntax in sql command

sfc-gh-pczajka:
### Pre-review checklist

### Changes description
* Add `--template-syntax` flag for `snow sql` command, configuring which template syntax should be used to resolve variables:
  * standard (`<% ... %>`), enabled by default
  * legacy (`&{ ... }`), enabled by default
  * jinja (new) (`{{ ... }}`), disabled by default

It is possible to choose multiple options, for example `--template-syntax all` or `--template-syntax standard --temaplte-syntax jinja`, or disable templating at all (`--template-syntax none`)


RELEASE-NOTES.md
```diff
@@ -22,6 +22,10 @@
 * Add support for passing OAuth token via `--token` flag.
 * Added feature to allow suppressing the new version warning message in the Snowflake CLI, configurable via a configuration file or environment variable.
 * Add support for `--format=CSV`
+* Add `--template-syntax` flag for `snow sql` command, configuring which template syntax should be used to resolve variables:
```
sfc-gh-daniszewski: that might be personal, but to me `--template-syntax` implies I should pick one while there are actually two enabled by default and generally we expect a List here. Additionally we support NONE or ALL which logically isn't a list. Not blocker, just my thoughts.

src/snowflake/cli/api/rendering/sql_templates.py
```diff
@@ -36,6 +38,16 @@
 RESERVED_KEYS = [CONTEXT_KEY, FUNCTION_KEY]
 
 
+@dataclass
+class SQLTemplateSyntaxConfig:
+    """Class defining which syntax should be used for the template resolution.
+    Jinja syntax is not recommended and should be disabled by default."""
```
sfc-gh-daniszewski: do we envision any other templating syntax to be added? This class and its usages can be nightmarish to maintain if list gets bigger

src/snowflake/cli/api/rendering/sql_templates.py
```diff
@@ -86,27 +106,53 @@ def choose_sql_jinja_env_based_on_template_syntax(
...
 
-def snowflake_sql_jinja_render(content: str, data: Dict | None = None) -> str:
+def snowflake_sql_jinja_render(
+    content: str,
+    enabled_syntax_config: SQLTemplateSyntaxConfig,
+    data: Dict | None = None,
+) -> str:
+    """
+    If both legacy and standard syntax are enabled, CLI chooses one basing on provided content.
+    If jinja syntax is enabled, it is resolved after standard and legacy syntax.
```
sfc-gh-daniszewski, line 116: is it anyhow useful to have three coexistent templating mechanisms at the same time? I wonder if there are real use-cases that someone uses i.e. jinja after legacy templating.

sfc-gh-daniszewski (commented):
Few small comments

RELEASE-NOTES.md
sfc-gh-pczajka: Changed to `--enbled-templating`, IMO sounds better :D

src/snowflake/cli/api/rendering/sql_templates.py
sfc-gh-pczajka: I've added a comment

src/snowflake/cli/api/rendering/sql_templates.py
sfc-gh-pczajka: We don't expect any - we have our own, SnowSQL and now I'm adding jinja requested by users for testing EXECUTE IMMEDIATE scripts locally

sfc-gh-daniszewski (approved):
thx

PR #2378, opened 2025-06-09
SNOW-2131889 Add CSV output format

sfc-gh-daniszewski:
### Pre-review checklist

### Changes description
Allow specifying `--format=CSV`


sfc-gh-astus (dismissed):

src/snowflake/cli/_app/printing.py
```diff
@@ -94,28 +95,60 @@ def _print_multiple_table_results(obj: CollectionResult):
...
+        writer.writerow(data)
+    elif isinstance(data, list):
+        header_written = False
+        writer = csv.DictWriter(sys.stdout, None, lineterminator="\n")  # type: ignore
+        for entry in data:
+            if not header_written:
+                writer.fieldnames = [*entry]
+                writer.writeheader()
+                header_written = True
+            writer.writerow(entry)
```
sfc-gh-jwilkowski: what do you think about splitting header and rows writing into separate steps? Something closer to
```suggestion
        if not data:
            return
        writer.fieldnames = [*data[0]]
        writer.writeheader()
        for entry in data[1:]:
            writer.writerow(entry)
```

sfc-gh-jwilkowski: Another question is if this will work nicely with StreamResults? By looking at https://github.com/snowflakedb/snowflake-cli/pull/2378/files#diff-caebdc6d163d129f10b8516c203cd9b51cdc51e268caa215be98051c43fb3d9bR118-R119 it seems that we'll be repeating headers for every chunk of the stream, which imho should be skipped?

But after looking at tests that we might want to support different streams from a single command I won't push for this change.

tests/output/test_printing.py
```diff
@@ -373,9 +381,91 @@ def g():
...
+
 @pytest.fixture
 def _empty_cursor(mock_cursor):
     return lambda: mock_cursor(
         columns=["string", "number", "array", "object", "date"],
         rows=[],
     )
+
+
+@pytest.fixture()
```
sfc-gh-jwilkowski: This might be a personal thing, but I believe that in most tests fixtures live at the top of the file, after imports and before tests

src/snowflake/cli/_app/printing.py
sfc-gh-daniszewski: I refactored it a bit differently, but in the same fashion of header first, then data.
As for repeating header - yes, we don't assume all "entries" in StreamResults be the same kind - TABLE output also duplicates the headers.

tests/output/test_printing.py
sfc-gh-daniszewski: I wonder if there is any standard approach for this. I've seen fixtures both on top and bottom in other projects - here I saw one fixture on the bottom so I added my own there as well :)

PR #2390, opened 2025-06-12
Snow 2150178 fix app teardown test

sfc-gh-pczajka:
### Pre-review checklist

### Changes description
Fix `app teardown` test after release channels got enabled by default

### DIFF
--- a/tests_integration/nativeapp/test_teardown.py
+++ b/tests_integration/nativeapp/test_teardown.py
@@ -227,22 +227,31 @@ def test_nativeapp_teardown_pkg_versions(
 
     with nativeapp_project_directory(test_project):
         result = runner.invoke_with_connection(["app", "version", "create", "v1"])
-        assert result.exit_code == 0
+        assert result.exit_code == 0, result.output
+        result = runner.invoke_with_connection(
+            ["app", "release-channel", "add-version", "--version", "v1", "DEFAULT"]
+        )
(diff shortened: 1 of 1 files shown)


tests_integration/nativeapp/test_teardown.py
```diff
@@ -227,22 +227,31 @@ def test_nativeapp_teardown_pkg_versions(
 
     with nativeapp_project_directory(test_project):
         result = runner.invoke_with_connection(["app", "version", "create", "v1"])
-        assert result.exit_code == 0
+        assert result.exit_code == 0, result.output
 
         # when setting a release directive, we will not have the ability to drop the version later
         if default_release_directive:
+            result = runner.invoke_with_connection(
```
sfc-gh-melnacouzi, line 234: any reason why not just do `snow app publish --version v1 --patch 0 ? This will replace both of these commands.

sfc-gh-melnacouzi: You can even run `snow app publich --version v1 --patch 0 --create-version` and it will replace all 3 commands ;)

sfc-gh-pczajka: Fixed

I went with the fist option (without `--create-version`), as it should be published only when [default_release_directive] parameter is `True` :)

sfc-gh-melnacouzi (approved):
lgtm - thanks for the fix.

PR #2421, opened 2025-06-17
feat: [SNOW-2154833] implement project drop-version command

sfc-gh-jwilkowski:
### Pre-review checklist

### Changes description
https://snowflakecomputing.atlassian.net/browse/SNOW-2154833

### DIFF
--- a/src/snowflake/cli/_plugins/project/commands.py
+++ b/src/snowflake/cli/_plugins/project/commands.py
@@ -25,6 +25,7 @@ from snowflake.cli._plugins.project.project_entity_model import (
 from snowflake.cli.api.cli_global_context import get_cli_context
 from snowflake.cli.api.commands.decorators import with_project_definition
 from snowflake.cli.api.commands.flags import (
+    IfExistsOption,
     IfNotExistsOption,
     OverrideableOption,
     PruneOption,
@@ -248,3 +249,24 @@ def list_versions(
     pm = ProjectManager()
     results = pm.list_versions(project_name=identifier)
     return QueryResult(results)
+
+
+@app.command(requires_connection=True)
+def drop_version(
+    identifier: FQN = project_identifier,
+    version_name: str = typer.Argument(help="Name or alias of the version to drop"),
+    if_exists: bool = IfExistsOption(help="Do nothing if the version does not exist."),
+    **options,
+):
+    """
+    Drops a version from the project.
+    """
+    pm = ProjectManager()
+    pm.drop_version(
+        project_name=identifier,
+        version_name=version_name,
+        if_exists=if_exists,
+    )
+    return MessageResult(
+        f"Version '{version_name}' dropped from project '{identifier}'"
+    )
--- a/src/snowflake/cli/_plugins/project/commands.py
+++ b/src/snowflake/cli/_plugins/project/commands.py
@@ -254,7 +254,9 @@ def list_versions(
 @app.command(requires_connection=True)
 def drop_version(
     identifier: FQN = project_identifier,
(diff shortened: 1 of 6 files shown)


src/snowflake/cli/_plugins/project/manager.py
```diff
@@ -72,7 +72,7 @@ def _create_version(
         stage_path = StagePath.from_stage_str(from_stage)
         query = f"ALTER PROJECT {project_name.identifier} ADD VERSION"
         if alias:
-            query += f' IF NOT EXISTS "{alias}"'
+            query += f" IF NOT EXISTS {alias}"
```
sfc-gh-jwilkowski, line 79: Interestingly those double quotes were causing issues with identifying versions

sfc-gh-rsawicki (approved):
lgtm

src/snowflake/cli/_plugins/project/commands.py
```diff
@@ -248,3 +249,34 @@ def list_versions(
...
+        show_default=False,
+    ),
+    if_exists: bool = IfExistsOption(help="Do nothing if the version does not exist."),
+    **options,
+):
+    """
+    Drops a version from the project.
+    """
+    # Detect potential shell expansion issues
+    if version_name and version_name.upper() == "VERSION":
```
sfc-gh-daniszewski, line 268: `$1` in bash is first positional argument passed to command, isn't it possible that it would expand to something unpredictable?

src/snowflake/cli/_plugins/project/commands.py
```diff
@@ -248,3 +249,34 @@ def list_versions(
...
+        cli_console.warning(
+            f"Version name '{version_name}' might be truncated due to shell expansion. "
+            f"If you meant to use a version like 'VERSION$1', try using single quotes: 'VERSION$1'."
+        )
+
+    pm = ProjectManager()
+    pm.drop_version(
```
sfc-gh-daniszewski, line 275: even if we detect expansion issues, we still do DROP. Probably it will fail, but is it possible that expansion will actually produce something droppable and we'll remove something we didn't want?

src/snowflake/cli/_plugins/project/commands.py
sfc-gh-jwilkowski: Yes, it's possible but with this warning I wanted to handle the (imo) most probable issue: default version names set by snowflake are `VERSION$x`, so users are most likely to try to delete using such version format and that's covered here. It is also possible that some users would create version with alias `VERSION` (which would cause a false positive warning).

Thinking about the alternatives, I could also warn whenever the version starts with `VERSION` but that might be too eager to show warnings, so I discarded it

src/snowflake/cli/_plugins/project/commands.py
sfc-gh-jwilkowski: Similar to my previous answer, I wanted to offer guidance for the most probable issue I can see here. If that won't change then we're limiting possible scenarios to: user wanted to delete `VERSION$2`, this got shortened to just `VERSION`, we showed the warning and tried to delete a version called `VERSION`. `VERSION` is not a version name that can be produced by Snowflake, so somebody would have to use `VERSION` as an alias, which seem to be rather odd.

Anyway, this is just me trying to image what is likely to succeed/fail, do you have any better ideas to guide users in here? This might become a more common issue, since this versioning convention will be shared by other entities as well

sfc-gh-daniszewski: OK, I think I'm convinced that possible blast radius is very narrow, let's go with it.

sfc-gh-daniszewski (approved):

PR #2448, opened 2025-06-25
SNOW-2162936 Add `--refresh` option to `snow stage copy`

### LINKED ISSUE #2436: SNOW-2162936: Refresh Stage after Copy
### SnowCLI version

Snowflake CLI version: 3.9.1

### Python version

Python 3.12.10

### Platform

macOS-15.5-arm64-arm-64bit

### What happened

When do `snow stage copy` the files get successfully uploaded but then I see Stage's metadata never get refreshed with copied files. I have to manually run `alter stage <stage_name> refresh` to trigger the metadata update and in-turn to picked up by downstream objects like `streams`. 

Note: When do upload files via Load Data in Snowsight you will see refresh happening automatically.

### Console output

```Shell

```

### How to reproduce

- Create stage 
- Copy files to stage 
- Create stream on the stage 
- When you select * from stream, no results.

Now run the command `alter stage <stage name> refresh` and then do `select * from stream` , you will see the added files.

sfc-gh-daniszewski:
### Pre-review checklist

### Changes description
Adding `--refresh` to `stage copy` to reload associated directory table.
Also adding `--enable-directory` to `stage create` to enable directory support for the stage

FIxes #2436

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -17,6 +17,15 @@
 ## Backward incompatibility
 
 ## Deprecations
+
+## New additions
+
+## Fixes and improvements
+
+
+# v3.10.0
+
+## Deprecations
 * Snowpark processor in Native Apps.
 
 ## New additions
@@ -29,8 +38,7 @@
   * jinja (new) (`{{ ... }}`), disabled by default
 * Added packages alias for artifact_repository_packages in the snowflake.yml schema.
 * Added support for copying files between named stages using `snow stage copy @src_stage @dst_stage` command.
-* Added `--enable-directory` flag for `snow stage create` to enable directory table on newly created stage.
-* Added `--refresh` flag for `snow stage copy` to refresh content of directory table associated with stage.
+* DBT commands: `deploy`, `execute` and `list` are now available in preview.
 
 ## Fixes and improvements
 * Fixed failing snow sql command when snowflake.yml is invalid and query has no templating.
--- a/scripts/packaging/delete_mac_certs.sh
+++ b/scripts/packaging/delete_mac_certs.sh
@@ -1,0 +1,13 @@
+#!/bin/bash -e -x
+
+# Set the Keychain path where the certificates will be imported
+KEYCHAIN_PATH=$HOME/Library/Keychains/login.keychain-db
+
+# Delete the Developer ID Installer certificate from the keychain
+security delete-certificate -c "Developer ID Installer: Snowflake Computing INC. (W4NT6CRQ7U)" "$KEYCHAIN_PATH"
+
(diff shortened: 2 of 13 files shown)


src/snowflake/cli/_plugins/stage/commands.py
```diff
@@ -282,4 +288,10 @@ def _put(
             parallel=parallel,
             auto_compress=auto_compress,
         )
-        return QueryResult(cursor)
+        output = QueryResult(cursor)
+
+    if refresh:
```
sfc-gh-jwilkowski, line 301: I'm wondering if it wouldn't be good to provide any info to the user that the refreshing happened. We could add it to the result, but that would feel a bit awkward... But what do you think about at least using CliConsole.message or logger.info?

sfc-gh-jwilkowski (commented):
Left one suggestion

src/snowflake/cli/_plugins/stage/commands.py
sfc-gh-daniszewski: Good idea, I added `log.info` in manager

tests_integration/test_stage.py
```diff
@@ -866,3 +915,15 @@ def test_create_encryption(runner, test_database, encryption):
...
+@pytest.mark.integration
+def test_create_directory_option(runner, test_database):
+    stage_name = "stage_with_directory"
+    result = runner.invoke_with_connection_json(
+        ["stage", "create", stage_name, "--enable-directory"]
+    )
+    assert result.exit_code == 0, result.output
+    assert result.json == {
+        "status": f"Stage area {stage_name.upper()} successfully created."
+    }
```
sfc-gh-jwilkowski, line 929: I'd like to ask for another assertion that would confirm that this stage has directory enabled

sfc-gh-mraba: Not sure if this is the place to test the internals.
Either the query was successful and the stage was created or it failed.

sfc-gh-jwilkowski: Imho it's not an internal - test scenario is about creating a stage with enabled directory, if somebody would delete only those lines https://github.com/snowflakedb/snowflake-cli/pull/2448/files#diff-a524d08d9ca9b1a20fbcfe4770bee955bd9bd33d3ecdec0fb83c318b00dc1000R557-R558 then current test would be still passing and imo it would wrong

sfc-gh-daniszewski: > Either the query was successful and the stage was created or it failed.

That was my initial thought as well, but I agree that this is not internal - it's wether or not it was actually enabled.
Added.

sfc-gh-jwilkowski (approved):

PR #2458, opened 2025-06-26
SNOW-2155042 fix json format for nested objects

### LINKED ISSUE #2393: `snow sql --format json` should represent OBJECT and ARRAY values as JSON data
### Description

When using `--format json`, `snow sql` would format structured and semi-structured data as nested objects/arrays, rather than strings containing JSON. 

For instance, the command:

```
snow sql --format json -q "
SELECT 
  ARRAY_CONSTRUCT(1, 2, 3) as a,
  OBJECT_CONSTRUCT('foo', 'XXXX', 'bar', 42) as o
"
```

would return:

```
[
    {
        "A": [
          1, 
          2, 
          3
        ],
        "O": {
            "foo": "XXXX",
            "bar": 42
        }
    }
]
```

rather than:

```json
[
    {
        "A": "[\n  1,\n  2,\n  3\n]",
        "O": "{\n  \"foo\": \"XXXX\",\n  \"bar\": 42\n}"
    }
]
```

### Context

This would make it much easier to fetch structured data from Snowflake.

sfc-gh-daniszewski:
### Pre-review checklist

### Changes description

Print nested JSONs as objects/arrays, not strings.
Fixes #2393

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -43,7 +43,7 @@
 ## Fixes and improvements
 * Fixed failing snow sql command when snowflake.yml is invalid and query has no templating.
 * Fix JSON serialization for `Decimal`, `time` and `binary`.
-* Nested JSONs are printed as JSONs, not strings.
+* Added `--format=JSON_EXT` option to return JSON objects as proper JSON structures rather than strings.
 * Refactored Streamlit app deployment (using `FROM <stage>` syntax); removed deprecated Streamlit features
 
 
--- a/src/snowflake/cli/api/cli_global_context.py
+++ b/src/snowflake/cli/api/cli_global_context.py
@@ -46,7 +46,6 @@ class _CliGlobalContextManager:
     silent: bool = False
     verbose: bool = False
     experimental: bool = False
-    expand_json: bool = False
     enable_tracebacks: bool = True
     is_repl: bool = False
 
@@ -177,10 +176,6 @@ class _CliGlobalContextAccess:
         return self._manager.experimental
 
     @property
-    def expand_json(self) -> bool:
-        return self._manager.expand_json
-
-    @property
     def project_definition(self) -> ProjectDefinition | None:
         return self._manager.project_definition
 
@@ -202,7 +197,7 @@ class _CliGlobalContextAccess:
     def _should_force_mute_intermediate_output(self) -> bool:
         """Computes whether cli_console output should be muted."""
         return (
-            self._manager.output_format.is_json()
+            self._manager.output_format.is_json
             or self._manager.output_format == OutputFormat.CSV
         )
 
--- a/src/snowflake/cli/api/commands/flags.py
+++ b/src/snowflake/cli/api/commands/flags.py
@@ -446,14 +446,6 @@ OutputFormatOption = typer.Option(
     rich_help_panel=_CLI_BEHAVIOUR,
 )
 
-ExpandJsonOption = typer.Option(
-    False,
-    "--expand-json",
-    help="Expand JSON strings in query results into structured JSON objects instead of keeping them as strings.",
-    callback=_context_callback("expand_json"),
-    is_flag=True,
-    rich_help_panel=_CLI_BEHAVIOUR,
-)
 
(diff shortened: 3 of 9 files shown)


src/snowflake/cli/api/output/types.py
```diff
@@ -69,13 +69,49 @@ def result(self):
...
+        - ARRAY (type code 10): Can contain JSON arrays
+        - OBJECT (type code 9): Can contain JSON objects
+        """
+        processed_row = {}
+        for i, (column_name, value) in enumerate(row_dict.items()):
+            # Check if this column can contain JSON data
+            if i < len(self.column_types) and self.column_types[i] in (
+                5,
+                9,
+                10,
```
sfc-gh-mraba: Maybe we should introduce an enum or consts in favour of values?
wdyt?

tests_integration/test_snowpark.py
```diff
@@ -280,13 +278,13 @@ def test_snowpark_flow(
         _snowpark_test_steps.snowpark_execute_should_return_expected_value(
             object_type="procedure",
             identifier="hello_procedure('foo')",
-            expected_value='"Hello foo"',
+            expected_value="Hello foo",
```
sfc-gh-mraba: Are we sure this won't be a BC?

sfc-gh-daniszewski: it is! I forgot to hide this behind feature flag 🤦

RELEASE-NOTES.md
```diff
@@ -43,6 +43,7 @@
 ## Fixes and improvements
 * Fixed failing snow sql command when snowflake.yml is invalid and query has no templating.
 * Fix JSON serialization for `Decimal`, `time` and `binary`.
+* Nested JSONs are printed as JSONs, not strings.
```
sfc-gh-jwilkowski: Since it has become a new format, maybe a different wording would be better? 
New additions:
 * Introduced new format option JSON_EXT which prints JSONs as JSONs, not strings

src/snowflake/cli/api/cli_global_context.py
```diff
@@ -46,6 +46,7 @@ class _CliGlobalContextManager:
     silent: bool = False
     verbose: bool = False
     experimental: bool = False
+    expand_json: bool = False
```
sfc-gh-jwilkowski: Shouldn't this be removed since we went with new format type?

src/snowflake/cli/api/commands/flags.py
```diff
@@ -446,6 +446,15 @@ def _diag_log_allowlist_path_callback(path: str):
     rich_help_panel=_CLI_BEHAVIOUR,
 )
 
+ExpandJsonOption = typer.Option(
```
sfc-gh-jwilkowski: Same, shouldn't this be removed?

src/snowflake/cli/api/cli_global_context.py
```diff
@@ -175,6 +176,10 @@ def verbose(self) -> bool:
     def experimental(self) -> bool:
         return self._manager.experimental
 
+    @property
```
sfc-gh-jwilkowski: and here

src/snowflake/cli/api/output/formats.py
```diff
@@ -18,4 +18,8 @@
 class OutputFormat(Enum):
     TABLE = "TABLE"
     JSON = "JSON"
+    JSON_EXT = "JSON_EXT"
     CSV = "CSV"
+
+    def is_json(self) -> bool:
```
sfc-gh-mraba, line 25:
```suggestion
    @property
    def is_json(self) -> bool:
```

RELEASE-NOTES.md
sfc-gh-daniszewski: correct, updated

src/snowflake/cli/api/cli_global_context.py
sfc-gh-daniszewski: yes, thx for pointing this

sfc-gh-jwilkowski (approved):

PR #2460, opened 2025-06-26
add sleep before checking async query results

sfc-gh-pczajka:
### Pre-review checklist

### Changes description
Wait short amount of time before checking async query results to avoid false-negative test results.
Reason: very rarely the result of the last async call is missing.

### DIFF
--- a/tests_integration/sql/test_async_queries.py
+++ b/tests_integration/sql/test_async_queries.py
@@ -1,4 +1,5 @@
 import pytest
+from time import sleep
 
 
 class TableForTesting:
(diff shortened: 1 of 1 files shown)


tests_integration/sql/test_async_queries.py
```diff
@@ -15,6 +16,9 @@ def add_value_query(self, value):
         return f"INSERT INTO {self.name} VALUES ('{value}')"
 
     def get_contents(self):
+        # as very rarely the result of the last async query is not yet visible in the test result,
+        # we wait short amount of time to avoid false-negative test results
+        sleep(0.1)
```
sfc-gh-mraba: Maybe we could add some incrementing backoff policy ? 0.1, 0.3, 0.9, Fail
WDYT

sfc-gh-pczajka: Good idea!

PR #2487, opened 2025-07-14
feat: [SNOW-2155083] spcs service: support auto-suspend-secs

sfc-gh-jwilkowski:
Added entry for new auto_suspend_secs parameter functionality in SPCS service commands (deploy, set, unset) that allows configuring automatic service suspension after inactivity period.

### Pre-review checklist

### Changes description
...


tests/spcs/test_services.py
```diff
@@ -1531,6 +1531,7 @@ def test_set_property(mock_execute_query):
     max_instances = 3
     query_warehouse = "test_warehouse"
     auto_resume = False
+    auto_suspend_secs = 600
```
sfc-gh-mraba, line 1534: Maybe this could be defined instead of magic number?

tests/spcs/test_services.py
```diff
@@ -1707,10 +1723,47 @@ def test_unset_property_no_properties_cli(mock_unset, runner):
...
         auto_resume=False,
+        auto_suspend_secs=False,
         comment=False,
     )
 
 
+@patch(EXECUTE_QUERY)
+def test_set_property_auto_suspend_secs_only(mock_execute_query):
+    service_name = "test_service"
+    auto_suspend_secs = 300
```
sfc-gh-mraba, line 1734: Here is another magic value. Why not 600 like earlier?

tests/spcs/test_services.py
sfc-gh-jwilkowski: None of those matter really, it's just an integer. This variable is local to this test only, I wouldn't see any point in extracting values of [service_name, min_instances, max_instances](https://github.com/snowflakedb/snowflake-cli/pull/2487/files/339a050854d9d126ef9aed2a4b3867591c42cb6f#diff-1a5e47d95ef7040973f1be31864ff7df482b551a85667681e0b4d624765e5b4bR1529-R1531) etc either

tests/spcs/test_services.py
sfc-gh-jwilkowski: as above, the exact value doesn't really matter, it just needs to be some positive integer

PR #2498, opened 2025-07-21
NO-SNOW Fix #595 - Add `snow connection remove` command

### LINKED ISSUE #595: SNOW-988145: Unable to delete or override existing connection in SnowCLI
### Description

When creating a connection using `snow connection add` it fails if the connection already exists. Please consider the following two features:
1. Add --override the will override a connection if already exists
2. Add `snow connection remove` that will remove a connection (and will return 4, not error, if the connection doesn't exist)

### Context

_No response_

sfc-gh-ggurlaskie:
### Pre-review checklist

### Changes description
A proposal to add a `snow connection remove` command discussed in #595.

Use case:
If a mistake or update is needed in `snow connection add`, the current solution seems to be for the CLI user to locate and edit the config file by hand.

Change:
Add a `snow connection remove` command to allow users to remove connection entries from their config programmatically.

Example Usage:
```
$  snow connection remove -n conn1
Removed connection conn1 from <...>/Library/Application Support/snowflake/config.toml.

$  snow connection remove -n default1
Removed connection default1 from <...>/Library/Application Support/snowflake/config.toml. It was the default connection, so default connection is now unset.

$  snow connection remove -n conn1
Usage: snow connection remove [OPTIONS]
Try 'snow connection remove --help' for help.
╭─ Error ────────────────────────────────────────────────────────────────╮
│ Connection conn1 does not exist.                                       │
╰────────────────────────────────────────────────────────────────────────╯
```


src/snowflake/cli/api/utils/dict_utils.py
```diff
@@ -71,3 +71,47 @@ def traverse(
...
+    key = keys[-1]
+    if key not in path_to_target[-1]:
+        return False
+
+    del path_to_target[-1][key]
+
+    # Remove parents that become empty
+    while len(path_to_target) > 1:
+        child = path_to_target.pop()
+        parent = path_to_target[-1]
```
sfc-gh-mraba: Would you consider adding const/var describing what `-1` is ie `KEY_TO_REMOVE` or unpacking the keys into 
```python
parent_keys, key_to_remove = keys[:-1], keys[-1]
```

for better readability?

sfc-gh-ggurlaskie: I agree, I think really the mutations make it hard to read. I pushed a new iteration that removed some duplicated code and should be more readable.

sfc-gh-mraba (dismissed):

sfc-gh-mraba (approved):

sfc-gh-ggurlaskie:
Fixed in #2517

PR #2511, opened 2025-07-28
SNOW-2229492 and SNOW-2229493 - cleanup versions in DCM

sfc-gh-daniszewski:
### Pre-review checklist

### Changes description
This PR addresses the following:
* Removing `snow dcm add-version`
* * `--prune` and `--alias` logic moved to `dcm plan` and `dcm deploy`
* Removing `--no-version` flag on `dcm create` and removed creating initial version whatsoever

### DIFF
--- a/src/snowflake/cli/_plugins/dcm/commands.py
+++ b/src/snowflake/cli/_plugins/dcm/commands.py
@@ -11,10 +11,11 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import logging
 from typing import List, Optional
 
 import typer
+from click import UsageError
 from snowflake.cli._plugins.dcm.dcm_project_entity_model import (
     DCMProjectEntityModel,
 )
@@ -22,6 +23,7 @@ from snowflake.cli._plugins.dcm.manager import DCMProjectManager
 from snowflake.cli._plugins.object.command_aliases import add_object_command_aliases
 from snowflake.cli._plugins.object.commands import scope_option
 from snowflake.cli._plugins.object.manager import ObjectManager
+from snowflake.cli.api.artifacts.upload import sync_artifacts_with_stage
 from snowflake.cli.api.cli_global_context import get_cli_context
 from snowflake.cli.api.commands.decorators import with_project_definition
 from snowflake.cli.api.commands.flags import (
@@ -37,7 +39,7 @@ from snowflake.cli.api.commands.snow_typer import SnowTyperFactory
 from snowflake.cli.api.commands.utils import get_entity_for_operation
 from snowflake.cli.api.console.console import cli_console
 from snowflake.cli.api.constants import ObjectType
-from snowflake.cli.api.exceptions import CliError
+from snowflake.cli.api.exceptions import CliError, NoProjectDefinitionError
 from snowflake.cli.api.feature_flags import FeatureFlag
 from snowflake.cli.api.identifiers import FQN
 from snowflake.cli.api.output.types import (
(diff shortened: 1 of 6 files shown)


src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -91,52 +87,79 @@
...
+        "--alias",
+        help="Alias for the deployment.",
+        show_default=False,
+    ),
+    prune: bool = typer.Option(
+        False,
+        "--prune",
+        help="Remove unused artifacts from the stage during sync. Mutually exclusive with --from.",
+        show_default=False,
+    ),
```
sfc-gh-mraba: I would suggest keeping options definitions above commands for readability of function signature.

src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -91,52 +87,79 @@
...
     **options,
 ):
     """
     Applies changes defined in DCM Project to Snowflake.
     """
-    if version and from_stage:
-        raise CliError("--version and --from are mutually exclusive.")
+    if prune and from_stage:
+        raise UsageError("--prune and --from are mutually exclusive.")
```
sfc-gh-mraba: For future.
AFAIK we have a `OverrideableOption / OverrideableArgument` for mutually exclusive options.

src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -280,3 +244,39 @@ def drop_version(
...
+        return None
+
+    try:
+        project_entity = get_entity_for_operation(
+            cli_context=cli_context,
+            entity_id=None,
+            project_definition=cli_context.project_definition,
+            entity_type="dcm",
+        )
+    except (NoProjectDefinitionError, UsageError) as e:
```
sfc-gh-mraba: Do we want to catch UsageError here?
Usually it should lead to the app termination.

src/snowflake/cli/_plugins/dcm/manager.py
```diff
@@ -50,23 +48,13 @@ def execute(
...
-        return self.execute_query(query)
-
-    def create(
-        self, project: DCMProjectEntityModel, initialize_version_from_local_files: bool
-    ) -> None:
-        self._create_object(project.fqn)
-        if initialize_version_from_local_files:
-            self.add_version(project=project)
+    def create(self, project: DCMProjectEntityModel) -> None:
+        query = dedent(f"CREATE DCM PROJECT {project.fqn.sql_identifier}")
```
sfc-gh-mraba: I don't thin this `dedent` is useful here. It's mostly for `"""` strings.

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-daniszewski: good point, moved

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-daniszewski: we do! refactored to use it

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-daniszewski: that's arguable - I added this catch as `get_entity_for_operation` does throw it when provided `entity_id` is not found or there are multiple of those. With the catch, the `_sync_local_files` would return `None` and we will display following error.
> Unable to plan DCM Project. Either provide a --from stage or ensure you're in a DCM project directory.

On the other hand, if we won't catch it, it would result in more specific error message.

src/snowflake/cli/_plugins/dcm/manager.py
sfc-gh-daniszewski: sure, thx, changed

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-daniszewski: After some thinking I'd remove this catch (and the if above) to rely on already used errors and simplify logic

PR #2520, opened 2025-07-31
rename `drop-version` to `drop-deployment`. SQL unchanged

sfc-gh-daniszewski:
### Pre-review checklist
   * [] I've confirmed that instructions included in README.md are still correct after my changes in the codebase.

### Changes description
...

### DIFF
--- a/src/snowflake/cli/_plugins/dcm/commands.py
+++ b/src/snowflake/cli/_plugins/dcm/commands.py
@@ -207,7 +207,7 @@ def list_deployments(
 
 
 @app.command(requires_connection=True)
-def drop_version(
+def drop_deployment(
     identifier: FQN = dcm_identifier,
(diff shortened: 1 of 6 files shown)


src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -207,7 +207,7 @@ def list_deployments(
 
 
 @app.command(requires_connection=True)
-def drop_version(
+def drop_deployment(
```
sfc-gh-turbaszek, line 210: Should old name be preserved to keep backward compatibility?

sfc-gh-daniszewski: not in DCM case, it's PrPr still and we're aligned on no backwards compatibility.

PR #2545, opened 2025-08-16
Add remote dev plugin Register the remote dev plugin and add remote dev container spec

sfc-gh-zzhu:
### Pre-review checklist

### Changes description
...


tests/remote/test_utils.py
```diff
@@ -0,0 +1,269 @@
...
+        ):
+            get_node_resources(mock_session, "nonexistent_pool")
+
+
+class TestStagePathUtils:
+    """Test stage path validation and formatting utilities."""
+
+    def test_validate_stage_path_valid(self):
+        """Test validate_stage_path with valid paths."""
+        valid_paths = [
```
sfc-gh-dhung: nit: consider adding some test cases with double quoted identifiers with special characters

src/snowflake/cli/_plugins/remote/utils.py
```diff
@@ -0,0 +1,134 @@
...
+        or CLOUD_INSTANCE_FAMILIES[cloud][instance_family]
+    )
+
+
+def validate_stage_path(path: str) -> bool:
+    """Validate if a string is a valid Snowflake stage path."""
+    return path.startswith("@")
+
+
+def format_stage_path(stage_path: str) -> str:
```
sfc-gh-dhung: What is expected behavior around:
- Invalid identifiers
- Trailing spaces
- SnowURLs

sfc-gh-dhung (approved):

src/snowflake/cli/_plugins/remote/utils.py
sfc-gh-zzhu: Added trailing spaces and simple snowurl handling. For invalid stage path, I prefer to pass it through for now just because it's a bit cumbersome to leave the regex matching in the client side. I will leave a todo and see if we need to do that in the future.

sfc-gh-dhung (approved):

PR #2553, opened 2025-08-20
SNOW-2247516-spsc-tests-warmup

sfc-gh-mraba:
### Pre-review checklist

### Changes description
...

### DIFF
--- a/tests_integration/tests_using_container_services/spcs/testing_utils/spcs_services_utils.py
+++ b/tests_integration/tests_using_container_services/spcs/testing_utils/spcs_services_utils.py
@@ -77,7 +77,6 @@ class SnowparkServicesTestSteps:
         )
 
     def create_second_service(self, service_name: str, database: str) -> None:
-
         result = self._setup.runner.invoke_with_connection_json(
             [
                 "spcs",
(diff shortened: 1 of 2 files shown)


tests_integration/tests_using_container_services/spcs/testing_utils/spcs_services_utils.py
```diff
@@ -342,6 +341,32 @@ def list_endpoints_should_show_endpoint(self, service_name: str):
...
+        instances_result = self._setup.runner.invoke_with_connection_json(
+            [
+                "spcs",
+                "service",
+                "list-instances",
+                service_name,
+                *self._database_schema_args(),
+            ]
+        )
+        assert instances_result.exit_code == 0
```
sfc-gh-jwilkowski: Do we really need to use assertions here? Wouldn't it be better to ie write logs if anything fails instead? Those commands might fail for various reasons and I don't think it's good to interrupt the actual test before it even begins.

sfc-gh-mraba: if they fail, the whole tests will fail

sfc-gh-daniszewski (commented):
Are you sure the systems will be resumed just by listing them?

tests_integration/tests_using_container_services/spcs/testing_utils/spcs_services_utils.py
sfc-gh-jwilkowski: but still, wouldn't it be better to fail on actual test rather than some preparation steps?

sfc-gh-mraba:
> Are you sure the systems will be resumed just by listing them?

It's based on manual action. Listing before scheduled tests worked.
I want to confirm it by making this action part of test flow.

tests_integration/tests_using_container_services/spcs/testing_utils/spcs_services_utils.py
sfc-gh-mraba: It will fail on the test using this fixture. I see no point on failing first test action since setup cannot confirm the system is ready.

sfc-gh-daniszewski (approved):
Let's give it a try :)

PR #2600, opened 2025-09-10
SNOW-2193769 Enable partial, case-insensitive matching for object names in `snow logs`

sfc-gh-daniszewski:
### Pre-review checklist

### Changes description
Add `--partial` option to change search from strict to partial, case-insensitive.

### DIFF
--- a/tests/logs/__snapshots__/test_logs.ambr
+++ b/tests/logs/__snapshots__/test_logs.ambr
@@ -64,6 +64,116 @@
   
   '''
 # ---
+# name: test_partial_match_query_construction[MyObject-False]
+  '''
+  SELECT
+                  timestamp,
+                  resource_attributes:"snow.database.name"::string as database_name,
+                  resource_attributes:"snow.schema.name"::string as schema_name,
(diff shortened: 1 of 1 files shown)


src/snowflake/cli/_plugins/logs/manager.py
```diff
@@ -80,10 +85,23 @@ def get_raw_logs(
...
+
+        # Build the object name condition based on partial_match flag
+        if partial_match:
+            # Use ILIKE for case-insensitive partial matching with wildcards
+            escaped_pattern = escape_like_pattern(escaped_object_name)
+            object_condition = f"object_name ILIKE '%{escaped_pattern}%'"
+        else:
+            # Use exact match (original behavior)
+            object_condition = f"object_name = '{escaped_object_name}'"
+
```
sfc-gh-mraba, line 104: Maybe this could be converted to parametrised query instead of manual escaping?

sfc-gh-daniszewski: it can, but that would require changes to the testing framework and would be against majority of CLI sql executions where f-strings are used.

PR #2638, opened 2025-10-01
Snow 2306184 ng config support 3

sfc-gh-mraba:
### Pre-review checklist

### Changes description
This PR refactors the Snowflake CLI configuration management by removing the module-level `CONFIG_MANAGER` singleton and replacing it with a factory-based approach managed through the `CliGlobalContext`. This change improves testability and eliminates test flakiness caused by shared global state.

### DIFF
--- a/src/snowflake/cli/api/config_ng/README.md
+++ b/src/snowflake/cli/api/config_ng/README.md
@@ -1,0 +1,91 @@
+<!--
+ Copyright (c) 2024 Snowflake Inc.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ -->
+
+## Config NG overview
+
+This package implements the "next generation" configuration system used by
+Snowflake CLI. It is a read-only resolver with explicit source ordering,
+full provenance tracking, and dedicated presentation/logging utilities.
+
+### Resolution flow
+- **Four phases in `resolver.py`**: file sources (connection-level replacement),
+  overlay sources (field-level merge), default connection creation, then history
+  finalization.
+- **Precedence is explicit**: the resolver consumes sources in the order they are
+  provided; later sources win.
+- **Default connection** is created when only root-level connection parameters
+  are provided (no `connections` section).
+
+### Source types
+- **FILE sources** (replace whole connections): `SnowSQLConfigFile`,
+  `CliConfigFile`, `ConnectionsConfigFile`.
+- **OVERLAY sources** (merge fields): `SnowSQLEnvironment`, `CliEnvironment`,
+  `ConnectionSpecificEnvironment`, `CliParameters`.
+- Root-level params are split via `extract_root_level_connection_params()` and
+  merged into connections via `merge_params_into_connections()`.
+
+### Observability and presentation
+- **Observers** (`observers.py`) collect telemetry stats and full per-key history.
+- **Presentation** (`presentation.py`) formats tables, chains, and exports masked
+  history to JSON.
+- **Logging helpers** (`resolution_logger.py`) expose high-level entry points
+  when the alternative config provider is enabled.
+- **Diagnostics**: sources can emit `SourceDiagnostic` messages that the resolver
+  collects and the presenter renders.
+- **Masking** is key-based: sensitive values are masked when keys include
+  sensitive fragments (see `masking.py`). Values under non-sensitive keys are
+  displayed as-is.
+
+### Extending the system
+- Add a new source by implementing `ValueSource` in `core.py`, returning nested
+  dicts from `discover()` and identifying `source_type` and `source_name`.
+- Wire new sources in `source_factory.py` / `source_manager.py` and update
+  `constants.py` with any new source names or env var keys.
+- Keep masking in mind for sensitive values (see `masking.py` and presenter
+  usage).
+
+### Example: snow helpers config resolution
+The `snow helpers show-config-sources` command renders config and connection
+resolution details when the alternative config provider is enabled:
+
+```
+export SNOWFLAKE_CLI_CONFIG_V2_ENABLED=true
+
+# Show a summary table for all keys
+snow helpers show-config-sources
+
+# Show detailed resolution for a single key in a specific connection
+snow helpers show-config-sources account --connection dev --show-details
+```
+
+Example output (abridged):
+
+```
+key                       value   params  global_envs  connections_env  snowsql_env  connections.toml  config.toml  snowsql
+connections.dev.account   acme    +       +                                         +                 -            -
+connections.dev.user      alice           +                                         +                 -            -
+connections.dev.role      analyst                                                        +              -            -
+
+Configuration Resolution History
+================================================================================
+
+Key: connections.dev.account
+Final Value: acme
+Resolution Chain:
+  1. [overridden] snowsql_config: legacy_acme
+  2. [overridden] cli_config_toml: acme
+  3. [SELECTED] cli_arguments: acme
+```


tests/nativeapp/codegen/test_compiler.py
```diff
@@ -148,7 +148,7 @@ def test_skips_disabled_processors(test_proj_def, test_compiler):
         {"dest": "./", "src": "app/*", "processors": ["test_processor"]}
     ]
     test_compiler = NativeAppCompiler(_get_bundle_context(pkg_model))
-    test_compiler.register(TestProcessor)
+    test_compiler.register(AcmeProcessor)
 
     # TestProcessor is never invoked, otherwise calling its methods will make the test fail
```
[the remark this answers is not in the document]
sfc-gh-mraba: The class was renamed to avoid warnings during tests collection.

src/snowflake/cli/_app/loggers.py
```diff
@@ -136,7 +137,11 @@ def _check_log_level(self, config):
 
     @property
     def filename(self):
-        return self.path.path / _DEFAULT_LOG_FILENAME
+        from snowflake.cli.api.utils.path_utils import path_resolver
```
sfc-gh-jwilkowski, line 140: nit: local import

src/snowflake/cli/api/cli_global_context.py
```diff
@@ -144,6 +156,75 @@ def _clear_definition_manager(self):
...
+            name="connections",
+            parse_str=tomlkit.parse,
+            default=dict(),
+        )
+
+        manager.add_option(
+            name="default_connection_name", parse_str=str, default="default"
+        )
+
+        # Add CLI-specific options (current lines 66-70 in config.py)
```
sfc-gh-jwilkowski: this comment will become useless quickly

src/snowflake/cli/api/cli_global_context.py
```diff
@@ -144,6 +156,75 @@ def _clear_definition_manager(self):
...
+            ConfigSliceOptions,
+        )
+        from snowflake.connector.constants import CONFIG_FILE
+
+        # Get current connections file path (handles test env changes)
+        connections_file = get_connections_file()
+
+        # Create ConfigSlice for connections.toml (same as singleton CONFIG_MANAGER)
+        connections_slice = ConfigSlice(
+            path=connections_file,
```
sfc-gh-jwilkowski, line 183: ?
```
path=self.connections_file_override or connections_file
```

sfc-gh-jwilkowski: actually is that override even needed? We don't seem to be using it anywhere?

sfc-gh-mraba: With connector based singleton this was created by default. As we are dropping the connector version of singleton we need to have our own support for connections.toml

src/snowflake/cli/_plugins/helpers/commands.py
```diff
@@ -317,3 +318,89 @@ def check_snowsql_env_vars(**options):
...
+        snow helpers show-config-sources account
+
+        # Show detailed resolution for a specific key
+        snow helpers show-config-sources account --show-details
+
+        # Export complete resolution history to file
+        snow helpers show-config-sources --export config_debug.json
+
+    Note: This command requires the enhanced configuration system to be enabled.
+    Set SNOWFLAKE_CLI_CONFIG_V2_ENABLED=true to enable it.
```
sfc-gh-jwilkowski: I recall that providing examples might not play well out of the box with our documentation generation tooling. Imo it's worth to create a sanity check ticket before final doc release. That being said, I really like and support the idea of providing examples for commands in CLI 👏

src/snowflake/cli/_plugins/helpers/commands.py
```diff
@@ -317,3 +318,89 @@ def check_snowsql_env_vars(**options):
...
+    if export_file:
+        success = export_resolution_history(export_file)
+        if not success:
+            return MessageResult(
+                f"❌ Failed to export resolution history to {export_file}"
+            )
+        return MessageResult(
+            f"✅ Resolution history exported to: {export_file}\n\n"
+            f"This file contains complete details about configuration resolution "
+            f"and can be attached to support tickets."
```
sfc-gh-jwilkowski: This feels useless? First line is good enough imo

src/snowflake/cli/_plugins/helpers/commands.py
sfc-gh-jwilkowski: 😅 
Image

src/snowflake/cli/_plugins/helpers/commands.py
```diff
@@ -317,3 +318,89 @@ def check_snowsql_env_vars(**options):
...
     return MultipleResults(results)
+
+
+@app.command(
+    name="show-config-sources",
+    requires_connection=False,
+    hidden=os.environ.get(ALTERNATIVE_CONFIG_ENV_VAR, "").lower()
+    not in ("1", "true", "yes", "on"),
+)
+def show_config_sources(
```
sfc-gh-jwilkowski, line 329: when I'm thinking about this command one of the primary use cases would be "I have connection x defined in different places, how can I see where it's taking it's values from". what do you think about providing --connection flag to serve that instead of key argument?

src/snowflake/cli/_plugins/helpers/commands.py
```diff
@@ -317,3 +318,89 @@ def check_snowsql_env_vars(**options):
...
+        None,
+        help="Specific configuration key to show resolution for (e.g., 'account', 'user'). If not provided, shows summary for all keys.",
+    ),
+    show_details: bool = typer.Option(
+        False,
+        "--show-details",
+        "-d",
+        help="Show detailed resolution chains for all sources consulted.",
+    ),
+    export_file: Optional[Path] = typer.Option(
```
sfc-gh-jwilkowski: from what I understood this option and related logic should be removed

src/snowflake/cli/_plugins/helpers/commands.py
sfc-gh-mraba: Maybe in future. This PR is already huge.

src/snowflake/cli/api/config_ng/merge_operations.py
```diff
@@ -0,0 +1,118 @@
...
+
+    Returns:
+        Dictionary of connections with params merged in
+
+    Example:
+        Input:
+            connections = {"dev": {"account": "dev_acc", "user": "dev_user"}}
+            params = {"user": "override_user", "password": "new_pass"}
+        Output:
+            {"dev": {"account": "dev_acc", "user": "override_user", "password": "new_pass"}}
```
sfc-gh-jwilkowski, line 86: This feels confusing, at least I interpreted it as if the top level settings would override what's defined on particular connection level

src/snowflake/cli/api/config_ng/parsers.py
```diff
@@ -0,0 +1,161 @@
...
+                    mapped_key = cls.SNOWSQL_KEY_MAP.get(key, key)
+                    result["connections"][conn_name][mapped_key] = value
+
+            elif section == "variables":
+                result["variables"] = dict(config[section])
+
+        return cls._apply_default_connection_fallback(result)
+
+    @classmethod
+    def _apply_default_connection_fallback(
```
sfc-gh-jwilkowski, line 109: @sfc-gh-mraba here's the logic that actually takes general settings and applies specifics overlay

src/snowflake/cli/_plugins/helpers/commands.py
sfc-gh-jwilkowski: then for now I'd only suggest to make key an option not an argument

src/snowflake/cli/api/config_ng/resolver.py
```diff
@@ -0,0 +1,716 @@
...
+
+        return observer_list, telemetry, history
+
+    def attach_observer(self, observer: ResolutionObserver) -> None:
+        """Attach a new observer at runtime."""
+        self._observers.append(observer)
+        if isinstance(observer, TelemetryObserver):
+            self._telemetry_observer = observer
+        if isinstance(observer, ResolutionHistoryTracker):
+            self._history_observer = observer
```
sfc-gh-jwilkowski, line 141: This method seems to be redundant. It's only used once in ensure_history_tracking to attach ResolutionHistoryTracker. Those 3 lines might just as well end up in ensure_history_tracking

src/snowflake/cli/api/config_ng/resolver.py
```diff
@@ -0,0 +1,716 @@
...
+
+    def ensure_history_tracking(self) -> bool:
+        """
+        Ensure a history tracker is attached.
+
+        Returns:
+            True if a new tracker was attached and observers should be reset.
```
sfc-gh-jwilkowski: seems that this return value is needed only for resetting - why not do it in this method? Feels weird when you call ensure_something method and it returns you false even though that thing is ensured

src/snowflake/cli/api/config_ng/resolver.py
```diff
@@ -0,0 +1,716 @@
...
+
+    def _reset_observers(self) -> None:
+        for observer in self._observers:
+            observer.reset()
+
+    def _notify(self, method_name: str, *args, **kwargs) -> None:
+        for observer in self._observers:
+            getattr(observer, method_name)(*args, **kwargs)
+
+    def add_source(self, source: "ValueSource") -> None:
```
sfc-gh-jwilkowski: dead code?

src/snowflake/cli/api/config_ng/resolver.py
```diff
@@ -0,0 +1,716 @@
...
+
+        Args:
+            source_type: Type of source to filter by
+
+        Returns:
+            List of sources matching the type
+        """
+        return [s for s in self._sources if s.source_type is source_type]
+
+    def _record_discoveries(self, source_values: Dict[str, ConfigValue]) -> None:
```
sfc-gh-jwilkowski: dead code

src/snowflake/cli/api/config_ng/__init__.py
```diff
@@ -0,0 +1,131 @@
...
+    "ResolutionPresenter",
+    "show_all_resolution_chains",
+    "show_resolution_chain",
+    "SnowSQLConfigFile",
+    "SnowSQLEnvironment",
+    "SnowSQLParser",
+    "SnowSQLSection",
+    "SourceManager",
+    "SourceType",
+    "SourceDiagnostic",
```
sfc-gh-jwilkowski, line 127: duplicate

src/snowflake/cli/api/config_ng/resolver.py
```diff
@@ -0,0 +1,716 @@
...
+            self._history_observer.record_discovery(k, config_value)
+
+    def _collect_source_diagnostics(self, source: "ValueSource") -> None:
+        diagnostics = []
+        consumer = getattr(source, "consume_diagnostics", None)
+        if callable(consumer):
+            diagnostics = consumer()
+        elif hasattr(source, "get_diagnostics"):
+            getter = getattr(source, "get_diagnostics")
+            diagnostics = getter() if callable(getter) else getter  # type: ignore[misc]
```
sfc-gh-jwilkowski, line 182: dead? nothing implements such method

src/snowflake/cli/api/config_ng/resolver.py
```diff
@@ -0,0 +1,716 @@
...
+        self.attach_observer(history)
+        return True
+
+    def _reset_observers(self) -> None:
+        for observer in self._observers:
+            observer.reset()
+
+    def _notify(self, method_name: str, *args, **kwargs) -> None:
+        for observer in self._observers:
+            getattr(observer, method_name)(*args, **kwargs)
```
sfc-gh-jwilkowski, line 157: This might be rather brittle if you pass a method_name with a typo

src/snowflake/cli/api/config_ng/merge_operations.py
sfc-gh-mraba: reorderd items in `source_factory.py` and updated the test cases

src/snowflake/cli/api/config_ng/parsers.py
sfc-gh-mraba: this is logic used when reading snowsql config. it uses its global values to fill only the missing keys in named connections

src/snowflake/cli/api/config_ng/resolver.py
sfc-gh-mraba: - `attach_observer`: Handles the mechanics of attaching any observer and updating cached references
- `ensure_history_tracking`: Ensures a history tracker exists, creating one if needed

`attach_observer` (or `add_observer`) is the standard method name in this pattern. Removing it would make the implementation less recognizable/idiomatic to developers familiar with the pattern.

This is a case where the separation serves design clarity and maintainability

src/snowflake/cli/api/config_ng/resolver.py
sfc-gh-mraba: refactored to avoid misleading return type

src/snowflake/cli/api/config_ng/resolver.py
sfc-gh-mraba: dropped

src/snowflake/cli/_plugins/helpers/commands.py
sfc-gh-mraba: added `--connection` and tests

PR #2643, opened 2025-10-03
SNOW-2187773 Allow configurable precision for Decimal

sfc-gh-daniszewski:
### Pre-review checklist

### Changes description
Allow modifying default precision for calculations on instances of `decimal.Decimal`.

### DIFF
--- a/tests_integration/sql/test_sql_decfloat.py
+++ b/tests_integration/sql/test_sql_decfloat.py
@@ -114,10 +114,8 @@ def test_decimal_precision_environment_variable(runner, reset_decimal_precision)
         assert len(result.json) == 1
         row = result.json[0]
 
-        test_value = str(row["TEST_VALUE"])
-        pi_value = str(row["PI_VALUE"])
-
-        print(f"Environment test - test_value: {test_value}, pi_value: {pi_value}")
+        assert row["TEST_VALUE"] == "1234.567890"
+        assert row["PI_VALUE"] == "3.141592654"
 
     finally:
         if original_env is not None:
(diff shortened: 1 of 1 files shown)


tests_integration/sql/test_sql_decfloat.py
```diff
@@ -0,0 +1,168 @@
...
+            result.exit_code == 0
+        ), f"Failed to execute SQL with env var precision: {result.output}"
+
+        assert len(result.json) == 1
+        row = result.json[0]
+
+        test_value = str(row["TEST_VALUE"])
+        pi_value = str(row["PI_VALUE"])
+
+        print(f"Environment test - test_value: {test_value}, pi_value: {pi_value}")
```
sfc-gh-jwilkowski: shouldn't we assert that test_value and pi_value are as expected? This test would work even if I removed support for this env variable

tests_integration/sql/test_sql_decfloat.py
```diff
@@ -0,0 +1,168 @@
...
+        pi_value = str(row["PI_VALUE"])
+
+        print(f"Precedence test - test_value: {test_value}, pi_value: {pi_value}")
+
+        assert (
+            len(test_value.replace(".", "")) <= 7
+        ), f"Expected CLI param precision (5) to override env var (25): {test_value}"
+        assert (
+            len(pi_value.replace(".", "")) <= 7
+        ), f"Expected CLI param precision (5) to override env var (25): {pi_value}"
```
sfc-gh-jwilkowski: aren't we able to use regular eq operator?

sfc-gh-jwilkowski (commented):
Leaving some comments about tests

tests_integration/sql/test_sql_decfloat.py
sfc-gh-daniszewski: oh snap - sure, added. Thx for spotting

tests_integration/sql/test_sql_decfloat.py
sfc-gh-daniszewski: We can - this approach allows formatting differences (`,` vs `.`) or rounding (although math rules are clear how to do it). But it seems to work fine with just comparing - changed.

PR #2680, opened 2025-10-28
NO-CONN Support decimal_precision in config file

sfc-gh-daniszewski:
### Pre-review checklist

### Changes description
Allow setting decimal_precision in config file as well

### DIFF
--- a/tests/__snapshots__/test_docs_generation_output.ambr
+++ b/tests/__snapshots__/test_docs_generation_output.ambr
@@ -174,7 +174,7 @@
     Differentiate exit error codes based on failure type. Default: False.
   
   :samp:`--decimal-precision {INTEGER}`
-    Number of decimal places to display for decimal values. Uses Python's default precision if not specified.
+    Number of decimal places to display for decimal values. Uses Python's default precision if not specified. [env var: SNOWFLAKE_DECIMAL_PRECISION].
   
   '''
 # ---
--- a/tests/api/commands/__snapshots__/test_snow_typer.ambr
+++ b/tests/api/commands/__snapshots__/test_snow_typer.ambr
@@ -128,8 +128,7 @@
   |                                                       decimal values. Uses   |
   |                                                       Python's default       |
   |                                                       precision if not       |
-  |                                                       specified.             |
-  |                                                       [env var:              |
+  |                                                       specified. [env var:   |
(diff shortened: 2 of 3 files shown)


src/snowflake/cli/api/commands/flags.py
```diff
@@ -516,6 +516,14 @@ def _diag_log_allowlist_path_callback(path: str):
 
 def _decimal_precision_callback(value: int | None):
     """Callback to set decimal precision globally when provided."""
+    from snowflake.cli.api.config import CLI_SECTION, get_config_value
+
+    if value is None:
+        try:
+            value = get_config_value(CLI_SECTION, key="decimal_precision", default=None)
```
sfc-gh-jwilkowski: This is getting weird, as in get_config_value we'd also be checking env variables, but called differently since you're using CLI_SECTION. IMO you could change either the env variable or remove CLI_SECTION to be consistent

```
def get_config_value(*path, key: str, default: Optional[Any] = Empty) -> Any:
    """Looks for given key under nested path in toml file."""
    env_variable = get_env_value(*path, key=key)
    if env_variable:
        return env_variable
    try:
        return get_config_section(*path)[key]
    except (KeyError, NonExistentKey, MissingConfigOptionError, ConfigSourceError):
        if default is not Empty:
            return default
        raise
```

tests_integration/sql/test_sql_decfloat.py
```diff
@@ -155,3 +155,81 @@ def test_decimal_precision_param_overrides_env(runner, reset_decimal_precision):
...
+
+    assert row["TEST_VALUE"] == "1234.56789012346"
+    assert row["PI_VALUE"] == "3.14159265358979"
+
+
+def test_decimal_precision_cli_overrides_config(
+    runner, reset_decimal_precision, temp_dir
+):
+    """Test that CLI parameter takes precedence over config file setting."""
+    from pathlib import Path
```
sfc-gh-jwilkowski: local import

sfc-gh-daniszewski: fixd

src/snowflake/cli/api/commands/flags.py
sfc-gh-daniszewski: good point, changed

PR #2708, opened 2025-12-04
Jw/snow 2326969 implement dcm refresh

sfc-gh-jwilkowski:
### Pre-review checklist

### Changes description
...

### DIFF
--- a/src/snowflake/cli/_plugins/dcm/reporters.py
+++ b/src/snowflake/cli/_plugins/dcm/reporters.py
@@ -20,6 +20,7 @@ from typing import Any, Dict, Generic, Iterator, List, Optional, TypeVar, Union
 
 from snowflake.cli._plugins.dcm import styles
 from snowflake.cli.api.console.console import cli_console
+from snowflake.cli.api.exceptions import CliError
 from snowflake.cli.api.sanitizers import sanitize_for_terminal
 from snowflake.connector.cursor import SnowflakeCursor
 
@@ -58,10 +59,17 @@ class Reporter(ABC, Generic[T]):
             cli_console.styled_message("No data.\n")
             return
 
-        result_data = row[0]
-        result_json = (
-            json.loads(result_data) if isinstance(result_data, str) else result_data
-        )
+        try:
+            result_data = row[0]
+            result_json = (
+                json.loads(result_data) if isinstance(result_data, str) else result_data
+            )
+        except IndexError:
+            log.debug("Unexpected response format: %s", row)
+            raise CliError("Could not process response.")
+        except json.JSONDecodeError as e:
+            log.debug("Could not decode response: %s", e)
+            raise CliError("Could not process response.")
 
         raw_data = self.extract_data(result_json)
         parsed_data: Iterator[T] = self.parse_data(raw_data)
@@ -179,7 +187,7 @@ class RefreshReporter(Reporter[RefreshRow]):
     def extract_data(self, result_json: Dict[str, Any]) -> List[Dict[str, Any]]:
         if not isinstance(result_json, dict):
             log.debug("Unexpected response type: %s, expected dict", type(result_json))
-            return []
+            raise CliError("Could not process response.")
 
         refreshed_tables = result_json.get(self._DATA_KEY, list())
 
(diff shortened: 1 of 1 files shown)


src/snowflake/cli/_app/printing.py
```diff
@@ -341,7 +342,9 @@ def _print_single_table(obj):
 
 def print_result(cmd_result: CommandResult, output_format: OutputFormat | None = None):
     output_format = output_format or _get_format_type()
-    if is_structured_format(output_format):
```
sfc-gh-mraba, line 344: This ifology could benefit from match

src/snowflake/cli/_plugins/dcm/reporters.py
```diff
@@ -0,0 +1,300 @@
...
+
+
+class Reporter(ABC, Generic[T]):
+    def __init__(self) -> None:
+        self.command_name = ""
+
+    @abstractmethod
+    def extract_data(self, result_json: Dict[str, Any]) -> List[Dict[str, Any]]:
+        """Extract the relevant data from the result JSON."""
+        pass
```
sfc-gh-mraba: `...` should be used in favour of `pass`

src/snowflake/cli/api/console/console.py
```diff
@@ -112,16 +112,11 @@ def spinner(self):
...
-            transient=False,
+            transient=True,
         ) as progress:
-            try:
-                yield progress
-            finally:
-                for task_id in progress.task_ids:
-                    if not progress.tasks[task_id].finished:
-                        progress.update(task_id, completed=1, total=1)
+            yield progress
```
sfc-gh-mraba: You need to make sure the task is finished. With plain yield you risk the Progress in unfinished state when task fail.

src/snowflake/cli/api/console/console.py
```diff
@@ -151,6 +146,12 @@ def panel(self, message: str):
         panel = Panel(message, style=style)
         self._print(panel)
 
+    def safe_print(self, message: str | Text):
+        """Displays an already formatted message.
+
+        It's callers responsibility to call sanitize_for_terminal on any external data"""
+        self._print(message)
+
```
sfc-gh-mraba: We shouldn't pass the sanitisation of output to others. 
This may lead to incidents.

src/snowflake/cli/_plugins/dcm/styles.py
sfc-gh-mraba, line 1: It's worth placing the style at api level.
I would even consider moving this up to the console module. Even if only single plugin is using it may spread

src/snowflake/cli/_app/printing.py
sfc-gh-jwilkowski: done

src/snowflake/cli/_plugins/dcm/reporters.py
sfc-gh-jwilkowski: done

src/snowflake/cli/api/console/console.py
sfc-gh-jwilkowski: This change was actually reverting to what we had before, because the finished progress bar was printed to stdout and that was messing with `invoke_with_connection_json` method, and with loading stdout as json in particular.

I've decided to keep this try/except for safety, but reverted the progress bar to be transient, which means it will be cleared out after it's done

src/snowflake/cli/api/console/console.py
sfc-gh-jwilkowski: Added `styled_message()` which takes care of sanitization and styling. For now the styles can be provided from external sources, but eventually we can incorporate them to main CLI styles

src/snowflake/cli/_plugins/dcm/reporters.py
```diff
@@ -0,0 +1,289 @@
...
+        if not row:
+            cli_console.styled_message("No data.\n")
+            return
+
+        result_data = row[0]
+        result_json = (
+            json.loads(result_data) if isinstance(result_data, str) else result_data
+        )
+
+        raw_data = self.extract_data(result_json)
```
sfc-gh-mraba, line 74: Is it guaranteed that the result_data will always be a valid JSON?

PR #2709, opened 2025-12-05
SNOW-2805230 headless auth

sfc-gh-mraba:
### Pre-review checklist

### Changes description
Fix `externalbrowser` auth for headless systems.

### DIFF
--- a/src/snowflake/cli/_app/snow_connector.py
+++ b/src/snowflake/cli/_app/snow_connector.py
@@ -54,6 +54,7 @@ log = logging.getLogger(__name__)
 ENCRYPTED_PKCS8_PK_HEADER = b"-----BEGIN ENCRYPTED PRIVATE KEY-----"
 UNENCRYPTED_PKCS8_PK_HEADER = b"-----BEGIN PRIVATE KEY-----"
 AUTHENTICATOR_EXTERNAL_BROWSER: Literal["externalbrowser"] = "externalbrowser"
+AUTHENTICATOR_PARAM: Literal["authenticator"] = "authenticator"
 
 # connection keys that can be set using SNOWFLAKE_* env vars
 SUPPORTED_ENV_OVERRIDES = [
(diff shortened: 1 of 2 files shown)


sfc-gh-asawicki (dismissed):

src/snowflake/cli/_app/snow_connector.py
```diff
@@ -88,6 +89,33 @@
...
 
 
+class _SilentStdStream(io.StringIO):
+    """Buffer connector chatter while optionally mirroring to another stream."""
+
+    def __init__(self, mirror: Optional[TextIO] = None) -> None:
+        super().__init__()
+        self._mirror = mirror
+
+    def write(self, text: str) -> int:  # type: ignore[override]
```
sfc-gh-jwilkowski:
```
def write(self, text: Optional[str]) -> int:  # type: ignore[override]
```

src/snowflake/cli/_app/snow_connector.py
```diff
@@ -186,12 +214,22 @@ def connect_to_snowflake(
 
     _update_internal_application_info(connection_parameters)
 
+    authenticator = str(connection_parameters.get("authenticator", "")).lower()
```
sfc-gh-jwilkowski: I have several comments to lines below:
1) The comment in L225 is getting outdated, as you're starting to pass messages to stdout
2) the whole thing lacks clarification what it is doing and why - why are you redirecting stderr to stdout only with this particular authenticator type?
3) I don't like the name _SilentStdStream as it's not silent, it rather redirects something else or mirrors it if we want to stick to that naming

src/snowflake/cli/_app/snow_connector.py
sfc-gh-mraba: The signature should be text: str.
Updated guard for the false values.

src/snowflake/cli/_app/snow_connector.py
sfc-gh-mraba: 1. Yupp - updated according to current state. The alternative output formats are not polluted so the piping and redirecting output works as intented.
2. Moved to dedicated private func with proper doc string
3. Updated as requested

PR #2718, opened 2025-12-12
feat: [SNOW-2869663] implement dcm test command

sfc-gh-jwilkowski:
### Pre-review checklist

### Changes description
https://snowflakecomputing.atlassian.net/browse/SNOW-2869663

### DIFF
--- a/src/snowflake/cli/_plugins/dcm/utils.py
+++ b/src/snowflake/cli/_plugins/dcm/utils.py
@@ -17,7 +17,6 @@ from functools import wraps
 from pathlib import Path
 from typing import Any
 
-import typer
 from snowflake.cli._plugins.dcm.reporters import RefreshReporter, TestReporter
 from snowflake.cli.api.output.types import EmptyResult
 
@@ -81,9 +80,7 @@ def mock_dcm_response(command_name: str):
             reporter_mapping = {"refresh": RefreshReporter, "test": TestReporter}
 
             reporter = reporter_mapping[command_name]()
-            has_errors = reporter.process(cursor)
-            if has_errors:
-                raise typer.Exit(code=1)
+            reporter.process(cursor)
             return EmptyResult()
 
         return wrapper


src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -307,6 +307,26 @@ def refresh(
...
+    Tests all expectations defined in DCM project.
+    """
+    with cli_console.spinner() as spinner:
+        spinner.add_task(description=f"Testing dcm project {identifier}", total=None)
+        result = DCMProjectManager().test(project_identifier=identifier)
+
+    reporter = TestReporter()
+    has_errors = reporter.process(result)
+    if has_errors:
+        raise typer.Exit(code=1)
```
sfc-gh-jwilkowski: I am a bit torn about it. This raises the question that sometimes we'd print that big error message (as we do elsewhere in CLI) and sometimes just the line with how many tests have failed.

But on the other hand I really like the minimalistic version of just printing 1 line without that boarder. Happy to hear your feedback

sfc-gh-mraba: Why not use CliError or other appropriate exception?

src/snowflake/cli/_plugins/dcm/manager.py
```diff
@@ -164,6 +164,10 @@ def refresh(self, project_identifier: FQN):
         query = f"EXECUTE DCM PROJECT {project_identifier.sql_identifier} REFRESH ALL"
         return self.execute_query(query=query)
 
+    def test(self, project_identifier: FQN):
```
sfc-gh-mraba, line 167: I would avoid using generic `test` as name for the method.

src/snowflake/cli/_plugins/dcm/reporters.py
```diff
@@ -53,11 +53,16 @@ def print_summary(self) -> None:
         """Print a summary."""
         ...
 
-    def process(self, cursor: SnowflakeCursor) -> None:
+    def process(self, cursor: SnowflakeCursor) -> bool:
+        """Process cursor data and print results.
+
+        Returns:
+            True if there were errors (e.g., failed tests), False otherwise.
+        """
```
sfc-gh-mraba: This is confusing. The name od the methods does not correlate with True on error.

src/snowflake/cli/_plugins/dcm/reporters.py
```diff
@@ -75,6 +80,154 @@ def process(self, cursor: SnowflakeCursor) -> None:
         parsed_data: Iterator[T] = self.parse_data(raw_data)
         self.print_renderables(parsed_data)
         self.print_summary()
+        return self.check_for_errors()
+
+    def check_for_errors(self) -> bool:
+        return False
```
sfc-gh-mraba: Why?

src/snowflake/cli/_plugins/dcm/reporters.py
```diff
@@ -75,6 +80,154 @@ def process(self, cursor: SnowflakeCursor) -> None:
...
+
+
+class TestStatus(Enum):
```
sfc-gh-mraba, line 101: I would avoid `test` in codebase. Maybe `Sample` would be a better name?
Is this for testing some integration?

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-jwilkowski: CliError would be captured by typer, styles would be stripped and whole exception would be printed as that irritating panel with border, which I think is less readable than a message with relevant exit code. I see the issues related with proposed approach as well, so if you feel strongly about using CliError I can change it

src/snowflake/cli/_plugins/dcm/reporters.py
sfc-gh-jwilkowski: I want to stick close to the actual domain of what we're doing - the sql command is called test, so using sample/expectation/etc would only add more mental tax.

src/snowflake/cli/_plugins/dcm/reporters.py
sfc-gh-jwilkowski: removed

src/snowflake/cli/_plugins/dcm/reporters.py
sfc-gh-jwilkowski: reverted that change in favour of just raising CliError

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-jwilkowski: I've refactor it as we discussed. Commands would raise regular CliErrors and in future we can think of less verbose styling of error panel

src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -307,6 +307,24 @@ def refresh(
     return EmptyResult()
 
 
+@app.command(requires_connection=True)
+@mock_dcm_response("test")
```
sfc-gh-mraba, line 311: Is this mocking of responses required? Especially as this is described as `testing utility`.

sfc-gh-jwilkowski: For the time being it's very helpful as I can easily check how the command output would look like without hitting the server. Eventually I'll remove this decorator, but for the time being I'd like to keep it since it's not affecting normal usage if you don't know how to trigger it

PR #2727, opened 2026-01-07
Jw/snow 2930627 fix dcm and decimal integration tests

sfc-gh-jwilkowski:
### Pre-review checklist

### Changes description
Fixes SNOW-2930191 and SNOW-2930627

### DIFF
--- a/src/snowflake/cli/api/utils/types.py
+++ b/src/snowflake/cli/api/utils/types.py
@@ -31,7 +31,7 @@ def try_cast_to_bool(value: Any) -> bool:
     know_booleans_mapping = {"true": True, "false": False, "1": True, "0": False}
 
     if value.lower() not in know_booleans_mapping:
-        raise ValueError(f"Could not case {value} to bool value")
+        raise ValueError(f"Could not cast {value} to bool value")
     return know_booleans_mapping[value.lower()]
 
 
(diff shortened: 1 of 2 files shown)


src/snowflake/cli/api/utils/types.py
```diff
@@ -33,3 +33,22 @@ def try_cast_to_bool(value: Any) -> bool:
     if value.lower() not in know_booleans_mapping:
         raise ValueError(f"Could not case {value} to bool value")
     return know_booleans_mapping[value.lower()]
+
+
+def try_cast_to_int(value: Any) -> int:
```
sfc-gh-asawicki, line 38: nit: it would be nice to have unit tests for functions like this one (but I see there are also no tests for `try_cast_to_bool` above). We could discuss it internally; it's not blocking.

src/snowflake/cli/api/utils/types.py
```diff
@@ -33,3 +33,22 @@ def try_cast_to_bool(value: Any) -> bool:
...
+    if isinstance(value, int) and not isinstance(value, bool):
```
sfc-gh-asawicki, line 39: question: why do we need `not isinstance(value, bool)`? if it's int already, then why do we care about it being bool?

sfc-gh-jwilkowski: Python being Python - bool is a subclass of int
```
>>> isinstance(False, int)
True
```

src/snowflake/cli/api/utils/types.py
sfc-gh-jwilkowski: I can't find a good excuse to not do it nowadays :)

PR #2758, opened 2026-02-09
Add no prompt exit option to repl

remisalmon:
### Pre-review checklist

### Changes description

This is adding a --no-prompt-exit option to the snow sql command to exit without getting prompted; tested locally.


src/snowflake/cli/_plugins/sql/repl.py
```diff
@@ -193,17 +195,6 @@ def repl_prompt(self, msg: str = " > ") -> str:
             if self._next_input == default_text:
                 self._next_input = None
 
-    def yn_prompt(self, msg: str) -> str:
```
remisalmon, line 196: This was only being called from the ask_yn function and can just be moved there?

sfc-gh-jwilkowski (commented):
Thanks for your contribution! I like the idea, but I think it would be more useful you could set that as in config.toml or as env variable. You can find a similar implementation in [_decimal_precision_callback](https://github.com/snowflakedb/snowflake-cli/blob/main/src/snowflake/cli/api/commands/flags.py#L523-L553). To make it clear that it's scoped only to repl, maybe it will be better to be explicit about it in the name, so something like `no_prompt_exit_repl`?

Please also add some unit tests for this change, here you'll find some [examples](https://github.com/snowflakedb/snowflake-cli/blob/main/tests/sql/test_repl.py#L49-L69)

src/snowflake/cli/_plugins/sql/commands.py
```diff
@@ -145,6 +153,9 @@ def execute_sql(
     single_transaction = bool(single_transaction)
     std_in = bool(std_in)
 
+    if no_prompt_exit_repl is None:
```
remisalmon, line 155: This  follows the same order of precedence documented in https://docs.snowflake.com/en/developer-guide/snowflake-cli/connecting/configure-connections:
- Command-line parameters
- Environment variables overriding specific config.toml parameters, such as SNOWFLAKE_CONNECTIONS_MYCONNECTION_PASSWORD
- Connections defined in config.toml file manually or using snow connection add command

src/snowflake/cli/_plugins/sql/commands.py
```diff
@@ -122,6 +123,13 @@ def execute_sql(
         help="Syntax used to resolve variables before passing queries to Snowflake.",
         case_sensitive=False,
     ),
+    no_prompt_exit_repl: Optional[bool] = typer.Option(
+        None,
+        "--no-prompt-exit-repl",
+        help="Do not prompt before exiting the REPL.",
+        envvar="SNOWFLAKE_NO_PROMPT_EXIT_REPL",
```
sfc-gh-jwilkowski: you can remove envvar, as get_config_value already takes care of reading from both env var and config file. In that case the envvar would be `SNOWFLAKE_CLI_NO_PROMPT_EXIT_REPL`

src/snowflake/cli/_plugins/sql/commands.py
```diff
@@ -145,6 +153,9 @@ def execute_sql(
     single_transaction = bool(single_transaction)
     std_in = bool(std_in)
 
+    if no_prompt_exit_repl is None:
+        no_prompt_exit_repl = get_config_value("cli", key="no_prompt_exit_repl", default=False)
```
sfc-gh-jwilkowski: now I see that it might be better to use `get_config_bool_value`, sorry for mixing that up earlier

sfc-gh-jwilkowski (commented):
I've got some last remarks. Please also add an entry into `RELEASE_NOTES.md` with short description of this change

RELEASE-NOTES.md
```diff
@@ -16,6 +16,7 @@
 # Unreleased version
 ## Backward incompatibility
 * **Configuration System (NG)**: File-based configuration sources (`snowsql_config`, `cli_config_toml`, `connections_toml`) now use **connection-level replacement** instead of field-level merging. When a later file source defines a connection, it completely replaces the entire connection from earlier file sources - fields are NOT inherited. Environment variables and CLI arguments continue to overlay per-field on top of the file-derived connection. This provides more predictable configuration behavior where file-defined connections are atomic units.
+* Added `--no-prompt-exit-repl` option and configuration setting to skip the exit confirmation prompt in the SQL REPL.
```
sfc-gh-jwilkowski, line 22: Please rebase or update the upstream changes. We've recently released 3.16.0 and release notes changed a bit since. Also this note should be under ##New additions section

sfc-gh-jwilkowski (commented):
@remisalmon please run unit tests and regenerate snapshots with `pytest tests/test_help_messages.py --snapshot-update`.

RELEASE-NOTES.md
remisalmon: Done and done.

remisalmon:
Hi @sfc-gh-jwilkowski this is ready for a re review whenever you have a chance. Thanks again.

sfc-gh-jwilkowski:
Hey @remisalmon sorry for the long wait! We appreciate your work on this and didn't want to ask you for another round of rebasing so I fast tracked it in https://github.com/snowflakedb/snowflake-cli/pull/3103. This change will be released in 3.21.0, thanks!

PR #2762, opened 2026-02-11
Jw/snow 2999995 dcm manifest v2

sfc-gh-jwilkowski:
### Pre-review checklist

### Changes description
https://snowflakecomputing.atlassian.net/browse/SNOW-3017490
https://snowflakecomputing.atlassian.net/browse/SNOW-2999995

### DIFF
--- a/src/snowflake/cli/_plugins/dcm/models.py
+++ b/src/snowflake/cli/_plugins/dcm/models.py
@@ -141,7 +141,7 @@ class DCMManifest:
 
         if self.manifest_version != SUPPORTED_MANIFEST_VERSION:
             raise InvalidManifestError(
-                f"Manifest version '{self.manifest_version}' is not supported. Expected version {SUPPORTED_MANIFEST_VERSION}."
+                f"Manifest version {self.manifest_version} is not supported. Expected version {SUPPORTED_MANIFEST_VERSION}."
             )
 
     def _validate_target_configuration_exists(self, target: DCMTarget):
--- a/tests/dcm/test_models.py
+++ b/tests/dcm/test_models.py
@@ -222,7 +222,7 @@ class TestDCMManifest:
 
         with pytest.raises(
             InvalidManifestError,
-            match=f"Manifest version '{version}' is not supported. Expected version 2.",
+            match=f"Manifest version {version} is not supported. Expected version 2.",
         ):
             DCMManifest.from_dict(data)
 
@@ -298,7 +298,7 @@ class TestLoadManifest:
                 yaml.dump({"manifest_version": 1, "type": "dcm_project"}, f)
             with pytest.raises(
                 InvalidManifestError,
-                match=r"Manifest version '1' is not supported. Expected version 2.",
+                match=r"Manifest version 1 is not supported. Expected version 2.",
             ):
                 DCMManifest.load(SecurePath(project_dir))
 


tests_integration/test_dcm_project.py
```diff
@@ -246,261 +246,27 @@ def test_project_drop_deployment(
...
-        ), "Table should exist after deploy"
-
-        # Clean up
-        result = runner.invoke_with_connection(["dcm", "drop", project_name])
-        assert result.exit_code == 0, result.output
-
-
-@pytest.mark.qa_only
-@pytest.mark.integration
-def test_project_plan_with_output_path(
```
sfc-gh-jwilkowski, line 378: restore this test

src/snowflake/cli/_plugins/dcm/exceptions.py
```diff
@@ -0,0 +1,25 @@
...
+class ManifestNotFoundError(Exception):
+    """Manifest file does not exist."""
+
+
+class InvalidManifestError(Exception):
+    """Manifest file is not valid (empty, wrong type, wrong version)."""
+
+
+class ManifestConfigurationError(Exception):
+    """Manifest is valid but has configuration issues (target not found, config doesn't exist)."""
```
sfc-gh-mraba, line 25: Why not derive from BaseCLIError?

src/snowflake/cli/_plugins/dcm/models.py
```diff
@@ -0,0 +1,182 @@
...
+DCM_PROJECT_TYPE = "dcm_project"
+
+
+def _is_valid_manifest_version(version: str) -> bool:
+    """Check if manifest version is valid (>= 2.0 and < 3.0)."""
+    try:
+        v = float(version)
+        return 2.0 <= v < 3.0
+    except ValueError:
+        return False
```
sfc-gh-mraba: Any case when 3.1.0-beta will fail and are handled by ValueError
This is simpler and more bulletproof
```python
version = '2.3.1-pre'
v = tuple(int(v) for v in version.split('.')[:2])
return (2,0) <= v < (3,0)
```

src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -94,6 +98,79 @@
...
+
+    Raises:
+        ManifestNotFoundError: When manifest.yml doesn't exist (caller decides handling)
+        CliError: When manifest is invalid or misconfigured
+    """
+    try:
+        manifest = DCMManifest.load(source_path)
+        effective_target = manifest.get_effective_target(target)
+    except ManifestNotFoundError:
+        raise
```
sfc-gh-mraba: It re-rises itself. Unnecessary.

src/snowflake/cli/_plugins/dcm/models.py
```diff
@@ -0,0 +1,182 @@
...
+                f"Manifest file type is undefined. Expected {DCM_PROJECT_TYPE}."
+            )
+
+        if self.project_type.lower() != DCM_PROJECT_TYPE:
+            raise InvalidManifestError(
+                f"Manifest file is defined for type {self.project_type}. Expected {DCM_PROJECT_TYPE}."
+            )
+
+        if not self.manifest_version:
+            raise InvalidManifestError(f"Manifest version is undefined.")
```
sfc-gh-mraba: f-string without interpolation

src/snowflake/cli/api/commands/flags.py
```diff
@@ -710,6 +713,19 @@ def convert(self, value, param, ctx):
         return value
 
 
+class LocalDirectoryType(click.ParamType):
+    """Click parameter type that converts a path string to SecurePath."""
+
+    name = "PATH"
+
+    def convert(self, value, param, ctx) -> SecurePath:
```
sfc-gh-mraba:
```suggestion
    def convert(self, value, param: click.Parameter | None, ctx: click.Context | None) -> SecurePath:
```

And the types for `value` are unknown

src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -94,6 +98,79 @@
...
+    None,
+    "--target",
+    help="Target profile from manifest.yml to use. Uses default_target if not specified.",
+    show_default=False,
+)
+
+save_output_option = typer.Option(
+    False,
+    "--save-output",
+    help="Download plan output files to local 'out/' directory.",
```
sfc-gh-lkyc, line 111: Are you sure it's only about "plan? files?

src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -94,6 +98,79 @@
...
+
+
+def _resolve_context_with_optional_manifest(
+    from_location: SecurePath, identifier: FQN | None, target: str | None
+) -> TargetContext:
+    try:
+        context = _resolve_target_context(identifier, target, from_location)
+    except ManifestNotFoundError:
+        if not identifier:
+            raise CliError("No project identifier specified and no manifest.yml found.")
```
sfc-gh-lkyc: Are you sure this error indicates both "no project identifier specified" and `no manifest.yml found' rather than the lack of the manifest only?

src/snowflake/cli/_plugins/dcm/models.py
```diff
@@ -0,0 +1,182 @@
...
+            project_name=data.get("project_name", ""),
+            templating_config=data.get("templating_config"),
+        )
+
+
+@dataclass
+class DCMManifest:
+    """DCM manifest v2 structure."""
+
+    manifest_version: str
```
sfc-gh-lkyc: It should be int - we never put a string here. The additional benefit would be easier implementation of the `_is_valid_manifest_version` method - it will check whether it's `2`.

src/snowflake/cli/_plugins/dcm/models.py
```diff
@@ -0,0 +1,182 @@
...
+                f"{MANIFEST_FILE_NAME} was not found in directory {source_path.path}."
+            )
+
+        with dcm_manifest_file.open(read_file_limit_mb=DEFAULT_SIZE_LIMIT_MB) as fd:
+            data = yaml.safe_load(fd)
+            if not data:
+                raise InvalidManifestError("Manifest file is empty or invalid.")
+
+            manifest = cls.from_dict(data)
+            manifest.validate()
```
sfc-gh-lkyc: We could validate it in the `from_dict` method. It will prevent the accidental creation of an invalid manifest.

tests/test_data/projects/dcm_project/manifest.yml
```diff
@@ -1,4 +1,8 @@
+manifest_version: 2.0
```
sfc-gh-lkyc: Please update also other test manifests
```suggestion
manifest_version: 2
```

src/snowflake/cli/_plugins/dcm/models.py
```diff
@@ -0,0 +1,182 @@
...
+            target.templating_config
+            and target.templating_config not in self.templating.configurations
+        ):
+            raise ManifestConfigurationError(
+                f"Target '{target.name}' references unknown configuration '{target.templating_config}'."
+            )
+
+    def get_target(self, target_name: str) -> DCMTarget:
+        """Get a specific target by name."""
+        target_name = target_name.lower()
```
sfc-gh-lkyc: In Snowflake, case-insensitive strings are converted to upper case, like a table name in SQL statement. I believe we should follow a similar approach.

sfc-gh-lkyc (commented):
In general, it looks good. I left a few comments.

src/snowflake/cli/_plugins/dcm/exceptions.py
sfc-gh-jwilkowski: Those are domain specific exceptions, so I don't want to couple them with click/cli concepts

src/snowflake/cli/_plugins/dcm/models.py
sfc-gh-jwilkowski: As suggested by @sfc-gh-lkyc manifest version is supposed to be an integer, so it was simplified even further

src/snowflake/cli/_plugins/dcm/models.py
sfc-gh-jwilkowski: removed

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-jwilkowski: removed

src/snowflake/cli/api/commands/flags.py
sfc-gh-jwilkowski: added

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-jwilkowski: for the time being it's only about plan. We need revisit storing raw jsons of commands

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-jwilkowski: yeah, some commands (ie describe, drop, create) can work without the manifest as long as identifier is provided as arg. If none of those things happened, then we need to raise an error

src/snowflake/cli/_plugins/dcm/models.py
sfc-gh-jwilkowski: fixed, thanks for pointing it out

tests/test_data/projects/dcm_project/manifest.yml
sfc-gh-jwilkowski: done

src/snowflake/cli/_plugins/dcm/models.py
sfc-gh-jwilkowski: that's a very good point, thanks for raising it. Fixed

src/snowflake/cli/_plugins/dcm/models.py
sfc-gh-jwilkowski: moved to from_dict

src/snowflake/cli/_plugins/dcm/models.py
```diff
@@ -0,0 +1,181 @@
...
+            )
+
+        if self.project_type.lower() != DCM_PROJECT_TYPE:
+            raise InvalidManifestError(
+                f"Manifest file is defined for type {self.project_type}. Expected {DCM_PROJECT_TYPE}."
+            )
+
+        if self.manifest_version != SUPPORTED_MANIFEST_VERSION:
+            raise InvalidManifestError(
+                f"Manifest version '{self.manifest_version}' is not supported. Expected version {SUPPORTED_MANIFEST_VERSION}."
```
sfc-gh-lkyc: nit: We quote the actual version, but don't quote the expected. We should be consistent, and probably don't quote it.

sfc-gh-lkyc: Or quote it every validation error message

PR #2772, opened 2026-02-19
SNPOW-3017496 raw_analyze command in the DCM plugin

sfc-gh-lkyc:
### Pre-review checklist

### Changes description
https://snowflakecomputing.atlassian.net/browse/SNOW-3017496

### DIFF
--- a/src/snowflake/cli/_plugins/dcm/utils.py
+++ b/src/snowflake/cli/_plugins/dcm/utils.py
@@ -18,7 +18,6 @@ from pathlib import Path
 from typing import Any
 
 from snowflake.cli._plugins.dcm.reporters import (
-    AnalyzeReporter,
     RefreshReporter,
     TestReporter,
 )
@@ -84,7 +83,6 @@ def mock_dcm_response(command_name: str):
             reporter_mapping = {
                 "refresh": RefreshReporter,
                 "test": TestReporter,
-                "analyze": AnalyzeReporter,
             }
 
             reporter = reporter_mapping[command_name]()


src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -291,6 +291,42 @@ def plan(
...
+            variables=variables,
+        )
+
+    row = result.fetchone()
+    if row and row[0]:
+        cli_console.styled_message(row[0])
+    else:
+        cli_console.styled_message("No data.")
+
+    return EmptyResult()
```
sfc-gh-jwilkowski: ~I'd suggest going for standard CommandResult subclasses as those would work better with --format flag. Check if QueryResult, SingleQueryResult or maybe QueryJsonValueResult would be better?~ The above won't rather work. I think we should go for some simplified Reporter class to at least detect if analysis reported any issues

src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -291,6 +291,42 @@ def plan(
...
+        result = manager.raw_analyze(
+            project_identifier=project_id,
+            configuration=context.configuration,
+            from_stage=effective_stage,
+            variables=variables,
+        )
+
+    row = result.fetchone()
+    if row and row[0]:
+        cli_console.styled_message(row[0])
```
sfc-gh-jwilkowski: I forgot about this - we should actually do some processing of analyze response because if the analysis detected any issues we should return exit code 1 to indicate that something went wrong. I forgot that we had that in original implementation.

If you pull the most recent changes from main, you'll see that we have Reporter classes that are responsible for parsing the response. Maybe you can use those? For sure TestReporter looks for failed tests

sfc-gh-lkyc: That makes sense. I will apply it.

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-lkyc: OK. I will extract the reported class. Initially, I tried using `QueryJsonValueResult`, but it didn't work. The reason is that it's designed to handle JSON arrays, but analyze returns a JSON object. I decided to go the simplest way and just dump what we receive, as one of the assumptions was to make the implementation quick and easy.

I will extract a separate reporter class and handle errors here. That's a fair point.

src/snowflake/cli/_plugins/dcm/reporters/refresh.py
```diff
@@ -353,3 +568,77 @@ def _generate_summary_renderables(self) -> List[Text]:
...
+    def print_renderables(self, data: Iterator[Dict[str, Any]]) -> None:
+        for _ in data:
+            pass
+        if self._raw_data is not None:
+            cli_console.styled_message(self._raw_data)
+            cli_console.styled_message("\n")
+
+    def _generate_summary_renderables(self) -> List[Text]:
+        if self._error_count == 0:
+            return [Text("Analysis completed successfully.")]
```
sfc-gh-lkyc: I'm not sure whether we need it. @sfc-gh-jwilkowski WDYT?

tests_integration/test_dcm_project.py
```diff
@@ -643,3 +655,218 @@ def test_dcm_end_to_end_workflow(
...
+        )
+        assert result.exit_code == 0, result.output
+
+        _extract_and_validate_raw_analyze_json(result.output)
+    finally:
+        os.chdir(original_cwd)
+
+    # Clean up
+    result = runner.invoke_with_connection(["dcm", "drop", project_name])
+    assert result.exit_code == 0, result.output
```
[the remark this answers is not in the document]
sfc-gh-lkyc: I followed patterns. Will create a separate PR to solve it globally for the dcm plugin.

src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -291,6 +295,39 @@ def plan(
     return _process_plan_result(result)
 
 
+@app.command(requires_connection=True, hidden=True)
+@mock_dcm_response("analyze")
```
sfc-gh-jwilkowski: not needed but also not harmful

tests_integration/test_dcm_project.py
```diff
@@ -643,3 +655,218 @@ def test_dcm_end_to_end_workflow(
...
+@pytest.mark.qa_only
+@pytest.mark.integration
+@pytest.mark.parametrize(
+    "target_args,expected_project_suffix",
+    [
+        pytest.param([], "dev", id="default_target"),
+        pytest.param(["--target", "test"], "test", id="explicit_target"),
+    ],
+)
+def test_dcm_raw_analyze_with_target(
```
sfc-gh-jwilkowski: I think you can also just add another step to https://github.com/snowflakedb/snowflake-cli/pull/2772/changes#diff-c83e9b85812729ca08e3fcf8ef4c90420857420a0dfec5544b1a65fce71e3f4bR617 instead of adding another test

tests_integration/test_dcm_project.py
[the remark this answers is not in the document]
sfc-gh-jwilkowski: you don't really need to cleanup, test_database fixture should take care of that

tests_integration/test_dcm_project.py
```diff
@@ -643,3 +655,218 @@ def test_dcm_end_to_end_workflow(
...
+                f"table_name='{test_database}.PUBLIC.AnalyzeFromDirTable'",
+                "--from",
+                f"{project_source_path}",
+            ]
+        )
+        assert result.exit_code == 0, result.output
+
+        _extract_and_validate_raw_analyze_json(result.output)
+    finally:
+        os.chdir(original_cwd)
```
sfc-gh-jwilkowski: integration tests take more time so I'd suggest to test the happy path (`test_dcm_raw_analyze_basic`) and unhappy path (`test_dcm_raw_analyze_with_errors`) + extend `test_dcm_end_to_end_workflow`.

`test_dcm_raw_analyze_with_variables` and `test_dcm_raw_analyze_from_another_directory` don't add that much value as those functionalities are already tested elsewhere and can be unit tested

src/snowflake/cli/_plugins/dcm/reporters/refresh.py
sfc-gh-jwilkowski: We probably don't but at the same time it does not hurt :)

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-jwilkowski: Ehhh I sorry for misleading you - Reporter abstraction is not helping here and instead it adds more complexity for no good reason. In the end I think it would be best to come back you your original idea, so to parse the data from cursor back in the command and detect if anything was failing - in such case raise an error so that we also return a non-0 exit code. 

If you prefer to keep AnalyzeReporter that's probably also fine, although I see that it's an abstraction for a different problem.

tests_integration/test_dcm_project.py
sfc-gh-lkyc: Done

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-lkyc: Done

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-lkyc: Removed

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-lkyc: I would keep it, as it keeps consistency with other commands.

tests_integration/test_dcm_project.py
sfc-gh-lkyc: This link doesn't point to any specific location in the file, so I guess you mean merging two tests into one. I merged `test_dcm_raw_analyze_with_target` and `test_dcm_raw_analyze_basic` and named it `test_dcm_raw_analyze_success`.

sfc-gh-jwilkowski: I was referring to `test_dcm_end_to_end_workflow`

src/snowflake/cli/_plugins/dcm/reporters/base.py
```diff
@@ -28,6 +28,7 @@
 
 class Reporter(ABC, Generic[T]):
     def __init__(self) -> None:
+        self.result_raw_data = None
```
sfc-gh-lkyc, line 31: I need it to print raw value in the `AnalyzeReporter`

tests_integration/test_dcm_project.py
sfc-gh-lkyc: Right, that makes sense. Done.

src/snowflake/cli/_plugins/dcm/utils.py
```diff
@@ -77,7 +81,11 @@ def wrapper(*args: Any, **kwargs: Any):
                 return func(*args, **kwargs)
 
             cursor = FakeCursor(data)
-            reporter_mapping = {"refresh": RefreshReporter, "test": TestReporter}
+            reporter_mapping = {
+                "refresh": RefreshReporter,
+                "test": TestReporter,
+                "analyze": AnalyzeReporter,
```
sfc-gh-jwilkowski: This is not needed. This utility was used for mocking server response with existing json files for faster tests of commands without hitting the server. Since we're not really playing with different formats of analyze, we don't need to mock it either

src/snowflake/cli/_plugins/dcm/utils.py
```diff
@@ -17,7 +17,11 @@
 from pathlib import Path
 from typing import Any
 
-from snowflake.cli._plugins.dcm.reporters import RefreshReporter, TestReporter
+from snowflake.cli._plugins.dcm.reporters import (
+    AnalyzeReporter,
```
sfc-gh-jwilkowski: Not needed

src/snowflake/cli/_plugins/dcm/utils.py
sfc-gh-lkyc: Removed. TY.

src/snowflake/cli/_plugins/dcm/utils.py
sfc-gh-lkyc: Removed. TY.

sfc-gh-jwilkowski (approved):

PR #2773, opened 2026-02-22
SNOW-3138822: Add `snow streamlit logs` command for Streamlit apps live log streaming

sfc-gh-svishnu:
SNOW-3138822: Add `snow streamlit logs` command for Streamlit apps live log streaming

### Pre-review checklist

#### Context

Developers building Streamlit apps on Snowflake currently have no CLI way to stream logs from a deployed app. To inspect runtime behavior, view errors, or debug issues they must leave the terminal and navigate the Snowsight UI. This PR adds a `snow streamlit logs` command that connects to the app's container runtime and streams log entries directly to the terminal in real time for Streamlit-in-Snowflake applications running on the SPCSv2 container runtime. This also enables cortex code or any other coding agents to access app logs and perform actions based on them.

#### What this enables

- **Live log streaming** — log entries from the deployed Streamlit container are printed to stdout as they are produced.
- **Historical tail** — the `--tail` / `-n` option fetches up to N historical log lines before switching to live streaming (default 100, max 1000).
- **Two ways to identify the app:**
  - From a **project definition** (`snowflake.yml`) — the standard Snowflake CLI workflow.
  - Via the **`--name` flag** — pass a fully qualified name directly without needing a project definition file.
- **Structured output (JSON / CSV / plain)** — the command returns a `StreamResult` that honors the global `--format` flag. `--format JSON` emits one JSON object per log entry (JSONL, pipe to `jq`); `--format CSV` emits a CSV stream with a header row; the default plain format prints the same human-readable line the reviewer would see in Snowsight.
- **Clean output stream** — status messages (token fetch, connect, streaming, stopped) go through `cli_console` on the side channel, so stdout stays valid JSONL / CSV with nothing to strip before piping downstream.
- **Safe log rendering** — every log line is passed through `sanitize_for_terminal()` before printing, so a malicious app cannot inject ANSI escape sequences or control characters into your terminal. The short-lived developer API token is wrapped in `SecretType` so it cannot leak via `repr()` / `str()` / exception traces.
- **SPCSv2 runtime validation** — the command validates that the target app runs on the SPCSv2 container runtime before attempting to connect. A clear error message is shown if the runtime is unsupported.

#### How it works (user-facing flow)

1. The user runs `snow streamlit logs` (with `snowflake.yml` present) or `snow streamlit logs --name db.schema.my_app`.
2. The CLI resolves the fully qualified name of the Streamlit app.
3. A `DESCRIBE STREAMLIT` call verifies the app exists and runs on SPCSv2.
4. The CLI calls `SYSTEM$GET_STREAMLIT_DEVELOPER_API_TOKEN` to obtain a short-lived authentication token and the app's resource URI.
5. The resource URI is converted to a WebSocket URL and the CLI opens a persistent WebSocket connection, authenticating with the token.
6. A `StreamLogsRequest` (protobuf) is sent to request historical + live logs.
7. Log entries arrive as binary protobuf frames, are decoded into `LogEntry` objects, sanitized, and yielded through a `StreamResult` so `print_result` renders them to stdout in the format requested by `--format`.
8. The user presses **Ctrl+C** to stop streaming.

#### Flow diagram

```mermaid
flowchart TD
    A["snow streamlit logs [--name ...]"] --> B{--name provided?}

    B -- Yes --> C[Resolve FQN from --name flag]
    B -- No --> D{snowflake.yml exists?}

    D -- No --> E[Error: no app specified]
    D -- Yes --> F[Read entity from project definition]
    F --> G[Resolve FQN from entity model]

    C --> H["DESCRIBE STREAMLIT <fqn>"]
    G --> H

    H --> I{SPCSv2 runtime?}
    I -- No --> J[Error: unsupported runtime]
    I -- Yes --> K["SYSTEM$GET_STREAMLIT_DEVELOPER_API_TOKEN(fqn)"]

    K --> L[Receive token + resource URI]
    L --> M[Open WebSocket connection to resource URI]
    M --> N[Send StreamLogsRequest with tail_lines]

    N --> O[Receive loop]
    O --> P{Frame type?}

    P -- Binary --> Q[Decode protobuf LogEntry]
    Q --> R[Print formatted log line to stdout]
    R --> O

    P -- Close --> S[Connection closed by server]
    P -- Timeout --> O

    O -- Ctrl+C --> T[Close WebSocket gracefully]
    S --> T
    T --> U["Done — 'Log streaming stopped.'"]
```

#### Demo

https://github.com/user-attachments/assets/40d19262-aa15-4ae9-bce3-efee33b7d250

#### Files added/modified
```
src/snowflake/cli/_plugins/streamlit/
    ├── commands.py                          # modified — added `logs` subcommand
    ├── log_streaming.py                     # new — WebSocket streaming client
    ├── proto_codec.py                       # new — protobuf encode/decode + LogEntry dataclass
    ├── proto/
    │   ├── __init__.py                      # new — package marker
    │   ├── README.md                        # new — protobuf regeneration instructions + runtime-version patch
    │   ├── developer/
    │   │   └── v1/
    │   │       └── logs_service.proto       # new — source proto definition (see proto/README.md for regen steps)
    │   └── generated/
    │       ├── __init__.py                  # new — package marker
    │       └── developer/
    │           ├── __init__.py              # new — package marker
    │           └── v1/
    │               ├── __init__.py          # new — package marker
    │               └── logs_service_pb2.py  # new — generated protobuf bindings

    tests/streamlit/
    ├── __init__.py                          # new — package marker
    └── test_streamlit_logs.py               # new — 39 unit tests

    pyproject.toml                           # modified — added protobuf + websocket-client deps
    pylock.toml                              # modified — lockfile update
    snyk/requirements.txt                    # modified — dependency snapshot
```

#### Testing

39 tests covering token retrieval, WebSocket URL construction, runtime validation (via `ObjectManager().describe`), protobuf encode/decode, the full recv-loop (binary frames, timeouts, close frames, pings, decode errors, structured JSON/CSV output), terminal-sanitization of log content, `SecretType` masking of the developer API token, and both command-level paths (`--name` flag and project definition) including error cases with the correct `CliError` subclass.


src/snowflake/cli/_plugins/streamlit/proto/generated/developer/v1/logs_service_pb2.py
```diff
@@ -0,0 +1,62 @@
+# -*- coding: utf-8 -*-
```
sfc-gh-jwilkowski, line 1: when this file needs to be generated? Do you have any process for that in mind?

snyk/requirements.txt
```diff
@@ -66,5 +66,6 @@ tzdata==2025.2 ; sys_platform == 'win32'
 tzlocal==5.3.1
 urllib3==2.6.3
 wcwidth==0.2.13
+websocket-client==1.9.0
```
sfc-gh-jwilkowski, line 69: I will need to ask around if we need a formal approval to add a new package

src/snowflake/cli/_plugins/streamlit/commands.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2024 Snowflake Inc.
+# Copyright (c) 2026 Snowflake Inc.
```
sfc-gh-jwilkowski: Is it needed to update the year? I'd rather keep it as previous value

src/snowflake/cli/_plugins/streamlit/commands.py
```diff
@@ -215,6 +220,84 @@ def get_url(
...
+    conn = cli_context.connection
+
+    if name is not None:
+        if entity_id is not None:
+            raise ClickException(
+                "Cannot specify both --name and an entity ID. "
+                "Use --name to identify the app directly, or use an "
+                "entity ID to reference a snowflake.yml definition."
+            )
+        # --name flag provided: resolve FQN and validate via server-side DESCRIBE
```
sfc-gh-jwilkowski: If you find those comments useful then maybe it would be better to convert those into log.debug messages? That way we could get additional data when debugging client issues

src/snowflake/cli/_plugins/streamlit/commands.py
```diff
@@ -215,6 +220,84 @@ def get_url(
...
+    WebSocket and prints log entries in real time. Press Ctrl+C to stop streaming.
+
+    Log streaming requires SPCSv2 runtime.
+    """
+    cli_context = get_cli_context()
+    conn = cli_context.connection
+
+    if name is not None:
+        if entity_id is not None:
+            raise ClickException(
```
sfc-gh-jwilkowski: Please use `CliError` and it's subclasses

src/snowflake/cli/_plugins/streamlit/log_streaming.py
```diff
@@ -0,0 +1,228 @@
...
+    except KeyboardInterrupt:
+        pass
+    finally:
+        try:
+            ws.close(status=websocket.STATUS_NORMAL)
+        except Exception as e:
+            log.debug("Error closing WebSocket: %s", e)
+        if streaming:
+            sys.stdout.write("\n--- Log streaming stopped.\n")
+            sys.stdout.flush()
```
sfc-gh-jwilkowski: You're not sanitizing the data coming server with `sanitize_for_terminal()`. Not sure if users can emit custom logs in their app code, but if yes, then a malicious app can cause issues

src/snowflake/cli/_plugins/streamlit/log_streaming.py
```diff
@@ -0,0 +1,228 @@
...
+    cursor = conn.cursor()
+    try:
+        # fqn is already validated by IdentifierType / FQN.using_connection —
+        # DESCRIBE uses identifier syntax, not string literals, so no
+        # single-quote injection risk.
+        cursor.execute(f"DESCRIBE STREAMLIT {fqn}")
+        row = cursor.fetchone()
+        description = cursor.description
+    finally:
+        cursor.close()
```
sfc-gh-jwilkowski: You might use something like this:
```
ObjectManager().describe(
            object_type=ObjectType.STREAMLIT.value.cli_name, fqn=name
        )
```

src/snowflake/cli/_plugins/streamlit/log_streaming.py
```diff
@@ -0,0 +1,228 @@
...
+# read deadline.  When no log entry arrives within this window, we re-issue
+# recv_data() so the loop stays responsive to KeyboardInterrupt.
+_WS_RECV_TIMEOUT_SECONDS = 90
+
+_HANDSHAKE_TIMEOUT_SECONDS = 10
+
+
+@dataclass
+class DeveloperApiToken:
+    token: str
```
sfc-gh-jwilkowski: I think it would be better to use SecretType

src/snowflake/cli/_plugins/streamlit/log_streaming.py
```diff
@@ -0,0 +1,228 @@
...
+            ws.connect(ws_url, header=header, timeout=_HANDSHAKE_TIMEOUT_SECONDS)
+        except Exception as e:
+            raise ClickException(f"Failed to connect to log stream: {e}") from e
+
+        # 4. Send StreamLogsRequest
+        ws.send_binary(encode_stream_logs_request(tail_lines))
+        log.debug("Sent StreamLogsRequest with tail_lines=%d", tail_lines)
+
+        cli_console.step(f"Streaming logs (tail={tail_lines}). Press Ctrl+C to stop.")
+        sys.stdout.write("---\n")
```
sfc-gh-jwilkowski: Please use CliConsole instead of printing straight to stdout.

src/snowflake/cli/_plugins/streamlit/log_streaming.py
```diff
@@ -0,0 +1,228 @@
...
+
+    except KeyboardInterrupt:
+        pass
+    finally:
+        try:
+            ws.close(status=websocket.STATUS_NORMAL)
+        except Exception as e:
+            log.debug("Error closing WebSocket: %s", e)
+        if streaming:
+            sys.stdout.write("\n--- Log streaming stopped.\n")
```
sfc-gh-jwilkowski: as discussed, you should probably use something like StreamResult so that the output is a valid json/csv ready to being processed by other tools

src/snowflake/cli/_plugins/streamlit/proto/generated/developer/v1/logs_service_pb2.py
sfc-gh-svishnu: Documented the regeneration command and the required post-generation patch (try/except around `ValidateProtobufRuntimeVersion` for protobuf<5 compat) in a new `src/snowflake/cli/_plugins/streamlit/proto/README.md`, plus a pointer comment at the top of `logs_service.proto`.

snyk/requirements.txt
sfc-gh-svishnu: Acknowledged — leaving to your call. No code change here; happy to swap `websocket-client` for a slimmer alternative if formal approval is denied.

src/snowflake/cli/_plugins/streamlit/commands.py
sfc-gh-svishnu: Reverted 2026 → 2024 across every touched file.

src/snowflake/cli/_plugins/streamlit/commands.py
sfc-gh-svishnu: Done — the narrative comments in `streamlit_logs` are now `log.debug(...)` calls (including `fqn` / `--name` context) so they're available when debugging client issues.

src/snowflake/cli/_plugins/streamlit/commands.py
sfc-gh-svishnu: Replaced with `CliArgumentError` (commands.py arg validation and fqn quote check), `CliSqlError` (token call / DESCRIBE failures), `CliConnectionError` (WS handshake), and `CliError` (wrong runtime). Tests updated to expect the matching subclass.

src/snowflake/cli/_plugins/streamlit/log_streaming.py
sfc-gh-svishnu: Applied `sanitize_for_terminal(...)` to `LogEntry.content` in both `format_line()` and `to_dict()`; added a unit test that feeds ANSI/control sequences and asserts they're stripped from both outputs.

src/snowflake/cli/_plugins/streamlit/log_streaming.py
sfc-gh-svishnu: Refactored to use `ObjectManager(connection=conn).describe(object_type=ObjectType.STREAMLIT.value.cli_name, fqn=fqn)`. `fqn` is now passed as `FQN` from the call site rather than re-stringified; tests patch `log_streaming.ObjectManager` instead of raw cursor execs.

src/snowflake/cli/_plugins/streamlit/log_streaming.py
sfc-gh-svishnu: Wrapped `DeveloperApiToken.token` in `SecretType`; unwrapped only at the `Authorization` header construction. Test asserts `token.value == "abc123"` and that the token value does not appear in `repr()` or `str()`.

src/snowflake/cli/_plugins/streamlit/log_streaming.py
sfc-gh-svishnu: All status lines (token fetch, connect, streaming, stopped) now go through `cli_console.step(...)` — `sys.stdout.write` is gone from `log_streaming.py`. Users can pass `--silent` to suppress them when piping JSON/CSV.

src/snowflake/cli/_plugins/streamlit/log_streaming.py
sfc-gh-svishnu: `stream_logs` is now a generator of `CommandResult` — `ObjectResult(entry.to_dict())` in JSON/CSV modes, `MessageResult(entry.format_line())` in plain/TABLE mode. The command returns `StreamResult(stream_logs(...))` so `print_result` renders valid JSONL / CSV / plain. Synthetic end-to-end smoke confirms no status chatter or token value leaks to stdout across all three formats.

sfc-gh-svishnu:
Hi @sfc-gh-jwilkowski — thank you so much for the thoughtful review comments, and apologies for the long delay in getting back to you. I was pulled into other work and couldn't carve out the time to address them sooner.

I've now gone through all of the feedback and pushed fixes for each of the 10 review threads (structured `StreamResult` output, `CliError` subclasses, `SecretType` for the developer token, `sanitize_for_terminal` on log content, `ObjectManager().describe` for the runtime check, `cli_console` for status messages, the protobuf regeneration README, copyright revert, etc.). The branch is also rebased onto the latest `main`, and all 39 unit tests pass locally.

Whenever you have a moment, would you mind taking another look? Happy to iterate on anything that still needs refinement. Thanks again for the careful review!

src/snowflake/cli/_plugins/streamlit/log_streaming.py
```diff
@@ -0,0 +1,251 @@
...
+    Calls SYSTEM$GET_STREAMLIT_DEVELOPER_API_TOKEN and returns a
+    DeveloperApiToken with the token and resource URI.
+    """
+    if "'" in fqn:
+        raise CliArgumentError(
+            f"Invalid Streamlit app name: {fqn}. Name must not contain single quotes."
+        )
+
+    query = f"CALL SYSTEM$GET_STREAMLIT_DEVELOPER_API_TOKEN('{fqn}', false);"
+    log.debug("Fetching developer API token for %s", fqn)
```
sfc-gh-jwilkowski, line 74: The connector's default `pyformat` paramstyle supports `%s` bind parameters for `CALL` statements — verified in the connector source and matching the pattern already used elsewhere in the CLI (e.g. `repl_commands.py:415`). Bind params handle escaping natively, so the single-quote guard above and the f-string construction can both be dropped:

```python
cursor.execute(
    "CALL SYSTEM$GET_STREAMLIT_DEVELOPER_API_TOKEN(%s, false);",
    (fqn,),
)
```

src/snowflake/cli/_plugins/streamlit/log_streaming.py
```diff
@@ -0,0 +1,251 @@
...
+                    entry = decode_log_entry(data)
+                except (DecodeError, ValueError) as e:
+                    log.warning("Failed to decode log entry: %s", e)
+                    continue
+                yield _result_for_entry(entry, output_format)
+            elif opcode == websocket.ABNF.OPCODE_CLOSE:
+                break
+            elif opcode == websocket.ABNF.OPCODE_PING:
+                ws.pong(data)
+
```
sfc-gh-jwilkowski, line 243: The `if/elif/elif` chain silently drops any opcode that isn't BINARY, CLOSE, or PING. The server could send a TEXT frame to report an auth error, and we'd swallow it with no trace. A debug log here would make those cases visible:

```python
else:
    log.debug("Ignoring unexpected WebSocket opcode %s from server", opcode)
```

tests/streamlit/test_streamlit_logs.py
```diff
@@ -0,0 +1,701 @@
...
+        entry.timestamp.seconds = seconds
+        entry.timestamp.nanos = nanos
+        entry.sequence = sequence
+        entry.level = level
+        return entry.SerializeToString()
+
+    def test_decode_app_log(self):
+        data = self._make_pb2_log_entry(
+            log_source=1,  # LOG_SOURCE_APP
+            content="Hello from app",
```
sfc-gh-jwilkowski, line 192: `log_source=1` / `level=2` are magic integers — `LOG_SOURCE_APP`, `LOG_LEVEL_INFO` etc. are already imported at the top of this file. Using the named constants makes intent clear and survives proto enum renumbering.

tests/streamlit/test_streamlit_logs.py
```diff
@@ -0,0 +1,701 @@
...
+
+        # Verify the sent bytes decode to a StreamLogsRequest with tail_lines=42
+        request = pb2.StreamLogsRequest()
+        request.ParseFromString(sent_bytes)
+        assert request.tail_lines == 42
+
+
+class TestValidateSpcsV2Runtime:
+    SPCS_V2 = "SYSTEM$ST_CONTAINER_RUNTIME_PY3_11"
+    FQN = FQN.from_string("DB.SCHEMA.MY_APP")
```
sfc-gh-jwilkowski, line 529: `SPCS_V2` is defined again at module level as `SPCS_V2_NAME` on line 607, but that module-level constant is never referenced. Remove the orphan or consolidate into one shared constant.

tests/streamlit/test_streamlit_logs.py
```diff
@@ -0,0 +1,701 @@
...
+        )
+        return mock_cursor
+
+    def _patch_object_manager(self, mock_cursor):
+        """Patch ObjectManager so describe(...) returns the given cursor."""
+        patcher = mock.patch(
+            "snowflake.cli._plugins.streamlit.log_streaming.ObjectManager"
+        )
+        mock_cls = patcher.start()
+        mock_cls.return_value.describe.return_value = mock_cursor
```
sfc-gh-jwilkowski: Manual `patcher.start()` / `patcher.stop()` with `try/finally` can be replaced with `with mock.patch(...) as mock_cls:` throughout this class — more idiomatic and removes the risk of a bare `stop()` call being missed during future refactors.

src/snowflake/cli/_plugins/streamlit/proto/generated/developer/v1/logs_service_pb2.py
```diff
@@ -0,0 +1,62 @@
...
+# Generated by the protocol buffer compiler.  DO NOT EDIT!
```
sfc-gh-jwilkowski, line 2: The file was generated with protobuf 6.32.1 (encoded in the `ValidateProtobufRuntimeVersion` call), but the declared dependency is `protobuf>=5.29.6,<6`. Please regenerate using a protobuf 5.x toolchain so the generated file matches the supported version range and the `try/except` shim isn't load-bearing for the normal case.

src/snowflake/cli/_plugins/streamlit/log_streaming.py
sfc-gh-svishnu: Done — switched to the connector's pyformat `%s` bind parameter, dropped the single-quote guard, and replaced the unit test with one that proves a hostile fqn is passed as a parameter tuple (so no quoting can escape).

src/snowflake/cli/_plugins/streamlit/log_streaming.py
sfc-gh-svishnu: Done — added an explicit `else` branch that `log.debug`s the unexpected opcode, so a stray TEXT/auth frame is at least visible under `--debug` instead of being silently dropped.

tests/streamlit/test_streamlit_logs.py
sfc-gh-svishnu: Done — replaced the `log_source=1 / level=2` magic ints with `LOG_SOURCE_APP` / `LOG_LEVEL_INFO` (and `LOG_SOURCE_MANAGER` / `LOG_LEVEL_WARN` in the sibling test) using the constants already imported at the top of the file.

tests/streamlit/test_streamlit_logs.py
sfc-gh-svishnu: Done — removed the unused module-level `SPCS_V2_NAME`. The class-level `SPCS_V2` constant is the only one referenced now.

tests/streamlit/test_streamlit_logs.py
sfc-gh-svishnu: Done — rewrote `TestValidateSpcsV2Runtime` to use `with mock.patch(...) as mock_cls:` blocks; removed the manual `patcher.start()` / `stop()` / try-finally scaffolding.

src/snowflake/cli/_plugins/streamlit/proto/generated/developer/v1/logs_service_pb2.py
sfc-gh-svishnu: Done — regenerated with a pinned protobuf 5.x toolchain (`protobuf==5.29.6` + `grpcio-tools<1.69`). The emitted `ValidateProtobufRuntimeVersion(... 5, 28, 1, ...)` now falls inside our `>=5.29.6,<6` pin, so the load-bearing try/except shim is gone. README updated with the 5.x regeneration recipe + the one residual manual edit (`timestamp_pb2` side-effect import — protoc emits it without a noqa marker and ruff strips it as unused, which would break descriptor-pool resolution at import time).

RELEASE-NOTES.md
```diff
@@ -41,6 +41,7 @@
   * `project_owner` is checked for `snow dcm create`
 * Added a `--secondary-roles` option (plus matching `SNOWFLAKE_SECONDARY_ROLES` env var and `secondary_roles` config key) to `snow connection add` and the global connection overrides. The value is forwarded to `snowflake-connector-python` and accepts `ALL` or `NONE`, so sessions can be pinned to the primary role without running an extra `USE SECONDARY ROLES` statement.
 * Added `--force` flag to `snow spcs service drop` to allow dropping services that contain block storage volumes.
+* Added `snow streamlit logs` command to stream live logs from a Streamlit-in-Snowflake app running on the SPCSv2 container runtime. Supports `--tail` for historical lines, `--name` to target apps without a project definition, and honors the global `--format` flag (plain / JSON / CSV) for downstream piping.
```
sfc-gh-jwilkowski: @sfc-gh-svishnu please move it to unreleased section and it's good to go 🚀

sfc-gh-svishnu: Done — moved to the `# Unreleased version` section in 1f1206fd. Thanks!

sfc-gh-jwilkowski (approved):

PR #2776, opened 2026-02-23
SNOW-3022692 Use database and schema name from project name if provided

sfc-gh-lkyc:
### Pre-review checklist

### Changes description
https://snowflakecomputing.atlassian.net/browse/SNOW-3022692

Database and schema names from the project FQN have higher precedence than the connection database and schema names.
Examples:
- project name contains database name and schema name - database and schema name override these used in the connection configuration
- project name does not contain database name or schema name - database and schema name should be provided in the connection configuration

Affected project types:
- DCM
- DBT

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -30,7 +30,7 @@
 * all authenticators (including `snowflake-jwt`, `username_password_mfa`, `workload_identity`) are now case-insensitive.
 * Fixed `snow streamlit deploy --prune` failing with incorrect stage path format for streamlit entities using versioned deployment. The `snow://` prefix is now correctly preserved through all stage path operations.
 * Fixed a bug with `snow dbt deploy` where dbt project would have files uploaded first and project properties updated afterwards. This could lead to failed deploys if for example project lacked external access integrations and dependencies were specified.
-* Changed how fully qualified name for temporary stages is established for `snow dbt deploy`. Database and schema from dbt project object's fully qualified name take precedence over database and schema from session.
+* Changed how the fully qualified name for temporary stages is established for `snow dbt deploy`. The database and schema from the DBT Project object's fully qualified name take precedence over those from the session.
 
 # v3.15.0
 


tests/api/test_fqn.py
```diff
@@ -256,3 +256,39 @@ def test_with_special_characters(self, mock_ctx, mock_time, name, expected_name)
...
+
+        assert result.database == "context_db"
+        assert result.schema == "context_schema"
+        assert result.name == "STAGE_RESOURCE_1234567890_TEST"
+
+    def test_partial_resource_fqn_with_schema_only(self, mock_ctx, mock_time):
+        mock_ctx().connection = MagicMock(
+            database="context_db", schema="context_schema"
+        )
+        resource_fqn = FQN(name="resource", database=None, schema="resource_schema")
```
sfc-gh-jwilkowski: out of curiosity: is that even possible to have a resource with defined schema, but not the db?

tests_integration/test_dbt.py
```diff
@@ -691,3 +691,54 @@ def test_execute_with_dbt_version(
         assert result.exit_code == 0, result.output
         assert "Running with dbt=1.10.15" in result.output
         assert "Done. PASS=2 WARN=0 ERROR=0 SKIP=0 NO-OP=0 TOTAL=2" in result.output
+
+
+@pytest.mark.integration
+def test_deploy_with_fqn_schema_takes_priority_over_connection(
```
sfc-gh-jwilkowski: I don't think this test covers what has been changed. Modification of `from_resource` is affecting where the *temp stage* is created, not the dbt project object. My bet would be that this test would also pass on current main branch. If we really wanted to test where the temp stage was created, then we would need to get it out of stdout from deploy command as we're printing it there.

But thinking about it more, I see this more as an implementation detail and I am not convinced it needs a dedicated integration test (which makes the whole suite longer). IMO it would be good as well if you altered any of the existing tests to create the dbt project specified by a full fqn. The session's db is randomized, schema is set to public, so you can reuse part of this logic's test.

tests_integration/test_dbt.py
```diff
@@ -691,3 +691,54 @@ def test_execute_with_dbt_version(
...
+        finally:
+            # Cleanup: drop the project and the alternative schema
+            runner.invoke_with_connection(
+                [
+                    "sql",
+                    "-q",
+                    f"DROP DBT PROJECT IF EXISTS {snowflake_session.database}.{alt_schema}.{name}",
+                ]
+            )
+            snowflake_session.execute_string(f"DROP SCHEMA IF EXISTS {alt_schema}")
```
sfc-gh-jwilkowski: no need for cleanup as `test_database` fixture drops the database as cleanup

tests/api/test_fqn.py
sfc-gh-lkyc: Yes. When:
- you use `my_db.my_schema.my_table` you provide the fully-qualified name and the current context does not matter
- you use `my_schema.my_table` you provide the schema and name, but the database is retrieved from the current context
- you use `my_table` you provide the name, but the database and the schema is retrieved from the current context

It is how it works in Snowflake SQL.

tests_integration/test_dbt.py
sfc-gh-lkyc: OK. I reverted the test.

tests_integration/test_dbt.py
sfc-gh-lkyc: I removed the est

tests_integration/test_dbt.py
sfc-gh-lkyc: I don't have a strong opinion here, as the number of integration tests usually depends heavily on the quality and quantity of unit tests in the project. @sfc-gh-jwilkowski, you're the expert, so I trust your judgement :)

tests/api/test_fqn.py
sfc-gh-jwilkowski: ahhh true, I forgot that `schema.name` is also supported pattern

RELEASE-NOTES.md
```diff
@@ -30,7 +30,7 @@
 * all authenticators (including `snowflake-jwt`, `username_password_mfa`, `workload_identity`) are now case-insensitive.
 * Fixed `snow streamlit deploy --prune` failing with incorrect stage path format for streamlit entities using versioned deployment. The `snow://` prefix is now correctly preserved through all stage path operations.
 * Fixed a bug with `snow dbt deploy` where dbt project would have files uploaded first and project properties updated afterwards. This could lead to failed deploys if for example project lacked external access integrations and dependencies were specified.
-
+* A project identifier with a schema name, or both the database and schema names, overrides the database and schema names used in the connection configuration.
```
sfc-gh-jwilkowski: I think the scope of this change is actually quite narrow. Currently it affects only where temp stages used while deploying a dbt project get created, so I'd start with that. Current wording makes me anxious about how big this change could be :D What do you think about:
```
*Changed how fully qualified name for temporary stages is established for `snow dbt deploy`. Database and schema from dbt project object's fully qualified name take precedence over database and schema from session
```
WDYT?

sfc-gh-lkyc: Looks good. Thanks for improving it. I've applied it.

tests_integration/test_dbt.py
sfc-gh-lkyc: I've updated the `test_deploy_and_execute_with_full_fqn` integration test to cover the scenario where there is no database and schema in the context.

Issue #2764, opened 2026-02-12
SNOW-3074550: Snow stage copy --recursive doesnt work with FQNs properly


sfc-gh-apgupta:
### SnowCLI version

3.14.0

### What happened

Hi team, does snow stage copy not support FQNs properly?

Setup:
Stage migrations exists in SNOWCONVERT_AI database, not apgupta.
aimig connection uses database apgupta by default.

Role is not the issue, its engineer in both cases.

```
% snow stage copy "@SNOWCONVERT_AI.PUBLIC.MIGRATIONS/ai-verification/JOB_20260211214614_KA6WCK2T/" "/Users/apgupta/WORK/migrations/ai-migration-outputs/spcs_prod/" --recursive -c aimig
╭─ Error ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ 002003 (02000): 01c259e2-0106-15d7-0003-4c3a2817c032: SQL compilation error:                                                                                             │
│ Stage 'APGUPTA.PUBLIC.MIGRATIONS' does not exist or not authorized. 

% snow stage copy @SNOWCONVERT_AI.PUBLIC.MIGRATIONS/ai-verification/JOB_20260211214614_KA6WCK2T/ /Users/apgupta/WORK/migrations/ai-migration-outputs/spcs_prod/ --recursive -c aimig --role ENGINEER --database SNOWCONVERT_AI
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| file                                                                                                                                    | size    | status     | message |
|-----------------------------------------------------------------------------------------------------------------------------------------+---------+------------+---------|
| JOB_20260211214614_KA6WCK2T/input/converted.zip                                                                                         | 2022867 | DOWNLOADED |         |
| JOB_20260211214614_KA6WCK2T/input/source.zip                                                                                            | 907850  | DOWNLOADED |         |
```

### Console output

```Shell

```

PR #2777, opened 2026-02-23
Fix `snow stage copy --recursive` dropping FQN qualifiers

sfc-gh-moczko:
### Pre-review checklist

### Changes description
https://snowflakecomputing.atlassian.net/browse/SNOW-3074550
fixes https://github.com/snowflakedb/snowflake-cli/issues/2764

## Summary
- Fixed `iter_stage()` in `StageManager` silently dropping database and schema qualifiers from fully-qualified stage names during recursive copy, causing GET commands to resolve against the connection's default database instead of the one specified in the FQN.

## Problem
When running `snow stage copy @db.schema.stage/path/ ./local --recursive`, the generated GET commands would use `@stage/...` instead of `@db.schema.stage/...`. This happened because Snowflake's `ls` command returns file names with only the unqualified stage name (e.g., `stage/path/file.txt`), and `iter_stage()` passed that directly to `build_path()`, which reconstructed a new `StagePath` without the original FQN qualifiers.

## Root cause
In `StageManager.iter_stage()`, the `else` branch called `self.build_path(file["name"])` where `file["name"]` from `ls` output contains only the bare stage name. `build_path` → `StagePath.from_stage_str()` parsed this as a new unqualified stage, discarding the database/schema from the original user input.

## Fix
Extract the relative file path from `ls` output by splitting off the unqualified stage name prefix, then join it with `stage_path.root_path()` which preserves the original FQN from user input.

## Test plan
- [x] Added 5 parametrized test cases to `test_copy_get_recursive` covering: FQN root (`@db.schema.exe`), FQN with trailing slash, FQN with subdirectory, FQN with specific file, and schema-qualified (`@schema.exe`)
- [x] All 17 recursive copy tests pass (8 existing + 5 new FQN + 4 user stage)
- [x] Full stage test suite passes (319 tests, 50 snapshots)
- [x] Verified tests catch regression: reverting the fix causes FQN test cases to fail (GET commands show `@exe/...` instead of `@db.schema.exe/...`)

### DIFF
--- a/tests/stage/test_stage.py
+++ b/tests/stage/test_stage.py
@@ -17,6 +17,7 @@ from unittest import mock
 from unittest.mock import MagicMock
 
 import pytest
+from snowflake.cli._plugins.git.manager import GitManager
 from snowflake.cli._plugins.stage.manager import StageManager, TemporaryDirectory
 from snowflake.cli.api.errno import DOES_NOT_EXIST_OR_NOT_AUTHORIZED
 from snowflake.cli.api.stage_path import StagePath
@@ -30,6 +31,7 @@ if IS_WINDOWS:
 
 
 STAGE_MANAGER = "snowflake.cli._plugins.stage.manager.StageManager"
+GIT_MANAGER = "snowflake.cli._plugins.git.manager.GitManager"
 
 
 @pytest.mark.parametrize(
@@ -560,9 +562,6 @@ def test_copy_get_recursive_from_user_stage(
     ]
 
 
-GIT_MANAGER = "snowflake.cli._plugins.git.manager.GitManager"
-
-
 @pytest.mark.parametrize(
     "stage_path, files_on_stage, expected_stage_path, expected_calls",
     [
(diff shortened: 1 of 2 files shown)


tests/stage/test_stage.py
```diff
@@ -514,6 +560,68 @@ def test_copy_get_recursive_from_user_stage(
     ]
 
 
+GIT_MANAGER = "snowflake.cli._plugins.git.manager.GitManager"
```
sfc-gh-jwilkowski: please move constants to the top of the file

tests/stage/test_stage.py
```diff
@@ -514,6 +560,68 @@ def test_copy_get_recursive_from_user_stage(
...
+def test_copy_get_recursive_from_git_repo(
+    mock_execute,
+    mock_cursor,
+    temporary_directory,
+    stage_path,
+    files_on_stage,
+    expected_stage_path,
+    expected_calls,
+):
+    from snowflake.cli._plugins.git.manager import GitManager
```
sfc-gh-jwilkowski: Is a local import really needed? If not, then please move it to imports at the section at the top of the file

PR #2781, opened 2026-02-24
Jw/snow 3017492 add terse plan reporter

sfc-gh-jwilkowski:
### Pre-review checklist

### Changes description
https://snowflakecomputing.atlassian.net/browse/SNOW-3017492
with v1 plan api:
image

with v2 plan api:
image

### DIFF
--- a/src/snowflake/cli/_plugins/dcm/reporters/plan.py
+++ b/src/snowflake/cli/_plugins/dcm/reporters/plan.py
@@ -21,7 +21,7 @@ from rich.style import Style
 from rich.text import Text
 from snowflake.cli._plugins.dcm import styles
 from snowflake.cli._plugins.dcm.reporters.base import Reporter, cli_console
-from snowflake.cli.api.exceptions import CliError
+from snowflake.cli.api.exceptions import CliError, FQNNameError
 from snowflake.cli.api.identifiers import FQN
 from snowflake.cli.api.project.util import unquote_identifier
 from snowflake.cli.api.sanitizers import sanitize_for_terminal
@@ -81,7 +81,7 @@ class PlanRow:
             domain = sanitize_for_terminal(entity.object_id.domain.upper())
             sanitized_fqn = sanitize_for_terminal(entity.object_id.fqn)
             fqn = FQN.from_string(sanitized_fqn)
-        except ValidationError as e:
+        except (ValidationError, FQNNameError) as e:
             # Forward-compatible fallback: if a future version changes the
             # changeset entry shape, the CLI degrades gracefully instead of crashing.
             log.debug(
@@ -100,7 +100,11 @@ class PlanRow:
             try:
                 if "fqn" in object_id:
                     fqn = FQN.from_string(sanitize_for_terminal(str(object_id["fqn"])))
-            except Exception:  # noqa: BLE001
+            except Exception as e:  # noqa: BLE001
+                log.debug(
+                    "Failed to read FQN from provided string: %s",
+                    e,
+                )
                 fqn = None
 
         return cls(
--- a/tests/dcm/test_reporters/test_plan_reporter.py
+++ b/tests/dcm/test_reporters/test_plan_reporter.py
@@ -301,6 +301,59 @@ class TestPlanRow:
         assert "\x1b" not in row.operation
         assert "\x1b" not in row.domain
 
+    def test_from_dict_fallback_defaults_when_entry_is_missing(self):
+        entry = {}
+
+        row = PlanRow.from_dict(entry)
+
+        assert row.operation == "UNKNOWN"
+        assert row.domain == "UNKNOWN"
+        assert row.fqn is None
+
+    def test_from_dict_fallback_unparseable_fqn_yields_none(self):
+        entry = {
+            "type": "CREATE",
+            "object_id": {
+                "domain": "TABLE",
+                "fqn": "completely invalid fqn!!!",
+            },
+        }
+
+        row = PlanRow.from_dict(entry)
+
+        assert row.operation == "CREATE"
+        assert row.domain == "TABLE"
+        assert row.fqn is None
+
+    def test_from_dict_fallback_no_fqn_key_yields_none(self):
+        entry = {
+            "type": "DROP",
+            "object_id": {
+                "domain": "ROLE",
+            },
+        }
+
+        row = PlanRow.from_dict(entry)
+
+        assert row.operation == "DROP"
+        assert row.domain == "ROLE"
+        assert row.fqn is None
+
+    def test_from_dict_fallback_all_keys_wrong(self):
+        entry = {
+            "type_v2": "CREATE",
+            "object_id": {
+                "DOMAIN_v2": "TABLE",
+                "FQN_v2": '"DB"."SCH"."ORDERS"',
+            },
+        }
+
+        row = PlanRow.from_dict(entry)
+
+        assert row.operation == "UNKNOWN"
+        assert row.domain == "UNKNOWN"
+        assert row.fqn is None
+
     def test_display_fqn_with_none(self):
         row = PlanRow(operation="CREATE", domain="TABLE", fqn=None)
 


src/snowflake/cli/_plugins/dcm/reporters/plan.py
```diff
@@ -0,0 +1,244 @@
...
+
+    model_config = {"populate_by_name": True}
+
+
+class PlanEntityChange(BaseModel):
+    """Top-level entity change in the changeset."""
+
+    type_: str = Field(None, alias="type")
+    object_id: PlanObjectId
+
```
sfc-gh-jwilkowski, line 47: We've agreed that for now we want to have a terse output and in future we'll add a flag to display more details. For that reason there's no need as of today to add handling the full schema, so `Change` type and others nested

src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -196,9 +203,11 @@ def _process_plan_result(cursor: SnowflakeCursor) -> CollectionResult:
 
     data = json.loads(first_value)
 
-    # Handle new format
+    # Handle new format with reporter
     if isinstance(data, dict) and data.get("version", 0) == 2:
-        return CollectionResult(data.get("changeset", list()))
+        reporter = PlanReporter(command_name=command_name)
+        reporter.process_payload(data)
```
sfc-gh-lkyc, line 211: Why not `process`?

src/snowflake/cli/_plugins/dcm/reporters/plan.py
```diff
@@ -0,0 +1,246 @@
...
+
+class PlanResponse(BaseModel):
+    """Top-level version 2 plan response."""
+
+    version: int
+    metadata: Dict[str, Any] = Field(default_factory=dict)
```
sfc-gh-lkyc: It's not used, so why do we need it?

src/snowflake/cli/_plugins/dcm/reporters/plan.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+class PlanObjectId(BaseModel):
+    """Object identifier within a changeset entry."""
+
+    domain: str
+    name: str
+    fqn: str
+    database: Optional[str] = None
+    schema: Optional[str] = None
```
sfc-gh-lkyc: From what I see, only `domain` and `fqn` are used, so why do we need other properties here?

src/snowflake/cli/_plugins/dcm/reporters/plan.py
```diff
@@ -0,0 +1,246 @@
...
+
+    operation: str
+    domain: str
+    fqn: Optional[FQN] = None
+
+    @property
+    def sort_key(self) -> tuple:
+        """Sort key: operation priority (CREATE < ALTER < DROP < unknown), then domain alphabetically."""
+        return (
+            _OPERATION_ORDER.get(self.operation, len(_OPERATION_ORDER)),
```
sfc-gh-lkyc, line 71: It would be good to add a debug log for a scenario when the operation does not exist in the `_OPERATION_ORDER` object.

src/snowflake/cli/_plugins/dcm/reporters/plan.py
```diff
@@ -0,0 +1,246 @@
...
+            operation = sanitize_for_terminal(
+                str(entry_dict.get("type", "UNKNOWN")).upper()
+            )
+            object_id = entry_dict.get("object_id", {})
+            object_id = object_id if isinstance(object_id, dict) else {}
+            domain = sanitize_for_terminal(str(object_id.get("domain", "UNKNOWN")))
+            fqn = None
+            try:
+                if "fqn" in object_id:
+                    fqn = FQN.from_string(sanitize_for_terminal(str(object_id["fqn"])))
```
sfc-gh-lkyc, line 102: Just out of curiosity, what is the scenario for it? I assume the changeset format is strict, and it always returns all the required properties.

src/snowflake/cli/_plugins/dcm/reporters/plan.py
```diff
@@ -0,0 +1,246 @@
...
+        """Format an FQN for human-friendly display (unquoted)."""
+        if self.fqn is None:
+            return "UNKNOWN"
+        parts = []
+        if self.fqn.database:
+            parts.append(unquote_identifier(self.fqn.database))
+        if self.fqn.schema:
+            parts.append(unquote_identifier(self.fqn.schema))
+        parts.append(unquote_identifier(self.fqn.name))
+        return ".".join(parts)
```
sfc-gh-lkyc, line 126: It can be a method of the `FQN` class. It will also make unit testing easy.

src/snowflake/cli/_plugins/dcm/reporters/plan.py
```diff
@@ -0,0 +1,246 @@
...
+
+        SummaryLabels = namedtuple(
+            "SummaryLabels", ["created", "altered", "dropped", "header"]
+        )
+        labels = {
+            "plan": SummaryLabels("to create", "to alter", "to drop", "Planned"),
+            "deploy": SummaryLabels("created", "altered", "dropped", "Deployed"),
+        }[self.command_name]
+
+        if self._summary.created > 0:
```
sfc-gh-lkyc: I wonder whether it's a good idea to print these items only when the value is greater than 0, rather than always.

src/snowflake/cli/_plugins/dcm/reporters/plan.py
sfc-gh-lkyc: This comment actually can be applied to any place where there are `if-else` blocks, and we go to the moment when the operation name is different from what is expected.

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-jwilkowski: I need to distinguish between v2 plan api from the old format so I don't have the cursor anymore to pass to process. That's why I'm passing decoded `data`. 

I could move v1 vs v2 logic to PlanReporter class, but this felt easier to cleanup in the future without making things more complicated

src/snowflake/cli/_plugins/dcm/reporters/plan.py
sfc-gh-jwilkowski: Yeah, that would also be my assumption, but I'd rather be more defensive here as contracts can break intentionally or not. 
This is also our best effort to prepare for potential v3 version of this api, which might be different and not all clients migrate to newer cli version in time.

src/snowflake/cli/_plugins/dcm/reporters/plan.py
sfc-gh-jwilkowski: removed

src/snowflake/cli/_plugins/dcm/reporters/plan.py
sfc-gh-jwilkowski: removed 👌

src/snowflake/cli/_plugins/dcm/reporters/plan.py
sfc-gh-jwilkowski: done

src/snowflake/cli/_plugins/dcm/reporters/plan.py
sfc-gh-jwilkowski: FQN has already properties such as identifier/url_identifier/sql_identifier, which all are quite similar. I don't want to introduce another method to those already available for such a narrow use case. If we end up using this more widely, then I'd be happy to

src/snowflake/cli/_plugins/dcm/reporters/plan.py
sfc-gh-jwilkowski: done

src/snowflake/cli/_plugins/dcm/reporters/plan.py
sfc-gh-lkyc: I'm not a fan of defensive programming, I rather like strict rules, but I won't block this PR because of it. However, I temporarily removed this code locally, ran all unit tests, and none failed. Does it have test coverage?

sfc-gh-lkyc: I see you added tests. Thanks. I'm resolving the comment.

PR #2786, opened 2026-02-26
Fix `snow stage copy --recursive` for vstage (snow://) paths

sfc-gh-moczko:
## Summary

Fixes two bugs that caused `snow stage copy --recursive` to fail when downloading from vstage (`snow://`) paths, particularly DCM deployment stages like `snow://project/.../deployments/DEPLOYMENT$1/`.

- **Bug 1 — `iter_stage()` loses `snow://` prefix**: When `ls` returns plain relative paths (e.g., `deployments/DEPLOYMENT$1/manifest.yml`), `build_path()` misinterprets them as `@`-prefixed stage paths, generating invalid SQL (`get @deployments/...` instead of `get 'snow://project/...'`). Fixed by adding a vstage-aware branch that reconstructs full paths via `stage_path.root_path() / file["name"]`.

- **Bug 2 — `StagePath.relative_to()` is case-sensitive**: Snowflake normalizes unquoted identifiers to lowercase in `ls` output (`DEPLOYMENT$1` becomes `deployment$1`), but `PurePosixPath.relative_to()` is case-sensitive, causing `ValueError` when computing local directory structure. Fixed with case-insensitive part-by-part comparison.

### Files changed

| File | Change |
|---|---|
| `src/snowflake/cli/_plugins/stage/manager.py` | 2-line `elif` branch in `iter_stage()` for `snow://` paths |
| `src/snowflake/cli/api/stage_path.py` | Replaced `relative_to()` with case-insensitive implementation |
| `tests/stage/test_stage.py` | 4 new parametrized test cases across 2 test functions |
| `tests/stage/test_stage_path.py` | 8 new parametrized test cases across 3 test functions |

## Test plan

- [x] New unit tests reproduce both bugs (5 failures before fix, 0 after)
- [x] Full stage test suite passes (327 passed, 50 snapshots)
- [x] All pre-commit hooks pass (ruff, black, mypy, codespell)
- [x] Manual verification against real DCM project on Snowflake (7 files downloaded with correct directory structure)
- [ ] CI passes

### DIFF
--- a/src/snowflake/cli/_plugins/stage/manager.py
+++ b/src/snowflake/cli/_plugins/stage/manager.py
@@ -567,10 +567,12 @@ class StageManager(SqlExecutionMixin):
                 path = StagePath.get_user_stage() / file["name"]
             elif stage_path.is_git_repo():
                 path = self.build_path(file["name"])
-            elif stage_path.absolute_path().startswith(SNOW_PREFIX):
+            elif stage_path.is_vstage():
                 path = stage_path.root_path() / file["name"]
             else:
-                # Snowflake `ls` returns unqualified names; re-attach the original FQN.
+                # For @-stage paths: Snowflake ls returns unqualified names (e.g.
+                # "stage_name/path/file.sql"); re-attach the original FQN prefix so
+                # the result uses the correct qualified stage name.
                 file_name = file["name"]
                 parts = file_name.split("/", maxsplit=1)
                 relative_path = parts[1] if len(parts) > 1 else ""
--- a/src/snowflake/cli/api/stage_path.py
+++ b/src/snowflake/cli/api/stage_path.py
@@ -11,6 +11,11 @@ from snowflake.cli.api.project.util import (
 USER_STAGE_PREFIX = "~"
 SNOW_PREFIX = "snow://"
 
+# Snowflake unquoted identifiers contain only [A-Za-z0-9_$] and are normalized
+# to lowercase in ls output.  Components outside this charset came from quoted
+# identifiers whose case Snowflake preserves — those must be compared exactly.
+_UNQUOTED_IDENTIFIER_RE = re.compile(r"^[A-Za-z0-9_$]+$")
+
 
 class StagePath:
     def __init__(
@@ -69,6 +74,9 @@ class StagePath:
 
     def is_git_repo(self) -> bool:
         return self._is_git_repo
+
+    def is_vstage(self) -> bool:
+        return self._is_snow_prefixed_stage
 
     @property
     def git_ref(self) -> str | None:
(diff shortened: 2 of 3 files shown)


src/snowflake/cli/_plugins/stage/manager.py
```diff
@@ -567,6 +567,8 @@ def iter_stage(self, stage_path: StagePath):
                 path = StagePath.get_user_stage() / file["name"]
             elif stage_path.is_git_repo():
                 path = self.build_path(file["name"])
+            elif stage_path.absolute_path().startswith(SNOW_PREFIX):
+                path = stage_path.root_path() / file["name"]
             else:
                 # Snowflake `ls` returns unqualified names; re-attach the original FQN.
```
sfc-gh-olorek: Nit: the `else` branch comment now only applies to `@`-stage paths.  what do you think about
" For @-stage paths: Snowflake ls returns... "

src/snowflake/cli/api/stage_path.py
```diff
@@ -287,7 +287,25 @@ def quoted_absolute_path(self) -> str:
...
+        # deployment$1). PurePosixPath.relative_to() is case-sensitive, so we
+        # do a case-insensitive prefix check and return the actual-cased tail.
+        self_parts = self.path.parts
+        other_parts = stage_path.path.parts
+        if len(self_parts) < len(other_parts):
+            raise ValueError(
+                f"{self.path!r} is not in the subpath of {stage_path.path!r}"
+            )
+        for s, o in zip(self_parts[: len(other_parts)], other_parts):
+            if s.lower() != o.lower():
```
sfc-gh-olorek: I am unsure here about being case-insensite. Snowflake preservers case sensitivity for quoted identifiers: `"Data"/file.txt"` should be different than `"data"/file.txt"`.   we should handle this case

src/snowflake/cli/_plugins/stage/manager.py
```diff
@@ -567,6 +567,8 @@ def iter_stage(self, stage_path: StagePath):
                 path = StagePath.get_user_stage() / file["name"]
             elif stage_path.is_git_repo():
                 path = self.build_path(file["name"])
+            elif stage_path.absolute_path().startswith(SNOW_PREFIX):
```
sfc-gh-olorek: Nit, not a blocker: this can be encapsulated with new `is_vstage` method on `StagePath`

src/snowflake/cli/_plugins/stage/manager.py
sfc-gh-moczko: Fixed. The comment on the `else` branch now reads: "For @-stage paths: Snowflake ls returns unqualified names (e.g. \`stage_name/path/file.sql\`); re-attach the original FQN prefix so the result uses the correct qualified stage name."

src/snowflake/cli/api/stage_path.py
sfc-gh-moczko: Fixed. `relative_to()` now uses per-component case rules:
- **Unquoted identifiers** (only `[A-Za-z0-9_$]`) — Snowflake normalises these to lowercase in `ls` output, so we compare case-insensitively.
- **Quoted identifiers** (any character outside that set, e.g. a dot or space) — Snowflake preserves their case, so we compare exactly.

The detection regex `_UNQUOTED_IDENTIFIER_RE = re.compile(r"^[A-Za-z0-9_$]+$")` is applied to each path component pair. If either component has characters outside the unquoted charset, we fall through to an exact match. Added `test_relative_to_quoted_identifier_success` and `test_relative_to_raises_for_non_subpath` to cover both cases.

src/snowflake/cli/_plugins/stage/manager.py
sfc-gh-moczko: Done. Added `is_vstage()` method to `StagePath` and updated `iter_stage()` to use it (`elif stage_path.is_vstage():`). Also added parametrised `test_is_vstage` and `test_is_vstage_preserved_through_operations` to verify the method returns the correct value through `joinpath`, `parent`, and `root_path`.

PR #2790, opened 2026-02-27
feat: [SNOW-3148148] save json responses of dcm commands

sfc-gh-jwilkowski:
### Pre-review checklist

### Changes description
https://snowflakecomputing.atlassian.net/browse/SNOW-3148140

### DIFF
--- a/src/snowflake/cli/_plugins/dcm/commands.py
+++ b/src/snowflake/cli/_plugins/dcm/commands.py
@@ -236,7 +236,7 @@ def _process_plan_result(
         log.info(
             "Detected DCM plan result version 2 format.",
         )
-        reporter = PlanReporter(command_name=command_name, save_output=save_output)
+        reporter = PlanReporter(save_output=save_output, command_name=command_name)
         reporter.process_payload(data)
         return EmptyResult()
 
--- a/src/snowflake/cli/_plugins/dcm/manager.py
+++ b/src/snowflake/cli/_plugins/dcm/manager.py
@@ -12,11 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import logging
-from contextlib import contextmanager
 from pathlib import Path
-from typing import Generator, List
+from typing import List
 
-from snowflake.cli._plugins.dcm.models import MANIFEST_FILE_NAME
+from snowflake.cli._plugins.dcm.models import MANIFEST_FILE_NAME, SOURCES_FOLDER
+from snowflake.cli._plugins.dcm.utils import collect_output
 from snowflake.cli._plugins.stage.manager import StageManager
 from snowflake.cli.api.artifacts.upload import sync_artifacts_with_stage
 from snowflake.cli.api.commands.utils import parse_key_value_variables
@@ -32,59 +32,10 @@ from snowflake.cli.api.secure_path import SecurePath
 from snowflake.cli.api.sql_execution import SqlExecutionMixin
 from snowflake.cli.api.stage_path import StagePath
 
-SOURCES_FOLDER = "sources"
-OUTPUT_FOLDER = "out"
 log = logging.getLogger(__name__)
 
 
 class DCMProjectManager(SqlExecutionMixin):
-    @contextmanager
-    def _collect_output(
-        self, project_identifier: FQN, command_name: str = "plan"
-    ) -> Generator[str, None, None]:
-        """
-        Context manager for handling command output artifacts - creates temporary stage,
-        downloads files to out/<command_name>/ folder after execution.
-
-        Args:
-            project_identifier: The DCM project identifier
-            command_name: Name of the command, used for the output subdirectory
-
-        Yields:
-            str: The effective output path to use in the DCM command
-        """
-        stage_manager = StageManager()
-        temp_stage_fqn = FQN.from_resource(
-            ObjectType.DCM_PROJECT, project_identifier, "OUTPUT_TMP_STAGE"
-        )
-        log.info(
-            "Creating temporary output stage for DCM %s artifacts (project_identifier=%s, stage=%s).",
-            command_name,
-            project_identifier,
-            temp_stage_fqn.identifier,
-        )
-        stage_manager.create(temp_stage_fqn, temporary=True)
-        effective_output_path = StagePath.from_stage_str(
-            temp_stage_fqn.identifier
-        ).joinpath("/outputs")
-        local_output_path = SecurePath(OUTPUT_FOLDER) / command_name
-
-        try:
-            yield effective_output_path.absolute_path()
-        finally:
-            log.info(
-                "Downloading DCM %s artifacts from stage to local path (project_identifier=%s, stage_path=%s, local_path=%s).",
-                command_name,
-                project_identifier,
-                effective_output_path.absolute_path(),
-                local_output_path.resolve(),
-            )
-            local_output_path.mkdir(parents=True, exist_ok=True)
-            stage_manager.get_recursive(
-                stage_path=effective_output_path.absolute_path(),
-                dest_path=local_output_path.path,
-            )
-
     def deploy(
         self,
         project_identifier: FQN,
@@ -148,7 +99,7 @@ class DCMProjectManager(SqlExecutionMixin):
         query += self._get_from_stage_query(from_stage)
 
         if save_output:
-            with self._collect_output(project_identifier) as output_stage:
+            with collect_output(project_identifier) as output_stage:
                 query += f" OUTPUT_PATH {output_stage}"
                 result = self.execute_query(query=query)
         else:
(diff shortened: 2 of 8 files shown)


src/snowflake/cli/_plugins/dcm/utils.py
```diff
@@ -12,17 +12,56 @@
...
+    """Save raw JSON response to out/<command>.json."""
+    output_dir = SecurePath(OUTPUT_FOLDER)
+    output_dir.mkdir(exist_ok=True)
+    json_file = output_dir / f"{command_name}.json"
+    try:
+        if isinstance(raw_data, str):
+            json_file.write_text(raw_data)
+        else:
+            json_file.write_text(json.dumps(raw_data))
+    except Exception as e:
```
sfc-gh-kolszewski, line 62: I think we could have a test for `save_command_response` testing that it finishes gracefully if writing file throws an Exception.

src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -304,6 +320,8 @@ def plan(
     """
     Plans a DCM Project deployment (validates without executing).
     """
+    clear_command_artifacts("plan")
```
sfc-gh-kolszewski, line 323: So the artifacts are cleared even if no new results are going to be saved (ie save_output is False) - that is intended?

src/snowflake/cli/_plugins/dcm/reporters/base.py
```diff
@@ -64,8 +66,15 @@ def print_summary(self) -> None:
             cli_console.styled_message(renderable.plain, style=renderable.style)
         cli_console.styled_message("\n")
 
+    def _try_save_response(self, result_json: Dict[str, Any]) -> None:
+        """Save raw JSON response if save_output is enabled and raw data is available."""
+        if self.save_output and isinstance(result_json, dict):
```
sfc-gh-lkyc: Do we need to check the instance type? The parameter type already indicates that it's dict

src/snowflake/cli/_plugins/dcm/utils.py
```diff
@@ -83,6 +122,13 @@ def wrapper(*args: Any, **kwargs: Any):
             if data is None:
                 return func(*args, **kwargs)
 
+            # Lazy imports to avoid circular dependency with reporters.
+            from snowflake.cli._plugins.dcm.reporters import (
+                PlanReporter,
+                RefreshReporter,
+                TestReporter,
+            )
```
sfc-gh-lkyc, line 184: I don't know the codebase and its design well enough, but I think testing utilities, like `mock_dcm_response`, should live in the test sources. I guess that because of the current design, you may have a circular dependency.

src/snowflake/cli/_plugins/dcm/manager.py
```diff
@@ -55,30 +58,32 @@ def _collect_output(self, project_identifier: FQN) -> Generator[str, None, None]
...
+            command_name,
             project_identifier,
             temp_stage_fqn.identifier,
         )
         stage_manager.create(temp_stage_fqn, temporary=True)
         effective_output_path = StagePath.from_stage_str(
             temp_stage_fqn.identifier
         ).joinpath("/outputs")
-        local_output_path = SecurePath(OUTPUT_FOLDER)
+        local_output_path = SecurePath(OUTPUT_FOLDER) / command_name
```
sfc-gh-lkyc: The knowledge of where the output files are saved is spread across `utils.py` and `manager.py` (maybe sth else as well?). It would be beneficial to extract the output folder management logic to a separate class in a separate file, encapsulating the logic and paths concatenation.

src/snowflake/cli/_plugins/dcm/reporters/base.py
```diff
@@ -27,9 +28,10 @@
 
 
 class Reporter(ABC, Generic[T]):
-    def __init__(self) -> None:
+    def __init__(self, save_output: bool = False) -> None:
```
sfc-gh-lkyc, line 31: We don't need default here I believe, as all the subclasses provide a value . In general, I see setting the default value for `save_output` in multiple places in the code, and it's worth reducing it if possible.

tests/dcm/test_commands.py
```diff
@@ -1340,6 +1392,32 @@ def test_refresh_with_target_flag(
...
+        mock_dcm_manager().refresh.return_value = mock_cursor(
+            rows=[(json.dumps(refresh_result),)], columns=("result",)
+        )
+
+        with change_directory(tmp_path):
+            result = runner.invoke(["dcm", "refresh", "my_project", "--save-output"])
+
+            assert result.exit_code == 0, result.output
+            json_file = tmp_path / OUTPUT_FOLDER / "refresh.json"
+            assert json_file.exists()
```
sfc-gh-lkyc: Let's check the file content, like in other tests in this file. Can we extract the assertion logic to a separate method? It looks very similar across all scenarios.
The same comment applies to the `TestDCMTest` class.

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-jwilkowski: yes. My motivation is that artifacts should represent the last executed command (split by command type). If we wouldn't do that, then if a command failed it would be unclear if the artifacts come from last successful execution or the failed execution. If we always clear those files, then we are unambiguous about it.

src/snowflake/cli/_plugins/dcm/utils.py
sfc-gh-jwilkowski: This isn't a testing (testing like in unit testing) utility but rather a development utility/helper to mock a particular backend response to see how this would get formatted. IMO it does not belong to test sources as we always import files from src to tests and this would invert it

sfc-gh-lkyc: OK. My assumption was based on this:

> testing utility to test different reporting styles on mocked responses without touching the backend

src/snowflake/cli/_plugins/dcm/utils.py
```diff
@@ -12,17 +12,56 @@
...
+            json_file.write_text(json.dumps(raw_data))
+    except Exception as e:
+        log.error("Failed to save command response: %s", e)
+        return
+    log.info(
+        "Saved raw JSON response for command '%s' in %s.",
+        command_name,
+        json_file.resolve(),
+    )
+    cli_console.step(f"Artifacts saved to: {output_dir.resolve()}")
```
sfc-gh-lkyc: This line prints
```
Artifacts saved to: SecurePath("/Users/lkyc/Downloads/tst5/out")
```
I would expect sth like this
```
Artifacts saved to: /Users/lkyc/Downloads/tst5/out
```

src/snowflake/cli/_plugins/dcm/utils.py
sfc-gh-jwilkowski: added a test 👌

src/snowflake/cli/_plugins/dcm/reporters/base.py
sfc-gh-jwilkowski: removed

src/snowflake/cli/_plugins/dcm/manager.py
sfc-gh-jwilkowski: So I've moved logic that handles cleanup of out/ dir, saving raw responses and fetching files from backend into utils. I kept those as functions/context managers for now, as of now I don't see the need to group them under a class

src/snowflake/cli/_plugins/dcm/reporters/base.py
sfc-gh-jwilkowski: Yes, this default gets repeated 4 times, but if we remove it, I'd have to set it in 40+ unit tests for reporter classes :) I'd prefer to keep it that way

tests/dcm/test_commands.py
sfc-gh-jwilkowski: done

src/snowflake/cli/_plugins/dcm/utils.py
sfc-gh-jwilkowski: done, good spot!

tests/dcm/test_commands.py
```diff
@@ -19,6 +23,12 @@ def _analyze_response(files=None):
     return json.dumps({"files": files})
 
 
+def _assert_json_dumped(command: str, api_result: dict[str, Any], tmp_path: Path):
+    json_file = tmp_path / OUTPUT_FOLDER / f"{command}.json"
```
sfc-gh-lkyc, line 27: It's better not to use production code in tests:
```suggestion
    json_file = tmp_path / "out" / f"{command}.json"
```

tests/dcm/test_utils.py
```diff
@@ -0,0 +1,94 @@
+import json
+from unittest import mock
+
+from snowflake.cli._plugins.dcm.utils import (
+    OUTPUT_FOLDER,
```
sfc-gh-lkyc, line 5: Same as here: https://github.com/snowflakedb/snowflake-cli/pull/2790/changes#r2882626360

PR #2791, opened 2026-02-27
fix: add configurable file encoding to resolve #2759 (UnicodeDecodeError on JP Windows)

### LINKED ISSUE #2759: SNOW-3066557: UnicodeDecodeError when executing SQL files with UTF-8 encoding on Japanese Windows
### SnowCLI version

3.15.0

### Python version

3.11 (embedded in PyApp)

### Platform

Windows 10/11 (Japanese locale, CP932 default encoding)

### What happened

Snowflake CLI fails to execute SQL files containing non-ASCII characters (Japanese comments) with UTF-8 encoding on Japanese Windows environment, despite setting PYTHONUTF8=1 and PYTHONIOENCODING=utf-8.

### Console output

```Shell
Actual Result:
UnicodeDecodeError: 'cp932' codec can't decode byte 0x86 in position 88: illegal multibyte sequence

Error Traceback:
File "...\snowflake\cli\_plugins\sql\statement_reader.py", line 233, in files_reader
    stmts = split_statements(io.StringIO(f.read()), remove_comments)
UnicodeDecodeError: 'cp932' codec can't decode byte 0x86 in position 88: illegal multibyte sequence
```

### How to reproduce

Steps to Reproduce:

Create a SQL file with UTF-8 encoding containing Japanese comments:
Save the file as test.sql (UTF-8 encoding)

Run the command:

Actual Result:

Error Traceback:

Expected Result:
SQL file should be read with UTF-8 encoding and executed successfully.

What I've Tried:

✅ Set PowerShell to UTF-8 (chcp 65001) - No effect
✅ Set environment variables PYTHONUTF8=1 and PYTHONIOENCODING=utf-8 - No effect
✅ Modified PowerShell profile with UTF-8 settings - No effect
✅ Removed Japanese comments from SQL file - Workaround successful
Root Cause Analysis:

The issue occurs in statement_reader.py:233 where SecurePath.read() is called without specifying encoding. On Japanese Windows, the default encoding is CP932, not UTF-8.

The environment variables PYTHONUTF8 and PYTHONIOENCODING don't affect PyApp-bundled Python executables, as they use their own embedded Python runtime.

Proposed Solution:

Explicitly specify UTF-8 encoding when opening SQL files in [statement_reader.py](vscode-file://vscode-app/c:/Users/10001200792/AppData/Local/Programs/Microsoft%20VS%20Code/591199df40/resources/app/out/vs/code/electron-browser/workbench/workbench.html):

Or add encoding parameter support to SecurePath.read_text() method.

Related Documentation:

[Python's open() encoding parameter](vscode-file://vscode-app/c:/Users/10001200792/AppData/Local/Programs/Microsoft%20VS%20Code/591199df40/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
Similar issue was resolved for icacls command outputs (SnowflakeCLI_EncodingError.md in user documentation)
Workaround:
Remove non-ASCII characters from SQL files or use SnowSQL instead of Snowflake CLI.

sfc-gh-moczko:
## Summary
- Adds a configuration-driven encoding system (SNOWFLAKE_CLI_ENCODING_FILE_IO env var or [cli.encoding] file_io in config.toml) to allow users on non-UTF-8 locales (e.g., Japanese Windows with CP932) to specify the file I/O encoding
- Passes configured encoding through SecurePath.read_text() and SecurePath.open() at all file-reading callsites: SQL statement reader, Jinja rendering, Cortex commands, Snowpark requirements, and entity script templates
- Returns None (platform default) when unconfigured, preserving existing behavior for all current users

## Changes
- **New**: src/snowflake/cli/api/encoding.py -- get_file_io_encoding(), get_subprocess_encoding(), detect_encoding_environment()
- **Modified**: statement_reader.py -- from_file() and files_reader() now pass encoding (root cause of #2759)
- **Modified**: rendering/jinja.py -- read_file_content() and procedure_from_js_file()
- **Modified**: entities/utils.py -- render_script_template()
- **Modified**: cortex/commands.py and cortex/manager.py -- 5 read_text() callsites
- **Modified**: snowpark/commands.py -- _read_snowflake_requirements_file()
- **Modified**: api/config.py -- added ENCODING_SECTION_PATH constant

## Test plan
- [x] 12 unit tests for encoding.py (env var, config, precedence, detection, mismatch warning)
- [x] 2 new tests in test_statement_reader.py verifying UTF-8 Japanese SQL files work through both files_reader() and ParsedStatement.from_file()
- [x] 2 new tests in test_secure_path.py verifying read_text(encoding=) and open(encoding=) pass-through
- [x] Bug reproduction confirmed: CP932 locale + UTF-8 Japanese SQL -> UnicodeDecodeError without fix, succeeds with SNOWFLAKE_CLI_ENCODING_FILE_IO=utf-8
- [x] All pre-commit hooks pass (ruff, black, mypy, check-ast, codespell, license headers)

Closes #2759

### DIFF
--- a/tests/api/test_encoding.py
+++ b/tests/api/test_encoding.py
@@ -28,7 +28,7 @@ from snowflake.cli.api.exceptions import CliError
 
 def detect_encoding_environment() -> Dict[str, str]:
     """Detect and log encoding environment information (test helper only)."""
-    log = logging.getLogger(__name__)
+    log = logging.getLogger("snowflake.cli.encoding")
 
     env_info = {
         "filesystem": sys.getfilesystemencoding(),
@@ -119,7 +119,7 @@ class TestDetectEncodingEnvironment:
         monkeypatch.setattr("locale.getpreferredencoding", lambda: "cp1252")
         monkeypatch.delenv("SNOWFLAKE_CLI_ENCODING_FILE_IO", raising=False)
 
-        with caplog.at_level(logging.WARNING):
+        with caplog.at_level(logging.WARNING, logger="snowflake.cli"):
             detect_encoding_environment()
         assert "Encoding mismatch detected" in caplog.text
 


src/snowflake/cli/_plugins/cortex/commands.py
```diff
@@ -168,7 +169,10 @@ def complete(
     if text:
         prompt = text
     elif file:
-        prompt = SecurePath(file).read_text(file_size_limit_mb=DEFAULT_SIZE_LIMIT_MB)
+        prompt = SecurePath(file).read_text(
+            file_size_limit_mb=DEFAULT_SIZE_LIMIT_MB,
+            encoding=get_file_io_encoding(),
+        )
```
sfc-gh-olorek, line 175: Thanks for adding this! Just to be complete I noticed that nativeapp/artifacts.py and spcs/services/manager.py still uses the old version. Both would still fail on JP Windows. Could you also change it there?

src/snowflake/cli/api/encoding.py
```diff
@@ -0,0 +1,92 @@
...
+    env_encoding = os.environ.get("SNOWFLAKE_CLI_ENCODING_FILE_IO")
+    if env_encoding:
+        return env_encoding
+
+    try:
+        from snowflake.cli.api.config import get_config_value
+
+        return get_config_value("cli", "encoding", key="file_io")
+    except Exception:
+        return None
```
sfc-gh-olorek: In case of invalid encoding for exmpale: `SNOWFLAKE_CLI_ENCODING_FILE_IO=not-a-codec` the error will be generated late during file read. Could you add as a last step a verfication that the encoding is valid?:

```python
try:
  codecs.lookup(encoding)
except LookupError:
   raise CliError(
                f"Unknown encoding '{encoding}'. Check SNOWFLAKE_CLI_ENCODING_FILE_IO "
                f"or [cli.encoding] file_io in config.toml."
            )

File: src/snowflake/cli/api/encoding.py [hunk 3]
```diff
@@ -0,0 +1,92 @@
...
+    env_encoding = os.environ.get("SNOWFLAKE_CLI_ENCODING_SUBPROCESS")
+    if env_encoding:
+        return env_encoding
+
+    try:
+        from snowflake.cli.api.config import get_config_value
+
+        return get_config_value("cli", "encoding", key="subprocess")
+    except Exception:
+        return None
```
sfc-gh-olorek: This method is never used in the non-test codebase. Can we move it to the test package if it is a helper method or delete it if it not needed at all?

src/snowflake/cli/api/config.py
```diff
@@ -91,6 +91,7 @@ class Empty:
 PLUGINS_SECTION_PATH = [CLI_SECTION, PLUGINS_SECTION]
 PLUGIN_ENABLED_KEY = "enabled"
 FEATURE_FLAGS_SECTION_PATH = [CLI_SECTION, "features"]
+ENCODING_SECTION_PATH = [CLI_SECTION, "encoding"]
```
sfc-gh-olorek, line 94: This variable is never used. Could we delete it or acutally use it? :  see comment  about: `get_config_value`

src/snowflake/cli/api/encoding.py
```diff
@@ -0,0 +1,92 @@
...
+    3. None (use platform default)
+    """
+    env_encoding = os.environ.get("SNOWFLAKE_CLI_ENCODING_FILE_IO")
+    if env_encoding:
+        return env_encoding
+
+    try:
+        from snowflake.cli.api.config import get_config_value
+
+        return get_config_value("cli", "encoding", key="file_io")
```
sfc-gh-olorek: We can use `ENCODING_SECTION_PATH` for the hardcoded constants.

src/snowflake/cli/api/encoding.py
sfc-gh-olorek: `except Exception` with `return None` swallows all exceptions. What do you think about making the catch more granural (for example: `KeyError`, `MissingConfiguration`) and log the problem?

src/snowflake/cli/api/encoding.py
```diff
@@ -0,0 +1,92 @@
...
+    Returns dict with filesystem, default, and locale encoding values.
+    Logs a warning if encodings are inconsistent.
+    """
+    env_info = {
+        "filesystem": sys.getfilesystemencoding(),
+        "default": sys.getdefaultencoding(),
+        "locale": locale.getpreferredencoding(),
+    }
+
+    encodings = {v.lower().replace("-", "") for v in env_info.values()}
```
sfc-gh-olorek: Nit: codec name can also contain underscore for exmple: `utf_8`. Could you use Python's `codecs.lookup(name).name` instead of replacing the `-`?

src/snowflake/cli/_plugins/cortex/commands.py
sfc-gh-moczko: Done — added `encoding=get_file_io_encoding()` to `SecurePath.open()` in both `nativeapp/artifacts.py` and `spcs/services/manager.py`.

src/snowflake/cli/api/encoding.py
sfc-gh-moczko: Done — added `_validate_encoding()` which calls `codecs.lookup()` and raises `CliError` for unknown codec names. Validation fires immediately when the env var or config value is read, not deferred to file open time.

src/snowflake/cli/api/encoding.py
sfc-gh-moczko: Done — `detect_encoding_environment()` is moved to the test file as a local helper. It is no longer in `encoding.py`.

src/snowflake/cli/api/config.py
sfc-gh-moczko: Done — `ENCODING_SECTION_PATH` is now imported and used in `encoding.py`, removing the duplicate hardcoded path.

src/snowflake/cli/api/encoding.py
sfc-gh-moczko: Done — both `get_file_io_encoding()` and `get_subprocess_encoding()` now call `get_config_value(*ENCODING_SECTION_PATH, key=...)`.

src/snowflake/cli/api/encoding.py
sfc-gh-moczko: Done — replaced `except Exception` with `_MISSING_ENCODING_EXCEPTIONS = (KeyError, NonExistentKey, MissingConfigOptionError, ConfigSourceError)`, matching exactly the exceptions `get_config_value` raises when a key is absent. Added a `log.debug()` call on the not-configured path.

src/snowflake/cli/api/encoding.py
sfc-gh-moczko: Done — `detect_encoding_environment()` (now in the test file) uses `codecs.lookup(v).name` for canonical comparison instead of string replacement.

PR #2792, opened 2026-02-27
feat: [SNOW-2457402] Unify encoding support in SnowCLI

sfc-gh-kolszewski:
### Pre-review checklist

### Changes description
See https://snowflakecomputing.atlassian.net/browse/SNOW-2683933

This change adds new encoding configuration section with following keys:
- file_io - which controls the encoding used on file operations (read, write, open). Uses platform default if not set.
- subprocess - which controls the encoding used on commands run with subprocess. Uses platform default if not set.
- stdout - which controls the encoding used when CLI command output is written to stdout
- show_warnings - which controls whether a warning should be displayed if SnowCLI discovers encoding mismatch in the environment. True by default.

Encoding settings are read from the configuration by SecurePath's read_text, write_text and open methods. As such,  most places in the code where SecurePath was not used for these purposes were refactored to use SecurePath. I did not apply this change where encoding was already explicitly set as 'utf-8'

The change should be transparent to Unix users. Windows users can expect the warning to be displayed.


src/snowflake/cli/_plugins/nativeapp/codegen/setup/native_app_setup_processor.py
```diff
@@ -43,9 +43,10 @@
     PathMapping,
     ProcessorMapping,
 )
+from snowflake.cli.api.secure_path import SecurePath
 
 DEFAULT_TIMEOUT = 30
-DRIVER_PATH = Path(__file__).parent / "setup_driver.py.source"
+DRIVER_PATH = SecurePath(__file__).parent / "setup_driver.py.source"
```
sfc-gh-olorek: Bug: `DRIVER_PATH` changed from `Path` to `SecurePath`, but `DRIVER_PATH.read_text()` at line 169 is called without the required `file_size_limit_mb` argument — this is a `TypeError` at runtime.

Since this is an internal bundled script with no encoding concern, reverting back to `Path`.

sfc-gh-olorek (commented):
Three issues found during review — fixes committed on the branch.

tests/nativeapp/test_manager.py
```diff
@@ -856,7 +856,7 @@ def test_given_no_existing_pkg_when_create_app_pkg_then_success_and_respect_rele
     workspace_context,
     feature_flag,
 ):
-    mock_get_config_value.return_value = feature_flag
+    mock_get_config_value.side_effect = [None, feature_flag]
```
sfc-gh-olorek: Fragile: `side_effect = [None, feature_flag]` is order-dependent. If any additional `get_config_value` call is added to the production code path (e.g. another encoding config read), the list shifts and the test silently returns wrong values.

Replacing with a function-based `side_effect` that matches on `key` name, so it's resilient to changes in the number of config reads.

src/snowflake/cli/api/config.py
```diff
@@ -571,13 +655,19 @@ def _bool_or_unknown(value):
 
 
 def _read_config_file_toml() -> dict:
-    return tomlkit.loads(get_config_manager().file_path.read_text()).unwrap()
+    return tomlkit.loads(
+        SecurePath(get_config_manager().file_path).read_text(
```
sfc-gh-olorek: TOML files are always UTF-8 by spec. Since `SecurePath.read_text()` now applies the user's `file_io` encoding setting, a user who configures `file_io = "cp1252"` would corrupt the CLI's own config file on re-read.

Same issue applies to `_read_connections_toml` and `_update_connections_toml` below.

Adding explicit `encoding="utf-8"` to all three functions.

src/snowflake/cli/api/config.py
```diff
@@ -198,6 +203,48 @@ def config_init(config_file: Optional[Path]):
...
     create_initial_loggers()
+    if should_show_encoding_warnings():
+        detect_encoding_environment()
+
+
+def detect_encoding_environment():
+    """Detect encoding configuration and warn about mismatches"""
+    fs_encoding = sys.getfilesystemencoding().lower().replace("_", "-")
+    default_encoding = sys.getdefaultencoding().lower().replace("_", "-")
+    locale_encoding = locale.getpreferredencoding().lower().replace("_", "-")
```
sfc-gh-olorek: Small thing that could bite us: we normalize these three with `.lower().replace("_", "-")`, but `_validate_encoding` (and codecs in general) think in canonical names. The manual munging misses aliases — e.g. a platform that reports `utf8` won't normalize to `utf-8`, so it falls into the mismatch branch below and warns someone whose setup is actually fine. Could we route them through `codecs.lookup(enc).name` instead? That gives the canonical name (`utf-8`, `cp1252`, …) and keeps detection consistent with how we validate.

src/snowflake/cli/api/config.py
```diff
@@ -198,6 +203,48 @@ def config_init(config_file: Optional[Path]):
         _initialise_config(config_manager.file_path)
     _read_config_file()
     create_initial_loggers()
+    if should_show_encoding_warnings():
+        detect_encoding_environment()
```
sfc-gh-olorek, line 209: Since this lives in `config_init`, a Windows / non-UTF-8 user sees the warning on *every* command rather than once. It does mirror the existing permission-warning pattern and `show_warnings=false` is an escape hatch, so I won't block on it — but the full multi-line "recommended actions" block on every invocation gets noisy fast. Might be worth a once-per-process dedupe, or only printing the long form the first time.

tests_integration/test_encoding.py
```diff
@@ -0,0 +1,274 @@
...
+            finally:
+                os.chdir(init_dir)
+
+    @pytest.mark.integration
+    def test_spcs_image_registry_login_uses_subprocess_encoding(self, runner):
+        result = runner.invoke_with_connection(
+            ["spcs", "image-registry", "login"],
+            env={"SNOWFLAKE_CLI_ENCODING_SUBPROCESS": "utf-8"},
+        )
+        assert "Login Succeeded" in result.output or result.exit_code != 0
```
sfc-gh-olorek: This one is effectively a tautology — `"Login Succeeded" in output` **or** `exit_code != 0` is true whether or not the subprocess encoding does anything, so it can't fail for the reason the test is named after. Could we assert on something encoding-specific, or at least split the success vs failure expectations so a real regression actually shows up?

tests_integration/test_encoding.py
```diff
@@ -0,0 +1,274 @@
...
+        self, runner, tmp_path
+    ):
+        sql_file = tmp_path / "default_encoding.sql"
+        _write_file_with_encoding(sql_file, JAPANESE_SQL_CONTENT, "utf-8")
+
+        result = runner.invoke_with_connection_json(
+            ["sql", "-f", str(sql_file)],
+        )
+
+        assert result.json != [{"RESULT": JAPANESE_EXPECTED_VALUE}]
```
sfc-gh-olorek: The `!=` check is a little risky — it'll also pass if the command failed for an unrelated reason (connection, parse error, …), so we could end up with a green test that never touched the encoding path. If the intent is "UTF-8 bytes decoded as cp1252 come out as mojibake," asserting it *equals* the expected mojibake would confirm it failed for the right reason.

src/snowflake/cli/_plugins/nativeapp/codegen/snowpark/python_processor.py
```diff
@@ -218,7 +219,7 @@ def process(
                 if grant_statements is not None:
                     collected_output.append(grant_statements)
 
-                with open(sql_file, "a") as file:
+                with SecurePath(sql_file).open("a") as file:
```
sfc-gh-olorek: Subtle side effect worth a sanity check: when `sql_file` doesn't exist yet, `SecurePath.open("a")` touches it into existence with strict 0600 perms instead of the usual umask default (existing files are left alone). Almost certainly fine for these generated `__generated.sql` artifacts — just flagging in case anything downstream expects them to be group/world-readable.

src/snowflake/cli/api/secure_path.py
```diff
@@ -167,31 +167,59 @@ def mkdir(
...
         self._assert_exists_and_is_file()
         self._assert_file_size_limit(file_size_limit_mb)
-        log.info("Reading file %s", self._path)
-        return self._path.read_text(*args, **kwargs)
 
-    def write_text(self, *args, **kwargs):
+        if encoding is None:
+            from snowflake.cli.api.config import get_file_io_encoding
+
+            encoding = get_file_io_encoding()
```
sfc-gh-olorek, line 186: Minor / non-blocking: every text read/write/open now does a `get_file_io_encoding()` → `get_config_value()` + `codecs.lookup()` hop (same pattern in `write_text` and `open` just below). It's all in-memory and codecs caches internally, so perf is a non-issue — but this is what nudged `get_config_value` call counts up enough to need the `test_manager.py` mock rework. If we ever want to neaten it, resolving the encoding once per run would keep it contained.

sfc-gh-olorek (commented):
Second pass — a few more inline notes. All non-blocking except as called out; the normalization one in config.py is the only real correctness nit.

src/snowflake/cli/api/config.py
sfc-gh-kolszewski: Will take a look into that

src/snowflake/cli/api/config.py
sfc-gh-kolszewski: That is true. I was thinking that we can display a shorter message in the warning - something in the line of `WARNING: Encoding mismatch detected. Run 'snow helpers detect-encoding' for more details.`. Then we can have a separate helper command `snow helpers detect-encoding` that would display the longer, multiline message - this command was also proposed in the original [design doc] - WDYT?(https://docs.google.com/document/d/1x35mk7GqW7u1cpbmAXXFeJerUyZnW1FL2dZde7Ep_28/edit?tab=t.0#heading=h.ilf9xurhqjas)

src/snowflake/cli/_plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-kolszewski: Good catch. The switch to SecurePath ocurred here and in a couple of other places. This is technically a behavioral change and as you mentioned it could affect some pipelines of our customers if the permissions of files created by snowcli are narrowed down. On the other hand, perhaps the fact that we are not using strict 0600 perms for these files is an inconsistency in the first place.
I'm wondering what should be our approach here - I think we should discuss this

tests_integration/test_encoding.py
sfc-gh-kolszewski: Changed to equal comparison

tests_integration/test_encoding.py
sfc-gh-kolszewski: I've left just the `assert "Login Succeeded" in result.output` assertion and moved the test to a different place. This test mostly aimes to check that encoding does not mess up the `docker login` subprocess.

src/snowflake/cli/api/config.py
sfc-gh-kolszewski: Fixed

src/snowflake/cli/api/config.py
sfc-gh-kolszewski: Added the `snow helpers detect-encoding` command and reduced the warning to one-liner

src/snowflake/cli/api/secure_path.py
sfc-gh-kolszewski: I'd leave that as is then to reduce the scope of this change

sfc-gh-kolszewski:
Recent changes:
- added encoding support to places in codebase where file operations and subprocess running was added since this PR was opened
- addressed PR comments

src/snowflake/cli/_plugins/nativeapp/codegen/snowpark/python_processor.py
sfc-gh-kolszewski: Reverted the usage of SecurePath in favor of providing the encoding explicitly to avoid changing the permission mask

sfc-gh-olorek (dismissed):
lgtm

sfc-gh-olorek (approved):
lgtm

PR #2795, opened 2026-03-04
feat: [SNOW-3193772] save artifacts of raw analyze

sfc-gh-jwilkowski:
### Pre-review checklist

### Changes description
https://snowflakecomputing.atlassian.net/browse/SNOW-3193772

### DIFF
--- a/src/snowflake/cli/_plugins/dcm/manager.py
+++ b/src/snowflake/cli/_plugins/dcm/manager.py
@@ -110,7 +110,9 @@ class DCMProjectManager(SqlExecutionMixin):
         query += self._get_from_stage_query(from_stage)
 
         if save_output:
-            with collect_output(project_identifier) as output_stage:
+            with collect_output(
+                project_identifier, command_name="plan"
+            ) as output_stage:
                 query += f" OUTPUT_PATH {output_stage}"
                 result = self.execute_query(query=query)
         else:
--- a/tests_integration/test_dcm_project.py
+++ b/tests_integration/test_dcm_project.py
@@ -59,6 +59,15 @@ def assert_json_response_saved(command_name: str, project_root: Path):
     assert json_file.exists(), f"{command_name}.json was not created."
 
 
+def assert_artifacts_downloaded(artifacts_dir: Path):
+    local_files = []
+    for root, dirs, files in os.walk(artifacts_dir):
+        for file in files:
+            relative_path = os.path.relpath(os.path.join(root, file), artifacts_dir)
+            local_files.append(relative_path)
+    assert len(local_files) > 0, f"No artifact files were downloaded to {artifacts_dir}"
+
+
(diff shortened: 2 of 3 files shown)


tests/dcm/test_commands.py
```diff
@@ -5,7 +5,6 @@
 
 import pytest
 from snowflake.cli._plugins.dcm.models import DCMManifest
-from snowflake.cli._plugins.dcm.utils import OUTPUT_FOLDER
```
sfc-gh-jwilkowski, line 8: @sfc-gh-lkyc applied suggestions you've mentioned in the other PR

src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -350,9 +350,12 @@ def raw_analyze(
     from_location: SecurePath = from_option,
     variables: Optional[List[str]] = variables_flag,
     target: Optional[str] = target_option,
+    save_output: bool = save_output_option,
     **options,
 ):
     """Analyzes a DCM Project."""
+    clear_command_artifacts("raw-analyze")
```
sfc-gh-jwilkowski, line 357: I'm raw-analyze name, for the same reason we called this command raw-analyze - I don't want to use term analyze if there's a chance it'll be a public command

tests_integration/test_dcm_project.py
```diff
@@ -700,6 +700,57 @@ def test_dcm_end_to_end_workflow(
...
+        for root, dirs, files in os.walk(raw_analyze_artifacts):
+            for file in files:
+                relative_path = os.path.relpath(
+                    os.path.join(root, file), raw_analyze_artifacts
+                )
+                local_files.append(relative_path)
+
+        assert (
+            len(local_files) > 0
+        ), "No artifact files were downloaded to ./out/raw-analyze/"
```
sfc-gh-lkyc: These ten lines can be combined into a single method call, also used for the `plan` command and possibly others.

src/snowflake/cli/_plugins/dcm/manager.py
```diff
@@ -67,17 +67,28 @@ def raw_analyze(
...
             len(variables or []),
+            save_output,
         )
         query = f"EXECUTE DCM PROJECT {project_identifier.sql_identifier} ANALYZE"
         query += self._get_configuration_and_variables_query(configuration, variables)
         query += self._get_from_stage_query(from_stage)
-        return self.execute_query(query=query)
+
+        if save_output:
+            with collect_output(
```
sfc-gh-lkyc, line 84: Please remove the default value for the `command_name` parameter from the `collect_output`. It's now used for `plan` and `raw-analyze`, so the value is always provided, so the default does not make sense here - it can be even harmful if for some potential future call you can't provide the value.

PR #2797, opened 2026-03-04
fix: use execute_async() for async queries to prevent session cancellation

sfc-gh-moczko:
## Summary
- Replace `cursor.execute(stmt, _no_results=True)` with `cursor.execute_async(stmt)` in the async SQL execution path (`sql/manager.py`)
- The `_no_results=True` parameter (internal SnowSQL compat) does not register the query ID in the connector `_async_sfqids` tracking set, so `connection.close()` destroys the session and cancels in-flight async queries
- This fixes flaky `test_only_async_queries` and `test_mix` failures in `tests-trusted-ng` (Config V2) where the heavier config resolution timing made the race condition between query submission and connection teardown reproducible

## Root Cause
1. `cursor.execute(stmt, _no_results=True)` fires the query but does NOT add it to `_async_sfqids`
2. After each CLI invocation, the test runner calls `connection_cache.clear()` then `connection.close()`
3. The connector checks `_async_sfqids` -- empty -- so it calls `delete_session()`
4. The INSERT, still in-flight on the server, gets cancelled and the table ends up completely empty
5. `cursor.execute_async()` properly registers the query ID, preventing session destruction while async queries are pending

## Test plan
- [x] Added unit test `test_async_query_uses_execute_async` verifying `execute_async()` is called instead of `execute(_no_results=True)`
- [x] All 190 existing `tests/sql/` tests pass
- [ ] Integration tests `test_only_async_queries` and `test_mix` should now pass in both `tests-trusted` and `tests-trusted-ng`

### DIFF
--- a/src/snowflake/cli/_plugins/sql/manager.py
+++ b/src/snowflake/cli/_plugins/sql/manager.py
@@ -120,7 +120,7 @@ class SqlManager(SqlExecutionMixin):
         for stmt in compiled_statements:
             if stmt.execute_async:
                 cursor = self._conn.cursor(cursor_class=cursor_class)
-                cursor.execute(stmt.statement, _no_results=True)
+                cursor.execute_async(stmt.statement)
                 # only log query ID for consistency with SnowSQL
                 logger.info("Async execution id: %s", cursor.sfqid)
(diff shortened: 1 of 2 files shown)


sfc-gh-olorek (approved):
lgtm!

PR #2802, opened 2026-03-06
Add support for computed defaults

sfc-gh-lkyc:
### Pre-review checklist

### Changes description
https://snowflakecomputing.atlassian.net/browse/SNOW-3203048

### DIFF
--- a/src/snowflake/cli/api/project/schemas/template.py
+++ b/src/snowflake/cli/api/project/schemas/template.py
@@ -43,6 +43,8 @@ class ComputedValueResolvers:
     _resolvers: ClassVar[Dict[str, Callable[[], Optional[str]]]] = {
         "connection.account": _make_connection_resolver("account"),
         "connection.role": _make_connection_resolver("role"),
+        "connection.db": _make_connection_resolver("database"),
+        "connection.schema": _make_connection_resolver("schema"),
     }
 
     @classmethod
--- a/tests/test_init.py
+++ b/tests/test_init.py
@@ -74,6 +74,8 @@ TEMPLATE_YML_COMPUTED = _computed_template_yml(
 _COMPUTED_RESOLVER_PARAMS = [
     ("connection.account", "account", "account_name", "test_account_xyz"),
     ("connection.role", "role", "role_name", "dev_role"),
+    ("connection.db", "database", "db_name", "my_database"),
+    ("connection.schema", "schema", "schema_name", "my_schema"),
 ]
 
 


src/snowflake/cli/api/project/schemas/template.py
```diff
@@ -14,22 +14,56 @@
...
+from pydantic import BaseModel, Field, model_validator
+from snowflake.cli.api.exceptions import InvalidTemplateError, MissingConfigurationError
 from snowflake.cli.api.secure_path import SecurePath
 
+log = logging.getLogger(__name__)
+
+
+def _make_connection_resolver(key: str) -> Callable[[], Optional[str]]:
+    def resolver() -> Optional[str]:
+        from snowflake.cli.api.config import get_default_connection_dict
```
sfc-gh-jwilkowski: do we need a local import here?

src/snowflake/cli/api/project/schemas/template.py
```diff
@@ -14,22 +14,56 @@
...
+        "connection.role": _make_connection_resolver("role"),
+    }
     name: str = Field(..., title="Variable identifier")
     type: Optional[Literal["string", "float", "int"]] = Field(  # noqa: A003
         title="Type of the variable", default=None
     )
     prompt: Optional[str] = Field(title="Prompt message for the variable", default=None)
     default: Optional[Any] = Field(title="Default value of the variable", default=None)
+    default_computed: Optional[str] = Field(
+        title="Compute the default value dynamically. Supported: 'connection.account', 'connection.role'",
```
sfc-gh-jwilkowski: we could generate the list of supported computed_defaults from _COMPUTED_VALUE_RESOLVERS.keys()

src/snowflake/cli/api/project/schemas/template.py
```diff
@@ -41,14 +75,33 @@ def python_type(self):
...
+
+    @staticmethod
+    def _resolve_computed_value(key: str) -> Optional[str]:
+        resolver = TemplateVariable._COMPUTED_VALUE_RESOLVERS.get(key)
+        if resolver is None:
+            raise InvalidTemplateError(
+                f"Unknown default_computed value: '{key}'. "
+                f"Supported values: {list(TemplateVariable._COMPUTED_VALUE_RESOLVERS.keys())}"
+            )
+        return resolver()
```
sfc-gh-jwilkowski: I'd probably add a wide try/except block when running resolver, we don't know what might go wrong

tests/test_init.py
```diff
@@ -533,3 +534,251 @@ def test_jinja_blocks(runner, temporary_directory, test_projects_path, value, ex
...
+        # no-interactive: should fail because resolved value is None
+        result = runner.invoke(
+            ["init", project, "--template-source", template_root, "--no-interactive"]
+        )
+        assert result.exit_code == 1
+        assert "Cannot determine value of variable account_name" in result.output
+
+
+def _raise_missing_configuration_exception():
+    from snowflake.cli.api.exceptions import MissingConfigurationError
```
sfc-gh-jwilkowski: is a local import needed?

tests/test_init.py
```diff
@@ -533,3 +534,251 @@ def test_jinja_blocks(runner, temporary_directory, test_projects_path, value, ex
...
+        project = "project"
+        # no-interactive: should fail because resolved value is None
+        result = runner.invoke(
+            ["init", project, "--template-source", template_root, "--no-interactive"]
+        )
+        assert result.exit_code == 1
+        assert "Cannot determine value of variable account_name" in result.output
+
+
+def _raise_missing_configuration_exception():
```
sfc-gh-jwilkowski: please move all the helpers, fixtures etc to top of the file, so that we see all the tooling before actual tests start

src/snowflake/cli/api/project/schemas/template.py
sfc-gh-lkyc: No, I can use a standard import

tests/test_init.py
sfc-gh-lkyc: No. I added top-level.

src/snowflake/cli/api/project/schemas/template.py
sfc-gh-lkyc: The local import in resolver() is intentional — it's needed so that monkeypatch.setattr("snowflake.cli.api.config.get_default_connection_dict", ...) works in tests. A top-level import would capture the reference at import time, making it
  immune to monkeypatching.

src/snowflake/cli/api/project/schemas/template.py
sfc-gh-lkyc: Done

tests/test_init.py
sfc-gh-lkyc: Done. 

It's interesting because it's the opposite approach we use in Java code to see the scenarios first, and the tooling is a details that is not the primary thing :)

src/snowflake/cli/api/project/schemas/template.py
sfc-gh-lkyc: How would you like to handle the exception? Tue `_make_connection_resolver` resolver already handles `MissingConfigurationError` gracefully.

sfc-gh-jwilkowski: Sure, today that works for `_make_connection_resolver`, but you're making a precedence for adding new resolvers in future that might not always capture their exceptions. I think in case of unexpected errors I'd log a warning and return None.

BTW, a future extension of this idea would be to add a registry of resolvers, so that plugin authors can register their own, but I don't think it's in scope of this change, it's a fairly immature idea

sfc-gh-lkyc: I changed the logic to log a warning and not provide a default in such a scenario. The prompt is required.

src/snowflake/cli/api/project/schemas/template.py
sfc-gh-lkyc: Done

PR #2803, opened 2026-03-06
SNOW-3198917 DCM Plugin respects the format option

sfc-gh-lkyc:
### Pre-review checklist

### Changes description
https://snowflakecomputing.atlassian.net/browse/SNOW-3198917
When the `--format` option has a value other than `TABLE`, the requested format is respected in the DCM plugin, and the result set received from the backend is converted to the requested format. For `TABLE`, which is the default format, the command result is printed in a pretty way for some commands. 

Key Changes

1. Extracted RespectingColumnTypesRowMapper (output/types.py)
  - Refactored the VARIANT/OBJECT/ARRAY JSON-parsing logic out of QueryResult._process_columns into a reusable RespectingColumnTypesRowMapper class (implements a new RowMapper ABC).
  - CollectionResult now accepts an optional RowMapper, so any collection (not just QueryResult) can parse JSON columns in JSON_EXT format.
  - QueryResult is simplified — it delegates row mapping to the shared mapper.

  2. DCM plan & deploy commands (commands.py)
  - When output format is TABLE, behavior is unchanged (custom reporter prints, returns EmptyResult).
  - When format is JSON/JSON_EXT, returns a CollectionResult with the raw cursor data so the CLI framework serializes it properly.

  3. DCM test & refresh commands (reporters/base.py)
  - Reporter.process() now returns a CommandResult instead of None.
  - Same logic: TABLE format → EmptyResult (reporter handles display); JSON formats → CollectionResult with raw data.

  4. Manager type hints (manager.py)

  5. Test fixture fix (conftest.py)
  - MockCursor.description changed from a generator (yield from) to a direct return, fixing issues when description is accessed multiple times.

  6. Tests (test_commands.py)
  - Fixed columns args (added missing trailing commas to make them tuples).
  - Added 4 new tests (test_*_with_json_format_returns_raw_response) verifying that deploy, plan, refresh, and test return raw JSON when --format json is used.

### DIFF
--- a/src/snowflake/cli/_plugins/dcm/commands.py
+++ b/src/snowflake/cli/_plugins/dcm/commands.py
@@ -377,8 +377,7 @@ def raw_analyze(
         )
 
     reporter = AnalyzeReporter(save_output=save_output)
-    reporter.process(result)
-    return EmptyResult()
+    return reporter.process(result)
 
 
 @app.command(requires_connection=True)
--- a/tests/dcm/test_commands.py
+++ b/tests/dcm/test_commands.py
@@ -28,6 +28,26 @@ def _assert_json_dumped(command: str, api_result: dict[str, Any], tmp_path: Path
     assert json.loads(json_file.read_text()) == api_result
 
 
+def _mock_cursor_for_format(mock_cursor, data: dict, format_name: str):
+    columns: list[str] | list[dict[str, object]]
+    if format_name == "json":
+        columns = ["result"]
+    elif format_name == "json_ext":
+        columns = [{"name": "result", "type_code": 5}]
+    else:
+        raise ValueError(f"Unsupported format: {format_name}")
+    return mock_cursor(rows=[(json.dumps(data),)], columns=columns)
+
+
+def _assert_format_result(payload, expected_data, format_name: str):
+    if format_name == "json":
+        assert payload == [{"result": json.dumps(expected_data)}]
+    elif format_name == "json_ext":
+        assert payload == [{"result": expected_data}]
+    else:
+        raise ValueError(f"Unsupported format: {format_name}")
+
+
 @pytest.fixture
 def mock_dcm_manager():
     with mock.patch(
@@ -460,13 +480,7 @@ class TestDCMDeploy:
             assert result.exit_code == 0, result.output
             _assert_json_dumped("deploy", plan_response, tmp_path)
 
-    @pytest.mark.parametrize(
-        "format_name, columns, expect_parsed",
-        [
-            ("json", ("result",), False),
-            ("json_ext", [{"name": "result", "type_code": 5}], True),
-        ],
-    )
+    @pytest.mark.parametrize("format_name", ["json", "json_ext"])
     def test_deploy_with_json_formats_returns_response(
         self,
         mock_dcm_manager,
@@ -476,12 +490,10 @@ class TestDCMDeploy:
         mock_connect,
         project_directory,
         format_name,
-        columns,
-        expect_parsed,
     ):
         plan_response = {"version": 2, "changeset": []}
-        mock_dcm_manager().deploy.return_value = mock_cursor(
-            rows=[(json.dumps(plan_response),)], columns=columns
+        mock_dcm_manager().deploy.return_value = _mock_cursor_for_format(
+            mock_cursor, plan_response, format_name
         )
         mock_dcm_manager().sync_local_files.return_value = "TMP_STAGE"
         mock_manifest_load.return_value = _manifest_without_config()
@@ -491,8 +503,7 @@ class TestDCMDeploy:
 
         assert result.exit_code == 0, result.output
         payload = json.loads(result.output)
-        expected_value = plan_response if expect_parsed else json.dumps(plan_response)
-        assert payload == [{"result": expected_value}]
+        _assert_format_result(payload, plan_response, format_name)
 
 
 class TestDCMPlan:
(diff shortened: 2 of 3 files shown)


src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -220,7 +222,11 @@ def _process_plan_result(
         return CollectionResult([])
 
     first_row = rows[0]
-    first_value = list(first_row)[0] if first_row else None
+    first_value = None
+    if "result" in first_row:
+        first_value = first_row["result"]
+    elif "operations" in first_row:
+        first_value = first_row["operations"]
```
sfc-gh-jwilkowski: Can you confirm why we check for those keys specifically? Is it part of contract of plan api response v1 and v2 respectively?

src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -238,7 +244,9 @@ def _process_plan_result(
         )
         reporter = PlanReporter(save_output=save_output, command_name=command_name)
         reporter.process_payload(data)
-        return EmptyResult()
+        if get_cli_context().output_format == OutputFormat.TABLE:
+            return EmptyResult()
+        return CollectionResult(rows)
```
sfc-gh-jwilkowski: My only complaint would be returning rows here instead of data. This makes the json output much harder to read as whole response is treated as a string and thus is compacted instead of nicely indented etc like dict would, see:
```
snow dcm plan --format json_ext
[
    {
        "result": "{\"version\":2,\"metadata\":{\"timestamp\":\"2026-03-09T06:06:51.384Z\",\"query_id\":\"01c2e80e-010f-7a7b-0000-0f2d17fd55ca\",\"project_name\":\
...
```

sfc-gh-lkyc: I fully agree that it can be printed prettier; however, I use the core feature here. To mitigate it, some core changes would be needed. 
Using `rows` meets the requirements that we agreed on, that we don't print specific columns, but the whole result set.

I can check how to approach it, but if that's the only concern, I would go with this approach and improve it in the upcoming weeks. WDYT?

sfc-gh-lkyc: I suspect that the core logic does not recognize that the `result` column is of type `OBJECT` - that's why it's treated as text.

tests/conftest.py
```diff
@@ -470,7 +473,7 @@ def rowcount(self):
 
     @property
     def description(self):
-        yield from self._columns
+        return self._columns
```
sfc-gh-lkyc, line 476: It's how the real cursor works. It returns a list, as described here: https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-api#description

> This attribute is set to a list of [ResultMetadata](https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-api#label-python-connector-resultmetadata-object) objects.

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-lkyc: I've fixed it. Please take a look.

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-lkyc: I reverted back to use SnowflakeCursor

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-jwilkowski, line 1: Please apply the same idea to raw-analyze as well

src/snowflake/cli/api/output/types.py
```diff
@@ -16,12 +16,19 @@
...
 from enum import IntEnum
 
 from snowflake.cli.api.cli_global_context import get_cli_context
 from snowflake.cli.api.output.formats import OutputFormat
 from snowflake.connector import DictCursor
-from snowflake.connector.cursor import SnowflakeCursor
+from snowflake.connector.cursor import ResultMetadata, SnowflakeCursor
+
+
+class RowMapper(ABC):
```
sfc-gh-jwilkowski, line 28: Do we really need an ABC parent class if there's only 1 child class in sight? Feels like a premature abstraction

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-lkyc: Are you sure? What would `raw-analyze` return for the `TABLE` format? A table - similar to i.e., preview? If we change the behavior now, we may break the contract with AI tools using this command. I would ask first, and then adjust. WDYT?

I regret we didn't do it at the beginning, but we weren't aware of the formatting issue yet.

src/snowflake/cli/api/output/types.py
sfc-gh-lkyc: My goal was to build a contract/interface of the RowMapper, where `*Result` classes don't have to rely on specific implementations, following the strategy pattern.

Pros:
  - The interface is clear and focused (single responsibility)
  - It's used polymorphically in CollectionResult
  - It provides extensibility for future row mapping strategies (e.g., custom transformations)
 
Cons:
  - I don't see any :)

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-jwilkowski: as discussed: I don't want to change TABLE format of `raw-analyze` (let's keep it as is), but would like to fix `--format json/json_ext/csv`. TABLE/JSON/JSON_EXT would then return pretty much the same data, but with slightly different formatting.

src/snowflake/cli/api/output/types.py
sfc-gh-jwilkowski: I'm not saying that it's wrong, I would just say that it's an easy refactoring job to extract this ABC base from a single class _when it's needed_. Ultimately it's only my preference and I won't push back. LGTM 👌

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-lkyc: done

PR #2805, opened 2026-03-06
fix: escape glob special chars in directory paths during stage copy

sfc-gh-mdrach:
Directories with square brackets (e.g., [id], [slug]) failed during recursive stage copy with error 'File doesn't exist: [.../[id]/*]'.

The bug occurred because [id]/* was interpreted as a glob pattern matching 'i/*' or 'd/*' instead of the literal '[id]/*' directory.

Fix: Use glob.escape() before appending /* to directory paths. This only affects real directories (checked via .is_dir()), so user-provided glob patterns like 'src/[abc]' or '*.py' continue to work correctly.

Changes:
- manager.py:406,369-372: Add glob.escape() when appending wildcards
- test_stage.py: Add 4 tests for brackets and glob pattern preservation

Before
Screenshot 2026-03-06 at 1 44 36 PM

After
```
snow stage copy '/tmp/test-bug-repro/' @demo.public.test_bracket_stage --recursive -c qa6-mdrach

     +------------------------------------------------------------------------------+
     |         |         |         |         | source_ | target_ |         |        |
     |         |         | source_ | target_ | compres | compres |         | messag |
     | source  | target  | size    | size    | sion    | sion    | status  | e      |
     |---------+---------+---------+---------+---------+---------+---------+--------|
     | campaig | @demo.p | 13      | 16      | NONE    | NONE    | UPLOADE |        |
     | ns/[id] | ublic.t |         |         |         |         | D       |        |
     | /test.t | est_bra |         |         |         |         |         |        |
     | xt      | cket_st |         |         |         |         |         |        |
     |         | age/cam |         |         |         |         |         |        |
     |         | paigns/ |         |         |         |         |         |        |
     |         | [id]/te |         |         |         |         |         |        |
     |         | st.txt  |         |         |         |         |         |        |
     +------------------------------------------------------------------------------+

```

### Pre-review checklist

### Changes description
Fixed bug that caused literal filepaths with special characters like `[``]` to be interpreted as globs.


src/snowflake/cli/_plugins/stage/manager.py
```diff
@@ -366,11 +366,10 @@ def put(
         and switch back to the original role for the next commands to run.
         """
         if "*" not in str(local_path):
-            local_path = (
-                os.path.join(local_path, "*")
-                if Path(local_path).is_dir()
-                else str(local_path)
-            )
+            if Path(local_path).is_dir():
+                local_path = glob.escape(str(local_path)) + "/*"
```
sfc-gh-gbloom: If `local_path` has a trailing `/`, wouldn't we end up with `/tmp/path//*`? It might work in practice, but probably better to avoid it

tests/stage/test_stage.py
```diff
@@ -1555,3 +1557,156 @@ def test_stage_create_with_encryption_and_directory_options(
...
+        # Verify the pattern doesn't exist as a literal directory
+        assert not pattern_path.is_dir()
+
+        # Simulate what copy_to_tmp_dir does
+        if pattern_path.is_dir():
+            # Would escape if it was a real directory
+            glob_pattern = glob.escape(str(pattern_path)) + "/**/*"
+        else:
+            # Preserves pattern for globbing
+            glob_pattern = str(pattern_path)
```
sfc-gh-gbloom: Not sure I understand this test. We don't actually have a reference to `StageManager` here, no?

src/snowflake/cli/_plugins/stage/manager.py
sfc-gh-mdrach: sure, added

sfc-gh-jwilkowski: I'd use dedicated functions for appending paths like
```python
escaped_local_path = glob.escape(str(local_path))
local_path = os.path.join(escaped_local_path, "*")
```

src/snowflake/cli/_plugins/stage/manager.py
```diff
@@ -404,7 +403,7 @@ def copy_to_tmp_dir(self, local_path: Path, tmp_dir: Path):
 
         if local_path.is_dir():
             root = local_path
-            glob_pattern = str(local_path / "**/*")
+            glob_pattern = glob.escape(str(local_path)) + "/**/*"
```
sfc-gh-jwilkowski: similar to above

tests/stage/test_stage.py
```diff
@@ -1555,3 +1557,156 @@ def test_stage_create_with_encryption_and_directory_options(
...
+        test_file = bracket_dir / "test.txt"
+        test_file.write_text("test content")
+
+        sm = StageManager()
+        # Test uploading the directory with square brackets
+        sm.put(bracket_dir, "stageName")
+
+        # The path should be escaped to handle square brackets correctly
+        # Without escaping: file://.../campaigns/[id]/* would match nothing
+        # With escaping: file://.../campaigns/[[]id]/* matches the directory
```
sfc-gh-jwilkowski: I don't think those tests add much value, please remove them if they are not required

tests/stage/test_stage.py
```diff
@@ -1555,3 +1557,156 @@ def test_stage_create_with_encryption_and_directory_options(
...
+    tester = RecursiveUploadTester(temporary_directory)
+    tester.prepare(structure=structure)
+    tmp_created_by_copy = tester.execute(local_path=temporary_directory)
+
+    # Verify that directories with square brackets are included in the upload
+    actual_paths = [call["local_path"] for call in tester.calls]
+
+    # The key test: directories with square brackets should be processed correctly
+    # Without the fix, these would fail to upload because [id] would be interpreted
+    # as a glob pattern matching 'i' or 'd' characters
```
sfc-gh-jwilkowski: as above

tests/stage/test_stage.py
sfc-gh-mdrach: Which tests?

sfc-gh-jwilkowski: comments, my bad!

PR #2808, opened 2026-03-10
Snowflake Apps init/deploy commands

sfc-gh-gbloom:
### Pre-review checklist

### Changes description

This PR introduces a new `snow apps` command group for managing Snowflake Apps, gated behind the `ENABLE_SNOWFLAKE_APPS` feature flag. These new commands are still under development and will move to be under the existing `snow app` command in a later PR.

**Sub-commands**
1. `snow apps init` - Generates a `snowflake.yml` project file for a Snowflake App, auto-detecting available compute pools, external access integrations, and connection context (warehouse, database)
2. `snow apps deploy` - End-to-end build-and-deploy workflow: bundles source artifacts, uploads them to a stage, executes a container image build job, creates/updates an SPCS service, and polls until the app endpoint is ready

**New entity type: `snowflake-app`**
| Field | Description |
| --- | --- |
| `identifier` | database, schema, and name for the app |
| `meta` | title, description, and icon metadata (stored as a JSON comment on the service) |
| `artifacts` | src/dest glob mappings (reuses the existing bundle_artifacts helper) |
| `query_warehouse` | warehouse used by the service |
| `build_compute_pool` / `service_compute_pool` | compute pools for the build job and the running service |
| `build_eai` / `service_eai` | external access integrations for build and runtime |
| `artifact_repository` | artifact repository reference (image repository is currently used until artifact repositories are supported) |
| `code_stage` | stage name and encryption type for uploaded code |

**Deploy workflow**
1. Create schema if it doesn't exist.
2. Clear (or create) the code stage.
3. Bundle artifacts via bundle_artifacts and recursively upload directories to the stage.
4. Resolve the image repository URL (with local-registry rewriting for dev environments).
5. Drop any previous build job service, then execute a new EXECUTE JOB SERVICE with the sf-image-build builder image.
6. Poll build job status until DONE (or raise on FAILED).
7. Create the SPCS service (suspended) with app metadata comment, then alter its spec to use the freshly-built image.
8. Resume the service and poll until RUNNING.
9. Poll the endpoint URL until ingress provisioning completes, then return the public URL.

### DIFF
--- a/tests/apps/test_commands.py
+++ b/tests/apps/test_commands.py
@@ -683,6 +683,16 @@ class TestSnowflakeAppManager:
         assert mock_execute.call_count == 2
 
     @patch(EXECUTE_QUERY)
+    def test_get_service_endpoint_url_adds_https_prefix(self, mock_execute):
+        result_cursor = Mock()
+        result_cursor.fetchone.return_value = ("my-endpoint.snowflakecomputing.app",)
+        mock_execute.return_value = result_cursor
+
+        fqn = FQN(database="DB", schema="SCHEMA", name="SVC")
+        url = SnowflakeAppManager().get_service_endpoint_url(fqn)
+        assert url == "https://my-endpoint.snowflakecomputing.app"
+
+    @patch(EXECUTE_QUERY)
     def test_get_service_endpoint_url_not_found(self, mock_execute):
         result_cursor = Mock()
         result_cursor.fetchone.return_value = None
--- a/src/snowflake/cli/_plugins/apps/manager.py
+++ b/src/snowflake/cli/_plugins/apps/manager.py
@@ -448,5 +448,8 @@ serviceRoles:
         row = result.fetchone()
 
         if row:
-            return row[0]
+            url = row[0]
+            if url and not url.startswith(("http://", "https://")):
+                url = f"https://{url}"
+            return url
         return None


src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -0,0 +1,774 @@
...
+
+# Feature flags
+IS_PERSONAL_DB_SUPPORTED = False  # Will be enabled in the future
+
+# Default resource names for Snowflake Apps
+SNOW_APPS_COMPUTE_POOL = "SNOW_APPS_DEFAULT_COMPUTE_POOL"
+DEFAULT_EXTERNAL_ACCESS = "SNOW_APPS_DEFAULT_EXTERNAL_ACCESS"
+DEFAULT_ARTIFACT_REPOSITORY = "SNOW_APPS_DEFAULT_ARTIFACT_REPOSITORY"
+# TODO: Replace with artifact_repository from entity config once supported
+DEFAULT_IMAGE_REPOSITORY = "SNOW_APPS_DEFAULT_IMAGE_REPOSITORY"
```
sfc-gh-kbienkowski: I thought the CLI uses the values from snowflake.yml rather than trying to guess them. Shouldn't we have the defaults in the CoCo skill? These defaults will also change really soon.

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -0,0 +1,774 @@
...
+        code_stage: str,
+        image_repo_url: str,
+        app_id: str,
+        external_access_integration: Optional[str] = None,
+    ) -> None:
+        """Execute a build job service."""
+        spec = f"""spec:
+  containers:
+  - name: main
+    image: "/snowflake/images/snowflake_images/sf-image-build:0.0.1"
```
sfc-gh-kbienkowski: Similarly to the previous comment, isn't there a better place for the defaults, say a project template?

src/snowflake/cli/_plugins/apps/commands.py
sfc-gh-gbloom: No, there are different flows that generate `snowflake.yml`. We don't want to duplicate the logic for every AI agent, and we also need to support humans.
Moving the defaults to a config fetching system function makes more sense to me, but it can be a future improvement.

src/snowflake/cli/_plugins/apps/commands.py
sfc-gh-gbloom: This will soon go away once the builder service is ready, we only need it temporarily to have a working end-to-end flow

src/snowflake/cli/_app/commands_registration/builtin_plugins.py
```diff
@@ -36,6 +37,7 @@
 # plugin name to plugin spec
 def get_builtin_plugin_name_to_plugin_spec():
     plugin_specs = {
+        "apps": apps_plugin_spec,
```
sfc-gh-olorek, line 41: What do you think about adding a TODO note that this module will be later renamed (app) or merged together into _plugins/natvieapp later? This captures what description of your PR says.

sfc-gh-olorek: I am also thinking whether here or __init__.py is the correct place for this todo note

src/snowflake/cli/_plugins/apps/__init__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) 2024 Snowflake Inc.
```
sfc-gh-olorek: Nit: 2026

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -0,0 +1,774 @@
...
+from snowflake.cli.api.exceptions import CliError
+from snowflake.cli.api.feature_flags import FeatureFlag
+from snowflake.cli.api.identifiers import FQN
+from snowflake.cli.api.output.types import CommandResult, MessageResult
+from snowflake.cli.api.project.project_paths import ProjectPaths
+from snowflake.cli.api.project.util import get_env_username
+from snowflake.cli.api.secure_path import SecurePath
+from snowflake.cli.api.sql_execution import SqlExecutionMixin
+
+app = SnowTyperFactory(
```
sfc-gh-olorek, line 42: Currently command.py  contains: the commands, all helper functions, SnowflakeAppManager class with its methods, and the YAML generation logic.  What do you think about splitting into: manager.py (this is the pattern used by oter services), commands.py (just the definitions of the commands, and generate.py (YAML generation).

src/snowflake/cli/_plugins/apps/snowflake_app_entity.py
```diff
@@ -0,0 +1,24 @@
...
+from snowflake.cli._plugins.apps.snowflake_app_entity_model import (
+    SnowflakeAppEntityModel,
+)
+from snowflake.cli.api.entities.common import EntityBase
+
+
+class SnowflakeAppEntity(EntityBase[SnowflakeAppEntityModel]):
+    """Entity class for Snowflake App (snowflake-app) type."""
+
+    pass
```
sfc-gh-olorek, line 30: leaving this empty will mean that `snow ws deploy --entity-id my_app` will not work for this entity. Could you describe in the code comment whether you want to add the parity for it later or if this is just a design decision - explain it?

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -0,0 +1,774 @@
...
+    ) -> None:
+        """Alter a service with the built image spec."""
+        spec = f"""spec:
+  containers:
+    - name: main
+      image: "{image_url}"
+  endpoints:
+    - name: app-endpoint
+      port: 3000
+      public: true
```
sfc-gh-olorek: Instead of having this hardcoded could you add this default port in `SnowflakeAppEntityModel` with a default of 3000?

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -0,0 +1,774 @@
...
+    ctx.connection_context.validate_and_complete()
+    ctx.connection_context.update_from_config()
+    warehouse = ctx.connection_context.warehouse
+    database = ctx.connection_context.database
+
+    project_file.write_text(_generate_snowflake_yml(app_name, warehouse, database))
+    return MessageResult(f"Initialized Snowflake App project in {DEFINITION_FILENAME}.")
+
+
+@app.command(requires_connection=True)
```
sfc-gh-olorek, line 80: Per the current setup `requires_connection=True` you cannot generate `snowflake.yml` without a live connection. This is just to check whether compute_pool and external_access exist.  

 What do you think about splitting this into two modes:
1. offline(default). generate yaml with placeholder values for compute_pool. snow init - for native apps doesn't require a connection currently
2. connected (flag such as --detect-defaults) checks the account for existing resources. This will be leveraged by coco.

current behaviour couples project creation with infra discovery.

src/snowflake/cli/_plugins/apps/commands.py
sfc-gh-olorek: Just a small nit: I would prefer to have the best-practices already in-place from day one - just in case the "temp" state would be longer than anticipated

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -0,0 +1,774 @@
...
+        repo_url = row["repository_url"]
+
+        # Prod:
+        # repo_url = "pm-nax-consumer.registry.snowflakecomputing.com/gbloom_test_db/snow_app_my_test_app_gbloom/snow_apps_default_image_repository"
+
+        # Convert to local registry URL (replace .registry-dev. or .registry. with .registry-local.)
+        local_url = repo_url.replace(".registry-dev.", ".registry-local.")
+        local_url = local_url.replace(".registry.", ".registry-local.")
+        local_url = local_url.replace(".awsuswest2qa6.", ".")
+        local_url = local_url.replace(".awsuswest2qa3.", ".")
```
sfc-gh-olorek: will this work in production? why `.registry-local.` is needed?

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -0,0 +1,774 @@
...
+
+        if endpoint_url and "provisioning in progress" not in endpoint_url.lower():
+            return MessageResult(f"App ready at https://{endpoint_url}")
+
+        if endpoint_url:
+            cli_console.step(f"Endpoint status: {endpoint_url}")
+        else:
+            cli_console.step("Endpoint URL not yet available")
+
+        time.sleep(5)
```
sfc-gh-olorek: I think the very similar pattern of max_attempts with sleep is repeated 3 times. Can we use a lambda here to avoid repetition

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -0,0 +1,774 @@
...
+    return snowflake_apps[entity_id]
+
+
+class SnowflakeAppManager(SqlExecutionMixin):
+    """Manager for Snowflake App operations."""
+
+    def create_schema_if_not_exists(self, database: str, schema: str) -> None:
+        """Create schema if it doesn't exist."""
+        fqn = f"{database}.{schema}"
+        self.execute_query(f"CREATE SCHEMA IF NOT EXISTS {fqn}")
```
sfc-gh-olorek: This provides a way to do SQL injection. this provides a risk if snowflake.yaml contains malicious identifiers. this can be bypassed by:
```

      from snowflake.cli.api.identifiers import FQN

      fqn = FQN(database=database, schema=schema)

```

sfc-gh-olorek: Add teh minimum you should do fqn = f'"{database}"."{schema}"'

tests/apps/test_commands.py
```diff
@@ -0,0 +1,643 @@
...
+    )
+    def test_raises_when_not_found(self, _):
+        with pytest.raises(CliError, match="Entity 'my_app' not found"):
+            _get_entity("my_app")
+
+
+# ── _generate_snowflake_yml tests ─────────────────────────────────────
+
+
+class TestGenerateSnowflakeYml:
```
sfc-gh-olorek, line 344: The class does not test whether the generated YAML is parable and a valid project definition. what do you think about adding a tests that runs `render_definition_template` on the output

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -0,0 +1,774 @@
...
+# Feature flags
+IS_PERSONAL_DB_SUPPORTED = False  # Will be enabled in the future
+
+# Default resource names for Snowflake Apps
+SNOW_APPS_COMPUTE_POOL = "SNOW_APPS_DEFAULT_COMPUTE_POOL"
+DEFAULT_EXTERNAL_ACCESS = "SNOW_APPS_DEFAULT_EXTERNAL_ACCESS"
+DEFAULT_ARTIFACT_REPOSITORY = "SNOW_APPS_DEFAULT_ARTIFACT_REPOSITORY"
+# TODO: Replace with artifact_repository from entity config once supported
+DEFAULT_IMAGE_REPOSITORY = "SNOW_APPS_DEFAULT_IMAGE_REPOSITORY"
+
```
sfc-gh-olorek: The proposal document https://docs.google.com/document/d/1Upm3xNRHRAfC9_UpWT3tv9SQ2VFKi5mIFovlJ7KFLT4 mentions `snow app deploy [--upgrade]` but I think this flag was not implemented.

sfc-gh-olorek: We should either implement it or add a follow up tood.

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -0,0 +1,774 @@
...
+    If --entity-id is not specified and the project contains exactly one snowflake-app
+    entity, that entity will be used automatically.
+    """
+    _check_feature_enabled()
+    resolved_entity_id = _resolve_entity_id(entity_id)
+    entity = _get_entity(resolved_entity_id)
+
+    # ── Extract entity configuration ──────────────────────────────────
+    identifier = getattr(entity, "identifier", {})
```
sfc-gh-olorek: why can't we do `entity.identifier.name` the validation and defaults should not live in the command layer but in the model (entity) layer?

tests/apps/test_commands.py
```diff
@@ -0,0 +1,643 @@
...
+    def test_init_fails_when_feature_disabled(self, runner, tmp_path):
+        from tests_common import change_directory
+
+        with change_directory(tmp_path):
+            result = runner.invoke(["apps", "init", "--app-name", "my_app"])
+            assert result.exit_code == 1
+            assert "not available" in result.output
+
+
+class TestDeployCommand:
```
sfc-gh-olorek, line 743: The descritption of the PR mentions that integration tests were added. I see that here we don't really do integration tests. (The tests mocks sql calls). Deploy flow is a 12-step procedure where the interactions between steps matters.

src/snowflake/cli/_app/commands_registration/builtin_plugins.py
sfc-gh-gbloom: Added a comment

src/snowflake/cli/_plugins/apps/commands.py
sfc-gh-gbloom: Yes, prod works. These are fixes for local deployments:
The server presents a certificate like `*.registry-local.snowflakecomputing.com`, which doesn't match URLs with an additional subdomain like `org-account.qa6.registry-local.snowflakecomputing.com`.

src/snowflake/cli/_plugins/apps/commands.py

src/snowflake/cli/_plugins/apps/snowflake_app_entity.py
sfc-gh-gbloom: Done

src/snowflake/cli/_plugins/apps/commands.py
sfc-gh-gbloom: Yes, will follow up on this in a separate PR

src/snowflake/cli/_plugins/apps/commands.py
sfc-gh-gbloom: Good call, separated the logic to the 3 files as suggested

src/snowflake/cli/_plugins/apps/commands.py
sfc-gh-gbloom: Nice suggestion, I'll add it as a follow up improvement. For now all of the deployment flows require a connection.

src/snowflake/cli/_plugins/apps/commands.py
sfc-gh-gbloom: Right - removed unnecessary defensive calls

tests/apps/test_commands.py
sfc-gh-gbloom: Added

src/snowflake/cli/_plugins/apps/__init__.py
sfc-gh-gbloom: Fixed in all new files

src/snowflake/cli/_plugins/apps/commands.py
sfc-gh-gbloom: Extracted to a polling utility

src/snowflake/cli/_plugins/apps/commands.py
sfc-gh-gbloom: Done, using FQN everywhere now

tests/apps/test_commands.py
sfc-gh-gbloom: Added an integration test

PR #2809, opened 2026-03-10
SNOW-3202210:  Implement snow validate-image CLI command

sfc-gh-ajiang:
### Pre-review checklist

#### Background
Today, ML Jobs execute workloads using a Snowflake-managed ML runtime image (Container Runtime), with limited customization through dependency files or inline package installation. While this works for experimentation, it falls short of enterprise production requirements in four key areas:
1. Reproducibility -- Runtime package installation is non-deterministic; resolution may change over time, upstream dependencies may shift, and installations may succeed or fail depending on external repository state.
2. Security compliance -- Dynamic installation introduces unscanned dependencies and potential CVEs, making it difficult for customers to enforce allowlisting and vulnerability management policies.
3. CI/CD integration -- Customers need to build, scan, validate, and approve container images through their existing processes before deploying to production.
4. Environment control -- Development and production environments must be consistent over time.
Custom ML Runtime Images address these gaps by letting customers build governed, pre-validated container images based on Snowflake base runtime images and reference them in ML Job submissions. 

The design establishes the end-to-end flow for how custom runtime images are created, registered and validated, and managed for use in ML Jobs. A new account-level object, the Custom Runtime Environment (CRE), provides a governed reference (`cre@<name>`) to a validated, immutable image in a user-owned SPCS image repository.

Tech design doc: https://docs.google.com/document/d/1zQWrq62KWWtRlXTkDRZki7A12lLHkoT4w--KJS0ilpg/edit?tab=t.0#heading=h.rrhs7zj9hhot

We will implement snow `custom-image` validate in this PR

- Validates a locally available container image against the ML Jobs runtime compatibility contract before it is pushed to a Snowflake image repository. This is a client-side, pre-push check that uses Docker primitives (docker inspect, docker run) to verify the image satisfies minimum requirements for ML Job execution.

#### Checks performed

> | Check | Method | Requirement |
> |-------|--------|-------------|
> | Image exists | docker primitives | Image must be available locally |
> | Entrypoint | docker primitives | Must be /usr/local/bin/entrypoint.sh |
> | Environment variables | docker primitives | Must include DASHBOARD_PORT=12003 |
> | Required packages | pip list inside container | snowflake-ml-python, ray, and others |
> | Dependency health | pip check | No broken dependencies |
> | Vulnerability scan | grype | Advisory only (does not block registration) |
>
> The validation uses a YAML-driven check definition that can be extended as the runtime contract evolves. Checks are categorized as blocking (image will not work with ML Jobs if they fail) or advisory (reported but do not prevent usage).
>

- Test report: https://docs.google.com/document/d/1zQWrq62KWWtRlXTkDRZki7A12lLHkoT4w--KJS0ilpg/edit?tab=t.re7qb17krmd4

### DIFF
--- a/src/snowflake/cli/_plugins/custom_images/commands.py
+++ b/src/snowflake/cli/_plugins/custom_images/commands.py
@@ -29,8 +29,10 @@ app = SnowTyperFactory(
     help="Manages custom images for Snowpark Container Services.",
 )
 
-# Required to prevent Typer from collapsing the single-subcommand group into a flat command.
-app.callback()(lambda: None)
+
+@app.callback()
+def _callback():
+    pass
 
 
 @app.command(requires_connection=False)
--- a/src/snowflake/cli/_plugins/custom_images/manager.py
+++ b/src/snowflake/cli/_plugins/custom_images/manager.py
@@ -21,8 +21,6 @@ from typing import Any, Optional
 import yaml
 from click import ClickException
 
-CONFIG_DIR = Path(__file__).parent / "config"
-
 _FAIL_SEVERITIES = {"high", "critical"}
 
 
@@ -165,7 +163,7 @@ class CustomImageManager:
 
     def _get_image_info(self, image_name: str) -> Optional[dict]:
         """Get Docker image inspection info."""
-        returncode, stdout, stderr = self._run_docker_command(
+        returncode, stdout, _stderr = self._run_docker_command(
             ["docker", "inspect", image_name]
         )
         if returncode != 0:
@@ -250,13 +248,10 @@ class CustomImageManager:
         """Check if the image entrypoint matches the expected value and file exists."""
         # Get the actual configured entrypoint
         entrypoint = context.image_info.get("Config", {}).get("Entrypoint")
-        actual = (
-            entrypoint[0]
-            if isinstance(entrypoint, list) and entrypoint
-            else entrypoint
-            if not isinstance(entrypoint, list)
-            else None
-        )
+        if isinstance(entrypoint, list):
+            actual = entrypoint[0] if entrypoint else None
+        else:
+            actual = entrypoint
 
         # Check if entrypoint matches expected
         if actual != expected:
--- a/tests/custom_images/test_commands.py
+++ b/tests/custom_images/test_commands.py
@@ -12,31 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import json
 from unittest import mock
 
-
-def make_docker_inspect_response(
-    entrypoint: list[str] | None = None,
-    env_vars: list[str] | None = None,
-) -> str:
-    """Helper to create a mock docker inspect JSON response."""
-    return json.dumps(
-        [
-            {
-                "Config": {
-                    "Entrypoint": entrypoint,
-                    "Env": env_vars or [],
-                    "Labels": {},
-                }
-            }
-        ]
-    )
-
-
-def make_pip_list_response(packages: list[dict]) -> str:
-    """Helper to create a mock pip list JSON response."""
-    return json.dumps(packages)
+from tests.custom_images.test_helpers import (
+    FULL_PACKAGE_LIST,
+    create_mock_side_effect,
+    make_docker_inspect_response,
+    make_pip_list_response,
+)
 
 
 class TestValidateCustomImageCommand:
@@ -49,46 +32,13 @@ class TestValidateCustomImageCommand:
             entrypoint=["/usr/local/bin/entrypoint.sh"],
             env_vars=["DASHBOARD_PORT=12003"],
         )
-        pip_list_response = make_pip_list_response(
-            [
-                {"name": "snowflake-ml-python", "version": "1.0"},
-                {"name": "ray", "version": "2.0"},
-                {"name": "ipykernel", "version": "6.0"},
-                {"name": "sqlparse", "version": "0.5"},
-                {"name": "jinja2", "version": "3.0"},
-                {"name": "notebook", "version": "7.0"},
-                {"name": "ipython", "version": "8.0"},
-                {"name": "psutil", "version": "5.0"},
-                {"name": "snowflake-snowpark-python", "version": "1.0"},
-                {"name": "jupyter-server", "version": "2.0"},
-                {"name": "lightgbm-ray", "version": "0.1"},
-                {"name": "xgboost-ray", "version": "0.1"},
-                {"name": "snowflake", "version": "1.0"},
-                {"name": "snowflake.core", "version": "1.0"},
-                {"name": "snowflake-connector-python", "version": "3.0"},
-            ]
+        pip_list_response = make_pip_list_response(FULL_PACKAGE_LIST)
+
+        mock_run.side_effect = create_mock_side_effect(
+            inspect_response=inspect_response,
+            pip_list_response=pip_list_response,
+            pip_check_result=(0, "No broken requirements found."),
         )
-
-        def side_effect(*args, **kwargs):
-            cmd = args[0]
-            cmd_str = " ".join(cmd)
-            if cmd[0] == "docker":
-                if "inspect" in cmd:
-                    return mock.Mock(returncode=0, stdout=inspect_response, stderr="")
-                elif "run" in cmd:
-                    if "pip list" in cmd_str:
-                        return mock.Mock(
-                            returncode=0, stdout=pip_list_response, stderr=""
-                        )
-                    elif "pip check" in cmd_str:
-                        return mock.Mock(
-                            returncode=0,
-                            stdout="No broken requirements found.",
-                            stderr="",
-                        )
-            return mock.Mock(returncode=0, stdout="", stderr="")
-
-        mock_run.side_effect = side_effect
 
         result = runner.invoke(["custom-image", "validate", "test-image:latest"])
 
@@ -108,22 +58,10 @@ class TestValidateCustomImageCommand:
             ]
         )
 
-        def side_effect(*args, **kwargs):
-            cmd = args[0]
-            cmd_str = " ".join(cmd)
-            if cmd[0] == "docker":
-                if "inspect" in cmd:
-                    return mock.Mock(returncode=0, stdout=inspect_response, stderr="")
-                elif "run" in cmd:
-                    if "pip list" in cmd_str:
-                        return mock.Mock(
-                            returncode=0, stdout=pip_list_response, stderr=""
-                        )
-                    elif "pip check" in cmd_str:
-                        return mock.Mock(returncode=0, stdout="", stderr="")
-            return mock.Mock(returncode=0, stdout="", stderr="")
-
-        mock_run.side_effect = side_effect
+        mock_run.side_effect = create_mock_side_effect(
+            inspect_response=inspect_response,
+            pip_list_response=pip_list_response,
+        )
 
         result = runner.invoke(["custom-image", "validate", "test-image:latest"])
 
@@ -179,42 +117,12 @@ class TestValidateCustomImageCommand:
             entrypoint=["/wrong/entrypoint.sh"],
             env_vars=["DASHBOARD_PORT=12003"],
         )
-        pip_list_response = make_pip_list_response(
-            [
-                {"name": "snowflake-ml-python", "version": "1.0"},
-                {"name": "ray", "version": "2.0"},
-                {"name": "ipykernel", "version": "6.0"},
-                {"name": "sqlparse", "version": "0.5"},
-                {"name": "jinja2", "version": "3.0"},
-                {"name": "notebook", "version": "7.0"},
-                {"name": "ipython", "version": "8.0"},
-                {"name": "psutil", "version": "5.0"},
-                {"name": "snowflake-snowpark-python", "version": "1.0"},
-                {"name": "jupyter-server", "version": "2.0"},
-                {"name": "lightgbm-ray", "version": "0.1"},
-                {"name": "xgboost-ray", "version": "0.1"},
-                {"name": "snowflake", "version": "1.0"},
-                {"name": "snowflake.core", "version": "1.0"},
-                {"name": "snowflake-connector-python", "version": "3.0"},
-            ]
+        pip_list_response = make_pip_list_response(FULL_PACKAGE_LIST)
+
+        mock_run.side_effect = create_mock_side_effect(
+            inspect_response=inspect_response,
+            pip_list_response=pip_list_response,
         )
-
-        def side_effect(*args, **kwargs):
-            cmd = args[0]
-            cmd_str = " ".join(cmd)
-            if cmd[0] == "docker":
-                if "inspect" in cmd:
-                    return mock.Mock(returncode=0, stdout=inspect_response, stderr="")
-                elif "run" in cmd:
-                    if "pip list" in cmd_str:
-                        return mock.Mock(
-                            returncode=0, stdout=pip_list_response, stderr=""
-                        )
-                    elif "pip check" in cmd_str:
-                        return mock.Mock(returncode=0, stdout="", stderr="")
-            return mock.Mock(returncode=0, stdout="", stderr="")
-
-        mock_run.side_effect = side_effect
 
         result = runner.invoke(["custom-image", "validate", "test-image:latest"])
 
--- a/tests/custom_images/test_manager.py
+++ b/tests/custom_images/test_manager.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import json
+import subprocess
 from pathlib import Path
 from unittest import mock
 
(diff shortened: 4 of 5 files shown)


src/snowflake/cli/_plugins/custom_images/commands.py
```diff
@@ -0,0 +1,52 @@
...
+        ...,
+        help="Local Docker image identifier (image ID / hash).",
+    ),
+    **options,
+) -> CommandResult:
+    """
+    Validates a Docker image against Snowflake custom image requirements.
+    """
+    manager = CustomImageManager(config_path=DEFAULT_CONFIG_PATH)
+    report, output = manager.validate(image_hash=image_name)
```
sfc-gh-jijiang: nit: The image_name is passed as image_hash. This is a bit confusing because it’s not clear whether we expect an image name (e.g. repo:tag) or an image hash/ID. Could we standardize the terminology end-to-end?

src/snowflake/cli/_plugins/custom_images/manager.py
```diff
@@ -0,0 +1,477 @@
...
+    """Manager for custom image validation operations."""
+
+    def __init__(self, config_path: Path):
+        self.config_path = config_path
+        self.config = self._load_config(self.config_path)
+        self._check_handlers: dict[str, Any] = {
+            "environment_variables": self._check_environment_variables,
+            "python_packages": self._check_python_packages,
+            "dependency_health": self._check_dependency_health,
+            "vulnerability_scan": self._check_vulnerabilities,
```
sfc-gh-estevenson: Why are we checking for vulnerabilities in the validate script? I thought we decided that the customer will own vuln scanning. If we provide this, it implies we are responsible for catching vulns?

sfc-gh-ajiang: Thanks for catching up this. I updated the vulnerability check to be the optional check. The default behavior is to skip this check

src/snowflake/cli/_plugins/custom_images/commands.py
```diff
@@ -0,0 +1,64 @@
...
+        ...,
+        help="Local Docker image to validate. Accepts image name (e.g., 'myimage:latest') or image ID/hash.",
+    ),
+    image_type: str = typer.Option(
+        "cpu",
+        "--image-type",
+        help="Base image type: 'cpu' or 'gpu'. Defaults to 'cpu'.",
+    ),
+    **options,
+) -> CommandResult:
```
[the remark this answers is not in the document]
sfc-gh-ajiang: `**options` this is required in snowflake-cli framework

src/snowflake/cli/_plugins/custom_images/config/grype_cpu.yaml
sfc-gh-ajiang, line 1: For the whitelisted packages, I hardcoded them here. This is a public repo. We cannot hardcoded the credentials here to access the snowflake-image-builder. This means we should update the whitelist manually. Any concerns or any suggestions @sfc-gh-wesong

src/snowflake/cli/_plugins/custom_images/manager.py
sfc-gh-ajiang, line 1: we cannot detect the base image via docker primitives. So we do not have any checks on the base image. And for the image type, GPU or CPU, we rely on users' input. If they do not specify the image type in the command, we will assume that the image is a CPU image

src/snowflake/cli/_plugins/custom_images/manager.py
```diff
@@ -0,0 +1,522 @@
...
+            )
+            return result.returncode, result.stdout.strip(), result.stderr.strip()
+        except FileNotFoundError:
+            raise ClickException("Docker is not installed.")
+        except subprocess.TimeoutExpired:
+            raise ClickException("Docker command timed out.")
+
+    def _run_grype_command(self, image_name: str) -> tuple[int, str, str]:
+        """Run grype vulnerability scan. --fail-on high: exit non-zero only for High/Critical CVEs."""
+        cmd = ["grype", image_name, "-o", "json", "--fail-on", "high"]
```
sfc-gh-olorek, line 151: This leaves a door open for shell injection. Malicious image name could trigger unexpected behaviour. Consider validating imag_name before the command.

src/snowflake/cli/_plugins/custom_images/commands.py
```diff
@@ -0,0 +1,62 @@
...
+DEFAULT_CONFIG_PATH = CONFIG_DIR / "image_validation.yaml"
+
+
+app = SnowTyperFactory(
+    name="custom-image",
+    help="Manages custom images for Snowpark Container Services.",
+)
+
+# Required to prevent Typer from collapsing the single-subcommand group into a flat command.
+app.callback()(lambda: None)
```
sfc-gh-olorek: The registration system in typer_registration.py:78-95 already handles this automatically.

sfc-gh-olorek: No other plugin in the codebase does this manually

src/snowflake/cli/_plugins/custom_images/manager.py
```diff
@@ -0,0 +1,522 @@
...
+    ) -> ValidationResult:
+        """Check if the image entrypoint matches the expected value and file exists."""
+        # Get the actual configured entrypoint
+        entrypoint = context.image_info.get("Config", {}).get("Entrypoint")
+        actual = (
+            entrypoint[0]
+            if isinstance(entrypoint, list) and entrypoint
+            else entrypoint
+            if not isinstance(entrypoint, list)
+            else None
```
sfc-gh-olorek: This is very hard to read. Consider refactoring

sfc-gh-olorek: There is no test for the None case

sfc-gh-olorek: and there is no test for empty list

tests/custom_images/test_commands.py
```diff
@@ -0,0 +1,238 @@
...
+                if "inspect" in cmd:
+                    return mock.Mock(returncode=0, stdout=inspect_response, stderr="")
+                elif "run" in cmd:
+                    if "pip list" in cmd_str:
+                        return mock.Mock(
+                            returncode=0, stdout=pip_list_response, stderr=""
+                        )
+                    elif "pip check" in cmd_str:
+                        return mock.Mock(returncode=0, stdout="", stderr="")
+            return mock.Mock(returncode=0, stdout="", stderr="")
```
sfc-gh-olorek: this part of the code is repeated 3 times. Can we extract?

src/snowflake/cli/_plugins/custom_images/manager.py
```diff
@@ -0,0 +1,522 @@
...
+            [
+                "docker",
+                "run",
+                "--rm",
+                "--platform",
+                "linux/amd64",
+                context.image,
+                "bash",
+                "-c",
+                "pip list --format json",
```
sfc-gh-olorek, line 357: The docker run commands for pip list and pip check don't use --entrypoint "", unlike _check_entrypoint which does. This may cause silent fails. Please add it.

tests/custom_images/test_manager.py
```diff
@@ -0,0 +1,328 @@
...
+checks:
+  entrypoint: "/usr/local/bin/entrypoint.sh"
+  environment_variables:
+    - name: DASHBOARD_PORT
+      value: "12003"
+  python_packages:
+    - snowflake-ml-python
+    - ray
+  dependency_health: true
+  vulnerability_scan: true
```
sfc-gh-olorek: I think vulnerability scan is now  flag-driven, not config driven. Isn't it a dead code?

sfc-gh-olorek: Also I think now the production configs look differently, could we change this to reflect that?

tests/custom_images/test_commands.py
```diff
@@ -0,0 +1,238 @@
...
+                    "Labels": {},
+                }
+            }
+        ]
+    )
+
+
+def make_pip_list_response(packages: list[dict]) -> str:
+    """Helper to create a mock pip list JSON response."""
+    return json.dumps(packages)
```
sfc-gh-olorek: Those are duplicated between `test_commands.py` and `test_manager.py`. Can we extract to a seperate test util class?

src/snowflake/cli/_plugins/custom_images/manager.py
```diff
@@ -0,0 +1,522 @@
...
+        """Run a docker command and return (returncode, stdout, stderr)."""
+        try:
+            result = subprocess.run(
+                cmd, capture_output=True, text=True, timeout=timeout
+            )
+            return result.returncode, result.stdout.strip(), result.stderr.strip()
+        except FileNotFoundError:
+            raise ClickException("Docker is not installed.")
+        except subprocess.TimeoutExpired:
+            raise ClickException("Docker command timed out.")
```
sfc-gh-olorek, line 147: This crash CLI immediately , but all other failures (bad exit codes, broken JSON, image not found, etc.) are
  captured as ValidationResult entries.

Consider unifying.

sfc-gh-ajiang: These are intentionally different. `ClickException` is raised for infrastructure prerequisites (Docker/Grype not installed, timeouts) — meaning validation can't even be attempted. `ValidationResult` is used for image findings (wrong entrypoint, missing packages, CVEs) — meaning we checked and here's what we found. Reporting "Docker not installed" as a failed validation result would be misleading since no checks actually ran.

sfc-gh-ajiang: The `ClickException` only fires when `subprocess.run` itself fails (tool not found / timeout), meaning we can't communicate with Docker at all. Once Docker responds, everything is a `ValidationResult`.

src/snowflake/cli/_plugins/custom_images/manager.py
sfc-gh-ajiang: Shell injection is not possible here. We pass the command as a list to subprocess.run (not a string), and we never use shell=True. Per Python's [official documentation ](https://docs.python.org/3/library/subprocess.html#security-considerations)

>  "Unlike some other popen functions, this library will not implicitly choose to call a system shell. This means that all characters, including shell metacharacters, can safely be passed to child processes."

The image name is passed as a single element in the args list, so characters like ;, &&, $(), and backticks are treated as literal strings by the OS — no shell interprets them. Added a test that round-trips malicious names through a real subprocess to confirm they arrive unchanged.

sfc-gh-ajiang: And we also check if the image exists or not - so we do not need to validate the image name

`docker inspect` acts as a gatekeeper -- if the `image name` is bogus, malicious, or doesn't exist locally, `docker inspect` fails and validation short-circuits before any other commands run

src/snowflake/cli/_plugins/custom_images/manager.py
sfc-gh-ajiang: Thanks for catching this. Updated.

tests/custom_images/test_commands.py
sfc-gh-ajiang: yes, created `test_helper.py`

tests/custom_images/test_manager.py
sfc-gh-ajiang: Thanks for catching this. Updated.

tests/custom_images/test_commands.py
sfc-gh-ajiang: Thanks for catching this. Updated.

src/snowflake/cli/_plugins/custom_images/manager.py
sfc-gh-ajiang: The asymmetry is intentional — `_check_entrypoint` uses `--entrypoint`  because it needs to run a raw `test -f` to check whether the entrypoint file exists. The pip checks omit it so they run through the image's entrypoint, which sets up the runtime environment (PATH, virtualenv, etc.) that pip needs.

 Three reasons we don't add `--entrypoint ` to the pip checks:

1. Bypassing the entrypoint could break pip — the entrypoint script may activate a virtualenv or set PATH. Without it, pip might not be found or might inspect the wrong Python environment.

2. Re-passing the entrypoint explicitly isn't equivalent either — [per the Docker docs](https://docs.docker.com/engine/containers/run/#default-entrypoint): "Passing --entrypoint clears out any default command set on the image. That is, any CMD instruction in the Dockerfile used to build it." . Omitting the flag entirely lets Docker handle ENTRYPOINT/CMD interaction as it would in production.

3. Failures are not silent — both methods check return codes and report errors (e.g. "Failed to list Python packages. pip may not be installed or accessible").

> The default Docker behavior when --entrypoint is omitted: "The entrypoint refers to the default executable that's invoked when you run a container. You need to pass an explicit flag to override the entrypoint."

src/snowflake/cli/_plugins/custom_images/manager.py
```diff
@@ -0,0 +1,517 @@
...
+                "--platform",
+                "linux/amd64",
+                "--entrypoint",
+                "",
+                context.image,
+                "test",
+                "-f",
+                expected,
+            ]
+        )
```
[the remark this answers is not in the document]
sfc-gh-ajiang: The suggested fix (sh -c "test -f {expected}") doesn't solve the problem — if test isn't available in a minimal image, `sh` almost certainly isn't either. It also introduces an actual shell injection risk by passing expected through sh -c for shell interpretation, which the current code avoids by passing each argument as a separate list element to subprocess.run.

More importantly, this tool targets Snowflake custom images, which are full Linux environments with Python, pip, and bash. If test were missing, all the other checks (pip list, pip check, etc.) would also fail. Scratch-based or ultra-minimal images are not in scope.

src/snowflake/cli/_plugins/custom_images/manager.py
sfc-gh-ajiang: test: https://github.com/snowflakedb/snowflake-cli/pull/2809/changes#diff-4f24f789cd5727827d790f459fe4c2c8e88ab93648a1c16fd8943ea9ba05aff6R298

src/snowflake/cli/_plugins/custom_images/commands.py
sfc-gh-ajiang: But if I remove it, the test will fail.
```
FAILED tests/test_help_messages.py::test_required_arguments_dont_show_default[validate] - assert 2 == 0
 +  where 2 = <Result SystemExit(2)>.exit_code
FAILED tests/test_help_messages.py::test_help_messages[validate] - AssertionError: Usage: root [OPTIONS] COMMAND [ARGS]...
```
`typer_registration.py:78-95` does handle this automatically during `CLI registration`, but the test `test_help_messages.py` uses `load_only_builtin_command_plugins()` which bypasses registration and works directly with raw plugin specs. The `spec.command` cached property is evaluated from the raw typer instance
  before any auto-fix is applied, causing the path to resolve to ["validate"] instead of ["custom-image", "validate"].

The `@app.callback()` in `commands.py` — it ensures the typer instance is structurally correct (a group, not a collapsed single command) from the start, rather than relying on the registration-time fixup.

It looks like `test_help_messages.py` handles the command validation. Not being a regular contributor here, I focused on ensuring my code passes these existing tests to stay consistent with the current logic.

src/snowflake/cli/_plugins/custom_images/manager.py
sfc-gh-olorek: agree! closing this thread

src/snowflake/cli/_plugins/custom_images/commands.py
sfc-gh-olorek: agree. closing

src/snowflake/cli/_plugins/custom_images/manager.py
sfc-gh-olorek: thanks! closing

tests/custom_images/test_commands.py
sfc-gh-olorek: thanks! resolving

tests/custom_images/test_manager.py
sfc-gh-olorek: thanks! resolving

tests/custom_images/test_commands.py
sfc-gh-olorek: resolving!

src/snowflake/cli/_plugins/custom_images/manager.py
sfc-gh-olorek: agree. resolving

src/snowflake/cli/_plugins/custom_images/manager.py
sfc-gh-olorek: agree! closing

PR #2817, opened 2026-03-16
feat: [SNOW-3228385] upgrade python connector version

sfc-gh-jwilkowski:
### Pre-review checklist

### Changes description
https://snowflakecomputing.atlassian.net/browse/SNOW-3228385


tests/stage/test_stage.py
```diff
@@ -1676,7 +1676,7 @@ def test_stage_put_with_square_brackets_in_directory_name(mock_execute, mock_cur
         sm = StageManager()
         sm.put(bracket_dir, "stageName")
 
-        expected_path = glob.escape(str(bracket_dir.resolve())) + "/*"
+        expected_path = glob.escape(str(bracket_dir)) + "/*"
```
sfc-gh-jwilkowski, line 1679: those 2 changes were not related to this change, apparently nobody is running tests on macs anymore :)

src/snowflake/cli/api/config.py
```diff
@@ -176,6 +176,7 @@ def config_init(config_file: Optional[Path]):
 
     if config_file:
         get_cli_context_manager().config_file_override = config_file
+        _check_custom_config_permissions(config_file)
```
sfc-gh-lkyc, line 179: How is it related to bumping up the Python connector version?

src/snowflake/cli/api/config.py
```diff
@@ -489,6 +530,34 @@ def _check_default_config_files_permissions() -> None:
...
+
+
+def _enforce_strict_config_permissions() -> bool:
+    env_var = get_env_variable_name(
+        *FEATURE_FLAGS_SECTION_PATH, key="ENFORCE_STRICT_CONFIG_PERMISSIONS"
+    )
+    try:
+        return try_cast_to_bool(os.environ.get(env_var, "").lower())
+    except ValueError:
+        return False
```
sfc-gh-lkyc, line 594: Shouldn't we log a debug message, including the `ValueError` error details, here?

src/snowflake/cli/api/config.py
```diff
@@ -255,35 +256,75 @@ def _config_file():
...
+            config_manager.read_config(
+                skip_file_permissions_check=skip_permissions_check
+            )
         except ConfigSourceError as exception:
+            error_detail = (
+                str(exception.__cause__) if exception.__cause__ else exception.msg
+            )
             raise ClickException(
-                f"Configuration file seems to be corrupted. {str(exception.__cause__)}"
+                f"Configuration file seems to be corrupted. {error_detail}"
```
sfc-gh-lkyc, line 299: This now catches permission errors from the connector. The error message "Configuration file seems to be corrupted" may be misleading if the actual issue is permissions.
Consider checking the error type or message to provide more specific guidance.

src/snowflake/cli/api/config.py
```diff
@@ -255,35 +256,75 @@ def _config_file():
...
         if IS_WINDOWS:
-            warnings.filterwarnings(
-                action="ignore",
-                message="Bad owner or permissions.*",
-                module="snowflake.connector.config_manager",
+            _handle_windows_permission_warning(config_manager.file_path)
+        elif is_custom_config:
+            skip_permissions_check = not _handle_unix_custom_config_permissions(
+                config_manager.file_path
             )
```
sfc-gh-lkyc: It's a little bit confusing to me.

Also, the first check in  `_handle_unix_custom_config_permissions` is not needed, because `_check_custom_config_permissions` already raises an error if strict mode + bad permissions

src/snowflake/cli/api/config.py
```diff
@@ -489,6 +530,34 @@ def _check_default_config_files_permissions() -> None:
...
+
+    if (
+        _enforce_strict_config_permissions()
+        and config_file.exists()
+        and not file_permissions_are_strict(config_file)
+    ):
+        raise ConfigFileTooWidePermissionsError(config_file)
+
+
+def _enforce_strict_config_permissions() -> bool:
```
sfc-gh-lkyc: This doesn't check if the feature flag is defined in the config file, only in environment variables. Is this intentional? Should users be able to set this in their config file as well?

sfc-gh-jwilkowski: this should work only with env variable - the idea is that you can opt in for more strict config file checks, but if that's defined in the config file itself, then it's too late

src/snowflake/cli/api/config.py
sfc-gh-jwilkowski: Permission checks on CLI side are as strict as on connector side, so I don't expect those kinds of errors being raised here. The message is quite generic, but I wanted to keep it as close to the original as possible

src/snowflake/cli/api/config.py
sfc-gh-jwilkowski: we're moving from 3.x to 4.x connector series, which included a breaking change in how strict permission checks are. In general, CLI had strict checks in place before, but there was an assumption that if you're providing a custom config file, you're making an intentional operation and you trust that config file, so it does not have to have strict permissions.

src/snowflake/cli/api/config.py
sfc-gh-jwilkowski: windows handling did not change afair, just moved to a separate method
the second branch is new: CLI did not check permissions on custom config files, because we assume that users know what they provide if it's outside of the defaults. So if there's a custom config file, we skip permission checks on them. However, we let opting into those checks. If you have any suggestions about how to make it more clear, please LMK

src/snowflake/cli/api/config.py
sfc-gh-jwilkowski: added

src/snowflake/cli/api/config.py
```diff
@@ -489,6 +530,36 @@ def _check_default_config_files_permissions() -> None:
...
```
sfc-gh-olorek: Worth dropping a comment here explaining why this reads the env var directly instead of going through `FeatureFlag.is_enabled()`. The discussion in this PR explains it well (checking the config file is too late if the config itself might have bad permissions), but that context will be lost to anyone reading the code later — they'll just see it as inconsistent with how other feature flags work and try to "fix" it.

src/snowflake/cli/api/config.py
```diff
@@ -489,6 +530,36 @@ def _check_default_config_files_permissions() -> None:
             raise ConfigFileTooWidePermissionsError(CONFIG_FILE)
 
 
+def _check_custom_config_permissions(config_file: Path) -> None:
```
sfc-gh-olorek, line 555: Heads up: when strict mode is on and the file has wide permissions, this function raises `ConfigFileTooWidePermissionsError` during `config_init` — which means `_read_config_file` (and the `_handle_unix_custom_config_permissions` strict-mode branch) never executes for that scenario. The strict path in `_handle_unix_custom_config_permissions` returning `True` is effectively dead code in the error case.

Not a bug, but it'd be nice to have a short comment here or there noting that this is the early-exit gate and `_read_config_file` only deals with the non-strict / happy path.

tests/test_config.py
```diff
@@ -590,6 +595,30 @@ def test_no_error_when_init_from_non_default_config(
...
+    """When ENFORCE_STRICT_CONFIG_PERMISSIONS is enabled, strict permissions still work."""
+    with NamedTemporaryFile(suffix=".toml") as tmp:
+        config_path = Path(tmp.name)
+        config_path.chmod(0o600)  # Only owner can read/write
+        config_init(config_file=config_path)  # Should not raise
+
+
 @pytest.mark.parametrize(
     "content", ["[corrupted", "[connections.foo]\n[connections.foo]"]
 )
```
sfc-gh-olorek, line 654: The strict-flag tests look good, but I think we're missing coverage for the warning path when the flag is *off*. The existing parameterized test `test_too_wide_permissions_on_custom_config_file_causes_warning` checks that a warning fires, but it doesn't verify the new messaging (the env var name, the "future versions" language, etc.). Might be worth a quick `assert` on the warning text — that's the main thing users will actually read.

sfc-gh-olorek (commented):
A few things I noticed that weren't already called out in the existing review comments.

src/snowflake/cli/api/config.py
sfc-gh-lkyc: It would be helpful to include more details indicating that it cannot be parsed as a bool. For users, it will be more actionable.

sfc-gh-olorek (dismissed):
lgtm

src/snowflake/cli/api/config.py
```diff
@@ -489,6 +537,44 @@ def _check_default_config_files_permissions() -> None:
...
+    # the flag would defeat the purpose.
+    env_var = get_env_variable_name(
+        *FEATURE_FLAGS_SECTION_PATH, key="ENFORCE_STRICT_CONFIG_PERMISSIONS"
+    )
+    value = os.environ.get(env_var, "")
+    try:
+        return try_cast_to_bool(value.lower())
+    except ValueError:
+        log.debug("Could not parse %s value %r, defaulting to False", env_var, value)
+        return False
```
sfc-gh-lkyc, line 594: Consider adding test scenarios for it

src/snowflake/cli/api/config.py
sfc-gh-lkyc: I think there are mixed responsibilities, like determining whether permissions check should be skipped, and handling warnings. The other problem is overwriting the skip_permissions_check property, which does not make the understanding easier. The other problem is misaligned API of methods, whereas `_handle_windows_permission_warning` returns None, but `_handle_unix_custom_config_permissions` returns boolean. 

My proposal is to extract two separate strategies for permissions handling - one for Windows, and the other for Unix. Each strategy has two methods:
- `setup_warnings`
- `determine_skip_check`

And here is the code snipped:
```
class _PermissionsStrategy:
     """Base class for platform-specific permission check strategies."""

     def setup_warnings(
         self, config_path: Path, is_custom_config: bool, enforce_strict: bool
     ) -> None:
         raise NotImplementedError

     def determine_skip_check(
         self, is_custom_config: bool, enforce_strict: bool
     ) -> bool:
         raise NotImplementedError


 class _WindowsPermissionsStrategy(_PermissionsStrategy):
     """Windows-specific permission check strategy."""

     def setup_warnings(
         self, config_path: Path, is_custom_config: bool, enforce_strict: bool
     ) -> None:
         _handle_windows_permission_warning(config_path)

     def determine_skip_check(
         self, is_custom_config: bool, enforce_strict: bool
     ) -> bool:
         return False


 class _UnixPermissionsStrategy(_PermissionsStrategy):
     """Unix-specific permission check strategy."""

     def setup_warnings(
         self, config_path: Path, is_custom_config: bool, enforce_strict: bool
     ) -> None:
         if is_custom_config and not enforce_strict:
             _warn_about_wide_permissions_on_custom_config(config_path)

     def determine_skip_check(
         self, is_custom_config: bool, enforce_strict: bool
     ) -> bool:
         if not is_custom_config:
             return False
         return not enforce_strict


 def _determine_permissions_strategy() -> _PermissionsStrategy:
     """
     Determine the appropriate permissions strategy based on platform.

     Returns the platform-specific strategy instance.
     """
     return _WindowsPermissionsStrategy() if IS_WINDOWS else _UnixPermissionsStrategy()


 def _read_config_file():
     from snowflake.cli.api.cli_global_context import get_cli_context_manager

     config_manager = get_config_manager()
     is_custom_config = get_cli_context_manager().config_file_override is not None
     enforce_strict = _enforce_strict_config_permissions()

     permissions_strategy = _determine_permissions_strategy()

     with warnings.catch_warnings():
         permissions_strategy.setup_warnings(config_manager.file_path, is_custom_config, enforce_strict)
         skip_permissions_check = permissions_strategy.determine_skip_check(is_custom_config, enforce_strict)

         try:
             config_manager.read_config(
                 skip_file_permissions_check=skip_permissions_check
             )
         except ConfigSourceError as exception:
             error_detail = (
                 str(exception.__cause__) if exception.__cause__ else exception.msg
             )
             raise ClickException(
                 f"Configuration file seems to be corrupted. {error_detail}"
             )
```

It is more code, but cleaner separation making the understanding, and future changes easier. WDYT?

sfc-gh-jwilkowski: Thanks for this suggestion and illustrative example. I agree with the point about mixed concerns and styling issues, so I've applied those, but in a slightly different way. Personally I think that class based strategy pattern is an overkill here, however I'd happily explore that if we were to add more platforms or types of file locations. I don't see that coming anytime soon, so I want to stick to simpler if/else

sfc-gh-olorek (approved):

sfc-gh-lkyc (approved):

PR #2818, opened 2026-03-16
SNOW-3066960 Improved help for DCM commands

sfc-gh-lkyc:
### Pre-review checklist

### Changes description
This PR clarifies the description of DCM commands.

### DIFF
--- a/tests_integration/test_object.py
+++ b/tests_integration/test_object.py
@@ -147,22 +147,25 @@ def test_list_with_scope(runner, test_database, snowflake_session):
 
 
 @pytest.mark.integration
-def test_list_with_in_account_flag(runner, test_database, snowflake_session):
+def test_list_with_in_account_flag(runner, snowflake_session):
     """Test that --in-account flag lists objects at account scope."""
     result = runner.invoke_with_connection_json(
-        ["object", "list", "database", "--in-account"]
+        ["object", "list", "database", "--limit", "100", "--in-account"]
     )
     assert result.exit_code == 0, result.output
 
     # Verify the SQL query executed is correct by comparing with direct SQL
-    curr = snowflake_session.execute_string("show databases in account")
+    curr = snowflake_session.execute_string("show databases in account limit 100")
     expected = row_from_cursor(curr[-1])
 
     # Should have same structure
     assert result.json[0].keys() == expected[0].keys()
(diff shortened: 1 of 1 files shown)


src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -279,7 +285,7 @@ def deploy(
     **options,
 ):
     """
-    Applies changes defined in DCM Project to Snowflake.
+    Syncs local project changes to Snowflake by creating, altering, or dropping objects to match your definition files.
```
sfc-gh-jwilkowski: I'd avoid both applies and syncs to not introduce new terminology. I think that rest of your change makes good job at explaining what do we mean by deploying
```suggestion
    Deploys local project changes to Snowflake by creating, altering, or dropping objects to match your definition files.
```

src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -319,7 +325,7 @@ def plan(
     **options,
 ):
     """
-    Plans a DCM Project deployment (validates without executing).
+    Previews the objects that would be created, altered, or dropped to sync your project to Snowflake, without applying the changes. Provides a safe preview of the `deploy` command.
```
sfc-gh-jwilkowski: I'd avoid phrase `previews` as we have a preview command that means something very different and this might be confusing. 
```suggestion
    Prepares a plan of changes to Snowflake objects that would be created, altered or dropped, without applying the changes. Provides a safe validation before the `deploy` command.
```

src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -539,7 +545,7 @@ def preview(
     **options,
 ):
     """
-    Returns rows from any table, view, dynamic table.
+    Returns rows from any table, view, dynamic table defined in DCM project.
```
sfc-gh-jwilkowski: Is that true? Can't I reference any table?

sfc-gh-lkyc: Great comment! I was pretty sure we can't, but I verified that we can. I'm reverting this change.

sfc-gh-jwilkowski (dismissed):

PR #2820, opened 2026-03-17
Snowflake Apps validate, bundle and open commands

sfc-gh-gbloom:
### Pre-review checklist

### Changes description

Implementing `bundle`, `validate` and `open` for Snowflake Apps command.

1. Extracted the bundle step to a reusable `bundle` command
2. Added a `validate` command:
  a. Verifying a Dockerfile exists and exposes a port
  b. If the role is not ACCOUNTADMIN, verify `BIND SERVICE ENDPOINT` privilege
3. Added an `open` command to print the deployed app URL
4. Renamed `snow apps` to `snow __app` to avoid confusion with the existing `snow app` command
5. Making the `__app` commands hidden but functional when the feature flag is disabled

### DIFF
--- a/src/snowflake/cli/_plugins/apps/commands.py
+++ b/src/snowflake/cli/_plugins/apps/commands.py
@@ -19,8 +19,10 @@ from typing import Optional
 import typer
 from snowflake.cli._plugins.apps.generate import _generate_snowflake_yml
 from snowflake.cli._plugins.apps.manager import (
+    _APP_COMMAND_NAME,
     DEFAULT_IMAGE_REPOSITORY,
     DEFINITION_FILENAME,
+    EXPOSE_UNSUPPORTED_SYNTAX,
     SnowflakeAppManager,
     _find_dockerfile_expose_port,
     _get_entity,
@@ -38,7 +40,7 @@ from snowflake.cli.api.identifiers import FQN
 from snowflake.cli.api.output.types import CommandResult, MessageResult
 
 app = SnowTyperFactory(
-    name="__app",
+    name=_APP_COMMAND_NAME,
     help="Manages Snowflake Apps.",
     is_hidden=FeatureFlag.ENABLE_SNOWFLAKE_APPS.is_disabled,
 )
@@ -117,9 +119,10 @@ def validate(
 
     warnings: list[str] = []
 
-    project_paths = perform_bundle(resolved_entity_id, entity)
-
+    project_paths = None
     try:
+        project_paths = perform_bundle(resolved_entity_id, entity)
+
         # Validate Dockerfile has an EXPOSE directive
         bundle_root = project_paths.bundle_root
         dockerfile = bundle_root / "Dockerfile"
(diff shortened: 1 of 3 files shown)


src/snowflake/cli/_plugins/apps/manager.py
```diff
@@ -204,9 +203,74 @@ def _get_entity(entity_id: str) -> SnowflakeAppEntityModel:
...
+
 class SnowflakeAppManager(SqlExecutionMixin):
     """Manager for Snowflake App operations."""
 
+    def role_has_bind_service_endpoint(self, role: str) -> bool:
+        """Return True if *role* has the account-level BIND SERVICE ENDPOINT privilege."""
+        from snowflake.connector.cursor import DictCursor
+
+        cursor = self.execute_query(
+            f"SHOW GRANTS TO ROLE IDENTIFIER('{role}')", cursor_class=DictCursor
```
sfc-gh-olorek: If `role` contains `'` character it will fail or we may allow sql injection. Could you do a validation of the `role` before executing the statment?

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -77,6 +74,136 @@ def init(
...
+    Bundles the project, checks that a Dockerfile with an EXPOSE directive
+    exists, and verifies that the current role has the BIND SERVICE ENDPOINT
+    privilege required for deployment.
+    """
+    resolved_entity_id = _resolve_entity_id(entity_id)
+    entity = _get_entity(resolved_entity_id)
+
+    warnings: list[str] = []
+
+    project_paths = perform_bundle(resolved_entity_id, entity)
```
sfc-gh-olorek: If perform_bundle raises and exception the bundle might be partially created, and not cleaned. Consider putting int in the try block

src/snowflake/cli/_plugins/apps/manager.py
```diff
@@ -204,9 +203,74 @@ def _get_entity(entity_id: str) -> SnowflakeAppEntityModel:
...
+    directive.  Only simple ``EXPOSE <number>`` lines are recognised (the
+    ``/tcp`` and ``/udp`` suffixes are stripped).  Multi-port (``EXPOSE 3000
+    8080``) and range (``EXPOSE 3000-3005``) syntax is not supported.
+    """
+    dockerfile = bundle_root / "Dockerfile"
+    if not dockerfile.exists():
+        return None
+
+    expose_re = re.compile(r"^\s*EXPOSE\s+(\d+)(?:/(?:tcp|udp))?\s*$", re.IGNORECASE)
+    for line in dockerfile.read_text().splitlines():
```
sfc-gh-olorek: The method does not support multi port range (which is correctly documented). The problem is that the returned error message would be "Dockerfile does not contain an EXPOSE directive". consider creating another error message that throws that we don't support multi-port

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -77,6 +74,136 @@ def init(
...
+    fqn = entity.fqn
+    service_fqn = FQN(database=fqn.database, schema=fqn.schema, name=fqn.name)
+
+    manager = SnowflakeAppManager()
+    endpoint_url = manager.get_service_endpoint_url(service_fqn)
+
+    if not endpoint_url:
+        raise CliError(
+            f"No endpoint URL found for service {service_fqn}. "
+            f"Is the app deployed? Run 'snow __app deploy' first."
```
sfc-gh-olorek: after we migrate from `__app` to app we have to change it. Could you extract to a constant or add a comment reminder to do it after migration?

src/snowflake/cli/_plugins/apps/manager.py
sfc-gh-gbloom: Added these checks, and also changed the validation to only emit a warning to avoid false positives

src/snowflake/cli/_plugins/apps/commands.py

src/snowflake/cli/_plugins/apps/commands.py

src/snowflake/cli/_plugins/apps/manager.py
sfc-gh-gbloom: Good call, using `to_string_literal()` for escaping

PR #2823, opened 2026-03-18
feat: add `snow run` command for project scripts

sfc-gh-mborins:
### Pre-review checklist

### Changes description
Add a new `snow run` command that allows users to define and execute project-specific scripts in snowflake.yml or manifest.yml. Features include:

- Load script from either snowflake.yml or manifest.yml with conflict detection
- Track script source file and show in --list output
- Script definitions with cmd, description, shell, cwd, and env options
- Composite scripts using `run:` to sequence multiple scripts
- Variable interpolation with ${env.VAR} syntax
- CLI overrides with -D flag
- Dry-run mode and continue-on-error for composite scripts
- Extensive unit and integration tests covering various scenarios and edge cases

### DIFF
--- a/src/snowflake/cli/_plugins/run/commands.py
+++ b/src/snowflake/cli/_plugins/run/commands.py
@@ -18,7 +18,7 @@ from pathlib import Path
 from typing import List, Optional
 
 import typer
-from click import ClickException
+
 from snowflake.cli._plugins.run.manager import ScriptManager
 from snowflake.cli.api.cli_global_context import get_cli_context
 from snowflake.cli.api.commands.decorators import with_project_definition
@@ -26,7 +26,13 @@ from snowflake.cli.api.commands.flags import variables_option
 from snowflake.cli.api.commands.snow_typer import SnowTyperFactory
 from snowflake.cli.api.commands.utils import parse_key_value_variables
 from snowflake.cli.api.console import cli_console as cc
-from snowflake.cli.api.output.types import CommandResult, MessageResult
+from snowflake.cli.api.exceptions import CliError
+from snowflake.cli.api.output.types import (
+    CollectionResult,
+    CommandResult,
+    MessageResult,
+)
+from snowflake.cli.api.sanitizers import sanitize_for_terminal
 
 app = SnowTyperFactory(
     name="run",
@@ -102,37 +108,43 @@ def run_script(
             return MessageResult("No scripts defined in snowflake.yml or manifest.yml")
 
         source = manager.scripts_source or "project"
-        cc.message(f"Available scripts (from {source}):")
-        for name, script in scripts.items():
-            desc = script.description or "(no description)"
-            cc.message(f"  {name:20} {desc}")
-        return MessageResult("")
+        return CollectionResult(
+            [
+                {
+                    "name": name,
+                    "description": script.description or "",
+                    "source": source,
+                }
+                for name, script in scripts.items()
+            ]
+        )
 
     if not script_name:
         scripts = manager.list_scripts()
-        if scripts:
-            source = manager.scripts_source or "project"
-            cc.message(f"Available scripts from {source} (use --list for details):")
-            for name in scripts:
-                cc.message(f"  {name}")
-            return MessageResult("\nSpecify a script name to run, or use --list")
-        return MessageResult(
-            "No scripts defined. Add a 'scripts' section to snowflake.yml or manifest.yml"
-        )
+        if not scripts:
+            return MessageResult(
+                "No scripts defined. Add a 'scripts' section to snowflake.yml or manifest.yml"
+            )
+        source = manager.scripts_source or "project"
+        cc.message(f"Available scripts from {source} (use --list for details):")
+        for name in scripts:
+            cc.message(f"  {sanitize_for_terminal(name)}")
+        return MessageResult("\nSpecify a script name to run, or use --list")
 
-    vars_dict = {}
-    if var_overrides:
-        for v in parse_key_value_variables(var_overrides):
-            vars_dict[v.key] = v.value
+    vars_dict = (
+        {v.key: v.value for v in parse_key_value_variables(var_overrides)}
+        if var_overrides
+        else {}
+    )
 
     script = manager.get_script(script_name)
     if not script:
         available = list(manager.list_scripts().keys())
         if available:
-            raise ClickException(
+            raise CliError(
                 f"Script '{script_name}' not found. Available scripts: {', '.join(available)}"
             )
-        raise ClickException(
+        raise CliError(
             f"Script '{script_name}' not found. No scripts defined in snowflake.yml or manifest.yml"
         )
 
@@ -146,5 +158,7 @@ def run_script(
     )
 
     if not result.success:
-        raise typer.Exit(result.exit_code)
+        raise CliError(
+            f"Script '{script_name}' failed with exit code {result.exit_code}"
+        )
     return MessageResult("")
--- a/src/snowflake/cli/_plugins/run/commands.py
+++ b/src/snowflake/cli/_plugins/run/commands.py
@@ -116,7 +116,9 @@ def run_script(
             for name in scripts:
                 cc.message(f"  {name}")
             return MessageResult("\nSpecify a script name to run, or use --list")
-        return MessageResult("No scripts defined. Add a 'scripts' section to snowflake.yml or manifest.yml")
+        return MessageResult(
+            "No scripts defined. Add a 'scripts' section to snowflake.yml or manifest.yml"
+        )
 
     vars_dict = {}
     if var_overrides:
--- a/src/snowflake/cli/_plugins/run/commands.py
+++ b/src/snowflake/cli/_plugins/run/commands.py
@@ -1,0 +1,148 @@
+# Copyright (c) 2024 Snowflake Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import List, Optional
+
+import typer
+from click import ClickException
+from snowflake.cli._plugins.run.manager import ScriptManager
+from snowflake.cli.api.cli_global_context import get_cli_context
+from snowflake.cli.api.commands.decorators import with_project_definition
+from snowflake.cli.api.commands.flags import variables_option
+from snowflake.cli.api.commands.snow_typer import SnowTyperFactory
+from snowflake.cli.api.commands.utils import parse_key_value_variables
+from snowflake.cli.api.console import cli_console as cc
+from snowflake.cli.api.output.types import CommandResult, MessageResult
+
+app = SnowTyperFactory(
+    name="run",
+    help="Execute project scripts defined in snowflake.yml or manifest.yml.",
+)
+
+
+@app.command(name="run", no_args_is_help=False)
+@with_project_definition(is_optional=True)
+def run_script(
+    script_name: Optional[str] = typer.Argument(
+        None,
+        help="Name of the script to run.",
+    ),
+    list_scripts: bool = typer.Option(
+        False,
+        "--list",
+        "-l",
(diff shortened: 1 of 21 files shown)


src/snowflake/cli/_plugins/run/commands.py
```diff
@@ -0,0 +1,150 @@
...
+    """
+    ctx = get_cli_context()
+
+    project_root = ctx.project_root
+    if project_root is None:
+        project_root = Path.cwd()
+
+    manager = ScriptManager(project_root)
+
+    if list_scripts:
```
sfc-gh-turbaszek, line 104: Will this work with `--format=json`?

src/snowflake/cli/_plugins/run/commands.py
```diff
@@ -0,0 +1,150 @@
...
+                cc.message(f"  {name}")
+            return MessageResult("\nSpecify a script name to run, or use --list")
+        return MessageResult(
+            "No scripts defined. Add a 'scripts' section to snowflake.yml or manifest.yml"
+        )
+
+    vars_dict = {}
+    if var_overrides:
+        for v in parse_key_value_variables(var_overrides):
+            vars_dict[v.key] = v.value
```
sfc-gh-turbaszek:
```suggestion
vars_dict = {v.key: v.value for v in parse_key_value_variables(var_overrides)}
```
More pythonic

src/snowflake/cli/_plugins/run/commands.py
```diff
@@ -0,0 +1,150 @@
...
+
+    script = manager.get_script(script_name)
+    if not script:
+        available = list(manager.list_scripts().keys())
+        if available:
+            raise ClickException(
```
sfc-gh-jwilkowski: Please raise `CliError`

src/snowflake/cli/_plugins/run/manager.py
```diff
@@ -0,0 +1,366 @@
...
+        manifest_scripts, manifest_source = self._load_manifest_scripts()
+
+        if snowflake_scripts and manifest_scripts:
+            raise ClickException(
+                "Scripts defined in both manifest.yml and snowflake.yml.\n"
+                "Scripts must be defined in only one file per directory.\n\n"
+                "Recommendation: Move all scripts to one file.\n"
+                "- Use manifest.yml for DCM-focused projects\n"
+                "- Use snowflake.yml for app-focused projects"
+            )
```
sfc-gh-jwilkowski: Thinking out loud - is that a bad thing really? Maybe we can merge scripts from both sources?

src/snowflake/cli/_plugins/run/commands.py
```diff
@@ -0,0 +1,150 @@
...
+        scripts = manager.list_scripts()
+        if not scripts:
+            return MessageResult("No scripts defined in snowflake.yml or manifest.yml")
+
+        source = manager.scripts_source or "project"
+        cc.message(f"Available scripts (from {source}):")
+        for name, script in scripts.items():
+            desc = script.description or "(no description)"
+            cc.message(f"  {name:20} {desc}")
+        return MessageResult("")
```
sfc-gh-jwilkowski: You can also use EmptyResult

src/snowflake/cli/_plugins/run/commands.py
sfc-gh-jwilkowski: Would be better to use CollectionResult/ObjectResult to have support for different formatting options. I don't think json/csv would be very useful in this case, but having some response in those formats is better than nothing

sfc-gh-jwilkowski: Another issue I see is that introduces an inconsistency with other cli commands. As raised before, we usually have a dedicated command for listing objects

sfc-gh-jwilkowski: with a simple script like 
```
scripts:
  show_pwd:
    cmd: "pwd"
    description: "run pwd"
```
we get this:
```
snow run show_pwd --format json
/Users/jwilkowski/snowflake/dcm_test/dcm_manifest_v2
{
    "message": ""
}
```

src/snowflake/cli/_plugins/run/commands.py
```diff
@@ -0,0 +1,150 @@
...
+    if list_scripts:
+        scripts = manager.list_scripts()
+        if not scripts:
+            return MessageResult("No scripts defined in snowflake.yml or manifest.yml")
+
+        source = manager.scripts_source or "project"
+        cc.message(f"Available scripts (from {source}):")
+        for name, script in scripts.items():
+            desc = script.description or "(no description)"
+            cc.message(f"  {name:20} {desc}")
```
sfc-gh-jwilkowski: Use `sanitize_for_terminal()` before printing anything that comes from un untrusted source

src/snowflake/cli/_plugins/run/commands.py
```diff
@@ -0,0 +1,150 @@
...
+        scripts = manager.list_scripts()
+        if scripts:
+            source = manager.scripts_source or "project"
+            cc.message(f"Available scripts from {source} (use --list for details):")
+            for name in scripts:
+                cc.message(f"  {name}")
+            return MessageResult("\nSpecify a script name to run, or use --list")
+        return MessageResult(
+            "No scripts defined. Add a 'scripts' section to snowflake.yml or manifest.yml"
+        )
```
sfc-gh-jwilkowski: You can first deal with simple case to reduce indentation levels
```suggestion
        if not scripts:
            return MessageResult(
                "No scripts defined. Add a 'scripts' section to snowflake.yml or manifest.yml"
            )
        source = manager.scripts_source or "project"
        cc.message(f"Available scripts from {source} (use --list for details):")
        for name in scripts:
            cc.message(f"  {name}")
         return MessageResult("\nSpecify a script name to run, or use --list")
```

src/snowflake/cli/_plugins/run/commands.py
```diff
@@ -0,0 +1,150 @@
...
+        script_name,
+        extra_args=extra_args,
+        var_overrides=vars_dict,
+        dry_run=dry_run,
+        verbose=ctx.verbose,
+        continue_on_error=continue_on_error,
+    )
+
+    if not result.success:
+        raise typer.Exit(result.exit_code)
```
sfc-gh-jwilkowski: I'm a bit biased here, as on one hand I'd prefer this minimalistic way of reporting failures, but on the other hand we're not telling what failed or why. Also it's not consistent with how we report errors

src/snowflake/cli/_plugins/run/manager.py
```diff
@@ -0,0 +1,366 @@
...
+
+    def _parse_manifest_scripts(self, scripts_data: dict) -> Dict[str, ScriptModel]:
+        """Parse scripts section from manifest.yml into ScriptModel objects."""
+        result = {}
+        for name, script_def in scripts_data.items():
+            if isinstance(script_def, str):
+                result[name] = ScriptModel(cmd=script_def)
+            else:
+                result[name] = ScriptModel(**script_def)
+        return result
```
sfc-gh-jwilkowski, line 127: No error handling for malformed manifest.yaml. We've recently added `DCMManifest` class and afaik that's the only manifest we have right now, so might be better to use that or even extract some base fields such as manifest_version, project_type and scripts to a base class

src/snowflake/cli/_plugins/run/manager.py
```diff
@@ -0,0 +1,366 @@
...
+
+        def replace_var(match: re.Match) -> str:
+            var_name = match.group(1)
+            if var_name in variables:
+                value = variables[var_name]
+                if shell_mode:
+                    return shlex.quote(value)
+                return value
+            log.warning("Variable '${%s}' not found, leaving as-is", var_name)
+            return match.group(0)
```
sfc-gh-jwilkowski: I'd rather raise an error here, as using the var name could lead to dangerous outcomes.

src/snowflake/cli/_plugins/run/manager.py
```diff
@@ -0,0 +1,366 @@
...
+from snowflake.cli.api.console import cli_console as cc
+from snowflake.cli.api.constants import DEFAULT_SIZE_LIMIT_MB
+from snowflake.cli.api.project.schemas.project_definition import DefinitionV20
+from snowflake.cli.api.project.schemas.scripts import ScriptModel
+from snowflake.cli.api.secure_path import SecurePath
+from snowflake.cli.api.utils.models import ProjectEnvironment
+
+log = logging.getLogger(__name__)
+
+VARIABLE_PATTERN = re.compile(r"\$\{([^}]+)\}")
```
sfc-gh-jwilkowski, line 39: I'm worried about introducing another syntax for variable interpolation. Our users might get confused with those different syntax versions living side by side https://github.com/snowflakedb/snowflake-cli-templates/blob/main/streamlit_vnext_multi_page/snowflake.yml#L6-L9

sfc-gh-jwilkowski (commented):
I'd like to request some changes. I think that some design decisions require more discussion with @sfc-gh-jbilek (like --list as flag rather than standalone command)

src/snowflake/cli/_plugins/run/manager.py
sfc-gh-mborins: Good point on both fronts. The try/except already catches malformed YAML and logs it, so we won't crash on bad manifest files. Extracting shared base fields (manifest_version, project_type, scripts) from DCMManifest into a base class is a solid idea, but I think it's better suited for a follow-up PR to keep scope manageable here. Would that work?

src/snowflake/cli/_plugins/run/commands.py
sfc-gh-mborins: Fair point on consistency. Scripts aren't Snowflake objects so `add_object_command_aliases` doesn't fit directly, but we could add a dedicated `snow run list` subcommand to follow the pattern. Worth discussing with @sfc-gh-jbilek as you suggested. For now, I've updated `--list` to return `CollectionResult` so `--format=json` works properly.

One callout... anything we put after "run" breaks a potential script. That is why I am hesitant on the subcommands

src/snowflake/cli/_plugins/run/manager.py
sfc-gh-mborins: Valid concern. The distinction is: `<% ctx.env.XXX %>` is Jinja-style templating resolved at project load time (for entity definitions, identifiers, etc.), while `${var}` is resolved at script execution time and supports `--var` overrides for runtime flexibility. `${var}` is also the natural syntax users expect in shell commands (matches bash, Makefile, docker-compose conventions). I think clear documentation of the two scopes would address confusion. Happy to discuss if you'd prefer aligning to one syntax though.

src/snowflake/cli/_plugins/run/commands.py
sfc-gh-mborins: Agreed, we should be more informative. Updated to raise `CliError` with the script name and exit code: `Script 'deploy' failed with exit code 1`. This is consistent with how other CLI commands report errors.

src/snowflake/cli/_plugins/run/manager.py
sfc-gh-mborins: Good thought, but I'd prefer to keep single-source here. Having one file own all scripts avoids ambiguity about which file defines what, makes debugging simpler ("where is this script defined?" has exactly one answer), and prevents subtle ordering/shadowing issues if both files define the same name. The error message already guides users to pick one file.

src/snowflake/cli/_plugins/run/commands.py
sfc-gh-mborins: Done, switched to `CliError` throughout.

src/snowflake/cli/_plugins/run/commands.py
sfc-gh-mborins: Applied, thanks.

src/snowflake/cli/_plugins/run/commands.py
sfc-gh-mborins: Went with `CollectionResult` to support `--format=json` and `--format=csv` as jwilkowski suggested.

src/snowflake/cli/_plugins/run/commands.py
sfc-gh-mborins: Added `sanitize_for_terminal()` for script names displayed in the fallback path.

src/snowflake/cli/_plugins/run/commands.py
sfc-gh-mborins: Applied the early-return pattern, thanks.

src/snowflake/cli/_plugins/run/manager.py
sfc-gh-mborins: Agreed. Changed to raise `CliError` with a message pointing the user to define the variable or use `--var`.

PR #2848, opened 2026-04-01
Snowflake Apps: Improvements to setup command

sfc-gh-gbloom:
### Pre-review checklist

### Changes description

Improvements to the `__app setup` command:

1. Validating all required values upfront and report all missing ones at once, instead of silently falling back to null. Also simplified `generate.py` into a pure template renderer. All resolution logic now lives in `commands.py`.
2. Added `--compute-pool`, and `--build-eai` flags to set specific values.
3. Added `--dry-run` flag to print the resolved values without writing to `snowflake.yml`.
4. Each resolved value (database, warehouse, etc.) reports where it came from (CLI flag, account default, connection config).
5. The generated template now omits null values instead of specifying them as null


sfc-gh-olorek (commented):
Left a few inline comments — overall direction looks good, but spotted some things worth addressing.

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -68,27 +93,92 @@ def setup(
...
+    def _resolve(flag_val, config_key, conn_key=None, builtin=None):
+        """Return (value, source) using: flag > account default > connection config > builtin."""
+        if flag_val is not None:
+            return flag_val, "flag"
+        table_val = config_table.get(config_key)
+        if table_val:
+            return table_val, "account default"
+        if conn_key is not None:
+            # ctx.connection_context is only populated (without update_from_config) when
+            # the user explicitly passed the flag on the CLI, so this signals "flag" provenance.
```
sfc-gh-olorek, line 124: `DEFAULT_SCHEMA` is still defined in `generate.py` but nothing references it anymore. This `_resolve` call for schema has no `builtin` fallback, so if schema isn't in the config table or connection config you'd get `schema: None` in the YAML. Tests don't catch it because `_BASE_RESOLVED` always hardcodes `"schema": "SNOW_APPS"`.

Probably just needs `builtin=DEFAULT_SCHEMA` passed through here.

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -68,27 +93,92 @@ def setup(
         )
 
     project_file = Path.cwd() / DEFINITION_FILENAME
-    if project_file.exists():
+    if not dry_run and project_file.exists():
         return MessageResult(
             f"{DEFINITION_FILENAME} already exists. Skipping initialization."
         )
 
     ctx = get_cli_context()
```
sfc-gh-olorek, line 103: Nit: `flag_val` is checked with `is not None` but everything else (`table_val`, `ctx_val`, `conn_val`) uses truthiness. So an empty string `""` from a flag would be accepted, but the same empty string from the config table or connection config would be silently skipped. Probably fine in practice but feels like a subtle footgun — worth making them consistent one way or the other.

tests/apps/test_commands.py
```diff
@@ -39,7 +39,11 @@
 EXECUTE_QUERY = "snowflake.cli._plugins.apps.manager.SnowflakeAppManager.execute_query"
 OBJECT_EXISTS = "snowflake.cli._plugins.apps.manager._object_exists"
 GET_CLI_CONTEXT = "snowflake.cli._plugins.apps.manager.get_cli_context"
-GET_ENV_USERNAME = "snowflake.cli._plugins.apps.generate.get_env_username"
+GET_ENV_USERNAME = "snowflake.cli._plugins.apps.commands.get_env_username"
+COMMANDS_GET_COMPUTE_POOL = "snowflake.cli._plugins.apps.commands._get_compute_pool"
+COMMANDS_GET_EXTERNAL_ACCESS = (
```
sfc-gh-olorek: These two constants (`COMMANDS_GET_COMPUTE_POOL`, `COMMANDS_GET_EXTERNAL_ACCESS`) are defined but never used in any test. Looks like leftovers from a previous iteration — safe to remove.

src/snowflake/cli/__about__.py
```diff
@@ -14,11 +14,33 @@
 
 from __future__ import annotations
 
+import os
```
sfc-gh-olorek, line 17: This is a nice little feature but it's totally unrelated to the setup command changes. Would be cleaner as its own PR — makes the diff easier to review and keeps the git history focused.

src/snowflake/cli/_plugins/apps/generate.py
```diff
@@ -96,12 +57,10 @@ def _generate_snowflake_yml(
             type: snowflake-app
             identifier:
               name: {app_id.upper()}
-              database: {db_yaml}
+              database: {database}
```
sfc-gh-olorek, line 58: Heads up — the old template emitted `service_eai: null`, `description: null`, and `icon: null`. They're all gone now. If that's intentional cleanup, might be worth mentioning in the PR description so reviewers don't have to wonder. If anything downstream expects those keys to exist (even as null), this could break them.

src/snowflake/cli/__about__.py
sfc-gh-gbloom: Oops, accidentally brought in from another branch. Removed.

tests/apps/test_commands.py
sfc-gh-gbloom: Removed.

src/snowflake/cli/_plugins/apps/commands.py
sfc-gh-gbloom: Removed `DEFAULT_SCHEMA`.

`_resolve()` is rewritten in a downstream PR - keeping it to reduce conflicts.

src/snowflake/cli/_plugins/apps/commands.py
sfc-gh-gbloom: Same here, `_resolve()` is rewritten in a downstream PR - keeping it to reduce conflicts

src/snowflake/cli/_plugins/apps/generate.py
sfc-gh-gbloom: Intentional - updated the PR

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -71,27 +95,92 @@ def setup(
...
+        return None, "missing"
+
+    if IS_PERSONAL_DB_SUPPORTED:
+        database_resolved = (f"USER${get_env_username().upper()}", "personal db")
+    else:
+        database_resolved = _resolve(None, "database", conn_key="database")
+
+    resolved = {
+        "database": database_resolved,
+        "schema": _resolve(None, "schema", conn_key="schema"),
```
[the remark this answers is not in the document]
sfc-gh-gbloom: Intentional

sfc-gh-olorek (approved):
lgtm

PR #2858, opened 2026-04-06
Snowflake Apps:  Application service

sfc-gh-gbloom:
### Pre-review checklist

### Changes description

When deploying a Snowflake App with an artifact repository, use the new `APPLICATION SERVICE` SQL instead of the temporary system function.

Creating an application service:
```sql
CREATE APPLICATION SERVICE <app_service_fqn>
  FROM [ ARTIFACT REPOSITORY <repo_name> ] PACKAGE <package_name>
```

If the application already exists, upgrade to keep the same URL:
```sql
ALTER APPLICATION SERVICE <app_service_fqn> UPGRADE
  [ TO VERSION { <version_alias> } ]
```


src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -384,98 +481,163 @@ def deploy(
...
-                poll_fn=lambda: manager.get_build_status(build_job_fqn),
-                done_states={"DONE"},
-                error_states={"FAILED", "IDLE"},
-                known_pending_states={"PENDING", "RUNNING"},
-                timeout_message=f"Build timed out. Check service logs: {build_job_fqn}",
-            )
+        endpoint_url = desc.get("url", "")
+        if endpoint_url and not endpoint_url.startswith(("http://", "https://")):
+            endpoint_url = f"https://{endpoint_url}"
+        return MessageResult(f"App ready at {endpoint_url}")
```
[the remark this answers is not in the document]
sfc-gh-gbloom: The upgrade flow keeps the existing URL, no need to wait for provisioning

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -628,35 +627,88 @@ def deploy(
...
+                timeout_message=(
+                    f"Endpoint provisioning timed out. "
+                    f"Check: DESCRIBE APPLICATION SERVICE {service_fqn.identifier}"
+                ),
+            )
+
+        endpoint_url = desc.get("url", "")
+        if endpoint_url and not endpoint_url.startswith(("http://", "https://")):
+            endpoint_url = f"https://{endpoint_url}"
+        return MessageResult(f"App ready at {endpoint_url}")
```
sfc-gh-olorek, line 776: I see your reply to the Graphite comment that the upgrade flow preserves the URL — makes sense. But if `DESCRIBE` ever returns a row without a `url` key (transient server state, schema change, etc.), this silently produces `"App ready at https://"` which would be pretty confusing for users. A small guard here seems like cheap insurance:

```python
if not endpoint_url:
    raise CliError("Upgrade succeeded but no endpoint URL was returned.")
```

src/snowflake/cli/_plugins/apps/manager.py
```diff
@@ -820,35 +820,75 @@ def build_app_artifact_repo(
...
+            eai_list = ", ".join(external_access_integrations)
+            parts.append(f"EXTERNAL_ACCESS_INTEGRATIONS = ({eai_list})")
+        if query_warehouse:
+            parts.append(f"QUERY_WAREHOUSE = {query_warehouse}")
+        if comment:
+            escaped = comment.replace("'", "''")
+            parts.append(f"COMMENT = '{escaped}'")
+
+        query = "\n".join(parts)
+        self.execute_query(query)
```
sfc-gh-olorek, line 892: The `comment` field gets proper single-quote escaping below (nice), but `package_name`, `artifact_repo_fqn`, `compute_pool`, `query_warehouse`, and the EAI names are all interpolated without quoting. I know this follows the existing pattern in the codebase (e.g. `execute_build_job`), but if any of these ever accept values with spaces or special characters this'll break in fun ways. Might be worth a quick note, or at least consistent identifier quoting.

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -628,35 +627,88 @@ def deploy(
...
+                is_done=lambda d: not _svc_is_upgrading(d),
+                is_error=_svc_has_failed,
+                format_status=lambda d: (
+                    "upgrading" if _svc_is_upgrading(d) else "ready"
+                ),
+                timeout_message=(
+                    f"Upgrade timed out. Check logs:\n"
+                    f"  CALL SYSTEM$GET_APPLICATION_SERVICE_LOGS('{app_name}')"
+                ),
+            )
```
sfc-gh-olorek, line 759: The build-status polls above use `done_states`/`error_states`/`known_pending_states` while these new ones use `is_done`/`is_error`/`format_status`. Both calling conventions hit the same `_poll_until` function. Works today, but it's a bit fragile — if someone refactors `_poll_until` later and only tests one calling convention, the other breaks silently. Maybe worth a note near the `_poll_until` definition documenting both supported signatures?

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -628,35 +627,88 @@ def deploy(
...
+                artifact_repo_fqn=artifact_repo_fqn_str,
+                package_name=app_name,
+                compute_pool=service_compute_pool,
+                version="LATEST",
+                query_warehouse=query_warehouse,
+                external_access_integrations=eai_list,
+                comment=app_comment,
+            )
+        except ProgrammingError as e:
+            if e.errno == 2002 or "already exists" in str(e).lower():
```
sfc-gh-olorek: The `"already exists" in str(e).lower()` is pretty broad — any `ProgrammingError` that happens to mention "already exists" anywhere in its message would trigger the upgrade path, even if it's about a different object (sub-resource, grant, etc.). If errno 2002 is the canonical "object already exists" code, might be safer to just rely on that. Otherwise maybe tighten the string match to something like `"application service" ... "already exists"` so it doesn't accidentally swallow unrelated errors.

src/snowflake/cli/_plugins/apps/commands.py
sfc-gh-gbloom: Per backend spec this is not possible, but I added an additional guard in the polling logic to also check that the URL is not provisioning (for the case an upgrade was triggered before the initial URL was provisioned)

src/snowflake/cli/_plugins/apps/manager.py
sfc-gh-gbloom: Done

src/snowflake/cli/_plugins/apps/commands.py
sfc-gh-gbloom: I'm not sure it's a real concern. The `_poll_until` utility intentionally works with different signatures. Isn't the docstring explains it well?

src/snowflake/cli/_plugins/apps/commands.py
sfc-gh-gbloom: Fixed the condition to check for BOTH 2002 and the "already exists" string. I don't think checking only for 2002 error is enough.

sfc-gh-olorek (approved):
lgtm

PR #2860, opened 2026-04-07
feat: [SNOW-3680799] add interface-first plugin framework

sfc-gh-olorek:
**Jira:** [SNOW-3680799](https://snowflakecomputing.atlassian.net/browse/SNOW-3680799) — under epic [SNOW-3445770](https://snowflakecomputing.atlassian.net/browse/SNOW-3445770) (Command scaffolding).

### Pre-review checklist
   * [ ] _N/A — no user-facing interface._ This PR adds no commands, flags, or output formats. It is internal plumbing for plugin authors, so the design sign-off for user-facing surfaces does not apply.
   * [ ] _N/A — integration tests._ The framework has no runtime CLI surface and never touches a Snowflake connection; correctness is fully covered by unit tests, and `testing.py` exists specifically so plugin authors can validate contracts without a connection.
   * [ ] _N/A — manual macOS test._ Pure library code with no platform-specific behaviour; covered by unit tests.
   * [ ] _N/A — manual Windows test._ Same as above.
   * [x] I've confirmed my changes are up-to-date with the target branch (rebased onto `main`).
   * [ ] _Using the `skip-release-notes` label._ Per [lifecycle.md](https://github.com/snowflakedb/snowflake-cli/blob/main/docs/contributing/lifecycle.md), release notes describe user-facing impact; this change has none.
   * [x] I've updated documentation if behavior changed — module docstrings document the new API; no behaviour changed.

### Changes description

**Interface-first plugin framework** — a declarative API that separates *what* a plugin's command surface looks like (the **spec**) from *how* it behaves (the **handler**).

#### Motivation

Today a plugin defines its commands and its business logic in one place, so the CLI surface (command names, flags, help text, preview/hidden status) can only be reviewed by reading through the implementation. This framework lets a plugin declare its surface as plain data that can be reviewed and locked in first, with the implementation supplied — and validated against that surface — separately. That makes the public CLI contract easy to review, easy to unit-test, and consistent across plugins.

The bridge emits a standard `CommandSpec`, so **the plugin loader and command registration are completely unchanged** — nothing downstream knows the spec/handler split exists.

#### What's added (all new, additive-only)

| Module | Purpose |
| --- | --- |
| `api/plugins/command/interface.py` | Frozen dataclasses `ParamDef`, `CommandDef`, `CommandGroupSpec`, `SingleCommandSpec`; the `ParamKind` enum; the `REQUIRED` sentinel; and the `CommandHandler` ABC. |
| `api/plugins/command/bridge.py` | `build_command_spec(spec, handler)` → standard `CommandSpec`. Load-time handler validation (`validate_interface_handler`) and a small, extensible decorator registry (`register_decorator`). |
| `api/plugins/command/testing.py` | `assert_interface_well_formed`, `assert_handler_satisfies`, `assert_builds_valid_spec` — let plugin authors unit-test their contracts with no Snowflake connection. |

#### Usage

```python
from snowflake.cli.api.plugins.command.interface import (
    CommandDef, CommandGroupSpec, CommandHandler, ParamDef, ParamKind,
)
from snowflake.cli.api.plugins.command.bridge import build_command_spec

SPEC = CommandGroupSpec(
    name="greeter",
    help="Greeting commands.",
    commands=(
        CommandDef(
            name="hello",
            help="Greet someone.",
            handler_method="hello",
            params=(ParamDef(name="name", type=str, kind=ParamKind.ARGUMENT, help="Who to greet"),),
        ),
    ),
)

class GreeterHandler(CommandHandler):
    def hello(self, name: str) -> CommandResult:
        return MessageResult(f"Hello, {name}!")

# Produces a standard CommandSpec the existing loader consumes unchanged.
command_spec = build_command_spec(SPEC, GreeterHandler())
```

#### Design notes

- **Zero startup cost / no coupling.** The new modules are imported only on demand by plugins that opt in; `api/plugins/command/__init__.py` is untouched, so the CLI startup import graph is unchanged.
- **Validation up front.** `build_command_spec` checks (by default) that the handler implements every declared command before building, raising `InterfaceValidationError` listing *all* violations.
- **Native Typer.** Boolean flags rely on the `bool` annotation rather than the deprecated `is_flag=`, so no Typer deprecation warnings are emitted.

#### Testing

- 38 unit tests across `test_interface.py`, `test_bridge.py`, `test_testing.py` (dataclass validation, the spec→`CommandSpec` bridge, nested groups, single commands, flags, decorator errors, and the assertion helpers).
- `ruff`, `black`, and `mypy` all clean.

[SNOW-3680799]: https://snowflakecomputing.atlassian.net/browse/SNOW-3680799?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SNOW-3445770]: https://snowflakecomputing.atlassian.net/browse/SNOW-3445770?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/snowflake/cli/api/plugins/command/bridge.py
+++ b/src/snowflake/cli/api/plugins/command/bridge.py
@@ -1,0 +1,289 @@
+# Copyright (c) 2024 Snowflake Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Bridge: converts a command spec + handler into a standard CommandSpec.
+
+The bridge is the only piece that knows about both the declarative interface
+layer and the Typer/Click runtime.  Downstream code (plugin loader, command
+registration) sees a plain ``CommandSpec`` and is completely unaware of the
+interface/handler split.
+"""
+
+from __future__ import annotations
+
+import inspect
+from functools import wraps
+from typing import Callable
+
+import typer
+
+from snowflake.cli.api.commands.decorators import with_project_definition
+from snowflake.cli.api.commands.snow_typer import SnowTyperFactory
+from snowflake.cli.api.plugins.command import (
+    CommandPath,
+    CommandSpec,
+    CommandType,
+)
+from snowflake.cli.api.plugins.command.interface import (
+    CommandDef,
+    CommandGroupSpec,
+    CommandHandler,
+    ParamDef,
+    ParamKind,
+    REQUIRED,
+    SingleCommandSpec,
+)
+
+
+# ---------------------------------------------------------------------------
+# Decorator registry — maps string names to factory callables
+# ---------------------------------------------------------------------------
+
+_DECORATOR_REGISTRY: dict[str, Callable] = {
+    "with_project_definition": lambda: with_project_definition(),
+}
+
+
+def register_decorator(name: str, factory: Callable) -> None:
+    """Register an additional decorator for use in ``CommandDef.decorators``.
+
+    *factory* is called with no arguments and must return a decorator function.
+    """
+    _DECORATOR_REGISTRY[name] = factory
+
+
+# ---------------------------------------------------------------------------
+# Public entry point
+# ---------------------------------------------------------------------------
+
+
+def build_command_spec(
+    spec: CommandGroupSpec | SingleCommandSpec,
+    handler: CommandHandler,
+    *,
+    validate: bool = True,
+) -> CommandSpec:
+    """Convert a declarative spec + handler into a standard ``CommandSpec``.
+
+    The returned object is identical to what a traditional ``plugin_spec.py``
+    would produce — the rest of the CLI infrastructure is unchanged.
+
+    Args:
+        spec: Declarative command surface.
+        handler: Concrete handler implementing the business logic.
+        validate: If ``True`` (default), verify the handler satisfies the spec
+            before building.  Disable in production for faster startup if the
+            interface tests already cover this.
+    """
+    if validate:
+        validate_interface_handler(spec, handler)
+
+    if isinstance(spec, SingleCommandSpec):
+        factory = _build_single_command(spec, handler)
+        command_type = CommandType.SINGLE_COMMAND
+    else:
+        factory = _build_command_group(spec, handler)
+        command_type = CommandType.COMMAND_GROUP
+
+    return CommandSpec(
+        parent_command_path=CommandPath(list(spec.parent_path)),
+        command_type=command_type,
+        typer_instance=factory.create_instance(),
+    )
+
+
+# ---------------------------------------------------------------------------
+# Validation
+# ---------------------------------------------------------------------------
+
+
+class InterfaceValidationError(Exception):
+    """Raised when a handler does not satisfy its interface spec."""
+
+    def __init__(self, errors: list[str]):
+        self.errors = errors
+        msg = "Interface validation failed:\n" + "\n".join(
+            f"  - {e}" for e in errors
+        )
+        super().__init__(msg)
+
+
+def validate_interface_handler(
+    spec: CommandGroupSpec | SingleCommandSpec,
+    handler: CommandHandler,
+) -> None:
+    """Check that *handler* has a callable method for every command in *spec*.
+
+    Raises ``InterfaceValidationError`` with all violations listed.
+    """
+    errors: list[str] = []
+    for cmd in _collect_commands(spec):
+        method = getattr(handler, cmd.handler_method, None)
+        if method is None:
+            errors.append(
+                f"Handler missing method '{cmd.handler_method}' "
+                f"for command '{cmd.name}'"
+            )
+        elif not callable(method):
+            errors.append(f"Handler.{cmd.handler_method} is not callable")
+    if errors:
+        raise InterfaceValidationError(errors)
+
+
+def _collect_commands(
+    spec: CommandGroupSpec | SingleCommandSpec,
+) -> list[CommandDef]:
+    """Recursively collect all ``CommandDef`` objects from a spec tree."""
+    if isinstance(spec, SingleCommandSpec):
+        return [spec.command]
+    commands = list(spec.commands)
+    for sub in spec.subgroups:
+        commands.extend(_collect_commands(sub))
+    return commands
+
+
+# ---------------------------------------------------------------------------
+# Typer construction
+# ---------------------------------------------------------------------------
+
+
+def _build_command_group(
+    spec: CommandGroupSpec,
+    handler: CommandHandler,
+) -> SnowTyperFactory:
+    factory = SnowTyperFactory(name=spec.name, help=spec.help)
+
+    for cmd_def in spec.commands:
+        _register_command(factory, cmd_def, handler)
+
+    for subgroup in spec.subgroups:
+        sub_factory = _build_command_group(subgroup, handler)
+        factory.add_typer(sub_factory)
+
+    return factory
+
+
+def _build_single_command(
+    spec: SingleCommandSpec,
+    handler: CommandHandler,
+) -> SnowTyperFactory:
+    factory = SnowTyperFactory(name=spec.command.name)
+
+    _register_command(factory, spec.command, handler)
+
+    return factory
+
+
+def _register_command(
+    factory: SnowTyperFactory,
+    cmd_def: CommandDef,
+    handler: CommandHandler,
+) -> None:
+    """Build a Typer command function from a ``CommandDef`` + handler method."""
+    handler_method = getattr(handler, cmd_def.handler_method)
+    spec_param_names = {p.name for p in cmd_def.params}
+
+    # Build signature parameters for Typer introspection
+    sig_params: list[inspect.Parameter] = []
+    annotations: dict[str, type] = {}
+
+    for p in cmd_def.params:
+        annotations[p.name] = p.type
+        default = _make_typer_default(p)
+        sig_params.append(
+            inspect.Parameter(
+                p.name,
+                inspect.Parameter.KEYWORD_ONLY,
+                default=default,
+                annotation=p.type,
+            )
+        )
+
+    # **options is required by SnowTyper's global_options decorator injection
+    sig_params.append(
+        inspect.Parameter("options", inspect.Parameter.VAR_KEYWORD)
+    )
+
+    # Create the wrapper that Typer will call at runtime
+    @wraps(handler_method)
+    def command_fn(**kwargs):
+        handler_kwargs = {k: v for k, v in kwargs.items() if k in spec_param_names}
+        return handler_method(**handler_kwargs)
+
+    # Attach the constructed signature so Typer/Click build the right CLI
+    command_fn.__signature__ = inspect.Signature(sig_params)
+    command_fn.__annotations__ = annotations
+    command_fn.__doc__ = cmd_def.help
+    command_fn.__name__ = cmd_def.handler_method
+    command_fn.__qualname__ = cmd_def.handler_method
+
+    # Apply registered decorators in reverse order (outermost first)
+    for dec_name in reversed(cmd_def.decorators):
+        dec_factory = _DECORATOR_REGISTRY.get(dec_name)
+        if dec_factory is None:
+            raise ValueError(
+                f"Unknown decorator '{dec_name}' in command '{cmd_def.name}'. "
+                f"Register it with register_decorator() first."
+            )
+        command_fn = dec_factory()(command_fn)
+
+    factory.command(
+        name=cmd_def.name,
+        requires_connection=cmd_def.requires_connection,
+        require_warehouse=cmd_def.require_warehouse,
+        preview=cmd_def.is_preview,
+        hidden=cmd_def.is_hidden,
+    )(command_fn)
+
+
+def _make_typer_default(p: ParamDef):
+    """Create the appropriate ``typer.Argument`` or ``typer.Option`` default."""
+    is_required = p.default is REQUIRED
+
+    # Extra kwargs for custom Click type parsing (e.g. IdentifierType for FQN)
+    extra: dict = {}
+    if p.click_type is not None:
+        extra["click_type"] = p.click_type
+
+    if p.kind == ParamKind.ARGUMENT:
+        return typer.Argument(
+            ... if is_required else p.default,
+            help=p.help,
+            show_default=p.show_default,
+            hidden=p.hidden,
+            **extra,
+        )
+
+    # OPTION
+    cli_names = p.cli_names or (f"--{p.name.replace('_', '-')}",)
+
+    if p.is_flag:
+        return typer.Option(
+            p.default if not is_required else False,
+            *cli_names,
+            help=p.help,
+            is_flag=True,
+            show_default=p.show_default,
+            hidden=p.hidden,
+            **extra,
+        )
+
+    return typer.Option(
+        ... if is_required else p.default,
+        *cli_names,
+        help=p.help,
+        show_default=p.show_default,
+        hidden=p.hidden,
+        **extra,
+    )
--- a/src/snowflake/cli/api/plugins/command/bridge.py
+++ b/src/snowflake/cli/api/plugins/command/bridge.py
@@ -222,7 +222,7 @@ def _register_command(
         return handler_method(**handler_kwargs)
 
     # Attach the constructed signature so Typer/Click build the right CLI
-    command_fn.__signature__ = inspect.Signature(sig_params)
+    command_fn.__signature__ = inspect.Signature(sig_params)  # type: ignore[attr-defined]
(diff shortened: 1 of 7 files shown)


src/snowflake/cli/api/plugins/command/bridge.py
```diff
@@ -0,0 +1,285 @@
...
+        require_warehouse=cmd_def.require_warehouse,
+        preview=cmd_def.is_preview,
+        hidden=cmd_def.is_hidden,
+    )(command_fn)
+
+
+def _make_typer_default(p: ParamDef):
+    """Create the appropriate ``typer.Argument`` or ``typer.Option`` default."""
+    is_required = p.default is REQUIRED
+
```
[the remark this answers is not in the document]
sfc-gh-olorek: Good catch — fixed in 494bd993. I removed the redundant `required` field entirely rather than gating on both, so `default` is now the single source of truth: a parameter is required iff `default is REQUIRED` (mirroring how Click/Typer model it).

Worth noting the `is_required = p.default is REQUIRED and p.required` variant would not fully resolve this — in the `default is REQUIRED, required=False` case the `REQUIRED` sentinel would then leak through as the Typer default. Updated the dataclass, docstring, and tests accordingly.

src/snowflake/cli/api/plugins/command/interface.py
```diff
@@ -0,0 +1,196 @@
...
+    Use this instead of ``CommandGroupSpec`` when the plugin contributes
+    exactly one command with no subcommands.
+
+    Attributes:
+        parent_path: Where to attach in the CLI tree.
+        command: The command definition.
+    """
+
+    parent_path: tuple[str, ...] = ()
+    command: CommandDef = None  # type: ignore[assignment]
```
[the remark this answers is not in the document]
sfc-gh-olorek: Addressed in this revision — `SingleCommandSpec.command` is now a required field with no `None` default, so the `_collect_commands` / `_build_single_command` paths can no longer encounter `None`.

src/snowflake/cli/api/plugins/command/testing.py
```diff
@@ -0,0 +1,112 @@
...
+        assert_handler_satisfies(MY_SPEC, MyHandlerImpl())
+
+    def test_full_build():
+        assert_builds_valid_spec(MY_SPEC, MyHandlerImpl())
+"""
+
+from __future__ import annotations
+
+from snowflake.cli.api.plugins.command.bridge import (
+    _collect_commands,
```
sfc-gh-kolszewski: `_collect_commands` is a private function (`_` prefix) in `bridge.py`, but `testing.py` is a public-facing module for plugin authors. Importing a private symbol here creates a hidden coupling: any rename or refactoring of `_collect_commands` silently breaks the testing module. Please either expose it as `collect_commands` (no underscore) in `bridge.py`, or move it to a shared internal module that both `bridge.py` and `testing.py` can import from.

src/snowflake/cli/api/plugins/command/testing.py
```diff
@@ -0,0 +1,112 @@
...
+    - Every command has a non-empty name, help text, and handler_method.
+    - No duplicate handler_method values across the entire tree.
+    - handler_method values are valid Python identifiers.
+
+    Raises ``AssertionError`` on the first violation found.
+    """
+    commands = _collect_commands(spec)
+    seen_methods: set[str] = set()
+
+    for cmd in commands:
```
sfc-gh-kolszewski: Duplicate `handler_method` values are caught here, but two commands with the same CLI `name` in the same group are not. Typer/Click would silently register both and the second would shadow the first. Could you add a `seen_names` check alongside the existing `seen_methods` one? Note it should be per-group (not global across the tree) so that two different subgroups may legitimately reuse the same command name.

src/snowflake/cli/api/plugins/command/testing.py
```diff
@@ -0,0 +1,112 @@
...
+
+        for param in cmd.params:
+            assert param.name, f"Command '{cmd.name}' has a param with empty name"
+            assert param.name.isidentifier(), (
+                f"Command '{cmd.name}': param name '{param.name}' "
+                f"is not a valid Python identifier"
+            )
+
+
+def assert_handler_satisfies(
```
sfc-gh-kolszewski, line 129: `assert_handler_satisfies` raises `InterfaceValidationError` (from `bridge.py`), while `assert_interface_well_formed` raises `AssertionError`. Plugin authors writing `pytest.raises(AssertionError)` across all three helpers will get an unexpected exception type from this one. Please either wrap the error:
```python
try:
    validate_interface_handler(spec, handler)
except InterfaceValidationError as e:
    raise AssertionError(str(e)) from e
```
…or at minimum update the docstring to call out that this raises `InterfaceValidationError` rather than `AssertionError` like its siblings.

src/snowflake/cli/api/plugins/command/interface.py
```diff
@@ -0,0 +1,197 @@
...
+    Use this instead of ``CommandGroupSpec`` when the plugin contributes
+    exactly one command with no subcommands.
+
+    Attributes:
+        command: The command definition.
+        parent_path: Where to attach in the CLI tree.
+    """
+
+    command: CommandDef
+    parent_path: tuple[str, ...] = ()
```
sfc-gh-kolszewski, line 184: `CommandHandler` extends `ABC` but defines no `@abstractmethod` methods, so Python's ABC machinery provides no enforcement — any subclass with an empty body can be instantiated freely. All actual enforcement is deferred to `validate_interface_handler` at bridge build time. Is the `ABC` base intentional as a marker/documentation signal, or was the plan to eventually add abstract declarations here? If it's purely a marker, a short comment explaining that would prevent future readers from assuming ABC enforcement is in play.

tests/api/plugins/command/test_interface.py
```diff
@@ -0,0 +1,155 @@
...
+
+class TestSingleCommandSpec:
+    def test_basic(self):
+        cmd = CommandDef(name="sql", help="Execute SQL.", handler_method="execute_sql")
+        spec = SingleCommandSpec(parent_path=(), command=cmd)
+        assert spec.command.name == "sql"
+
+
+class TestCommandHandler:
+    def test_is_abstract(self):
```
sfc-gh-kolszewski: `issubclass(CommandHandler, CommandHandler)` is always `True` for any class and can never fail, so this assertion tests nothing. The comment says "Cannot instantiate directly if subclass has abstract methods", but `CommandHandler` has no `@abstractmethod` declarations, so ABC enforcement doesn't apply here either. What was the intended contract being verified? Should this be deleted, or was there a specific invariant you had in mind (e.g. that `CommandHandler` can't be directly instantiated, or that an incomplete concrete subclass fails bridge validation)?

src/snowflake/cli/api/plugins/command/bridge.py
```diff
@@ -0,0 +1,285 @@
...
+        spec: Declarative command surface.
+        handler: Concrete handler implementing the business logic.
+        validate: If ``True`` (default), verify the handler satisfies the spec
+            before building.  Disable in production for faster startup if the
+            interface tests already cover this.
+    """
+    if validate:
+        validate_interface_handler(spec, handler)
+
+    if isinstance(spec, SingleCommandSpec):
```
sfc-gh-kolszewski, line 94: *(Minor)* The docstring says "Disable in production for faster startup if the interface tests already cover this", which implies `validate=False` skips all error checking. However, the unknown-decorator `ValueError` inside `_register_command` fires unconditionally regardless of `validate`. Worth clarifying what `validate` actually guards (handler method presence only) and that decorator resolution errors always surface at build time.

src/snowflake/cli/api/plugins/command/bridge.py
```diff
@@ -0,0 +1,285 @@
...
+    command_fn.__name__ = cmd_def.handler_method
+    command_fn.__qualname__ = cmd_def.handler_method
+
+    # Apply registered decorators in reverse order (outermost first)
+    for dec_name in reversed(cmd_def.decorators):
+        dec_factory = _DECORATOR_REGISTRY.get(dec_name)
+        if dec_factory is None:
+            raise ValueError(
+                f"Unknown decorator '{dec_name}' in command '{cmd_def.name}'. "
+                f"Register it with register_decorator() first."
```
sfc-gh-kolszewski, line 237: *(Minor)* "Apply registered decorators in reverse order (outermost first)" is ambiguous without a concrete example. A short inline example would make the intent unambiguous and prevent misuse:
```python
# decorators=("A", "B") → A(B(fn)), list them outermost-first
```

src/snowflake/cli/api/plugins/command/bridge.py
```diff
@@ -0,0 +1,285 @@
...
+        # Click's ``is_flag`` directly is deprecated. A single declaration
+        # (e.g. ``--dry-run``) yields a flag with no ``--no-*`` counterpart.
+        return typer.Option(
+            p.default if not is_required else False,
+            *cli_names,
+            help=p.help,
+            show_default=p.show_default,
+            hidden=p.hidden,
+            **extra,
+        )
```
sfc-gh-kolszewski, line 293: *(Minor)* When `is_flag=True` and `default=REQUIRED`, the code silently substitutes `False`. A required boolean flag is semantically nonsensical, so this combination is almost certainly a spec authoring error. Rather than masking it, consider raising a `ValueError` (either here or in `assert_interface_well_formed`) when `is_flag=True, default=REQUIRED` is detected.

tests/api/plugins/command/test_bridge.py
```diff
@@ -0,0 +1,264 @@
...
+                ),
+            ),
+        )
+
+        class Handler(CommandHandler):
+            def run(self) -> CommandResult:
+                return MessageResult("ok")
+
+        with pytest.raises(ValueError, match="Unknown decorator"):
+            build_command_spec(spec, Handler())
```
sfc-gh-kolszewski, line 298: Could you add a test verifying decorator application order? For example, register two decorators that each prepend a marker string, then assert that `decorators=("outer", "inner")` produces `outer(inner(fn))`. Without it the `reversed()` logic in `_register_command` is exercised but not verified.

tests/api/plugins/command/test_testing.py
```diff
@@ -0,0 +1,152 @@
...
+        class Empty(CommandHandler):
+            pass
+
+        with pytest.raises(InterfaceValidationError):
+            assert_handler_satisfies(_valid_spec(), Empty())
+
+
+class TestAssertBuildsValidSpec:
+    def test_valid_pair_passes(self):
+        assert_builds_valid_spec(_valid_spec(), _ValidHandler())
```
sfc-gh-kolszewski, line 261: `TestAssertBuildsValidSpec` has only the happy-path test. Could you add a case with an invalid pair — e.g. a handler missing a declared method — and assert it raises? This would actually exercise the `validate=True` path through `assert_builds_valid_spec`.

src/snowflake/cli/api/plugins/command/testing.py
sfc-gh-olorek: Resolved in e7654576. `assert_interface_well_formed` now walks the spec group-by-group (needed for the per-group name check you flagged below), so the public `testing.py` no longer imports `_collect_commands` at all — the hidden coupling is gone. `_collect_commands` stays a private helper used internally by `bridge.py` and exercised by `test_bridge.py`, which is normal for a same-package test.

src/snowflake/cli/api/plugins/command/testing.py
sfc-gh-olorek: Done in e7654576. Added a per-group `seen_names` check: sibling commands in the same group can't collide, but two different subgroups may legitimately reuse a name. Handler-method uniqueness stays global (one method per handler instance). Covered by `test_duplicate_command_name_in_group_fails` and `test_same_command_name_across_subgroups_ok`.

src/snowflake/cli/api/plugins/command/testing.py
sfc-gh-olorek: Fixed in e7654576 by wrapping: `assert_handler_satisfies` now re-raises `InterfaceValidationError` as `AssertionError`. Since your point was about consistency *across all three* helpers, I did the same for `assert_builds_valid_spec` (it also surfaced `InterfaceValidationError`/`ValueError` from the build). Now `pytest.raises(AssertionError)` works uniformly for all three.

src/snowflake/cli/api/plugins/command/interface.py
sfc-gh-olorek: It's purely a marker — clarified in e7654576. The class docstring now states there are no `@abstractmethod` members and that the real contract is enforced by `validate_interface_handler` at bridge build time (so the spec stays the single source of truth for required methods), with a short inline comment on the `pass`. Also fixed the misleading `@abstractmethod` in the module-level usage example.

tests/api/plugins/command/test_interface.py
sfc-gh-olorek: Good catch — the old assertion was a tautology. Replaced in e7654576 with two meaningful checks: `test_is_marker_base_with_no_abstract_methods` (asserts `CommandHandler.__abstractmethods__ == frozenset()`, documenting the marker semantics from the comment above) and `test_subclass_is_instantiable`. The "incomplete subclass fails bridge validation" invariant is already covered by `test_missing_method_raises` in `test_bridge.py`.

src/snowflake/cli/api/plugins/command/bridge.py
sfc-gh-olorek: Clarified in e7654576. The docstring now states that `validate` guards *only* the handler-method presence check, and that unknown-decorator resolution (and any other spec-shape error) still raises at build time regardless of the flag.

src/snowflake/cli/api/plugins/command/bridge.py
sfc-gh-olorek: Added the example in e7654576:

```python
# decorators=("A", "B")  ->  A(B(command_fn))
```

with a note that we iterate in `reversed()` so the innermost decorator wraps first.

src/snowflake/cli/api/plugins/command/bridge.py
sfc-gh-olorek: Agreed — masking it was wrong. In e7654576 `_make_typer_default` now raises `ValueError` when `is_flag=True` and `default=REQUIRED`, and I dropped the `else False` coercion entirely. I also added the same rule to `assert_interface_well_formed` so authors catch it in the cheaper interface test rather than only at build. Covered by `test_required_flag_raises` (bridge) and `test_required_boolean_flag_fails` (testing).

tests/api/plugins/command/test_bridge.py
sfc-gh-olorek: Added `TestDecoratorOrder::test_decorators_applied_outermost_first` in e7654576. It registers two recording decorators and asserts that `decorators=("rec_outer", "rec_inner")` applies inner-then-outer — i.e. `outer(inner(fn))` — which exercises and verifies the `reversed()` logic.

tests/api/plugins/command/test_testing.py
sfc-gh-olorek: Added `test_invalid_pair_fails` in e7654576: a handler missing a declared method now asserts the helper raises, exercising the `validate=True` path through `assert_builds_valid_spec`.

src/snowflake/cli/api/plugins/command/testing.py
```diff
@@ -0,0 +1,155 @@
...
+        # Command names need only be unique among siblings — two different
+        # subgroups may legitimately expose the same command name.
+        seen_names: set[str] = set()
+        for cmd in group.commands:
+            check_command(cmd)
+            assert (
+                cmd.name not in seen_names
+            ), f"Duplicate command name '{cmd.name}' in group '{group.name}'"
+            seen_names.add(cmd.name)
+        for sub in group.subgroups:
```
sfc-gh-kolszewski, line 116: This loop recurses into each subgroup but there is no check that sibling subgroups have distinct names. Two subgroups with the same `name` in the same parent would cause Typer/Click to register one over the other, silently shadowing the first — the same class of bug you already guard against for commands (line 106–108). Consider adding a `seen_subgroup_names` set here, mirroring the `seen_names` pattern above:
```python
seen_subgroup_names: set[str] = set()
for sub in group.subgroups:
    assert sub.name not in seen_subgroup_names, (
        f"Duplicate subgroup name '{sub.name}' in group '{group.name}'"
    )
    seen_subgroup_names.add(sub.name)
    check_group(sub)
```

sfc-gh-olorek: Done in 6aba5bcd — good catch, thanks.

I went one step broader than a dedicated `seen_subgroup_names` set. Since Typer/Click registers leaf commands and subgroups into the **same** Click group namespace (`SnowTyperFactory.create_instance` does `app.command(...)` and `app.add_typer(...)` onto one app), a name can actually collide three ways within a group: command/command (already guarded), subgroup/subgroup (your case), and command/subgroup. So rather than add a second set, I extended the existing per-group `seen_names` to cover the *union* of command and subgroup names — one check catches all three. Names stay scoped per group, so `test_same_command_name_across_subgroups_ok` still holds.

Covered by `test_duplicate_subgroup_name_in_group_fails` (your subgroup/subgroup case) and `test_command_and_subgroup_name_collision_fails` (the command/subgroup variant). Docstring updated to describe the shared namespace.

sfc-gh-kolszewski (approved):
LGTM

PR #2862, opened 2026-04-07
chore: add interface-first plugin cookiecutter template

sfc-gh-olorek:
### Pre-review checklist

_Template/tooling PR: `plugin-template/` is excluded from repo CI (Jinja placeholders, not valid Python/TOML), so repo unit/integration tests and manual platform testing don't apply. Instead the generated project was validated end-to-end (see below). `skip-release-notes` applied._

### Changes description

Split from the original scope of this PR; the contributor guide is now in #3132.

Adds a cookiecutter template under `plugin-template/` for scaffolding
**interface-first** Snowflake CLI plugins:

- Supports `group` (multiple subcommands) and `single` command plugins.
- Generated project: `interface.py` (Phase 1 reviewable command surface + handler
  ABC), `handler.py` (Phase 2 implementation), `plugin_spec.py` (wiring via
  `build_command_spec`), and a contract test using the
  `snowflake.cli.api.plugins.command.testing` helpers.
- Generated files carry the standard Apache license header (matching
  `scripts/LICENSE.txt`), so a plugin contributed back into snowflake-cli as a
  built-in passes the repo's `insert-license` check unchanged.

Because the template **sources** contain Jinja placeholders that are not valid
Python/TOML, they are kept out of the repo's own tooling:
- pre-commit excludes `plugin-template/` globally (black/ruff/check-ast cannot
  parse the Jinja sources).
- `plugin-template/conftest.py` keeps the repo-wide pytest runs that collect from
  the project root (`pytest -m performance` / `-m e2e`) from importing the Jinja
  template tests.

**Validated against current `main`:** `cookiecutter plugin-template/` renders both
the `group` and `single` variants, the generated project builds and installs
(`pip install`), and its contract tests pass (`4 passed`).

**Updated from the original draft to match current `main`:**
- `build_command_spec` is now imported from `...command.bridge` (the package
  `__init__` no longer re-exports it); the old import would have failed at runtime.
- `pyproject.toml` declares `[tool.hatch.build.targets.wheel] packages` so the
  generated `src/`-layout project actually builds.

Pairs with the guide in #3132.

### DIFF
--- a/plugin-template/cookiecutter.json
+++ b/plugin-template/cookiecutter.json
@@ -1,0 +1,13 @@
+{
+    "plugin_name": "my-snow-plugin",
+    "plugin_module": "{{ cookiecutter.plugin_name | replace('-', '_') }}",
+    "package_namespace": "snowflakecli_plugins",
+    "cli_command_name": "{{ cookiecutter.plugin_name }}",
+    "cli_parent_path": "",
+    "plugin_description": "A Snowflake CLI plugin.",
+    "author_name": "Your Name",
+    "author_email": "you@example.com",
+    "requires_connection": true,
+    "command_type": ["group", "single"],
+    "python_requires": ">=3.10"
+}
--- a/plugin-template/cookiecutter.json
+++ b/plugin-template/cookiecutter.json
@@ -1,6 +1,7 @@
 {
     "plugin_name": "my-snow-plugin",
     "plugin_module": "{{ cookiecutter.plugin_name | replace('-', '_') }}",
+    "plugin_class_name": "{{ cookiecutter.plugin_module | replace('_', ' ') | title | replace(' ', '') }}",
     "package_namespace": "snowflakecli_plugins",
     "cli_command_name": "{{ cookiecutter.plugin_name }}",
     "cli_parent_path": "",
--- a/plugin-template/{{cookiecutter.plugin_name}}/src/{{cookiecutter.package_namespace}}/{{cookiecutter.plugin_module}}/interface.py
+++ b/plugin-template/{{cookiecutter.plugin_name}}/src/{{cookiecutter.package_namespace}}/{{cookiecutter.plugin_module}}/interface.py
@@ -1,0 +1,116 @@
+"""{{ cookiecutter.plugin_description }}
+
+This file defines the command surface for ``snow {{ cookiecutter.cli_command_name }}``.
+Submit this file for review (Phase 1) before writing the implementation.
+
+Commands
+--------
+{%- if cookiecutter.command_type == "group" %}
+- ``snow {{ cookiecutter.cli_command_name }} hello <name>``  -- Say hello.
+- ``snow {{ cookiecutter.cli_command_name }} status``        -- Show status.
+{%- else %}
+- ``snow {{ cookiecutter.cli_command_name }} <name>``  -- Run the command.
+{%- endif %}
+"""
+
+from __future__ import annotations
+
+from abc import abstractmethod
+
+from snowflake.cli.api.output.types import CommandResult
+from snowflake.cli.api.plugins.command.interface import (
+    CommandDef,
+{%- if cookiecutter.command_type == "group" %}
+    CommandGroupSpec,
+{%- else %}
+    SingleCommandSpec,
+{%- endif %}
+    CommandHandler,
+    ParamDef,
+    ParamKind,
+    REQUIRED,
+)
+
+# ---------------------------------------------------------------------------
+# Command surface (reviewable spec)
+# ---------------------------------------------------------------------------
+
+{% if cookiecutter.command_type == "group" -%}
+PLUGIN_SPEC = CommandGroupSpec(
+    name="{{ cookiecutter.cli_command_name }}",
+    help="{{ cookiecutter.plugin_description }}",
+    parent_path=({{ cookiecutter.cli_parent_path | default("", true) }}),
+    commands=(
+        CommandDef(
+            name="hello",
+            help="Say hello to someone.",
+            handler_method="hello",
+            requires_connection={{ cookiecutter.requires_connection }},
+            params=(
+                ParamDef(
+                    name="name",
+                    type=str,
+                    kind=ParamKind.ARGUMENT,
+                    help="Name to greet.",
+                ),
+            ),
+            output_type="MessageResult",
+        ),
+        CommandDef(
+            name="status",
+            help="Show plugin status.",
+            handler_method="status",
+            requires_connection={{ cookiecutter.requires_connection }},
+            output_type="MessageResult",
+        ),
+    ),
+)
+{%- else -%}
+PLUGIN_SPEC = SingleCommandSpec(
+    parent_path=({{ cookiecutter.cli_parent_path | default("", true) }}),
+    command=CommandDef(
+        name="{{ cookiecutter.cli_command_name }}",
+        help="{{ cookiecutter.plugin_description }}",
+        handler_method="run",
+        requires_connection={{ cookiecutter.requires_connection }},
+        params=(
+            ParamDef(
+                name="name",
+                type=str,
+                kind=ParamKind.ARGUMENT,
+                help="Name argument.",
+            ),
+        ),
+        output_type="MessageResult",
+    ),
+)
+{%- endif %}
+
+
+# ---------------------------------------------------------------------------
+# Handler contract (ABC)
+# ---------------------------------------------------------------------------
+
+
+class {{ cookiecutter.plugin_module | replace('_', ' ') | title | replace(' ', '') }}Handler(CommandHandler):
+    """Handler contract for {{ cookiecutter.cli_command_name }} commands.
+
+    Each abstract method corresponds to a ``CommandDef`` above via
+    ``handler_method``.
+    """
+{% if cookiecutter.command_type == "group" %}
+    @abstractmethod
+    def hello(self, name: str) -> CommandResult:
+        """Say hello."""
+        ...
+
+    @abstractmethod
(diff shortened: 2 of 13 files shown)


plugin-template/{{cookiecutter.plugin_name}}/src/{{cookiecutter.package_namespace}}/{{cookiecutter.plugin_module}}/interface.py
```diff
@@ -0,0 +1,130 @@
...
+
+# ---------------------------------------------------------------------------
+# Command surface (reviewable spec)
+# ---------------------------------------------------------------------------
+
+{% if cookiecutter.command_type == "group" -%}
+PLUGIN_SPEC = CommandGroupSpec(
+    name="{{ cookiecutter.cli_command_name }}",
+    help="{{ cookiecutter.plugin_description }}",
+    parent_path=({{ cookiecutter.cli_parent_path | default("", true) }}),
```
sfc-gh-kolszewski: **Bug: `parent_path` renders an unquoted identifier for non-empty values**

The template expression on this line:

```
parent_path=({{ cookiecutter.cli_parent_path | default("", true) }}),
```

renders correctly when `cli_parent_path` is empty (`parent_path=(),` — a valid empty tuple) but breaks for any non-empty value. With `cli_parent_path = "snow"` the generated code becomes:

```python
parent_path=(snow),
```

Python interprets `(snow)` as a parenthesized reference to the name `snow`, not a string — `NameError: name 'snow' is not defined` at import time. The empty-string default in `cookiecutter.json` hides this during default-value testing.

**Fix** — quote the value and add a trailing comma:

```
parent_path=({%- if cookiecutter.cli_parent_path %}"{{ cookiecutter.cli_parent_path }}",{%- endif %}),
```

plugin-template/{{cookiecutter.plugin_name}}/src/{{cookiecutter.package_namespace}}/{{cookiecutter.plugin_module}}/interface.py
```diff
@@ -0,0 +1,130 @@
...
+            help="Show plugin status.",
+            handler_method="status",
+            requires_connection={{ cookiecutter.requires_connection }},
+            output_type="MessageResult",
+        ),
+    ),
+)
+{%- else -%}
+PLUGIN_SPEC = SingleCommandSpec(
+    parent_path=({{ cookiecutter.cli_parent_path | default("", true) }}),
```
sfc-gh-kolszewski: Same `parent_path` rendering bug as line 56 — same fix applies here.

plugin-template/{{cookiecutter.plugin_name}}/src/{{cookiecutter.package_namespace}}/{{cookiecutter.plugin_module}}/interface.py
```diff
@@ -0,0 +1,130 @@
...
+    CommandDef,
+{%- if cookiecutter.command_type == "group" %}
+    CommandGroupSpec,
+{%- else %}
+    SingleCommandSpec,
+{%- endif %}
+    CommandHandler,
+    ParamDef,
+    ParamKind,
+    REQUIRED,
```
sfc-gh-kolszewski: **Unused import: `REQUIRED` is dead code in every generated plugin**

`REQUIRED` is imported here but never referenced by name in the template body. `ParamDef.default` already defaults to `REQUIRED` at the dataclass level, so the sentinel only needs to be imported explicitly when writing `default=REQUIRED` in a `ParamDef` — which no `ParamDef` in this template does.

Every generated `interface.py` will carry this unused import and fail `ruff --select F401` / `flake8 F401` on first run.

Either remove `REQUIRED` from the import, or add a `ParamDef` example that demonstrates its use (e.g. an optional option with `default=None`).

plugin-template/cookiecutter.json
```diff
@@ -0,0 +1,13 @@
+{
+    "plugin_name": "my-snow-plugin",
+    "plugin_module": "{{ cookiecutter.plugin_name | replace('-', '_') }}",
```
sfc-gh-kolszewski, line 3: **Simplification: class-name filter chain repeated 12 times across 4 files**

The expression `{{ cookiecutter.plugin_module | replace('_', ' ') | title | replace(' ', '') }}` appears 12 times in `handler.py` (3×), `interface.py` (2×), `plugin_spec.py` (1×), and `test_interface.py` (6×). cookiecutter supports derived variables — adding a `plugin_class_name` key in `cookiecutter.json` after `plugin_module` collapses all 12 repetitions to `{{ cookiecutter.plugin_class_name }}`:

```json
"plugin_class_name": "{{ cookiecutter.plugin_module | replace('_', ' ') | title | replace(' ', '') }}"
```

sfc-gh-olorek: Done in 6750a574 — added a derived `plugin_class_name` key right after `plugin_module` in `cookiecutter.json` and collapsed all 12 occurrences to `{{ cookiecutter.plugin_class_name }}`.

Verified behavior-preserving: rendered the template before and after with cookiecutter for both `command_type` values (`group` and `single`) — generated output is byte-identical. Also sanity-checked a non-default name (`cool-data-tool` → `CoolDataTool`).

plugin-template/{{cookiecutter.plugin_name}}/src/{{cookiecutter.package_namespace}}/{{cookiecutter.plugin_module}}/interface.py
sfc-gh-olorek: Fixed in 84365bb4 — removed the `REQUIRED` import. Confirmed nothing else in the template references it, and since `ParamDef.default` already defaults to `REQUIRED`, the example params stay required.

plugin-template/{{cookiecutter.plugin_name}}/src/{{cookiecutter.package_namespace}}/{{cookiecutter.plugin_module}}/interface.py
sfc-gh-olorek: Good catch — fixed in 84365bb4. I went slightly beyond the single-string form: since `parent_path` is `tuple[str, ...]` (a path of ancestor groups, per the framework docstring), I split the input on whitespace so multi-level attachments also render correctly:

- `""` → `parent_path=()`
- `"snowpark"` → `parent_path=("snowpark",)` (your case, now a valid 1-tuple)
- `"app bundle"` → `parent_path=("app", "bundle",)` (nested; the single-quoted form would mangle this into one `"app bundle"` segment)

```jinja
parent_path=({% for p in cookiecutter.cli_parent_path.split() %}"{{ p }}",{% if not loop.last %} {% endif %}{% endfor %}),
```

Rendered all three inputs across both `command_type` variants and confirmed the generated `.py` compile. Happy to drop back to the literal single-segment form if you'd rather keep it minimal.

plugin-template/{{cookiecutter.plugin_name}}/src/{{cookiecutter.package_namespace}}/{{cookiecutter.plugin_module}}/interface.py
sfc-gh-olorek: Fixed here too in 84365bb4 — same split-on-whitespace approach as line 56.

plugin-template/{{cookiecutter.plugin_name}}/src/{{cookiecutter.package_namespace}}/{{cookiecutter.plugin_module}}/interface.py
sfc-gh-olorek: Done in `84365bb4` — replaced the unquoted value with a split-loop that emits a quoted tuple:

```
parent_path=({% for p in cookiecutter.cli_parent_path.split() %}"{{ p }}",{% if not loop.last %} {% endif %}{% endfor %}),
```

Renders `()` for empty, `("snow",)` for a single segment, and `("snow", "app",)` for multi-segment paths — so space-separated parent paths work too, not just the single-string case.

Verified with cookiecutter: the old template rendered `parent_path=(snow)` (ruff `F821` undefined-name) for `cli_parent_path=snow`; the new output is clean under `ruff --select F401,F821` and `py_compile` across group/single × empty/single/multi-segment.

plugin-template/{{cookiecutter.plugin_name}}/src/{{cookiecutter.package_namespace}}/{{cookiecutter.plugin_module}}/interface.py
sfc-gh-olorek: Done in `84365bb4` — same split-loop fix applied to the `SingleCommandSpec` branch here. Verified by rendering both the `group` and `single` variants with a non-empty `cli_parent_path`.

plugin-template/{{cookiecutter.plugin_name}}/src/{{cookiecutter.package_namespace}}/{{cookiecutter.plugin_module}}/interface.py
sfc-gh-olorek: Done in `84365bb4` — removed `REQUIRED` from the import. No `ParamDef` in the template sets `default=REQUIRED` (it defaults at the dataclass level), so the sentinel was dead weight. Generated `interface.py` now passes `ruff --select F401` cleanly for both variants.

PR #2863, opened 2026-04-07
Snowflake Apps: Add --upload-only, --build-only, --deploy-only flags to deploy command

sfc-gh-gbloom:
### Pre-review checklist

### Changes description

Split the `__app deploy` flow into three individually runnable phases — upload, build, and deploy — via new `--upload-only`, `--build-only`, and `--deploy-only` flags.

Validation is now phase-aware: `build_compute_pool` is only required when the build phase runs, and `service_compute_pool`/`query_warehouse` are only required when the deploy phase runs.

### DIFF
--- a/src/snowflake/cli/_plugins/apps/commands.py
+++ b/src/snowflake/cli/_plugins/apps/commands.py
@@ -313,26 +313,55 @@ def deploy(
         "--entity-id",
         help="ID of the snowflake-app entity to deploy. Required if multiple snowflake-app entities exist.",
     ),
+    upload_only: bool = typer.Option(
+        False,
+        "--upload-only",
+        help="Bundle and upload source artifacts to the stage, then stop. "
+        "Skips the build and deploy phases.",
+    ),
+    build_only: bool = typer.Option(
+        False,
+        "--build-only",
+        help="Run only the build phase (assumes artifacts have already been uploaded). "
+        "Skips the upload and deploy phases.",
+    ),
+    deploy_only: bool = typer.Option(
+        False,
+        "--deploy-only",
+        help="Run only the deploy phase (assumes the container image has already been built). "
+        "Skips the upload and build phases.",
+    ),
     skip_build: bool = typer.Option(
         False,
         "--skip-build",
-        help="Skip the build phase and go straight to deploying the service. "
-        "Assumes the container image has already been built.",
+        hidden=True,
+        help="[Deprecated: use --deploy-only] Skip the build phase.",
     ),
     **options,
 ) -> CommandResult:
     """
     Builds and deploys a Snowflake App.
 
-    Uploads source artifacts, builds a container image, creates (or updates)
-    a service, and waits for it to become ready.
+    The deploy pipeline has three phases: upload, build, and deploy.
+    By default all three phases run in sequence. Use --upload-only,
+    --build-only, or --deploy-only to run a single phase.
 
     If --entity-id is not specified and the project contains exactly one snowflake-app
     entity, that entity will be used automatically.
-
-    Use --skip-build to skip the build phase and redeploy using the existing
-    container image (e.g. after a configuration-only change).
     """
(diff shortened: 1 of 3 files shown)


src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -250,26 +250,46 @@ def deploy(
...
+    By default all three phases run in sequence. Use --upload-only,
+    --build-only, or --deploy-only to run a single phase.
 
     If --entity-id is not specified and the project contains exactly one snowflake-app
     entity, that entity will be used automatically.
-
-    Use --skip-build to skip the build phase and redeploy using the existing
-    container image (e.g. after a configuration-only change).
     """
+    phase_flags = sum([upload_only, build_only, deploy_only])
```
sfc-gh-olorek: nit: this creates a throwaway list just to sum three bools — a tuple would be slightly cheaper:
```python
phase_flags = sum((upload_only, build_only, deploy_only))
```
Doesn't really matter, just a style thing.

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -351,17 +371,15 @@ def deploy(
 
     stage_manager = StageManager()
 
-    if not use_artifact_repo:
+    if (run_build or run_deploy) and not use_artifact_repo:
```
sfc-gh-olorek, line 438: `image_repo_url` only gets assigned inside this conditional. Right now it's safe because the early returns prevent it from being referenced when unset, but if someone later tweaks the flow they'll hit an `UnboundLocalError` with no obvious clue why. Might be worth initializing `image_repo_url = None` a few lines above so there's a clear contract.

tests/apps/test_commands.py
```diff
@@ -2504,3 +2503,304 @@ def test_deploy_artifact_repo_path(
         mock_mgr.alter_service_spec.assert_not_called()
         mock_mgr.execute_build_job.assert_not_called()
         assert mock_poll.call_count == 2
+
+    # ── Phase flag tests ──────────────────────────────────────────────
```
sfc-gh-olorek, line 2912: Would be nice to have a test in this section that invokes deploy with *no* phase flags and asserts all three phases run (upload, build, deploy). The existing tests further up probably cover it, but having one explicitly here would make the expected default behavior obvious and prevent regressions if someone touches the flag logic.

tests/apps/test_commands.py
```diff
@@ -2504,3 +2503,304 @@ def test_deploy_artifact_repo_path(
...
+        entity.artifact_repository = None
+        entity.image_repository = None
+        mock_get_entity.return_value = entity
+
+        with with_feature_flags({FeatureFlag.ENABLE_SNOWFLAKE_APPS: True}):
+            from tests_common import change_directory
+
+            with change_directory(tmp_path):
+                result = runner.invoke(["__app", "deploy", "--upload-only"])
+                assert "build_compute_pool is required" not in result.output
```
sfc-gh-olorek, line 3268: nit: this test doesn't check `result.exit_code` — if the command blows up for some unrelated reason, the `not in result.output` assertions would still pass (empty/error output won't contain those strings either). An `assert result.exit_code == 0` or even just checking it's *something* expected would tighten this up.

src/snowflake/cli/_plugins/apps/commands.py

src/snowflake/cli/_plugins/apps/commands.py
sfc-gh-gbloom: Done

tests/apps/test_commands.py
sfc-gh-gbloom: Test added

tests/apps/test_commands.py
sfc-gh-gbloom: Done

sfc-gh-olorek (approved):
All review comments addressed — looks good, thanks Guy!

PR #2865, opened 2026-04-08
fix: [SNOW-3251927] teardown dbs in nativeapp integration tests

sfc-gh-jwilkowski:
### Pre-review checklist

### Changes description
https://snowflakecomputing.atlassian.net/browse/SNOW-3251927

### DIFF
--- a/tests_integration/nativeapp/test_teardown.py
+++ b/tests_integration/nativeapp/test_teardown.py
@@ -146,6 +146,14 @@ def test_nativeapp_teardown_cascade(
         if expected_error is not None:
             assert result.exit_code == 1
             assert expected_error in result.output
+            if orphan_app:
+                # The package was already dropped manually, so the fixture-level
+                # `app teardown --force --cascade` won't reach the orphaned app
+                # or its owned database. Drop the app explicitly to cascade-drop
+                # db_name and prevent database leaks.
+                snowflake_session.execute_string(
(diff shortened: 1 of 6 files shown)


tests_integration/nativeapp/test_teardown.py
```diff
@@ -146,6 +146,14 @@ def test_nativeapp_teardown_cascade(
         if expected_error is not None:
             assert result.exit_code == 1
             assert expected_error in result.output
+            if orphan_app:
```
sfc-gh-lkyc, line 147: It should be in the try-finally block. Otherwise, if the assertion line above fails, then the `DROP APPLICATION ...` statement won't be executed.

sfc-gh-lkyc (commented):
Left one comment

tests_integration/nativeapp/test_teardown.py
sfc-gh-olorek, line 147: Nit: if either of the asserts above blows up, we skip this DROP and still leak the orphaned app + its database — which is the exact thing we're trying to fix here. Since `DROP APPLICATION IF EXISTS` is a no-op when there's nothing to drop, could we just move this block above the assertions? That way cleanup always happens regardless of test outcome.

sfc-gh-olorek (approved):
lgtm

PR #2866, opened 2026-04-08
Add SnowApps parameters tier to defaults resolution

sfc-gh-kbienkowski:
## Summary
- Read `DEFAULT_SNOWFLAKE_APPS_*` parameters via `SHOW PARAMETERS LIKE 'DEFAULT_SNOWFLAKE_APPS_%' IN USER`
- New precedence: **snowflake.yml > SnowApps parameters > config table > defaults > current session**
- **Breaking change**: config table now wins over current session values (intentional — explicit admin config should override session)
- **Breaking change (intentional)**: `IMAGE_REPO` is now assumed under `<resolved-db>.<resolved-schema>` instead of the previously hardcoded `APPS.PUBLIC`. This means if a user's resolved database is e.g. `USER$JOHN_DOE`, the image repo will be `USER$JOHN_DOE.<schema>.IMAGE_REPO`. This is intended — apps should use an image repo co-located with the app, not a shared `APPS.PUBLIC` location.
- Add `fetch_snow_apps_parameters()` to `SnowflakeAppManager`
- Remove hardcoded defaults (`_get_compute_pool`, `_get_external_access`)
- Resolve build and service compute pools separately (account parameters can differ)
- Provenance labels in `--dry-run`: `user input`, `account parameter`, `config table`, `default`, `current session`

## Test plan
- [x] All 174 unit tests pass
- [x] Tests for `fetch_snow_apps_parameters()`, precedence ordering, setup validation
- [x] Manually tested on Snowhouse with account parameters set

Companion skill PR: https://github.com/snowflake-eng/cortex-code-skills/pull/1318
Previous review on apps-dev: https://github.com/snowflakedb/snowflake-cli/pull/2856

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/snowflake/cli/_plugins/apps/commands.py
+++ b/src/snowflake/cli/_plugins/apps/commands.py
@@ -25,6 +25,7 @@ from snowflake.cli._plugins.apps.generate import (
 )
 from snowflake.cli._plugins.apps.manager import (
     _APP_COMMAND_NAME,
+    DEFAULT_IMAGE_REPOSITORY,
     DEFINITION_FILENAME,
     EXPOSE_UNSUPPORTED_SYNTAX,
     SnowflakeAppManager,
@@ -52,6 +53,14 @@ from snowflake.cli.api.output.types import (
 )
 from snowflake.cli.api.project.util import get_env_username, identifier_for_url
 from snowflake.connector.errors import ProgrammingError
+
+# ── Source provenance labels ──────────────────────────────────────────
+SOURCE_USER_INPUT = "user input"
+SOURCE_ACCOUNT_PARAM = "account parameter"
+SOURCE_CONFIG_TABLE = "config table"
+SOURCE_CURRENT_SESSION = "current session"
+SOURCE_DEFAULT = "default"
+SOURCE_MISSING = "missing"
 
 app = SnowTyperFactory(
     name=_APP_COMMAND_NAME,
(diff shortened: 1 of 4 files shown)


src/snowflake/cli/_plugins/apps/manager.py
```diff
@@ -715,6 +709,28 @@ def get_service_endpoint_url(
...
+            )
+            result: Dict[str, str] = {}
+            for row in cursor:
+                param_name = (row.get("key") or row.get("KEY") or "").upper()
+                param_value = row.get("value") or row.get("VALUE") or ""
+                mapped_key = _SNOW_APPS_PARAM_MAP.get(param_name)
+                if mapped_key and param_value:
+                    result[mapped_key] = param_value
+            return result
+        except Exception:
```
sfc-gh-olorek: Bare `Exception` here is pretty broad — could silently swallow something unexpected (like a bug in the cursor iteration logic). Might be worth narrowing to the Snowflake connector's error types (`ProgrammingError`, `DatabaseError`) so genuinely unexpected stuff still bubbles up. The `exc_info=True` is great though.

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -81,10 +81,15 @@ def setup(
     database = ctx.connection_context.database
 
     manager = SnowflakeAppManager()
+
+    # Parameter defaults: params > config table
     config_overrides = {}
     role = manager.current_role()
     if role:
         config_overrides = manager.fetch_config_table_defaults(role)
+    param_vals = manager.fetch_snow_apps_parameters()
```
sfc-gh-olorek: Tiny thing — this sits right below the `if role:` gate for config table fetching, so at first glance it looks like it should also be guarded. A one-line comment like `# params are per-user, not role-gated` would save the next reader a double-take.

src/snowflake/cli/_plugins/apps/generate.py
sfc-gh-gbloom: Do we need to check the `query_warehouse` param here too?

src/snowflake/cli/_plugins/apps/commands.py
sfc-gh-kbienkowski: fixed

src/snowflake/cli/_plugins/apps/manager.py
sfc-gh-kbienkowski: `fetch_config_table_defaults()` below also catches the Exception in the same way. It feels quite reasonable to me as it's a pretty safe behaviour to just move on without bothering the customer. maybe we should print the error more notably than as `debug`. wdyt @sfc-gh-gbloom ?

src/snowflake/cli/_plugins/apps/generate.py
sfc-gh-kbienkowski: Fixed

src/snowflake/cli/_plugins/apps/manager.py
sfc-gh-gbloom: The parameters query should be more stable. The config table might not exist or be accessible, and we are doing JSON parsing that might fail.
Here I would go with `ProgrammingError` which is what we use for failed queries. Consider telling the user that something is wrong with `cli_console.warning()`, even though I'm not sure it's that helpful in this case.

And FYI we soon plan to add telemetry spans to Snowflake Apps, so these exceptions will be tracked.

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -107,62 +116,119 @@ def setup(
...
+        if user_input:
+            return user_input, SOURCE_USER_INPUT
+        if account_param:
+            return account_param, SOURCE_ACCOUNT_PARAM
+        if config_table:
+            return config_table, SOURCE_CONFIG_TABLE
+        if default_value:
+            return default_value, SOURCE_DEFAULT
+        if current_session:
+            return current_session, SOURCE_CURRENT_SESSION
```
sfc-gh-olorek: These branches all use truthiness (`if user_input:`) instead of `if user_input is not None:`. This means falsy values like `""` or `0` are silently skipped rather than treated as an explicit value.

`_resolve_deploy_defaults` in `manager.py` correctly uses `if val is not None:` — these should match.

While `fetch_snow_apps_parameters` strips empty strings, the `config_table` and `current_session` sources do not, so an explicit empty string from either of those tiers would be silently dropped.

```suggestion
        if user_input is not None:
            return user_input, SOURCE_USER_INPUT
        if account_param is not None:
            return account_param, SOURCE_ACCOUNT_PARAM
        if config_table is not None:
            return config_table, SOURCE_CONFIG_TABLE
        if default_value is not None:
            return default_value, SOURCE_DEFAULT
        if current_session is not None:
            return current_session, SOURCE_CURRENT_SESSION
```

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -107,62 +116,119 @@ def setup(
...
-        return None, "missing"
-
-    if IS_PERSONAL_DB_SUPPORTED:
-        database_resolved = (f"USER${get_env_username().upper()}", "personal db")
-    else:
-        database_resolved = _resolve(None, "database", conn_key="database")
+    def _resolve(
+        user_input=None,
+        account_param=None,
+        config_table=None,
```
sfc-gh-olorek: nit: this parameter shadows the outer `config_table` dict (line 120). It works today because every call site passes `config_table=config_table.get(...)` explicitly, but if someone adds a call without that kwarg they'll silently get `None` instead of the fetched table. Consider renaming to `config_table_val` or `table_value`.

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -107,62 +116,119 @@ def setup(
...
-                return conn_val, "connection config"
-        if builtin is not None:
-            return builtin, "default"
-        return None, "missing"
-
-    if IS_PERSONAL_DB_SUPPORTED:
-        database_resolved = (f"USER${get_env_username().upper()}", "personal db")
-    else:
-        database_resolved = _resolve(None, "database", conn_key="database")
+    def _resolve(
```
sfc-gh-olorek, line 125: This inline `_resolve` and `_resolve_deploy_defaults` in `manager.py` implement the same 5-tier precedence with divergent semantics:

- Here: keyword args + truthiness checks (`if val:`)
- `manager.py`: dict iteration + `if val is not None:`

The same inputs can resolve differently depending on whether a source returns `""`. Worth extracting a shared resolution function (or at minimum aligning the truthiness checks now and converging later).

src/snowflake/cli/_plugins/apps/manager.py
sfc-gh-kbienkowski: kk, implementing more narrow exception handling

src/snowflake/cli/_plugins/apps/commands.py
sfc-gh-kbienkowski: kk, I'll make "" and None handling more explicit, converting "" to None when needed. I'm not sure it makes the code better, though.

src/snowflake/cli/_plugins/apps/commands.py
sfc-gh-kbienkowski: changing to config_table_val

src/snowflake/cli/_plugins/apps/commands.py
sfc-gh-kbienkowski: The semantics are now aligned, no more truthiness checks. Would you like to refactor to extract common logic now, a follow up PR or leave it as-is @sfc-gh-olorek ?

sfc-gh-olorek: I would prefer a follow up PR , to keep this PR focused

sfc-gh-olorek (approved):
lgtm

sfc-gh-gbloom (approved):

PR #2868, opened 2026-04-09
SNOW-2422874 Add snow connection generate-workload-identity-token command

sfc-gh-olorek:
### Pre-review checklist

### Changes description
Add a new `generate-workload-identity-token` command under `snow connection` that wraps the Python connector's `create_attestation` function.

- Supports all WIF providers (AWS, GCP, Azure, OIDC) via `--workload-identity-provider` flag or connection config
- For OIDC, resolves tokens from `--token-file-path` or connection config's `token` field
- Added 6 unit tests covering: basic usage, config-based usage, missing provider error, CLI override of config, OIDC with token file, error handling
- Updated 2 existing connection list tests to include the new `wif` test connection

### DIFF
--- a/src/snowflake/cli/_plugins/connection/commands.py
+++ b/src/snowflake/cli/_plugins/connection/commands.py
@@ -466,7 +466,7 @@ def generate_jwt(
 def generate_workload_identity_token(
     **options,
 ) -> CommandResult:
-    """Generate a workload identity token for authenticating to Snowflake via a WIF provider (AWS, GCP, Azure, or OIDC). The token is printed to stdout."""
+    """Generate a workload identity token for authenticating to Snowflake via a Workload Identity Federation (WIF) provider (AWS, AZURE, GCP, or OIDC). The token is printed to stdout."""
     from snowflake.connector.wif_util import AttestationProvider, create_attestation
 
     connection_details = get_cli_context().connection_context.update_from_config()
@@ -502,5 +502,5 @@ def generate_workload_identity_token(
         return MessageResult(attestation.credential)
     except (UsageError, ClickException):
         raise
-    except (ValueError, TypeError) as err:
+    except (ValueError, TypeError, ProgrammingError) as err:
         raise ClickException(str(err))
--- a/src/snowflake/cli/_plugins/connection/commands.py
+++ b/src/snowflake/cli/_plugins/connection/commands.py
@@ -64,13 +64,14 @@ from snowflake.cli.api.config import (
 )
 from snowflake.cli.api.config_ng.masking import mask_sensitive_value
 from snowflake.cli.api.console import cli_console
-from snowflake.cli.api.constants import ObjectType
+from snowflake.cli.api.constants import DEFAULT_SIZE_LIMIT_MB, ObjectType
 from snowflake.cli.api.output.types import (
     CollectionResult,
     CommandResult,
     MessageResult,
     ObjectResult,
 )
+from snowflake.cli.api.secure_path import SecurePath
 from snowflake.connector import ProgrammingError
 from snowflake.connector.constants import CONNECTIONS_FILE
 
@@ -465,7 +466,7 @@ def generate_jwt(
 def generate_workload_identity_token(
     **options,
 ) -> CommandResult:
-    """Generate a workload identity token, which will be printed out and displayed."""
+    """Generate a workload identity token for authenticating to Snowflake via a WIF provider (AWS, GCP, Azure, or OIDC). The token is printed to stdout."""
     from snowflake.connector.wif_util import AttestationProvider, create_attestation
 
     connection_details = get_cli_context().connection_context.update_from_config()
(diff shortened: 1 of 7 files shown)


sfc-gh-jwilkowski:
**Critical Issues**

1. **`Path.read_text()` is outside the try block** (`commands.py`): If `token_file_path` points to a nonexistent or unreadable file, `FileNotFoundError`/`PermissionError` produces a raw Python traceback. Either validate the path first or move the read inside the try block — see how `connection add` validates file existence at line 308.

2. **OIDC with no token source silently passes `None`** (`commands.py`): When `provider=OIDC` and neither `token` nor `token_file_path` is configured, `create_attestation(provider=OIDC, token=None)` is called silently. Add an explicit `UsageError("OIDC provider requires a token. Set --token-file-path or configure 'token' in the connection.")` — consistent with how `generate_jwt` validates its required fields.

**Recommended Improvements**

3. **`AttestationProvider.from_string()` is outside the try block**: An invalid `--workload-identity-provider` value also produces a raw traceback. Wrap in try/except or validate upfront with a message listing valid providers.

4. **`except Exception` is too broad**: `generate_jwt` uses `except (ValueError, TypeError)`. Narrow this to match, and avoid inadvertently surfacing connector internals (credentials, internal URLs, cloud metadata) in error messages.

5. **Use `SecurePath` instead of `Path` for token file reading**: The codebase uses `SecurePath` for security-sensitive file reads (e.g., `snow_connector.py` for private keys) to enforce file-size limits and consistent error handling. Same pattern should apply here.

6. **`test_generate_workload_identity_token_uses_config` is underspecified**: The test only checks `exit_code` and token in output, but doesn't verify `mock_create_attestation.assert_called_once_with(provider=AttestationProvider.GCP, token=None)`. Compare with `test_generate_workload_identity_token_cli_overrides_config` which correctly asserts call args — that's the core thing this test is meant to verify.

7. **Missing test: OIDC with `token` from config** (not `token_file_path`): The `if connection_details.token: token = connection_details.token` branch is never exercised by any test.

8. **Improve docstring**: `"Generate a workload identity token, which will be printed out and displayed."` → `"Generate a workload identity token for authenticating to Snowflake via a WIF provider (AWS, GCP, Azure, or OIDC). The token is printed to stdout."`

src/snowflake/cli/_plugins/connection/commands.py
```diff
@@ -459,3 +460,47 @@ def _decrypt(passphrase: str | None):
...
+                raise UsageError(
+                    "OIDC provider requires a token. "
+                    "Set --token-file-path or configure 'token' in the connection."
+                )
+
+        attestation = create_attestation(provider=provider, token=token)
+        return MessageResult(attestation.credential)
+    except (UsageError, ClickException):
+        raise
+    except (ValueError, TypeError) as err:
```
sfc-gh-jwilkowski: **[Critical]** `ProgrammingError` is missing from this handler.

`AttestationProvider.from_string()` raises `snowflake.connector.errors.ProgrammingError` (e.g. on an invalid provider string), and so does `create_attestation()` for credential failures. Neither maps to `ValueError` or `TypeError`, so this clause never fires for connector errors — `--workload-identity-provider INVALID` currently produces a raw stack trace instead of a clean error message.

`ProgrammingError` is already imported at line 75. Fix:
```python
except (ValueError, TypeError, ProgrammingError) as err:
    raise ClickException(str(err))
```

src/snowflake/cli/_plugins/connection/commands.py
```diff
@@ -459,3 +460,47 @@ def _decrypt(passphrase: str | None):
         return MessageResult(token)
     except (ValueError, TypeError) as err:
         raise ClickException(str(err))
+
+
+@app.command(requires_connection=True)
+def generate_workload_identity_token(
+    **options,
+) -> CommandResult:
+    """Generate a workload identity token for authenticating to Snowflake via a WIF provider (AWS, GCP, Azure, or OIDC). The token is printed to stdout."""
```
sfc-gh-jwilkowski: The `WIF` acronym is undefined here — users unfamiliar with the technology won't know what it stands for. Suggest expanding on first mention, and also normalising provider casing to all-caps to match the accepted enum values:
```python
"""Generate a workload identity token for authenticating to Snowflake via a Workload Identity Federation (WIF) provider (AWS, AZURE, GCP, or OIDC). The token is printed to stdout."""
```

tests/test_connection.py
```diff
@@ -1444,6 +1474,161 @@ def test_generate_jwt_raises_error_if_required_parameter_is_missing(
...
+        [
+            "connection",
+            "generate-workload-identity-token",
+            "--workload-identity-provider",
+            "GCP",
+        ],
+    )
+
+    assert result.exit_code == 0, result.output
+    assert "test-wif-token" in result.output
```
sfc-gh-jwilkowski, line 1491: **[Recommended]** Missing `assert_called_once_with` — this test only checks that the output contains the token string, but doesn't verify that `create_attestation` was called with the correct provider and token arguments. All sibling tests (`test_generate_workload_identity_token_uses_config`, `test_generate_workload_identity_token_cli_overrides_config`, `test_generate_workload_identity_token_oidc_with_token`, `test_generate_workload_identity_token_oidc_with_config_token`) assert call args. Add:
```python
from snowflake.connector.wif_util import AttestationProvider
mock_create_attestation.assert_called_once_with(
    provider=AttestationProvider.GCP, token=None
)
```

tests/test_connection.py
```diff
@@ -1444,6 +1474,161 @@ def test_generate_jwt_raises_error_if_required_parameter_is_missing(
...
+            "OIDC",
+        ],
+    )
+
+    assert result.exit_code != 0
+    assert "OIDC provider requires a token" in result.output
+
+
+@mock.patch("snowflake.connector.wif_util.create_attestation")
+def test_generate_workload_identity_token_error_handling(
```
sfc-gh-jwilkowski, line 1633: **[Recommended]** This test covers `ValueError` from `create_attestation`, but there is no test for an **invalid provider string** (e.g. `--workload-identity-provider INVALID`). `AttestationProvider.from_string("INVALID")` raises `ProgrammingError`, which the current handler does not catch — a test for this case would expose that bug and document the expected clean-error behaviour. Suggested test:
```python
def test_generate_workload_identity_token_invalid_provider_string(runner):
    result = runner.invoke(
        [
            "connection",
            "generate-workload-identity-token",
            "--workload-identity-provider",
            "INVALID",
        ],
    )
    assert result.exit_code != 0
    assert "Error" in result.output
```

sfc-gh-jwilkowski (changes requested):

src/snowflake/cli/_plugins/connection/commands.py
sfc-gh-olorek: Done

src/snowflake/cli/_plugins/connection/commands.py
sfc-gh-olorek: Thanks, done!

tests/test_connection.py
sfc-gh-olorek: Thanks! done!

tests/test_connection.py
sfc-gh-olorek: Thanks! done

sfc-gh-jwilkowski (dismissed):

sfc-gh-jwilkowski (approved):

PR #2869, opened 2026-04-09
SNOW-3202214: add custom image register command

sfc-gh-ajiang:
### Pre-review checklist

### Changes description
1. create a new command to register the custom runtime environment
 `snow custom-image register [local docker image name/hash] [spcs image registry url] [--skip-validarion] [--base-image-type] [--name]`

- `--skip-validation`: if users add this flag, we only push the local images to the spcs image repo; if not, after pushing the local images to the spcs image repo, we will have validation on that image
- `--base-image-type`: this is required if users do not have  the flag `--skip-validation`
2. add metrics for the `snow custom-images validate` and require snowflake connection for `snow custom-images validate`

### DIFF
--- a/src/snowflake/cli/_plugins/custom_images/commands.py
+++ b/src/snowflake/cli/_plugins/custom_images/commands.py
@@ -86,7 +86,7 @@ def register(
     return MessageResult(message)
 
 
-@app.command(requires_connection=False)
+@app.command(requires_connection=True)
 def validate(
     image: str = typer.Argument(
         ...,
--- a/src/snowflake/cli/_plugins/custom_images/commands.py
+++ b/src/snowflake/cli/_plugins/custom_images/commands.py
@@ -43,7 +43,7 @@ def _callback():
     pass
 
 
-@app.command(requires_connection=False)
+@app.command(requires_connection=True)
 def register(
     image: str = typer.Argument(
         ...,
--- a/src/snowflake/cli/_plugins/custom_images/commands.py
+++ b/src/snowflake/cli/_plugins/custom_images/commands.py
@@ -12,13 +12,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from enum import Enum
 from pathlib import Path
+from typing import Optional
 
 import typer
 from click import ClickException
 from snowflake.cli._plugins.custom_images.manager import CustomImageManager
 from snowflake.cli.api.commands.snow_typer import SnowTyperFactory
 from snowflake.cli.api.output.types import CommandResult, MessageResult
+
+
+class BaseImageType(str, Enum):
+    cpu = "cpu"
+    gpu = "gpu"
+
 
 CONFIG_DIR = Path(__file__).parent / "config"
 DEFAULT_CONFIG_PATH = CONFIG_DIR / "image_validation.yaml"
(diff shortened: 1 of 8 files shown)


src/snowflake/cli/_plugins/custom_images/commands.py
```diff
@@ -36,6 +44,49 @@ def _callback():
 
 
 @app.command(requires_connection=False)
+def register(
```
sfc-gh-olorek, line 47: Should this be `requires_connection=True`? When `--skip-validation` is not set, `register()` ends up calling `self.execute_query()` to create the CRE, which needs an active Snowflake connection. As-is I think users will hit a runtime error on that path.

src/snowflake/cli/_plugins/custom_images/commands.py
```diff
@@ -36,6 +44,49 @@ def _callback():
...
+        image=image,
+        registry=registry,
+        skip_validation=skip_validation,
+        base_image_type=base_image_type.value if base_image_type else None,
+        cre_name=cre_name,
+    )
+    return MessageResult(message)
+
+
+@app.command(requires_connection=True)
```
sfc-gh-olorek, line 89: Was this change intentional? `validate` used to be `requires_connection=False`. The `validate()` method itself doesn't call `execute_query()` — if this is needed so the CLI framework can flush the new metrics back to Snowflake on teardown, that makes sense, but it's a behavioral change for anyone currently running `snow custom-image validate` without connection params. Might be worth calling out in the PR description.

src/snowflake/cli/_plugins/custom_images/manager.py
```diff
@@ -561,6 +601,78 @@ def _check_vulnerabilities(self, context: ValidationContext) -> ValidationResult
...
+        if skip_validation:
+            return f"Successfully pushed '{image}' to '{registry}'."
+
+        # Create the Custom Runtime Environment
+        if not cre_name:
+            cre_name = f"mlruntimes_{uuid.uuid4().hex[:8]}"
+
+        image_path = self._parse_image_path(registry)
+        sql = (
+            f"CREATE CUSTOM RUNTIME ENVIRONMENT {cre_name} "
```
sfc-gh-olorek: nit: `cre_name` comes straight from `--name` user input and gets interpolated directly into the SQL string. A name with spaces or special characters will produce malformed SQL. Might want to validate it against something like `[a-zA-Z0-9_]+` or use identifier quoting.

src/snowflake/cli/_plugins/custom_images/manager.py
```diff
@@ -561,6 +601,78 @@ def _check_vulnerabilities(self, context: ValidationContext) -> ValidationResult
...
+            ["docker", "tag", image, registry]
+        )
+        if returncode != 0:
+            raise ClickException(
+                f"Failed to tag image '{image}' as '{registry}': {stderr}"
+            )
+
+        # Push the tagged image to the registry
+        returncode, _, stderr = self._run_docker_command(
+            ["docker", "push", registry], timeout=600
```
sfc-gh-olorek, line 654: Does `_run_docker_command` actually accept a `timeout` kwarg? I don't see it forwarding to `subprocess.run` in the existing code — if it doesn't, this is either silently ignored or blows up with a TypeError.

src/snowflake/cli/_plugins/custom_images/manager.py
```diff
@@ -561,6 +601,78 @@ def _check_vulnerabilities(self, context: ValidationContext) -> ValidationResult
...
         )
 
+    def _parse_image_path(self, registry: str) -> str:
+        """Strip the hostname from a registry URL to get the image path for CRE.
+
+        Example: 'host.snowflakecomputing.com/db/schema/repo/image:tag'
+                 -> '/db/schema/repo/image:tag'
+        """
+        slash_idx = registry.find("/")
+        if slash_idx == -1:
```
sfc-gh-olorek, line 612: If the registry has no `/` (bare hostname or something malformed), this returns the whole string as the image path, which would produce bogus SQL downstream. Probably worth raising a `ClickException` here instead of silently returning garbage.

src/snowflake/cli/_plugins/custom_images/commands.py
sfc-gh-ajiang: Yes, this change is intentional. I added it to ensure we can collect metrics. It looks like there haven’t been any recent releases, so `snow custom-image validate` has not been published yet.

src/snowflake/cli/_plugins/custom_images/manager.py
sfc-gh-ajiang: I think we have forwarded it
https://github.com/snowflakedb/snowflake-cli/blob/main/src/snowflake/cli/_plugins/custom_images/manager.py#L143

src/snowflake/cli/_plugins/custom_images/commands.py
sfc-gh-ajiang: Good catch. Thanks!

PR #2870, opened 2026-04-10
DCM Purge gated by the parameter

sfc-gh-lkyc:
### Pre-review checklist

### Changes description
- A new feature flag name `ENABLE_DCM_EARLY_ACCESS` was added. It enables or disables PrPr/early-access features of DCM.
- The `dcm purge` command is gated by the new parameter


src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -345,7 +345,9 @@ def _confirm_purge(project_id: FQN) -> None:
             )
 
 
-@app.command(requires_connection=True, hidden=True)
+@app.command(
+    requires_connection=True, is_enabled=FeatureFlag.ENABLE_DCM_EARLY_ACCESS.is_enabled
```
sfc-gh-jwilkowski: just fyi: so far we've been using `hidden`, not `is_enabled`. The difference is that hidden commands still can be executed (which makes it easier to use them without any prior setup steps)

tests/dcm/test_commands.py
```diff
@@ -506,6 +507,11 @@ def test_deploy_with_json_formats_returns_response(
         _assert_format_result(payload, plan_response, format_name)
 
 
+class TestDCMEarlyAccessFeatureFlag:
```
sfc-gh-jwilkowski: Imo this test adds very little value, consider removing it

tests/conftest.py
```diff
@@ -155,6 +155,16 @@ def isolate_snowflake_home(snowflake_home):
...
 
 
+@pytest.fixture(autouse=True, scope="session")
+def enable_dcm_early_access_for_tests():
+    import os
+
+    from snowflake.cli.api.feature_flags import FeatureFlag
+
+    os.environ[FeatureFlag.ENABLE_DCM_EARLY_ACCESS.env_variable()] = "true"
+    yield
```
sfc-gh-jwilkowski: why is this fixture needed? the only test that I see could use it is `test_purge_visibility_depends_on_feature_flag`, but it is parametrized itself? My main concern is that the feature flag defaults to false, but all tests will work as if it was enabled, which could be misleading. I'd be in favour of explicitly using with_feature_flag in every tests that depends on it, even if it's a bit repetitive (you can avoid the repetitions by using fixtures or grouping tests in classes that do pretty much this logic)

sfc-gh-jwilkowski (commented):
I'd consider removing the autouse fixture and set that flag explicitly only where it's needed

tests/dcm/test_commands.py
sfc-gh-lkyc: Removed. I added `test_purge_visibility_depends_on_feature_flag` under `TestDCMPurge` that checks the default value behavior and depending on the feature flag value.

tests/conftest.py
sfc-gh-lkyc: It's to make `purge` enabled in `test_help_messages.py.` tests. I removed the fixture and enabled the flag in `test_help_messages.py`. Do you think it looks better?

tests/dcm/test_commands.py
sfc-gh-jwilkowski: Personally I think that the actual change you're making (using a feature flag for dcm purge) does not need that much testing. `hidden` attribute comes from Typer, FeatureFlags are tested, we've used combination of those 2 functionalities many times, so I don't see what's the gain from those tests. I could justify `test_purge_hidden_from_help` as this could prevent accidental change of the feature flag default or something like that, but rest of the tests feel like an overkill to me

src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -345,7 +345,9 @@ def _confirm_purge(project_id: FQN) -> None:
             )
 
 
-@app.command(requires_connection=True, hidden=True)
+@app.command(
+    requires_connection=True, hidden=FeatureFlag.ENABLE_DCM_EARLY_ACCESS.is_disabled
```
sfc-gh-jwilkowski, line 349: as discussed, I think the only thing needed in this PR is removal of hidden=true os that the command shares visibility rules from the whole command group

PR #2872, opened 2026-04-10
Validate account_identifier and project_owner in DCM manifest targets

sfc-gh-lkyc:
### Pre-review checklist

### Changes description
Validate `account_identifier` and `project_owner` if they are provided in DCM manifest targets, and print warnings if any mismatch is found. The validation does not stop the CLI execution. When a manifest is used, the CLI validates these against the current Snowflake session before executing commands. 

Commands invoked with a direct `--identifier` (no manifest) skip validation - there is no target context in such a scenario.

#### `account_identifier` 
Is validated for all commands that resolve a manifest. The account identifier  

The comparison uses upper-case string representations, not requiring any additional quoting . It's safe because:
- organization name has to match the `[A-Z][A-Z0-9]*` regex
- account name has to match the `[A-Z][A-Z0-9_]+` regex

#### `project_owner` 
Is validated for `create` command only. The reason is that incorrect role used in the `create` command would result in a new dcm project created with possibly unexpected role - not the one mentioned in the manifest.  For the other commands we are more elastic, and don't have to validate, because even of the session role is different, there may be secondary roles configured for the user, that satisfy DCM Project privileges model.

The comparison logic/steps:
Role retrieved from session and from the target a normalized by converting to a valid identifier, adding quotes if needed. Unquotted identifiers are compared using case-insensitive strategy, whereas quotted identifiers use case-sensitive strategy.


tests_integration/test_data/projects/dcm_project/manifest.yml
```diff
@@ -6,3 +6,5 @@ default_target: dev
 targets:
   dev:
     project_name: "project_descriptive_name"
+    account_identifier: "MY_ORG-MY_ACCOUNT"
```
sfc-gh-jwilkowski, line 9: These placeholder values will cause all existing integration tests that use this project directory to fail against any real Snowflake account. `_validate_account_identifier` is called unconditionally on every manifest-resolving command, so tests like `test_dcm_deploy`, `test_dcm_plan_and_deploy_from_another_directory`, etc. will error with "Account mismatch" before doing any real work.

The two new tests (`test_dcm_account_identifier_validation`, `test_dcm_project_owner_validation`) correctly handle this by writing the manifest dynamically from the live session. The fixture-based manifests here need the same treatment — ideally via a `conftest.py` fixture that rewrites `account_identifier` and `project_owner` from `SELECT CURRENT_ORGANIZATION_NAME()` / `SELECT CURRENT_ACCOUNT_NAME()` before running.

src/snowflake/cli/_plugins/connection/util.py
```diff
@@ -193,6 +193,15 @@ def get_account(conn: SnowflakeConnection) -> str:
         raise MissingConnectionAccountError(conn)
 
 
+def get_account_identifier(conn: SnowflakeConnection) -> str:
+    *_, cursor = conn.execute_string(
+        "SELECT CURRENT_ORGANIZATION_NAME()", cursor_class=DictCursor
+    )
+    org_name = cursor.fetchone()["CURRENT_ORGANIZATION_NAME()"]
+    account_name = get_account(conn)
+    return f"{org_name}-{account_name}".upper()
```
sfc-gh-jwilkowski: Two issues here:

1. **Two round-trips**: `CURRENT_ORGANIZATION_NAME()` is queried separately, then `get_account()` executes `select current_account_name()` internally — two SQL calls per manifest-resolving command. A single `SELECT CURRENT_ORGANIZATION_NAME() AS org, CURRENT_ACCOUNT_NAME() AS acct` (with column aliases, which also avoids the fragile function-name-as-dict-key) would halve the latency.

2. **No error handling**: if `CURRENT_ORGANIZATION_NAME()` returns `NULL` or the query fails, this crashes with an unhelpful `TypeError`/`KeyError`. `get_account()` (just above) wraps its query in a try/except with a descriptive fallback — `get_account_identifier` should do the same.

src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -140,10 +143,29 @@ def _from_option_callback(value: Optional[SecurePath]) -> SecurePath:
...
+    if current_account != target.account_identifier:
+        raise CliError(
+            f"Account mismatch: manifest target specifies account_identifier '{target.account_identifier}', "
+            f"but the current session account is '{current_account}'."
+        )
+
+
+def _validate_project_owner(target: DCMTarget) -> None:
+    current_role = SqlExecutor().current_role()
+    if current_role and current_role.upper() != target.project_owner:
```
sfc-gh-jwilkowski: This guard silently passes when `current_role()` returns `None` (which can happen when `SELECT current_role()` returns NULL — e.g. a session whose default role was dropped, or certain OAuth edge cases). For a check protecting destructive write commands (`deploy`, `purge`, `plan`, `create`), failing open is the wrong default. Consider raising a `CliError` when the role cannot be determined, rather than skipping the check.

sfc-gh-jwilkowski:
**`drop` and `drop_deployment` — owner validation**

[commands.py, `drop` (~line 547)](https://github.com/snowflakedb/snowflake-cli/blob/lkyc/Validate-account_identifier-and-project_owner-in-DCM-manifest-targets/src/snowflake/cli/_plugins/dcm/commands.py#L547): `drop` drops the entire DCM project but doesn't pass `validate_owner=True`, while `plan` (a dry-run) does validate the owner. Was this intentional? `drop` and `drop_deployment` seem more destructive than `plan`, so I'd expect them to also enforce the owner check.

[commands.py, `drop_deployment` (~line 613)](https://github.com/snowflakedb/snowflake-cli/blob/lkyc/Validate-account_identifier-and-project_owner-in-DCM-manifest-targets/src/snowflake/cli/_plugins/dcm/commands.py#L613): Same question — was skipping owner validation here intentional? `drop_deployment` deletes a deployment, which feels like it warrants the same protection as `deploy` or `purge`.

tests_integration/test_data/projects/dcm_project/manifest.yml
sfc-gh-lkyc: Fair point. I'm not sure why there were no CI failures.

I addressed it

src/snowflake/cli/_plugins/connection/util.py
sfc-gh-lkyc: #1 DONE

#2
I don't think we should do any fallback. The rule is that the account identifier should match the account identifier in the current session.  Retrieved account identifier would have different formats depending on success or failure (ORG-ACCT vs ACCT). It might hide issues rather than surfacing them. What are the scenarios where `CURRENT_ORGANIZATION_NAME()` returns `null`? From what I know, there is always context of the organization available.
One thing that maybe is worth doing is handling the SQL execution exception and show a pretty message, like `Could not determine account identifier` and log the exception details.

sfc-gh-lkyc: I implemented handling of exceptions by printing a warning. The exception while retrieving account identifier does not impact on the standard flow.

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-lkyc: Implemented, but instead of raising CliError a warning is printed. I assumed that the check, which does not change the flow, but only prints warnings, should cause this flow to fail if the check fails.

src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -142,10 +146,83 @@ def _from_option_callback(value: Optional[SecurePath]) -> SecurePath:
...
+        _warn_cannot_validate(
+            "account identifier",
+            f"current account could not be determined: {e}",
+            _ACCOUNT_GUIDANCE,
+        )
+        return
+
+    if not current_account.matches(target.account_identifier):
+        cli_console.warning(
+            f"⚠️  Account mismatch: manifest target specifies account_identifier '{target.account_identifier}', "
```
sfc-gh-jwilkowski: please call `sanitize_for_terminal`

src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -142,10 +146,83 @@ def _from_option_callback(value: Optional[SecurePath]) -> SecurePath:
...
+        _warn_cannot_validate(
+            "project owner",
+            "current role could not be determined",
+            _ROLE_GUIDANCE,
+        )
+        return
+
+    if not same_identifiers(current_role, target.project_owner):
+        cli_console.warning(
+            f"⚠️  Role mismatch: manifest target specifies project_owner '{target.project_owner}', "
```
sfc-gh-jwilkowski: as above, please call sanitize_for_terminal

src/snowflake/cli/api/identifiers.py
```diff
@@ -221,3 +221,30 @@ def using_context(self) -> "FQN":
...
+        self.organization_name = organization_name.upper()
+        self.account_name = account_name.upper()
+
+    def as_hyphen_form(self) -> str:
+        return f"{self.organization_name}-{self.account_name}"
+
+    def as_dot_form(self) -> str:
+        return f"{self.organization_name}.{self.account_name}"
+
+    def matches(self, identifier: str) -> bool:
```
sfc-gh-jwilkowski: Can we reliably implement `AccountIdentifier.from_string` which would construct an instance that could later be compared to others with __eq__? Having both marches and __eq__ feels confusing as to when use which method? Or why should I use matches instead of more common __eq__? So what I'm getting at is if we can reliably decide if a string is a dot or hyphen form?

src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -477,7 +568,9 @@ def create(
     """
     Creates a DCM Project in Snowflake.
     """
-    context = _resolve_context_with_optional_manifest(from_location, identifier, target)
+    context = _resolve_context_with_optional_manifest(
+        from_location, identifier, target, validate_owner=True
```
sfc-gh-jwilkowski, line 580: Would be great to document why this is the only command that validates the owner

RELEASE-NOTES.md
```diff
@@ -25,6 +25,9 @@
 * Added `snow connection generate-workload-identity-token` command to generate a workload identity token for the current environment. Supports AWS, GCP, Azure, and OIDC providers via `--workload-identity-provider` flag or connection configuration.
 * Added `snow custom-image validate` command to validate custom Docker images against configured rules (entrypoint, environment variables, Python packages, dependency health). Supports an optional `--scan-vulnerabilities` flag to run Grype vulnerability scanning.
 * Added `snow dcm purge` command to drop all the objects managed by the specified DCM Project
+* Added checking values mismatch, printing warnings, of fields in DCM manifest target profiles:
+  * Validate `account_identifier` field against the current session account identifier for all manifest-based commands
+  * Validate `project_owner` field against the current session role for the `snow dcm create` command
```
sfc-gh-jwilkowski: * DCM manifest targets now validate `account_identifier` and `project_owner` fields.
    CLI validates these against the current session and prints a warning on mismatch:
    * `account_identifier` is checked for all manifest-based commands
    * `project_owner` is checked for `snow dcm create`

tests_integration/test_dcm_project.py
```diff
@@ -893,3 +897,103 @@ def test_dcm_purge(
...
+                    "account_identifier": account_id,
+                    "project_owner": "NONEXISTENT_ROLE_XYZ",
+                }
+            },
+        }
+        (project_root / "manifest.yml").write_text(yaml.dump(manifest))
+
+        result = runner.invoke_with_connection(["dcm", "create", "--target", "dev"])
+        assert result.exit_code == 1
+        assert "Role mismatch" in result.output
```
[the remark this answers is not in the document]
sfc-gh-lkyc: Fixed

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-lkyc: Applied

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-lkyc: Applied

src/snowflake/cli/api/identifiers.py
sfc-gh-lkyc: Yes, that makes sense. Initially, I didn't want to parse the account identifier, because I was afraid of doing it incorrectly. I found recently what are the allowed characters for organization and account name, and given the type of characters are strictly limited, I agree that we can parse reliably.

Fixed.

RELEASE-NOTES.md
sfc-gh-lkyc: Done

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-lkyc: I added some explanation

sfc-gh-jwilkowski (approved):

PR #2876, opened 2026-04-13
 DCM: Speed up file upload to temporary stage by parallelizing it

sfc-gh-lkyc:
### Pre-review checklist

#### Summary
- Replace per-file sequential PUT uploads in dcm plan/dcm deploy with directory-level `put_recursive`, matching the pattern already used by SPCS services
- Simplify `_collect_artifacts` to pass the `sources/` directory as a single artifact instead of enumerating individual files
- Remove unused `use_temporary_stage` parameter from `sync_artifacts_with_stage` (DCM was the only caller)

#### Performance
The previous implementation issued one `PUT SQL` command per file (102 round-trips in sample benchmark). The new implementation uses `put_recursive` which batches files by directory into a single `PUT file://dir/*` command, reducing round-trips and enabling parallel file transfer within each command.
Also, the previous implementation executed `LIST @stage` command after the stage creation, and computed MD5 hashes for every local file (never compared against anything) .

##### Benchmarks
Input
Benchmarked with 100 definition files (~10KB each) uploaded to a temporary stage:

##### Scenario 1
100 flat files in `sources/definitions`
> Before: 129s
> After: 9.37s
> Speedup: x14

##### Scenario 2
5 subdirs x 20 files in `sources/definitions`
> Before: 189s
> After: 12s
> Speedup: x16

There is still room for improvement, as the current implementation uses a level of parallelism of 4 (the default). I experimented with 16, and we can reduce the time by 30%.

### DIFF
--- a/src/snowflake/cli/_plugins/dcm/manager.py
+++ b/src/snowflake/cli/_plugins/dcm/manager.py
@@ -18,7 +18,7 @@ from typing import List
 from snowflake.cli._plugins.dcm.models import MANIFEST_FILE_NAME, SOURCES_FOLDER
 from snowflake.cli._plugins.dcm.utils import collect_output
 from snowflake.cli._plugins.stage.manager import StageManager
-from snowflake.cli.api.artifacts.upload import sync_artifacts_with_stage
+from snowflake.cli.api.artifacts.utils import bundle_artifacts
 from snowflake.cli.api.commands.utils import parse_key_value_variables
 from snowflake.cli.api.console.console import cli_console
 from snowflake.cli.api.constants import (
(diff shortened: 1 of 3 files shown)


sfc-gh-olorek (commented):
Nice speedup! Left a couple of inline comments — nothing blocking, mostly cleanup opportunities.

src/snowflake/cli/_plugins/dcm/manager.py
```diff
@@ -258,14 +258,31 @@ def sync_local_files(
...
+            project_paths = ProjectPaths(project_root=source_path.path)
+            project_paths.remove_up_bundle_root()
+            SecurePath(project_paths.bundle_root).mkdir(parents=True, exist_ok=True)
+            bundle_artifacts(
+                project_paths, artifacts, pattern_type=PatternMatchingType.GLOB
             )
 
+            try:
+                stage_manager = StageManager()
+                stage_manager.create(
```
sfc-gh-olorek, line 271: nit: `bundle_artifacts` sits outside the `try`/`finally`, so if it throws, the bundle root created two lines above won't get cleaned up. Moving it inside the `try` block would make the cleanup airtight.

(The old code had the same gap fwiw, so not a regression — just a nice-to-fix while you're in here.)

src/snowflake/cli/_plugins/dcm/manager.py
```diff
@@ -258,14 +258,31 @@ def sync_local_files(
...
+                )
+                for result in stage_manager.put_recursive(
+                    local_path=project_paths.bundle_root,
+                    stage_path=stage_fqn.identifier,
+                    temp_directory=project_paths.bundle_root,
+                ):
+                    log.info(
+                        "Uploaded %s to %s",
+                        result["source"],
+                        result["target"],
```
sfc-gh-olorek, line 282: Heads up: `put_recursive` calls `shutil.rmtree()` on subdirectories as it walks bottom-up, so the bundle root ends up in a partially-destroyed state during the upload. Doesn't matter right now since `clean_up_output()` nukes the whole thing in the `finally`, but it could trip someone up later if they ever need to inspect the bundle after upload. Might be worth a short comment here.

sfc-gh-olorek (dismissed):
lgtm

sfc-gh-jwilkowski (dismissed):

sfc-gh-olorek (approved):

PR #2884, opened 2026-04-14
SNOW-3202214: add metrics to custom image commands

sfc-gh-ajiang:
### Pre-review checklist

### Changes description
add metrics for the snow custom-images validate and require snowflake connection for snow custom-images validate


src/snowflake/cli/_plugins/custom_images/commands.py
```diff
@@ -35,7 +35,7 @@ def _callback():
     pass
 
 
-@app.command(requires_connection=False)
+@app.command(requires_connection=True)
```
sfc-gh-wesong: One concern about the connection requirement change.
requires_connection=False → requires_connection=True (commands.py)
This changes user-facing behavior. Currently snow custom-images validate works completely offline — users point it at a local Docker image and get results without needing a Snowflake account configured. Requiring a connection just to emit telemetry is a UX regression, especially for users who are still iterating on their image build before they've set up their Snowflake config.
Looking at the implementation, CLIMetrics is a plain in-memory dataclass (_CliGlobalContextManager.metrics is created via field(default_factory=CLIMetrics)). The _record_validate_metrics method only calls set_counter() / set_counter_default() — no connection needed for the recording itself.
Suggestion: keep requires_connection=False and let the telemetry layer handle the case where no connection is available (either skip sending or buffer for later). This preserves the offline validation UX while still collecting metrics when a connection exists.

sfc-gh-wesong (dismissed):

src/snowflake/cli/_plugins/custom_images/commands.py
sfc-gh-ajiang: Thanks for the suggestion. Just to confirm — if we keep requires_connection=False, telemetry will be silently skipped when no connection is   configured (via @ignore_exceptions). The metrics are collected in memory but never sent. Are you OK with losing telemetry data for users who run validation without a Snowflake connection set up?

src/snowflake/cli/_plugins/custom_images/manager.py
```diff
@@ -180,6 +182,38 @@ def _get_image_info(self, image_name: str) -> Optional[dict]:
...
+        # Initialize all failure counters to 0 (check ran, no failure)
+        for counter in check_counters.values():
+            metrics.set_counter_default(counter, 0)
+
+        overall_failed = False
+        for result in report.results:
+            if not result.passed:
+                overall_failed = True
+                for prefix, counter in check_counters.items():
+                    if result.check_name.startswith(prefix):
```
sfc-gh-olorek: The prefix matching here is order-sensitive and a bit fragile — today the prefixes are disjoint so it works, but a future check named e.g. `env_xyz` or `script_foo` that's meant to map elsewhere would silently land in the wrong bucket. Could we either map exact `check_name` → counter, or carry the counter on `ValidationResult` itself so we don't have to re-derive it here?

tests/custom_images/test_manager.py
```diff
@@ -31,6 +33,13 @@
 class TestCustomImageManager:
     """Tests for CustomImageManager validation."""
 
+    @pytest.fixture(autouse=True)
+    def reset_cli_context(self):
+        """Reset the CLI context before each test to get fresh metrics."""
+        token = _CLI_CONTEXT_MANAGER.set(None)
```
sfc-gh-olorek: Reaching into `_CLI_CONTEXT_MANAGER` from a test feels heavy — I don't think we do this anywhere else in the repo. Is there a reason `runner.invoke_with_connection` / the existing conftest fixtures wouldn't give you fresh metrics per test? If this pattern is genuinely needed, worth promoting to a shared conftest so other plugins can reuse it.

src/snowflake/cli/_plugins/custom_images/manager.py
```diff
@@ -180,6 +182,38 @@ def _get_image_info(self, image_name: str) -> Optional[dict]:
             pass
         return None
 
+    def _record_validate_metrics(self, report: ValidationReport) -> None:
```
sfc-gh-olorek, line 186: Intentional that `vulnerability_scan` failures aren't tracked here? `--scan-vulnerabilities` is user-facing, so a failure-rate counter for that check seems useful too.

src/snowflake/cli/_plugins/custom_images/commands.py
sfc-gh-olorek: +1 to keeping `requires_connection=False`. For users still iterating on their image locally, forcing a Snowflake connection just to run a local check is a meaningful regression. Best-effort telemetry (skip when no connection is configured) seems like the right tradeoff — better to lose some data than push users through connection setup for what's otherwise an offline command.

src/snowflake/cli/_plugins/custom_images/manager.py
sfc-gh-ajiang: Resolved — we added an optional counter_field to ValidationResult so each result carries its own counter. _record_validate_metrics now reads result.counter_field directly instead of deriving it from prefix   matching on check_name.

src/snowflake/cli/_plugins/custom_images/commands.py
sfc-gh-ajiang: make sense. I reverted this change

tests/custom_images/test_manager.py
sfc-gh-ajiang: Resolved — replaced the direct _CLI_CONTEXT_MANAGER.set(None) call with fork_cli_context(), which is the existing shared helper from cli_global_context.py already used in other test conftest files.

src/snowflake/cli/_plugins/custom_images/manager.py
sfc-gh-ajiang: Resolved — _check_vulnerabilities now sets counter_field=CustomImageCounterField.CUSTOM_IMAGE_VALIDATE_FAIL_VULNERABILITY_SCAN on its ValidationResult, so failures are tracked via the same counter_field mechanis as all other checks.

sfc-gh-olorek (approved):
All review comments addressed — thanks!

PR #2892, opened 2026-04-17
Add snow snowpark execute-file command for local Python/SQL execution

sfc-gh-mborins:
## Summary

- Adds `snow snowpark execute-file <path>` command that lets users execute a local `.py` or `.sql` file on Snowflake without manually uploading it to a stage first
- Supports single files, directories, and glob patterns (e.g. `./scripts/*.py`)
- Automatically uploads a `requirements.txt` from the same directory if present, so Snowpark package dependencies are picked up
- Uses a session-scoped temporary stage that auto-drops on disconnect; no cleanup required
- Supports the same `--on-error` and `-D` variable flags as `snow stage execute`

## How it works

`snow snowpark execute-file ./my_script.py` does the following under the hood:

1. Resolves the local path/glob to a list of `.py`/`.sql` files
2. Creates a temporary Snowflake stage (`snowflake_cli_tmp_stage_<timestamp>`)
3. Uploads the matched files (and `requirements.txt` if present) via `StageManager.put()`
4. Calls the existing `StageManager.execute()` against the temp stage, which handles Python bootstrapping, variable substitution, and error handling exactly as `snow stage execute` does

## Files changed

- `src/snowflake/cli/_plugins/stage/manager.py`: New `execute_from_local_path()` method on `StageManager`
- `src/snowflake/cli/_plugins/snowpark/commands.py`: New `execute-file` command with `--on-error` and `-D` options
- `tests/__snapshots__/test_help_messages.ambr`: Snapshot updated for new command (398 tests passing)

## Test plan

- [ ] `snow snowpark execute-file ./script.py` executes a single local Python file
- [ ] `snow snowpark execute-file ./script.sql` executes a single local SQL file
- [ ] `snow snowpark execute-file ./scripts/` executes all `.py`/`.sql` files in a directory
- [ ] `snow snowpark execute-file ./scripts/*.py` works with a glob pattern
- [ ] A `requirements.txt` in the same directory is picked up automatically
- [ ] An unsupported file type produces a clear `CliError`
- [ ] `--on-error continue` continues past errors in multi-file runs
- [ ] `snow snowpark execute-file --help` shows correct usage

### DIFF
--- a/src/snowflake/cli/_plugins/snowpark/commands.py
+++ b/src/snowflake/cli/_plugins/snowpark/commands.py
@@ -16,7 +16,7 @@ from __future__ import annotations
 
 import logging
 from collections import defaultdict
-from typing import Dict, Optional, Set, Tuple
+from typing import Dict, List, Optional, Set, Tuple
 
 import typer
 from click import ClickException, UsageError
@@ -70,8 +70,13 @@ from snowflake.cli.api.cli_global_context import (
 from snowflake.cli.api.commands.decorators import (
     with_project_definition,
 )
+from snowflake.cli.api.commands.common import (
+    OnErrorType,
+)
 from snowflake.cli.api.commands.flags import (
+    ExecuteVariablesOption,
     ForceReplaceOption,
+    OnErrorOption,
     IfExistsOption,
     PruneOption,
     ReplaceOption,
@@ -438,6 +443,25 @@ def execute(
         execution_identifier=execution_identifier, object_type=object_type
     )
     return SingleQueryResult(cursor)
+
+
+@app.command("execute-file", requires_connection=True)
+def execute_file(
+    local_path: str = typer.Argument(
+        ...,
+        help="Path to a local .py or .sql file, directory, or glob pattern to execute on Snowflake. "
+        "For example: ./script.py, ./scripts/, ./scripts/*.py",
+        show_default=False,
+    ),
+    on_error: OnErrorType = OnErrorOption,
+    variables: Optional[List[str]] = ExecuteVariablesOption,
+    **options,
+) -> CommandResult:
+    """Uploads a local Python or SQL file to a temporary stage and executes it on Snowflake."""
+    results = StageManager().execute_from_local_path(
+        local_path_str=local_path, on_error=on_error, variables=variables
+    )
+    return CollectionResult(results)
 
 
 @app.command("list", requires_connection=True)
--- a/src/snowflake/cli/_plugins/stage/manager.py
+++ b/src/snowflake/cli/_plugins/stage/manager.py
@@ -666,6 +666,64 @@ class StageManager(SqlExecutionMixin):
 
         return results
 
+    def execute_from_local_path(
+        self,
+        local_path_str: str,
+        on_error: OnErrorType,
+        variables: Optional[List[str]] = None,
+    ) -> list:
+        """
+        Uploads local file(s) to a temporary stage and executes them on Snowflake.
+        Accepts a plain file path, a directory, or a glob pattern (e.g. ./scripts/*.py).
+        Only .sql and .py files are executed.
+        """
+        local_path = Path(local_path_str)
+
+        if glob.has_magic(local_path_str):
+            matched_files = [Path(f) for f in glob.glob(local_path_str, recursive=True)]
+        elif local_path.is_dir():
+            matched_files = [f for f in local_path.rglob("*") if f.is_file()]
+        else:
+            matched_files = [local_path]
+
+        supported_files = [
+            f
+            for f in matched_files
+            if f.is_file() and f.suffix in EXECUTE_SUPPORTED_FILES_FORMATS
+        ]
+
+        if not supported_files:
+            raise CliError(
+                f"No supported files found at '{local_path_str}'. "
+                f"Only {', '.join(EXECUTE_SUPPORTED_FILES_FORMATS)} files are supported."
+            )
+
+        if glob.has_magic(local_path_str):
+            idx = next(i for i, c in enumerate(local_path_str) if c in "*?[")
+            base_dir = Path(local_path_str[:idx]).parent
+        elif local_path.is_dir():
+            base_dir = local_path
+        else:
+            base_dir = local_path.parent
+
+        tmp_stage_name = f"snowflake_cli_tmp_stage_{int(time.time())}"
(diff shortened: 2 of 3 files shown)


src/snowflake/cli/_plugins/stage/manager.py
```diff
@@ -666,6 +666,64 @@ def execute(
...
+
+        tmp_stage_name = f"snowflake_cli_tmp_stage_{int(time.time())}"
+        tmp_stage_fqn = FQN.from_stage(tmp_stage_name).using_connection(conn=self._conn)
+        tmp_stage = tmp_stage_fqn.identifier
+        self.create(tmp_stage_fqn, temporary=True)
+
+        for file in supported_files:
+            self.put(local_path=file, stage_path=f"@{tmp_stage}/")
+
+        requirements_file = base_dir / "requirements.txt"
```
sfc-gh-olorek, line 717: Does uploading `requirements.txt` to the stage actually cause Snowpark to install the deps at runtime? AFAIK just `PUT`-ing it next to the script doesn't feed it to the Snowpark bootstrapper that `StageManager.execute` uses for `.py` files — `EXECUTE IMMEDIATE` on a `.py` doesn't resolve requirements on its own.

Have you verified end-to-end that packages actually get picked up? If not, I'd rather drop the requirements bit from this PR than ship a feature that silently does nothing. The PR description sells this pretty confidently so I want to make sure it holds up.

src/snowflake/cli/_plugins/stage/manager.py
```diff
@@ -666,6 +666,64 @@ def execute(
...
+
+        if glob.has_magic(local_path_str):
+            idx = next(i for i, c in enumerate(local_path_str) if c in "*?[")
+            base_dir = Path(local_path_str[:idx]).parent
+        elif local_path.is_dir():
+            base_dir = local_path
+        else:
+            base_dir = local_path.parent
+
+        tmp_stage_name = f"snowflake_cli_tmp_stage_{int(time.time())}"
```
sfc-gh-olorek, line 709: 1-second resolution — two invocations in the same second will collide on the stage name. Not super likely in interactive use but easy to hit from scripts or CI matrices.

Can we use `uuid.uuid4().hex[:8]` (either appended to the timestamp or replacing it)?

src/snowflake/cli/_plugins/stage/manager.py
```diff
@@ -666,6 +666,64 @@ def execute(
...
+        Uploads local file(s) to a temporary stage and executes them on Snowflake.
+        Accepts a plain file path, a directory, or a glob pattern (e.g. ./scripts/*.py).
+        Only .sql and .py files are executed.
+        """
+        local_path = Path(local_path_str)
+
+        if glob.has_magic(local_path_str):
+            matched_files = [Path(f) for f in glob.glob(local_path_str, recursive=True)]
+        elif local_path.is_dir():
+            matched_files = [f for f in local_path.rglob("*") if f.is_file()]
```
sfc-gh-olorek, line 685: `rglob("*")` walks the entire tree — `.git/`, `.venv/`, `node_modules/`, `__pycache__/`, … — before we filter by suffix. If someone points this at a repo root it's a lot of `stat()` calls for nothing.

Prefer `rglob("*.py")` + `rglob("*.sql")` — same result without the extra traversal.

src/snowflake/cli/_plugins/stage/manager.py
```diff
@@ -666,6 +666,64 @@ def execute(
...
+            if f.is_file() and f.suffix in EXECUTE_SUPPORTED_FILES_FORMATS
+        ]
+
+        if not supported_files:
+            raise CliError(
+                f"No supported files found at '{local_path_str}'. "
+                f"Only {', '.join(EXECUTE_SUPPORTED_FILES_FORMATS)} files are supported."
+            )
+
+        if glob.has_magic(local_path_str):
```
sfc-gh-olorek, line 701: Same `glob.has_magic(local_path_str)` check as line 682 — can we compute it once at the top of the method and reuse? Also would let us collapse the two if/elif/else ladders into one pass if we wanted.

src/snowflake/cli/_plugins/stage/manager.py
```diff
@@ -666,6 +666,64 @@ def execute(
 
         return results
 
+    def execute_from_local_path(
+        self,
+        local_path_str: str,
+        on_error: OnErrorType,
+        variables: Optional[List[str]] = None,
+    ) -> list:
```
sfc-gh-olorek, line 674: Can we use the same return type as `self.execute()` rather than bare `list`? Helps type-check callers and documents the shape.

src/snowflake/cli/_plugins/stage/manager.py
```diff
@@ -666,6 +666,64 @@ def execute(
 
         return results
 
+    def execute_from_local_path(
+        self,
+        local_path_str: str,
```
sfc-gh-olorek, line 671: Nit: accept `str | Path` here? Callers often have a `Path` already and the `Path(local_path_str)` conversion on line 680 is a no-op in that case.

src/snowflake/cli/_plugins/snowpark/commands.py
```diff
@@ -70,8 +70,13 @@
 from snowflake.cli.api.commands.decorators import (
     with_project_definition,
 )
+from snowflake.cli.api.commands.common import (
+    OnErrorType,
+)
```
sfc-gh-olorek, line 75: Nit: `OnErrorType` lives in `snowflake.cli.api.commands.common`, but we're already importing from `snowflake.cli.api.commands.flags` right below. Two adjacent `from snowflake.cli.api.commands.*` blocks feels a bit scattered — could `OnErrorType` live in `flags` alongside `OnErrorOption` instead, or is there a reason it's split?

sfc-gh-olorek (commented):
Thanks for pushing this through — nice reuse of `StageManager.execute` instead of reimplementing the enumeration/variable-substitution logic.

Main blocker for me: no tests. The help snapshot is the only test change, and that's not enough coverage for ~250 lines of new behaviour. `tests/stage/test_stage.py` has the right pattern for `execute_from_local_path` (mock `StageManager.execute_query` the way the existing `execute` tests do), and something under `tests/snowpark/` would cover the CLI wiring.

A few inline comments below.

PR #2898, opened 2026-04-20
Snowflake Apps: Merge `snow __app` commands into `snow app`

sfc-gh-gbloom:
### Pre-review checklist

### Changes description

Unifies the Native App and Snowflake Apps Deploy command groups so users get a single `snow app` namespace. The CLI auto-detects which flow to run based on the entity type in `snowflake.yml`:

- `application` / `application package` entities → **Native App** flow (existing behavior)
- `snowflake-app` entities → **Snowflake Apps Deploy** flow (previously hidden under `snow __app`)

Shared subcommands (`bundle`, `deploy`, `validate`, `open`, `events`, `teardown`) automatically pick the correct behavior. Native-App-only commands (`run`, `diff`, `publish`) and the Snowflake-App-only `setup` command produce clear errors when invoked against the wrong project type. Cross-flow option misuse (e.g. `--upload-only` against a Native App project) is rejected with explicit error messages naming the offending option.

### Core changes
- **New `@with_app_flow_routing()` decorator** in `nativeapp/v2_conversions/compat.py` — injects `--entity-id`, `--package-entity-id`, `--app-entity-id`; detects flow at runtime (hybrid: entity-id lookup falls back to entity-type scanning); injects `app_flow` kwarg into the wrapped handler. V1 project definitions are auto-routed to Native App. Mixed-type projects without disambiguation produce a clear error.
- **New `@native_app_only()` decorator** for `run`, `diff`, `publish` — produces a clear error on `snowflake-app`-only projects before entity resolution complains.
- **Shared commands merged** in `nativeapp/commands.py`: `bundle`, `deploy`, `validate`, `open`, `events`, `teardown` branch on `app_flow` and reject wrong-flow options.
- **`snow app setup`** added to the `snow app` group (Snowflake Apps Deploy only; no `@with_project_definition()` since it creates the project file).
- **Snowflake Apps Deploy implementations extracted** in `apps/commands.py` as plain Python functions (no Typer coupling) called from the unified handlers.
- Renamed "Snowflake Apps" to "Snowflake Apps Deploy" as per latest branding decision.

### Cleanup
- Deleted `apps/plugin_spec.py` (no separate command group anymore).
- Removed `"__app"` entry from `builtin_plugins.py`.
- Deleted `ENABLE_SNOWFLAKE_APPS` from `feature_flags.py` (the flag is no longer needed).
- Removed `_APP_COMMAND_NAME` constant from `apps/manager.py`; error messages now say `snow app`.

### Help text
- Group help: `"Manages Snowflake Native Apps and Snowflake Apps Deploy."`
- Shared command docstrings have explicit per-flow sections.
- Flow-specific options labeled `(Native App only)` / `(Snowflake Apps Deploy only)`.


src/snowflake/cli/_plugins/nativeapp/commands.py
```diff
@@ -66,7 +79,7 @@
 
 app = SnowTyperFactory(
     name="app",
-    help="Manages a Snowflake Native App",
+    help="Manages Snowflake Native Apps and Snowflake Apps.",
```
sfc-gh-olorek: Just flagging: this string is now visible on every `snow app --help` invocation. Before this PR, the whole Snowflake App surface was hidden behind a feature flag, so this is effectively the public announcement. Fine if the rollout has been coordinated, but wanted to make sure that's intentional.

src/snowflake/cli/_plugins/nativeapp/commands.py
```diff
@@ -75,15 +88,98 @@
...
+        joined = ", ".join(sorted(set_options))
+        raise ClickException(
+            f"'snow app {command}' was invoked against a Native App entity "
+            f"(application / application package), but the following "
+            f"Snowflake App options were provided: {joined}. "
+            f"Remove them and try again."
+        )
+
+
+@app.command("setup", requires_connection=True)
```
sfc-gh-olorek, line 135: `setup` is Snowflake-App-only and the docstring calls that out, but the command itself is registered unconditionally on the `snow app` group. Before this PR it lived under the hidden `__app` group and was gated by `ENABLE_SNOWFLAKE_APPS`. If Snowflake Apps is still preview, this probably wants either `hidden=FeatureFlag.ENABLE_SNOWFLAKE_APPS.is_disabled` or a flag check inside the body so it doesn't show up in `snow app --help` for GA users.

sfc-gh-olorek (commented):
Overall direction of unifying the command groups makes sense, but I'm a bit concerned about the gating side of this. On main, the Snowflake App surface is hidden two ways: the `snow __app` double-underscore convention, and `is_hidden=FeatureFlag.ENABLE_SNOWFLAKE_APPS.is_disabled` on the command group. This PR removes both, and it also deletes `ENABLE_SNOWFLAKE_APPS` from `feature_flags.py`. If Snowflake Apps is now GA, then great, ignore this. But if it isn't, we're shipping the whole surface (help text, flags, `setup`) to everyone on the next CLI release.

Could you confirm GA status with whoever owns the Snowflake Apps rollout? If we're still in preview, I'd suggest keeping `ENABLE_SNOWFLAKE_APPS` and having `with_app_flow_routing` raise a clear "not yet GA, set this flag to opt in" error when `flow == SNOWFLAKE_APP` and the flag is disabled. You could also keep `setup` hidden via `hidden=FeatureFlag.ENABLE_SNOWFLAKE_APPS.is_disabled`.

Also — heads up — `tests/apps/test_commands.py` still references `FeatureFlag.ENABLE_SNOWFLAKE_APPS` in dozens of places; once the enum entry is gone, that file won't import. Worth a CI check.

Nice work on `@with_app_flow_routing` otherwise; the shape is clean and the error messages are genuinely helpful.

src/snowflake/cli/_plugins/nativeapp/v2_conversions/compat.py
```diff
@@ -235,3 +266,205 @@ def wrapper(*args, **kwargs):
...
+            entity_type = str(getattr(entity, "type", "")).lower()
+            if entity_type in NATIVE_APP_ENTITY_TYPES:
+                return AppFlow.NATIVE_APP
+            if entity_type in SNOWFLAKE_APP_ENTITY_TYPES:
+                return AppFlow.SNOWFLAKE_APP
+            raise ClickException(
+                f"Entity '{entity_id}' has type '{entity_type}', which is not "
+                f"supported by 'snow app' commands."
+            )
+        # Entity id was passed but does not exist -- let the per-flow handler
```
sfc-gh-olorek: Minor UX thing: if the project contains only `snowflake-app` entities and the user mistypes `--entity-id`, we fall through here and route to Native App, which then errors with something like "no application entity found" — which is misleading. Could we peek at the project entity types here and route to Snowflake App if that's all we have? The per-flow handler would still produce the specific "entity X not found" error, just in the right flow.

src/snowflake/cli/_plugins/nativeapp/commands.py
```diff
@@ -404,73 +609,111 @@ class RecordType(Enum):
...
+                "--until": until or None,
+                "--type": record_types if record_types else None,
+                "--scope": scopes if scopes else None,
+                "--consumer-org": consumer_org or None,
+                "--consumer-account": consumer_account or None,
+                "--consumer-app-hash": consumer_app_hash or None,
+                "--first": first if first >= 0 else None,
+                "--follow": True if follow else None,
+                "--follow-interval": (
+                    follow_interval if follow_interval != 10 else None
```
sfc-gh-olorek: The other options in this block use `None`-vs-default sentinels, but this one compares against the literal `10`, which is also the typer default. Side effect: if a user explicitly passes `--follow-interval 10` against a `snowflake-app` project, we silently accept a Native-App-only flag instead of rejecting it. Tiny, but worth making consistent — either switch to a `None` default on the typer option, or check against the declared default in one place.

src/snowflake/cli/_plugins/nativeapp/commands.py
sfc-gh-gbloom: Yes, this is intentional and makes Snowflake Apps PuPr. The feature flag was making the command hidden, but now that we are merging with the existing command it would be confusing to have the flag only control the help messages.

src/snowflake/cli/_plugins/nativeapp/commands.py
sfc-gh-gbloom: Same as above, the command should be registered unconditionally

src/snowflake/cli/_plugins/nativeapp/v2_conversions/compat.py
sfc-gh-gbloom: The fall-through logic already routes to the correct flow when only one flow type is present in the project, but the inline comment was misleading. Updated the comment to explain the behavior explicitly, and added tests.

src/snowflake/cli/_plugins/nativeapp/commands.py
sfc-gh-gbloom: Good call, extracted the default to a constant and changed the Typer default to None

sfc-gh-olorek (approved):
lgtm

PR #2899, opened 2026-04-21
feat: use ALTER STREAMLIT instead of CREATE OR REPLACE for existing apps

sfc-gh-zblackwood:
## Summary

- When `snow streamlit deploy --replace` is called on an app that **already exists**, use `ALTER STREAMLIT ... SET` instead of `CREATE OR REPLACE STREAMLIT`
- This preserves existing grants and permissions on the app, which `CREATE OR REPLACE` would drop
- Behavior is unchanged when the app does not yet exist (`CREATE OR REPLACE` is still used for new apps)

## What changed

**`streamlit_entity.py`**
- Added `get_alter_sql()` method that generates `ALTER STREAMLIT <name> SET <properties>` — mirrors `get_deploy_sql()` but without the CREATE prefix
- `_deploy_versioned` and `_deploy_legacy` now receive `object_exists` and call `get_alter_sql()` instead of `get_deploy_sql(replace=True)` when the app already exists

**`test_streamlit_commands.py`**
- Added `test_deploy_versioned_replace_existing_uses_alter` — verifies ALTER is used for existing versioned apps
- Added `test_deploy_legacy_replace_existing_uses_alter` — verifies ALTER is used for existing legacy apps

## Test plan

- [x] All existing streamlit tests pass (`tests/streamlit/`)
- [x] New tests cover the `--replace` + existing app path for both versioned and legacy deployments
- [ ] Manual integration test: deploy an app, grant usage to a role, redeploy with `--replace`, verify grant is preserved

.... Generated with [Cortex Code](https://go/cortex-cli)


sfc-gh-zblackwood:
Closing to re-open under zachary-blackwood_snow account.

sfc-gh-wschmitt (approved):
blessing this from the streamlit side

src/snowflake/cli/_plugins/streamlit/streamlit_entity.py
```diff
@@ -158,7 +169,102 @@ def get_add_live_version_sql(
...
+                clauses.append(self.model.get_imports_sql())
+
+        desired_wh = self.model.query_warehouse
+        if not desired_wh:
+            self._workspace_ctx.console.warning(
+                "[Deprecation] In next major version we will remove default query_warehouse='streamlit'."
+            )
+            desired_wh = "streamlit"
+        if not current or _id(cur.get("query_warehouse")) != _id(desired_wh):
+            clauses.append(f"QUERY_WAREHOUSE = {desired_wh}")
```
sfc-gh-mraba: `desired_wh` is taken directly from `self.model.query_warehouse` (user's `snowflake.yml)`. There are no quotes and no `IDENTIFIER()` wrapper. A malicious value like:
```
    query_warehouse: "wh; DROP STREAMLIT my_app;"
```
would produce:
```
    ALTER STREAMLIT IDENTIFIER('my_app') SET
    QUERY_WAREHOUSE = wh; DROP STREAMLIT my_app;;
```

Note: `get_deploy_sql` has the same bug, so this is pre-existing, but the PR propagates it.
`get_deploy_sql` quotes the fallback `'streamlit' `with single quotes but `get_alter_sql` does not — though both unquote user-supplied values.

If `snowflake.yml` values are partially derived from external inputs (PR titles, user-supplied parameters, environment variables), injection is reachable by a third party. Any future code path that populates these model fields from a non-local source (e.g., fetching config from an API) would become exploitable without code changes.

How about ` to_string_literal` , `to_identifier` and `to_quoted_identifier` from `snowflake.cli.api.project.util` ?

PR #2907, opened 2026-04-23
Support non-interactive shells in the `dcm purge` command

sfc-gh-lkyc:
### Pre-review checklist

### Changes description
SNOW-3382389
Support non-interactive shells in the `dcm purge` command.

Adds `--force`, `--interactive`, and `--no-interactive` to `snow dcm purge` so the command can run non-interactively, like in CI/CD pipelines. Previously, it always prompted the user to type `purge <identifier>`, blocking automation.

Behavior matches the nativeapp convention:

  | `--force` | `--interactive` | Result |
  |---|---|---|
  | false | true (default in TTY) | Show the existing typed-identifier confirmation prompt |
  | false | false (default in non-TTY, or via `--no-interactive`) | Fail with `Cannot purge the DCM project non-interactively without --force.` |
  | true | (any) | Skip the prompt and proceed |

  The existing `_confirm_purge` (which asks for the full project identifier) is preserved in interactive mode — it's a safety check for a highly destructive operation and doesn't fit the simple yes/no model used by nativeapp's `Policy` classes.

#### Shared-code refactor
DCM cannot import from other plugins, so the primitives that DCM needs were extracted to `src/snowflake/cli/api/`:
  - `ForceOption`, `InteractiveOption` → `src/snowflake/cli/api/commands/flags.py`
  - `is_tty_interactive()` → `src/snowflake/cli/api/utils/tty.py` (also consolidates a duplicate copy that lived in `_app/telemetry.py`)

### DIFF
--- a/src/snowflake/cli/_app/telemetry.py
+++ b/src/snowflake/cli/_app/telemetry.py
@@ -33,6 +33,7 @@ from snowflake.cli.api.commands.execution_metadata import ExecutionMetadata
 from snowflake.cli.api.config import get_feature_flags_section
 from snowflake.cli.api.output.formats import OutputFormat
 from snowflake.cli.api.utils.error_handling import ignore_exceptions
+from snowflake.cli.api.utils.tty import is_tty_interactive
 from snowflake.connector import ProgrammingError
 from snowflake.connector.telemetry import (
     TelemetryData,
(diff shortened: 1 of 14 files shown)


sfc-gh-jwilkowski (commented):
I think you need to also update test_dcm_purge integration test as it's likely to fail without --force option. otherwise lgtm, thank you especially for refactoring and moving shared utilities to api

sfc-gh-lkyc:
> I think you need to also update test_dcm_purge integration test as it's likely to fail without --force option. otherwise lgtm, thank you especially for refactoring and moving shared utilities to api

Good catch — fixed.

sfc-gh-jwilkowski (approved):

PR #2909, opened 2026-04-25
Fixed dcm commands crashing when `sources/` contains hidden files

sfc-gh-lkyc:
### Pre-review checklist

### Changes description
Fixes every `snow dcm ...` command that syncs the project (`plan`, `deploy`, `analyze`, `preview`, `purge`, `test`) crashing when `sources/` contains, or any subdirectory contain, a hidden file (`.gitignore`, `.DS_Store`, etc.): 

> 253006: File doesn't exist: ['/Users/.../output/bundle/sources/dbt/*']

 #### Root cause
Commit 22eb51ab switched DCM to `bundle_artifacts` +  `StageManager.put_recursive(temp_directory=bundle_root)`. `put_recursive` walks the bundle bottom-up and issues `PUT file://.../dir/*` per directory. The Snowflake Python connector expands that glob via `glob.glob(...)`, which does not match files starting with `.`. When a directory ends up containing only a dotfile (after its subdirs have been processed and `rmtree`'d), the glob returns an empty list and the connector raises 253006.

v3.16 did not have this problem because it enumerated source files individually and uploaded each by exact path — no wildcard expansion, dotfiles included.

#### Fix
A small, targeted change to `DCMProjectManager.sync_local_files` that preserves the `bundle_artifacts` + `put_recursive(temp_directory=bundle_root)` path introduced by 22eb51ab.

  `_build_upload_plan` returns an `UploadPlan` that is the single source of truth for what gets uploaded. One walk of the source tree populates three fields:

  - **`artifacts`** — `PathMapping`s for `bundle_artifacts`. `SOURCES_FOLDER` is added with `ignore=[".*"]` so dotfiles never reach the bundle and can't trigger the 253006 crash.
  - **`individual_files`** — list of `FileUpload(file, dest)` entries for dotfiles, uploaded per-file with exact paths (which bypass `glob.glob`'s wildcard expansion entirely).
  - **`relative_paths_to_upload`** — every file the user will see on the stage, used by the "Local changes to be deployed" banner.

  `sync_local_files` consumes the plan: `bundle_artifacts(plan.artifacts)` → `put_recursive(bundle_root, temp_directory=bundle_root)` for the batched upload of visible files → per-file `stage_manager.put(...)` for each `FileUpload` in `plan.individual_files`. `_report_files_to_be_deployed` reads directly from `plan.relative_paths_to_upload` — no filtering, no walking, no risk of drift between what's reported and what's uploaded.

#### Behavior changes visible to users
- **CLI output restored to pre-[22eb51ab](https://github.com/snowflakedb/snowflake-cli/commit/22eb51ab47c53c4afb43f7cc1d30d73e95e5d9ca) format.** `Creating temporary stage <FQN>.` step, `Local changes to be deployed:` banner with `added: <rel> -> <rel>` lines for every file (including dotfiles), and `Uploading files from local <bundle_root> directory to temporary stage.` step. The banner is driven by `plan.relative_paths_to_upload`, so it always matches what actually uploads.

### DIFF
--- a/src/snowflake/cli/_plugins/dcm/manager.py
+++ b/src/snowflake/cli/_plugins/dcm/manager.py
@@ -12,11 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import logging
+from dataclasses import dataclass, field
 from pathlib import Path
 from typing import List
 
 from snowflake.cli._plugins.dcm.models import MANIFEST_FILE_NAME, SOURCES_FOLDER
 from snowflake.cli._plugins.dcm.utils import collect_output
+from snowflake.cli._plugins.stage.diff import _to_diff_line
 from snowflake.cli._plugins.stage.manager import StageManager
 from snowflake.cli.api.artifacts.utils import bundle_artifacts
 from snowflake.cli.api.commands.utils import parse_key_value_variables
@@ -34,6 +36,19 @@ from snowflake.cli.api.stage_path import StagePath
 from snowflake.connector.cursor import SnowflakeCursor
 
 log = logging.getLogger(__name__)
+
+
+@dataclass
+class FileUpload:
+    file: Path
+    dest: str
+
+
+@dataclass
+class UploadPlan:
+    artifacts: List[PathMapping] = field(default_factory=list)
+    individual_files: List[FileUpload] = field(default_factory=list)
+    relative_paths_to_upload: List[str] = field(default_factory=list)
 
 
 class DCMProjectManager(SqlExecutionMixin):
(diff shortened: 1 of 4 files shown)


tests_integration/test_dcm_project.py
```diff
@@ -112,9 +112,24 @@ def test_dcm_deploy(
...
+    with project_directory("dcm_project") as project_root:
+        definitions = project_root / "sources" / "definitions"
+        (definitions / ".hidden_def.sql").write_text(
+            f"define table identifier('{hidden_file_table_fqn}') (fooBar string);\n"
+        )
+        hidden_folder = definitions / ".hidden"
+        hidden_folder.mkdir()
+        (hidden_folder / "in_hidden.sql").write_text(
+            f"define table identifier('{hidden_folder_table_fqn}') (fooBar string);\n"
+        )
```
sfc-gh-jwilkowski, line 131: great that you're extending this case with this case 💪

src/snowflake/cli/_plugins/dcm/manager.py
```diff
@@ -12,11 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import logging
+from dataclasses import dataclass, field
 from pathlib import Path
 from typing import List
 
 from snowflake.cli._plugins.dcm.models import MANIFEST_FILE_NAME, SOURCES_FOLDER
 from snowflake.cli._plugins.dcm.utils import collect_output
+from snowflake.cli._plugins.stage.diff import _to_diff_line
```
sfc-gh-jwilkowski, line 21: nit: looks like a candidate to be moved to console's api, as we're accessing private function from stage plugin

src/snowflake/cli/_plugins/dcm/manager.py
```diff
@@ -292,14 +321,52 @@ def sync_local_files(
...
+    def _build_upload_plan(source_path: Path, stage_root: str) -> UploadPlan:
+        plan = UploadPlan()
+        DCMProjectManager._add_manifest(plan, source_path)
+        DCMProjectManager._add_sources(plan, source_path, stage_root)
+        return plan
 
-        artifacts.append(PathMapping(src=MANIFEST_FILE_NAME))
+    @staticmethod
+    def _add_manifest(plan: UploadPlan, source_path: Path) -> None:
+        if (source_path / MANIFEST_FILE_NAME).exists():
```
sfc-gh-jwilkowski: all the commands that upload files assert that manifest exists in `_resolve_context_with_required_manifest`. At this point we either trust that we applied that check correctly (os the if is not needed) or we don't trust ourselves in which case I'd add an else branch that would produce a meaningful error about missing manifest

src/snowflake/cli/_plugins/dcm/manager.py
sfc-gh-lkyc: In the near future, we have to revisit what we print on the screen while uploading files. I used this mechanism only to restore the v3.16 UX, but I think we can do it better, or at least don't use the diff that prints added files in red font, which was a concern raised some time ago. As we no longer use the file-syncing mechanism because we always add new files, we can rework it for DCM and maybe for other plugins.

sfc-gh-lkyc: But your point is valid - I missed that I used a "private" function

src/snowflake/cli/_plugins/dcm/manager.py
sfc-gh-lkyc: Fair point! Fixed.

src/snowflake/cli/_plugins/dcm/manager.py
sfc-gh-jwilkowski: agree. That's not critical for now, just wanted to flag it

sfc-gh-jwilkowski (approved):

PR #2917, opened 2026-04-28
feat: [SNOW-3392936] escape single quotes in FQN.sql_identifier

sfc-gh-jwilkowski:
### Pre-review checklist

### Changes description
https://snowflakecomputing.atlassian.net/browse/SNOW-3392936


tests/api/test_fqn.py
```diff
@@ -12,14 +12,111 @@
...
+
+    @pytest.mark.parametrize(
+        "payload",
+        [
+            # Single-quote breakout with DROP TABLE
+            '"app_name\'); DROP TABLE sensitive_data --"',
+            # Privilege escalation
+            '"x\'); GRANT ALL PRIVILEGES ON ACCOUNT TO ROLE attacker --"',
+            # Data exfiltration via UNION
+            '"t\'); SELECT * FROM secrets; --"',
```
sfc-gh-olorek: Nice set of cases here. One thing I think we might be missing: identifiers containing backslashes or non-ASCII chars now get run through `codecs.encode(..., 'unicode-escape')` inside `to_string_literal`, which changes the bytes that end up between the outer quotes. It should still round-trip through Snowflake's parser, but it'd be reassuring to cover e.g. `"café"` and `"a\\b"` so we don't silently regress that if `to_string_literal` ever gets tweaked. Not a blocker — happy to follow up in a separate PR if easier.

sfc-gh-olorek (dismissed):
LGTM — fix is in the right place and using `to_string_literal` matches the rest of the codebase. Left one non-blocking thought on test coverage.

sfc-gh-olorek (commented):
lgtm

sfc-gh-olorek (approved):
lgtm 👍

PR #2918, opened 2026-04-28
feat: [SNOW-3317601] fix column names on joins with duplicates

sfc-gh-jwilkowski:
Fixes: #2846 

### Pre-review checklist

### Changes description
https://snowflakecomputing.atlassian.net/browse/SNOW-3317601

### DIFF
--- a/src/snowflake/cli/api/output/types.py
+++ b/src/snowflake/cli/api/output/types.py
@@ -126,7 +126,9 @@ class StreamResult(CommandResult):
 
 class QueryResult(CollectionResult):
     def __init__(self, cursor: SnowflakeCursor | DictCursor):
-        self.column_names = [col.name for col in cursor.description]
+        self.column_names = self._uniquify_column_names(
+            [col.name for col in cursor.description]
+        )
         super().__init__(
             elements=self._prepare_payload(cursor),
             row_mapper=RespectingColumnTypesRowMapper(cursor.description),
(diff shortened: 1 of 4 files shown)


src/snowflake/cli/api/output/types.py
```diff
@@ -138,6 +140,29 @@ def _prepare_payload(self, cursor: SnowflakeCursor | DictCursor):
             return (k for k in cursor)
```
sfc-gh-olorek, line 145: Heads up: this `DictCursor` branch yields row dicts straight from the connector, which has already collapsed duplicate keys before `QueryResult` ever sees them. So for `DictCursor` callers, `self.column_names` can now say `['id', 'id_2']` while the actual row dicts still have only a single `id` key — `column_names` and row payloads silently disagree.

In practice `DictCursor` is only used for metadata queries (`show objects`, `describe`, etc.) where collisions don't happen, so it's unlikely to bite. But worth either extending the fix to rebuild the dicts here using the uniquified names, or dropping a short comment calling out that the asymmetry is intentional.

src/snowflake/cli/api/output/types.py
```diff
@@ -138,6 +140,29 @@ def _prepare_payload(self, cursor: SnowflakeCursor | DictCursor):
...
         return ({k: v for k, v in zip(self.column_names, row)} for row in cursor)
 
+    @staticmethod
+    def _uniquify_column_names(column_names: list[str]) -> list[str]:
+        """Disambiguate duplicate column names by appending a counter suffix."""
+        unique_names: list[str] = []
+        used_names: set[str] = set(column_names)
```
sfc-gh-olorek, line 152: Nit: consider a one-line comment here explaining why `used_names` is pre-seeded with the full input. This is the subtle bit that makes the `['id', 'id', 'id_2']` → `['id', 'id_3', 'id_2']` case work (the real `id_2` further down reserves that name so the duplicate `id` skips past it). The test at `test_query_result_uniquify_does_not_rename_real_column_matching_generated_suffix` locks the behavior in, but a reader looking at this function in isolation has to reconstruct the reasoning. Not blocking.

sfc-gh-olorek (commented):
Nice focused fix. Two inline notes — the DictCursor one is worth thinking about before merge, the other is a nit.

src/snowflake/cli/api/output/types.py
sfc-gh-jwilkowski: added

src/snowflake/cli/api/output/types.py
sfc-gh-jwilkowski: I'd not go into explaining such details. Overall intent of this function is to make columns unique, and that's what the name and docstring hint at. Tests will catch potential issues

sfc-gh-olorek (dismissed):
Thanks for the comment on DictCursor — addresses the asymmetry concern. Fine leaving the uniquify internals as-is; tests cover the edge case.

sfc-gh-olorek (approved):

PR #2919, opened 2026-04-28
Snowflake Apps: default --app-name to current directory in setup

sfc-gh-gbloom:
### Pre-review checklist

### Changes description
- Make `snow app setup --app-name` optional for Snowflake Apps Deploy
- Default app name to the current project directory name when `--app-name` is omitted

### DIFF
--- a/src/snowflake/cli/_plugins/apps/commands.py
+++ b/src/snowflake/cli/_plugins/apps/commands.py
@@ -73,7 +73,7 @@ SOURCE_MISSING = "missing"
 
 
 def snowflake_app_setup(
-    app_name: str,
+    app_name: Optional[str],
     dry_run: bool,
     compute_pool: Optional[str],
     build_eai: Optional[str],
@@ -83,9 +83,16 @@ def snowflake_app_setup(
     See the ``snow app setup`` command in
     :mod:`snowflake.cli._plugins.nativeapp.commands` for the CLI surface.
     """
-    if not re.fullmatch(r"[a-zA-Z0-9_]+", app_name):
+    resolved_app_name = app_name or Path.cwd().name
+    if not resolved_app_name:
         raise ClickException(
-            f"Invalid app name '{app_name}'. "
+            "Could not derive app name from the current directory. "
+            "Please provide --app-name."
+        )
+
+    if not re.fullmatch(r"[a-zA-Z0-9_]+", resolved_app_name):
+        raise ClickException(
+            f"Invalid app name '{resolved_app_name}'. "
             "Only letters, digits, and underscores are allowed."
         )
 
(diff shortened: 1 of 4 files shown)


src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -85,9 +85,16 @@ def snowflake_app_setup(
     See the ``snow app setup`` command in
     :mod:`snowflake.cli._plugins.nativeapp.commands` for the CLI surface.
     """
-    if not re.fullmatch(r"[a-zA-Z0-9_]+", app_name):
+    resolved_app_name = app_name or Path.cwd().name
```
sfc-gh-jfloyd: Do you think we should do some basic formatting on the directory name? e.g. replace spaces with "_"?

sfc-gh-gbloom: Good call, normalizing the directory name now

sfc-gh-jfloyd (approved):

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -83,9 +83,25 @@ def snowflake_app_setup(
     See the ``snow app setup`` command in
     :mod:`snowflake.cli._plugins.nativeapp.commands` for the CLI surface.
     """
-    if not re.fullmatch(r"[a-zA-Z0-9_]+", app_name):
+    resolved_app_name = app_name
+    if resolved_app_name is None:
+        derived_app_name = Path.cwd().name
+        # For implicit names, normalize directory strings into a valid
+        # identifier by mapping common separators to "_" and stripping
+        # all other disallowed characters.
```
sfc-gh-olorek, line 91: nit: you could collapse this into one regex — something like `re.sub(r"[^a-zA-Z0-9_]+", "_", derived_app_name).strip("_")`. Side benefit is it'd also map `.` / `/` / etc. to `_`, so `my.app.v2` lands as `my_app_v2` instead of `myappv2`. Not blocking, just caught my eye.

tests/apps/test_commands.py
```diff
@@ -1204,6 +1204,72 @@ def test_init_creates_file(self, mock_mgr_cls, mock_gen, runner, tmp_path):
...
+            "schema": "PARAM_SCHEMA",
+            "query_warehouse": "PARAM_WH",
+        }
+        invalid_project = tmp_path / "!@#"
+        invalid_project.mkdir()
+
+        with change_directory(invalid_project):
+            result = runner.invoke(["app", "setup"])
+
+        assert result.exit_code == 1
```
sfc-gh-olorek, line 1270: Might be worth one more test that locks in the "explicit name is *not* normalized" behavior — e.g. `--app-name "my-app"` should still error out with the invalid-name message. The inline comment in `commands.py` says it's intentional, but nothing in the tests would catch a future refactor that silently starts normalizing explicit input too.

sfc-gh-olorek (approved):
LGTM — just a couple of non-blocking thoughts inline.

PR #2923, opened 2026-04-30
fix: [SNOW-3388311] stop duplicate LOGIN_HISTORY events when snow is rejected by an auth policy

sfc-gh-olorek:
### Pre-review checklist

### Changes description

Resolves [SNOW-3388311](https://snowflakecomputing.atlassian.net/browse/SNOW-3388311).

A single `snow` invocation that is blocked by an authentication policy was emitting **3 `AUTHN_POLICY_AUTHN_ATTEMPT_REJECTED`** events plus an **`OVERFLOW_FAILURE_EVENTS_ELIDED`** into `LOGIN_HISTORY`, while SnowSQL emits only 1. This makes it hard for customers to correlate a policy rejection with an actual user action, and inflates the apparent attack surface in security dashboards.

#### Root cause

The duplicates are not driver retries — the Python connector only dials once per `connect()`. They come from the CLI accessing `get_cli_context().connection` from several code paths per command:

1. pre-command telemetry (`pre_execute`)
2. the command body itself
3. the global `exception_handler`
4. post-command telemetry flush (`post_execute` → `flush_telemetry`)

Every access goes through `OpenConnectionCache`. The cache cached *successful* connections but deliberately skipped caching failures (`logger.debug("... not caching")`), so each of the 4 accesses independently called `build_connection()` → `connect_to_snowflake()` → `snowflake.connector.connect()`. Each dial produces its own `AUTHN_POLICY_AUTHN_ATTEMPT_REJECTED` event; GS elides the 4th once the per-account/per-window cap is hit, adding `OVERFLOW_FAILURE_EVENTS_ELIDED`. Net: 3 rejected + 1 overflow = the customer-visible pattern.

(A couple of the accesses are wrapped in `@ignore_exceptions()` for telemetry, so the user never sees the extra failures — but GS still records them.)

#### Fix

Cache the exception on the first failure in `OpenConnectionCache._insert()` and short-circuit subsequent `__getitem__` calls for the same `ConnectionContext`:

- new `failures: dict[str, Exception]` alongside `connections`;
- `__getitem__` re-raises the cached exception instead of redialing;
- `_insert` stores the exception on failure before re-raising;
- new `clear_failures()` for callers that want to retry without tearing down cached *open* connections;
- existing `clear()` also wipes failures.

After the fix a rejected CLI invocation dials Snowflake exactly once — matching SnowSQL and the customer's expectation — regardless of how many internal accesses of the shared connection the command makes.

Scope notes:
- The cache is per-process, so this does **not** change cross-invocation retry behavior. Re-running `snow` still produces a new connect attempt (as intended).
- `snow sql` REPL is unaffected: `_initialize_connection` runs first at startup and `sys.exit`s on auth failure, so the REPL loop never sees a cached failure.
- I considered a narrower fix keyed on the specific error code (390202 → 250001), but the driver rewrites 390202 into a generic `ER_FAILED_TO_CONNECT_TO_DB` before the CLI sees it, so a narrow fix would require brittle message-string matching. Caching *all* deterministic failures per context is the right shape — any auth-time error would have emitted the same duplicates.

#### Tests

Added 3 unit tests in `tests/api/test_connections.py`:

- `test_connection_cache_caches_failures` — 3 accesses after an auth-policy-like `DatabaseError`, asserts only 1 underlying `connect()` call and that the same exception instance is re-raised each time.
- `test_connection_cache_clear_failures_allows_retry` — after `clear_failures()`, a subsequent access redials.
- `test_connection_cache_clear_also_forgets_failures` — the existing `clear()` also forgets failures.

Ran wider suites locally with no regressions:
- `tests/api/test_connections.py`: 8 passed.
- `tests/api/` + `tests/app/` + `tests/test_connection.py` + `tests/test_common_global_context.py`: 692 passed.
- `tests/sql/` + `tests/api/test_sql_execution.py`: 224 passed.

[SNOW-3388311]: https://snowflakecomputing.atlassian.net/browse/SNOW-3388311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -21,6 +21,7 @@
 ## New additions
 
 ## Fixes and improvements
+* Fixed duplicate `LOGIN_HISTORY` events (and `OVERFLOW_FAILURE_EVENTS_ELIDED`) emitted when a `snow` command is rejected by an authentication policy. The internal connection cache now remembers a failed connect attempt and re-raises it on subsequent accesses within the same process, instead of re-dialing Snowflake once per access of the shared connection.
 
 
 # v3.17.0
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -25,6 +25,7 @@
 ## Fixes and improvements
 * Fixed `SELECT *` output being corrupted when joined tables share column names. Duplicate column names are now disambiguated by appending a numeric suffix (e.g. `NAME`, `NAME_2`).
 * Fixed `snow connection generate-jwt` and `snow connection generate-workload-identity-token` failing with `Connection None is not configured` when used with `--temporary-connection`.
+* Fixed duplicate `LOGIN_HISTORY` events (and `OVERFLOW_FAILURE_EVENTS_ELIDED`) emitted when a `snow` command is rejected by an authentication policy. The internal connection cache now remembers a failed connect attempt and re-raises it on subsequent accesses within the same process, instead of re-dialing Snowflake once per access of the shared connection.
 
 
 # v3.17.0
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -25,7 +25,7 @@
 ## Fixes and improvements
 * Fixed `SELECT *` output being corrupted when joined tables share column names. Duplicate column names are now disambiguated by appending a numeric suffix (e.g. `NAME`, `NAME_2`).
 * Fixed `snow connection generate-jwt` and `snow connection generate-workload-identity-token` failing with `Connection None is not configured` when used with `--temporary-connection`.
-* Fixed duplicate `LOGIN_HISTORY` events (and `OVERFLOW_FAILURE_EVENTS_ELIDED`) emitted when a `snow` command is rejected by an authentication policy. The internal connection cache now remembers a failed connect attempt and re-raises it on subsequent accesses within the same process, instead of re-dialing Snowflake once per access of the shared connection.
+* The internal connection cache now remembers failed connect attempts and re-raises the original exception on subsequent accesses within the same process, instead of re-dialing Snowflake every time a command accesses the shared connection. This fixes, among other cases, the customer-visible duplicate `LOGIN_HISTORY` events (and `OVERFLOW_FAILURE_EVENTS_ELIDED`) previously emitted when a `snow` invocation was rejected by an authentication policy.
 
 
 # v3.17.0
--- a/tests/api/test_connections.py
+++ b/tests/api/test_connections.py
@@ -98,3 +98,99 @@ def test_connection_cache_caches(
         password="dummy_password",
         application_name="snowcli",
     )
+
+
+@mock.patch("snowflake.connector.connect")
+@mock.patch("snowflake.cli._app.snow_connector.command_info")
+def test_connection_cache_caches_failures(
+    mock_command_info, mock_connect, local_connection_cache, test_snowcli_config
+):
+    """Once a connect() call fails, subsequent accesses must re-raise without
+    re-dialing — otherwise auth-policy rejection logs duplicate LOGIN_HISTORY
+    events (one per access of the CLI's global connection: pre-command
+    telemetry, command body, error handler, post-command telemetry).
+    """
+    from snowflake.cli.api.exceptions import InvalidConnectionConfigurationError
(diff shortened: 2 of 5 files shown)


tests/api/test_connections.py
```diff
@@ -98,3 +98,99 @@ def test_connection_cache_caches(
...
+@mock.patch("snowflake.cli._app.snow_connector.command_info")
+def test_connection_cache_caches_failures(
+    mock_command_info, mock_connect, local_connection_cache, test_snowcli_config
+):
+    """Once a connect() call fails, subsequent accesses must re-raise without
+    re-dialing — otherwise auth-policy rejection logs duplicate LOGIN_HISTORY
+    events (one per access of the CLI's global connection: pre-command
+    telemetry, command body, error handler, post-command telemetry).
+    """
+    from snowflake.cli.api.exceptions import InvalidConnectionConfigurationError
```
sfc-gh-jwilkowski: Please avoid local imports and move them to top of the file instead

sfc-gh-olorek:
Follow-up commit `cfa90fd4` addresses a gap in the original fix:

The PR description claimed the REPL was unaffected because `_initialize_connection` runs at startup and `sys.exit`s on auth failure. That's true for **startup-time** failures, but not for failures that happen **mid-REPL**:

1. REPL starts, `_initialize_connection` succeeds → connection cached, cleanup timer set for 10 min.
2. User idles >10 min → `_cleanup(key)` closes and pops the connection.
3. Next query → `__getitem__` redials → if *this* redial fails (network blip, IdP flap, expired token), the exception gets cached.
4. Every subsequent query in the same REPL session re-raises the cached failure. The user has to quit and relaunch to recover.

Pre-PR, step 4 would just redial again and (often) succeed. The follow-up calls `connection_cache.clear_failures()` after any exception in the REPL loop, so the next query re-attempts. Added a unit test covering the recovery path.

tests/api/test_connections.py
```diff
@@ -98,3 +98,99 @@ def test_connection_cache_caches(
...
+    from snowflake.connector.errors import DatabaseError
```
sfc-gh-jwilkowski: please avoid local imports and prefer top of the file imports

tests/api/test_connections.py
```diff
@@ -98,3 +98,99 @@ def test_connection_cache_caches(
...
+        local_connection_cache[ctx]
+    assert mock_connect.call_count == 1
+
+    local_connection_cache.clear_failures()
+
+    local_connection_cache[ctx]
+    assert mock_connect.call_count == 2
+
+
+@mock.patch("snowflake.connector.connect")
```
sfc-gh-jwilkowski: Please check if those explicit mocks are really needed. There's a similar fixture already defined elsewhere in the codebase and it might be better to reuse it:
```
@pytest.fixture
def mock_connect(mock_ctx):
    with mock.patch("snowflake.connector.connect") as _fixture:
        ctx = mock_ctx()
        _fixture.return_value = ctx
        _fixture.mocked_ctx = _fixture.return_value
        yield _fixture
```

RELEASE-NOTES.md
```diff
@@ -21,6 +21,7 @@
 ## New additions
 
 ## Fixes and improvements
+* Fixed duplicate `LOGIN_HISTORY` events (and `OVERFLOW_FAILURE_EVENTS_ELIDED`) emitted when a `snow` command is rejected by an authentication policy. The internal connection cache now remembers a failed connect attempt and re-raises it on subsequent accesses within the same process, instead of re-dialing Snowflake once per access of the shared connection.
```
sfc-gh-jwilkowski: I think this release note covers a very specific failure mode, but the code change is wider (it will cache connection failures originating from other reasons as well). Please start with general description what this change does and then explain the specific case being solved

sfc-gh-jwilkowski (changes requested):

sfc-gh-olorek:
@sfc-gh-jwilkowski thanks for the reivew! I have addressed your comments

sfc-gh-jwilkowski (approved):

sfc-gh-olorek:
Superseded by #2962 — same commits (`2ce2a491`), re-opened from an upstream branch so branch protection on `main` can be satisfied. Closing here.

PR #2928, opened 2026-05-04
Avoid opening a Snowflake connection just to emit command telemetry (SNOW-3450376)

sfc-gh-olorek:
### Pre-review checklist

### Changes description

Fixes [SNOW-3450376](https://snowflakecomputing.atlassian.net/browse/SNOW-3450376).

#### Problem

Every command emits a per-command telemetry event via `CLITelemetryClient`, which walks through `cli_context.connection._telemetry`. Accessing `.connection` lazily opens a Snowflake connection — TCP + TLS + login — the first time it's touched.

That means commands declared with \`requires_connection=False\` (e.g. \`snow --version\`, \`snow --help\`, \`snow config list\`, \`snow connection list\`) were triggering an authentication round-trip purely to push a telemetry event onto the connector batcher. In restricted environments (offline, no default connection, bad credentials), these no-connection commands would fail or hang even though the command body itself never needed Snowflake.

#### Fix

Decouple telemetry emission from connection dialing:

- **\`OpenConnectionCache.peek(ctx)\`** — new accessor that returns the cached connection if one is already open, and \`None\` otherwise. No dial, no cache mutation, no lifetime refresh.
- **\`CLITelemetryClient\`** now uses \`peek()\` instead of \`.connection\`. When no connection exists yet, telemetry data is staged on an internal buffer (\`self._pending\`) and drained opportunistically the next time a connection becomes available — either via a later \`send()\` on the same client or via \`flush()\`.
- **Connection-dependent commands still get telemetry.** For commands declared \`requires_connection=True\`, \`flush_telemetry(force_dial=True)\` dials on post-execute so buffered events reach the batcher. This preserves the pre-fix guarantee that telemetry is delivered for those commands, and \`@ignore_exceptions()\` on the caller still swallows any dial failure.
- **\`pre_execute\` now calls \`connection_context.validate_and_complete()\` when \`requires_connection=True\`**, preserving the resolution of the default connection name that used to happen implicitly when telemetry opened the connection.

Net effect:
- Commands that don't need Snowflake no longer trigger a login just to emit telemetry. Their telemetry events are discarded instead of forcing an expensive (and often failing) round-trip.
- Commands that do need Snowflake still send their telemetry exactly once per command, through the same connector batcher, in the same order as today.

#### Tests

- \`tests/api/test_connections.py\`:
  - \`test_connection_cache_peek_returns_none_without_dialing\` — asserts \`peek\` returns \`None\` and does not call \`snowflake.connector.connect\` or mutate cache state.
  - \`test_connection_cache_peek_returns_existing_connection\` — asserts \`peek\` returns the same instance as \`__getitem__\` once a connection has been dialed, without a second dial.
  - \`test_connection_cache_peek_rejects_non_context\` — validates the \`ConnectionContext\` type check.
- \`tests/app/test_telemetry.py\`:
  - \`test_no_connection_required_command_does_not_dial_for_telemetry\` — invokes \`snow connection list\` (which is \`requires_connection=False\`) and asserts \`snowflake.connector.connect\` is not called at all. This is the direct regression guard called out in the Jira acceptance criteria.
  - \`test_pending_telemetry_drains_when_connection_opens\` — invokes \`snow connection test\` and asserts that both the usage event (buffered during \`pre_execute\` before a connection existed) and the result event reach the connector batcher, in order, and that the internal buffer is empty after \`flush\`.
- \`tests/api/commands/test_snow_typer.py\` updated for the new \`requires_connection\` kwarg on \`pre_execute\` / \`post_execute\` and the new \`force_dial\` kwarg on \`flush_telemetry\`.

Ran \`tests/api/ tests/app/ tests/test_connection.py tests/test_common_global_context.py tests/sql/ tests/api/test_sql_execution.py\` locally — 918 passed, 1 pre-existing xfail, 0 failures.

[SNOW-3450376]: https://snowflakecomputing.atlassian.net/browse/SNOW-3450376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -21,6 +21,7 @@
 ## New additions
 
 ## Fixes and improvements
+* Commands that do not require a Snowflake connection (e.g. `snow --version`, `snow --help`, `snow config list`, `snow connection list`) no longer open a Snowflake connection as a side-effect of emitting telemetry. Telemetry events for such commands are now discarded instead of forcing a login, matching what users expect offline and in minimal environments.
 
 
 # v3.17.0
--- a/src/snowflake/cli/_app/telemetry.py
+++ b/src/snowflake/cli/_app/telemetry.py
@@ -385,21 +385,25 @@ class CLITelemetryClient:
             self._pending.append(telemetry_data)
 
     def flush(self, force_dial: bool = False):
-        telemetry = self._telemetry
-        if telemetry is None and force_dial and self._pending:
-            # Command declared requires_connection=True (or otherwise expects
-            # telemetry to be delivered even on early failure). Dial now so
-            # buffered events still reach the batcher, matching the pre-
-            # SNOW-3450376 behaviour. ``@ignore_exceptions()`` on the caller
-            # swallows any dial failure, same as before.
-            telemetry = self._ctx.connection._telemetry  # noqa: SLF001
-        if telemetry:
-            self._drain_pending(telemetry)
-            telemetry.send_batch()
-        else:
-            # No connection, no dial; pending events are dropped, matching
-            # the pre-fix behaviour where ``@ignore_exceptions()`` swallowed
-            # any telemetry dial failure on no-connection commands.
+        try:
+            telemetry = self._telemetry
+            if telemetry is None and force_dial and self._pending:
+                # Command declared requires_connection=True (or otherwise
+                # expects telemetry to be delivered even on early failure).
+                # Dial now so buffered events still reach the batcher,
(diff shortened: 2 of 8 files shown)


sfc-gh-olorek:
Closing this stale draft. Internal development moved to the private repository in July 2026 and PR head branches are not mirrored outbound, so this draft cannot land here — anything still wanted has to be re-opened there. The work itself is still tracked in SNOW-3450376, which stays the source of truth. The branch is preserved.

PR #2930, opened 2026-05-04
Fix SQL injection via single-quote escaping in util helpers (SNOW-3411343, SNOW-3416954)

sfc-gh-olorek:
### Pre-review checklist

### Changes description

Fixes two related SQL injection findings in `src/snowflake/cli/api/project/util.py`:

**SNOW-3411343** — `identifier_to_show_like_pattern()` wrapped the escaped identifier in a bare `f"'{...}'"`. `unquote_identifier()` strips only the outer `"`-delimiters and preserves embedded `'` verbatim, and `escape_like_pattern()` only escapes `%`/`_`. A quoted identifier such as `"stream'lit); GRANT ROLE ACCOUNTADMIN TO USER attacker;--"` read from a malicious `snowflake.yml` would emit:

```
show streamlits like 'stream'lit); GRANT ROLE ACCOUNTADMIN TO USER attacker;--'
```

`execute_stream` splits on `;`, so the `GRANT` runs as a second statement under the victim's session role. This protects all five call sites — `sql_execution.show_specific_object`, `sf_sql_facade` stage existence / generic show, `application_package` version query, `apps/manager` image repository lookup.

**SNOW-3416954** — `to_string_literal()` escaped single quotes with a backslash (`\'`). Snowflake's default `STANDARD_ESCAPE_SEQUENCES=FALSE` treats `\` as a literal character, so `'x\';` is parsed as the literal `'x\'` followed by `; GRANT ...` — the same vulnerability on the `snow app version create` / `add-patch` flow via `manifest.yml` `version.label`.

**Fix:** both helpers now use Snowflake's standard single-quote doubling (`''`). `identifier_to_show_like_pattern` delegates to the fixed `to_string_literal`, so the SHOW-LIKE fix automatically tracks the string-literal helper. This avoids `codecs.encode(..., \"unicode-escape\")`, but the old behaviour of encoding control characters into `\xNN` / `\n` etc. was itself unsound under `STANDARD_ESCAPE_SEQUENCES=FALSE` — treating non-quote bytes as literal is the correct behaviour. Tests updated to reflect the new canonical encoding.

### Links

- Jira: [SNOW-3411343](https://snowflakecomputing.atlassian.net/browse/SNOW-3411343), [SNOW-3416954](https://snowflakecomputing.atlassian.net/browse/SNOW-3416954)
- CWE-89: Improper Neutralization of Special Elements used in an SQL Command
- Related (separate fix): [SNOW-3392936](https://snowflakecomputing.atlassian.net/browse/SNOW-3392936) — `FQN.sql_identifier` uses `IDENTIFIER(...)` wrapper, different sink.

[SNOW-3411343]: https://snowflakecomputing.atlassian.net/browse/SNOW-3411343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/nativeapp/codegen/snowpark/test_extension_function_utils.py
+++ b/tests/nativeapp/codegen/snowpark/test_extension_function_utils.py
@@ -116,6 +116,9 @@ def test_ensure_string_literal():
     # Unterminated leading/trailing single quotes are escaped by doubling ('').
     assert ef_utils.ensure_string_literal("'abc") == "'''abc'"
     assert ef_utils.ensure_string_literal("abc'") == "'abc'''"
+    # Regression guard: SINGLE_QUOTED_STRING_LITERAL_REGEX's '' alternation
+    # must keep this a valid literal so ensure_string_literal is idempotent.
+    assert ef_utils.ensure_string_literal("'abc''def'") == "'abc''def'"
 
 
 def test_ensure_all_string_literals():
--- a/tests/nativeapp/codegen/snowpark/test_extension_function_utils.py
+++ b/tests/nativeapp/codegen/snowpark/test_extension_function_utils.py
@@ -113,8 +113,9 @@ def test_ensure_string_literal():
     assert ef_utils.ensure_string_literal("abc") == "'abc'"
     assert ef_utils.ensure_string_literal("'abc'") == "'abc'"
     assert ef_utils.ensure_string_literal("'abc def'") == "'abc def'"
-    assert ef_utils.ensure_string_literal("'abc") == r"'\'abc'"
-    assert ef_utils.ensure_string_literal("abc'") == r"'abc\''"
+    # Unterminated leading/trailing single quotes are escaped by doubling ('').
+    assert ef_utils.ensure_string_literal("'abc") == "'''abc'"
+    assert ef_utils.ensure_string_literal("abc'") == "'abc'''"
 
 
 def test_ensure_all_string_literals():
--- a/tests/project/test_util.py
+++ b/tests/project/test_util.py
@@ -265,11 +265,11 @@ def test_escape_like_pattern(raw_string, escaped):
         # must not close the SQL string literal. The embedded single quote is escaped
         # by being doubled ('') so the entire payload ends up as a harmless LIKE pattern.
         (
-            "\"stream'lit); GRANT ROLE ACCOUNTADMIN TO USER attacker;--\"",
+            '"stream\'lit); GRANT ROLE ACCOUNTADMIN TO USER attacker;--"',
             "'stream''lit); GRANT ROLE ACCOUNTADMIN TO USER attacker;--'",
         ),
         # A bare single quote in the identifier is doubled.
-        ("\"a'b\"", "'a''b'"),
+        ('"a\'b"', "'a''b'"),
         # Multiple embedded single quotes are each doubled.
         ("\"a'b'c\"", "'a''b''c'"),
     ],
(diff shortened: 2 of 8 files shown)


tests/nativeapp/codegen/snowpark/test_extension_function_utils.py
```diff
@@ -113,8 +113,9 @@ def test_ensure_string_literal():
     assert ef_utils.ensure_string_literal("abc") == "'abc'"
     assert ef_utils.ensure_string_literal("'abc'") == "'abc'"
     assert ef_utils.ensure_string_literal("'abc def'") == "'abc def'"
-    assert ef_utils.ensure_string_literal("'abc") == r"'\'abc'"
-    assert ef_utils.ensure_string_literal("abc'") == r"'abc\''"
+    # Unterminated leading/trailing single quotes are escaped by doubling ('').
+    assert ef_utils.ensure_string_literal("'abc") == "'''abc'"
+    assert ef_utils.ensure_string_literal("abc'") == "'abc'''"
```
sfc-gh-jwilkowski, line 122: Consider adding an idempotency case for a literal that already contains `''` internally, e.g.:
```python
assert ef_utils.ensure_string_literal("'abc''def'") == "'abc''def'"
```
`SINGLE_QUOTED_STRING_LITERAL_REGEX` does accept `''` (via the `''` alternation), so `ensure_string_literal` correctly short-circuits and returns the value unchanged — but there's no test for this. A future change that accidentally drops the `''` alternation from the regex would silently cause re-escaping of already-valid literals.

tests/project/test_util.py
```diff
@@ -241,6 +251,33 @@ def test_escape_like_pattern(raw_string, escaped):
...
+@pytest.mark.parametrize(
+    "identifier, expected",
+    [
+        # Unquoted identifiers are upper-cased by unquote_identifier, then wrapped.
+        ("streamlit", "'STREAMLIT'"),
+        ("my_table", "'MY\\\\_TABLE'"),
+        # Quoted identifiers preserve case and special chars, with LIKE wildcards escaped.
+        ('"streamlit"', "'streamlit'"),
+        ('"my_table"', "'my\\\\_table'"),
+        ('"pct%table"', "'pct\\\\%table'"),
```
sfc-gh-jwilkowski, line 272: Worth adding an integration test (even a lightweight one) for the SHOW-LIKE injection path. Since the vulnerability depends on how Snowflake actually parses the SQL at runtime, a unit test can only verify the shape of the string — it can't confirm the server treats the doubled quote as data rather than a literal delimiter. A test that issues `SHOW STREAMLITS LIKE '...'` with an injected payload and asserts zero rows (or no error) would give stronger confidence than unit tests alone.

sfc-gh-jwilkowski (commented):
Two small test suggestions (no blocking issues):

tests/nativeapp/codegen/snowpark/test_extension_function_utils.py
sfc-gh-olorek: Added in 87fd0aa — `test_ensure_string_literal` now asserts idempotency on `"'abc''def'"`. If a future change drops the `''` alternation from `SINGLE_QUOTED_STRING_LITERAL_REGEX`, that case will fail loudly.

tests/project/test_util.py
sfc-gh-olorek: Added in 87fd0aa — `tests_integration/test_object.py::test_show_like_rejects_sql_injection_payload` runs `SHOW STREAMLITS LIKE <identifier_to_show_like_pattern(payload)>` against the live session for three payloads (the SNOW-3411343 `; GRANT ...` one, a bare single quote, and a `\\';DROP TABLE...` backslash variant). Asserts single-cursor + zero rows, so a regression that left the `''` as a delimiter would raise or leak a chained statement.

sfc-gh-jwilkowski (approved):

PR #2933, opened 2026-05-04
fix: support private_key_file_pwd in connections.toml (SNOW-3012806)

### LINKED ISSUE #2738: SNOW-3012806: Snowflake CLI does not use the `private_key_file_pwd` key in toml
### SnowCLI version

main

### Python version

Python 3.13.5

### Platform

Linux-6.14.0-37-generic-x86_64-with-glibc2.39

### What happened

The snowflake cli does not respect the `private_key_file_pwd` key in toml that is supported in the snowflake python connector 

https://github.com/snowflakedb/snowflake-connector-python/blob/fef4999534be5180aa0e601a2a4cc2a95d1c2c0e/src/snowflake/connector/connection.py#L236-L237

The snow cli implementation is here 

https://github.com/snowflakedb/snowflake-cli/blob/78174c60f5ef5ae8459e2e900c51d50f045ec549/src/snowflake/cli/_app/snow_connector.py#L351-L359.

### Console output

```Shell
NA
```

### How to reproduce

Create a snowflake `connections.toml` file with `private_key_file` and `private_key_file_pwd` set and not have `PRIVATE_KEY_PASSPHRASE` set.

sfc-gh-olorek:
### Pre-review checklist

### Changes description

Fixes https://github.com/snowflakedb/snowflake-cli/issues/2738 / SNOW-3012806.

`snowflake-connector-python` supports storing the private-key passphrase in `connections.toml` under `private_key_file_pwd`, but Snowflake CLI ignored that field and only read `PRIVATE_KEY_PASSPHRASE` from the environment. A user whose connector-python config already worked would hit:

> Encrypted private key, you must provide the passphrase in the environment variable PRIVATE_KEY_PASSPHRASE.

and be forced to also set an env var. This PR makes the CLI read the passphrase from `connections.toml` while keeping the env var as a precedence override so existing setups are not broken.

**What the fix does**

- **TOML alias** (`src/snowflake/cli/api/config.py`): `LEGACY_CONNECTION_SETTING_ALIASES` now maps `private_key_file_pwd` → `private_key_passphrase`, so entries in `connections.toml` populate the existing CLI-native field used throughout the codebase (`ConnectionContext.private_key_passphrase`).
- **Connector-override alias** (`src/snowflake/cli/_app/snow_connector.py`): `CONNECTION_KEY_ALIASES` gets the same mapping so `--private-key-file-pwd`-style overrides and env vars are normalized.
- **Env-var surface**: Both `private_key_file_pwd` and `private_key_passphrase` added to `SUPPORTED_ENV_OVERRIDES`, so `SNOWFLAKE_PRIVATE_KEY_FILE_PWD` and `SNOWFLAKE_PRIVATE_KEY_PASSPHRASE` work as expected.
- **Private-key loading** (`_load_pem_to_der`): accepts an optional `config_passphrase` parameter. Resolution order: `PRIVATE_KEY_PASSPHRASE` env var (wins for back-compat) → config passphrase → error.
- **`snow connection generate-jwt`**: same precedence logic (env var → config-sourced `connection_details.private_key_passphrase` → interactive prompt).
- **Leak guard**: `connection_parameters.pop(\"private_key_passphrase\", None)` after key loading so the passphrase does not leak into `snowflake.connector.connect(**kwargs)`.
- **Error message** updated to mention both the config key and the env var.
- **Masking**: no code change needed — `mask_sensitive_value` already masks any key containing `passphrase` or `pwd`.

**Tests added (`tests/test_connection.py`):**

- `test_key_pair_authentication_passphrase_from_config` — parametrized over `private_key_file_pwd` and `private_key_passphrase`; writes a real encrypted PKCS#8 key, loads via config, asserts the DER the connector receives matches what `load_pem_private_key` produces with the correct passphrase, and that neither alias leaks into connector kwargs.
- `test_generate_jwt_with_pass_phrase_from_config` — parametrized over both aliases; asserts `generate-jwt` passes the config-sourced passphrase to `get_token_from_private_key`.
- `test_generate_jwt_env_passphrase_takes_precedence_over_config` — sets both env var and config, asserts env wins.

Full `tests/test_connection.py` (88 tests) and `tests/test_config.py` (80 tests) pass locally.

### Related

- Issue: #2738
- Jira: SNOW-3012806

### DIFF
--- a/src/snowflake/cli/_app/snow_connector.py
+++ b/src/snowflake/cli/_app/snow_connector.py
@@ -68,6 +68,8 @@ SUPPORTED_ENV_OVERRIDES = [
     "workload_identity_provider",
     "private_key_file",
     "private_key_path",
+    "private_key_file_pwd",
+    "private_key_passphrase",
     "private_key_raw",
     "database",
     "schema",
@@ -91,7 +93,12 @@ SUPPORTED_ENV_OVERRIDES = [
 ]
 
 # mapping of found key -> key to set
-CONNECTION_KEY_ALIASES = {"private_key_path": "private_key_file"}
+CONNECTION_KEY_ALIASES = {
+    "private_key_path": "private_key_file",
+    # `private_key_file_pwd` is the name used by the snowflake-connector-python
+    # TOML config; we normalize it to the CLI-native `private_key_passphrase`.
+    "private_key_file_pwd": "private_key_passphrase",
+}
 
 
 class _BufferedMirrorStream(io.StringIO):
@@ -308,9 +315,13 @@ def _load_private_key(connection_parameters: Dict, private_key_var_name: str) ->
         private_key_pem = _load_pem_from_file(
             connection_parameters[private_key_var_name]
         )
-        private_key = _load_pem_to_der(private_key_pem)
+        private_key = _load_pem_to_der(
+            private_key_pem,
+            config_passphrase=connection_parameters.get("private_key_passphrase"),
+        )
         connection_parameters["private_key"] = private_key.value
         del connection_parameters[private_key_var_name]
+        connection_parameters.pop("private_key_passphrase", None)
     else:
         raise CliError(
             "Private Key authentication requires authenticator set to SNOWFLAKE_JWT"
@@ -324,9 +335,13 @@ def _load_private_key_from_parameters(
         private_key_pem = _load_pem_from_parameters(
             connection_parameters[private_key_var_name]
         )
-        private_key = _load_pem_to_der(private_key_pem)
+        private_key = _load_pem_to_der(
+            private_key_pem,
+            config_passphrase=connection_parameters.get("private_key_passphrase"),
+        )
         connection_parameters["private_key"] = private_key.value
         del connection_parameters[private_key_var_name]
+        connection_parameters.pop("private_key_passphrase", None)
     else:
         raise CliError(
             "Private Key authentication requires authenticator set to SNOWFLAKE_JWT"
@@ -363,8 +378,9 @@ def _load_pem_from_parameters(private_key_raw: str) -> SecretType:
 def _validate_passphrase(passphrase: SecretType) -> None:
     if passphrase.value is None:
         raise CliError(
-            "Encrypted private key, you must provide the "
-            "passphrase in the environment variable PRIVATE_KEY_PASSPHRASE."
+            "Encrypted private key, you must provide the passphrase via the "
+            "`private_key_file_pwd` key in your connection config or the "
+            "`PRIVATE_KEY_PASSPHRASE` environment variable."
         )
     if passphrase.value == "":
         raise CliError(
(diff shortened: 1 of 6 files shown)


src/snowflake/cli/_app/snow_connector.py
```diff
@@ -308,9 +315,13 @@ def _load_private_key(connection_parameters: Dict, private_key_var_name: str) ->
         private_key_pem = _load_pem_from_file(
             connection_parameters[private_key_var_name]
         )
-        private_key = _load_pem_to_der(private_key_pem)
+        private_key = _load_pem_to_der(
+            private_key_pem,
+            config_passphrase=connection_parameters.get("private_key_passphrase"),
```
sfc-gh-jwilkowski: `config_passphrase` feels confusing to me - it's a passphrase for your private key file, not the config

src/snowflake/cli/_app/snow_connector.py
```diff
@@ -373,12 +389,20 @@ def _validate_passphrase(passphrase: SecretType) -> None:
         )
 
 
-def _load_pem_to_der(private_key_pem: SecretType) -> SecretType:
+def _load_pem_to_der(
+    private_key_pem: SecretType, config_passphrase: Optional[str] = None
```
sfc-gh-jwilkowski: as above, `config_passphrase` is imo confusing

src/snowflake/cli/_app/snow_connector.py
```diff
@@ -68,6 +68,8 @@
     "workload_identity_provider",
     "private_key_file",
     "private_key_path",
+    "private_key_file_pwd",
+    "private_key_passphrase",
     "private_key_raw",
```
sfc-gh-jwilkowski, line 71: I'd suggest removing both `"private_key_file_pwd"` and `"private_key_passphrase"` from `SUPPORTED_ENV_OVERRIDES`.

The env-override loop resolves every key through `_resolve_alias`, so both entries map to the same `connection_key = "private_key_passphrase"`. Because `private_key_file_pwd` appears first in the list, if `SNOWFLAKE_PRIVATE_KEY_FILE_PWD` is set it populates `connection_parameters["private_key_passphrase"]`, and then the `if connection_key not in connection_parameters` guard silently skips `SNOWFLAKE_PRIVATE_KEY_PASSPHRASE` — so the alias beats the canonical name. Wrong precedence.

More fundamentally though: `PRIVATE_KEY_PASSPHRASE` (the non-prefixed env var read directly in `_load_pem_to_der`) already covers the env-var use case and is the documented override. Adding `SNOWFLAKE_PRIVATE_KEY_PASSPHRASE` via `SUPPORTED_ENV_OVERRIDES` creates a second env var that does the same thing but with lower and order-dependent precedence — it can never actually win over `PRIVATE_KEY_PASSPHRASE`. `SNOWFLAKE_PRIVATE_KEY_FILE_PWD` would be a third.

The TOML aliases in `LEGACY_CONNECTION_SETTING_ALIASES` and `CONNECTION_KEY_ALIASES` are the right place to handle the `private_key_file_pwd` ↔ `private_key_passphrase` normalization. No new env var surface is needed.

src/snowflake/cli/_app/snow_connector.py
sfc-gh-olorek: Renamed to `passphrase` in a003e0b4 (both call sites and the definition).

src/snowflake/cli/_app/snow_connector.py
sfc-gh-olorek: Renamed to `passphrase` in a003e0b4.

src/snowflake/cli/_app/snow_connector.py
sfc-gh-olorek: Agreed — removed both entries from `SUPPORTED_ENV_OVERRIDES` in a003e0b4. The TOML aliases in `LEGACY_CONNECTION_SETTING_ALIASES` / `CONNECTION_KEY_ALIASES` stay as the single normalization point; `PRIVATE_KEY_PASSPHRASE` remains the only env override (read directly in `_load_pem_to_der`).

sfc-gh-jwilkowski (approved):

PR #2940, opened 2026-05-05
fix: [SNOW-3417058] guard `snow snowpark build` against zip-slip in wheels

sfc-gh-olorek:
### Pre-review checklist

### Changes description

**Ticket:** [SNOW-3417058](https://snowflakecomputing.atlassian.net/browse/SNOW-3417058) — *HIGH — Zip Slip Vulnerability in Wheel File Extraction*

#### The bug

`snow snowpark build` downloads dependency wheels with `pip wheel` and then extracts each `.whl` into the build directory via `RequirementWithWheel.extract_files`:

```python
def extract_files(self, destination: Path) -> None:
    if self.wheel_path is not None:
        zipfile.ZipFile(self.wheel_path).extractall(destination)
```

`zipfile.extractall` on Python 3.10 and 3.11 performs **no** path validation, and the hardening filter introduced in Python 3.12 (`extractall(filter=...)`) is **opt-in**. A malicious or corrupt wheel containing entries like `../../etc/something` or an absolute path would be written wherever the process has write permission — the classic **zip-slip** pattern (CWE-22).

#### The fix

Before calling `extractall`, walk `ZipFile.infolist()` and for each member verify that `(destination / member.filename).resolve()` is contained within the resolved destination. If any entry escapes, raise `UnsafeWheelEntryError` and abort — *before* writing any file, so a wheel that mixes safe and unsafe entries leaves the destination clean.

Containment is checked with `Path.resolve(strict=False).relative_to(destination)`, which correctly handles `..` segments and absolute paths regardless of platform.

#### Tests

New unit tests in `tests/snowpark/test_models.py` that build crafted wheels with `zipfile.ZipFile(..., 'w', ZIP_STORED)` and `writestr` (so traversal sequences are preserved verbatim):

* `test_extract_files_is_a_noop_when_wheel_path_is_none`
* `test_extract_files_extracts_safe_wheel` — legitimate wheels still extract
* `test_extract_files_rejects_zip_slip_relative_paths` — parametrized over `../evil.py`, `../../evil.py`, `subdir/../../evil.py`
* `test_extract_files_rejects_absolute_path_entry`
* `test_extract_files_aborts_before_writing_any_file` — safe + unsafe mix leaves destination empty

All 21 tests in `tests/snowpark/test_models.py` pass locally.

[SNOW-3417058]: https://snowflakecomputing.atlassian.net/browse/SNOW-3417058?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -23,6 +23,7 @@
 ## Fixes and improvements
 * Fixed `SELECT *` output being corrupted when joined tables share column names. Duplicate column names are now disambiguated by appending a numeric suffix (e.g. `NAME`, `NAME_2`).
 * Fixed `snow connection generate-jwt` and `snow connection generate-workload-identity-token` failing with `Connection None is not configured` when used with `--temporary-connection`.
+* Hardened wheel extraction during `snow snowpark build` against zip-slip: wheel archives whose entries resolve outside the extraction directory are now rejected with a clear error instead of writing files to arbitrary paths on disk.
 
 
 # v3.17.0
--- a/src/snowflake/cli/_plugins/snowpark/models.py
+++ b/src/snowflake/cli/_plugins/snowpark/models.py
@@ -20,7 +20,29 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import List
 
+from click.exceptions import ClickException
 from requirements import requirement
+
+
+class UnsafeWheelEntryError(ClickException):
+    """Raised when a wheel archive contains an entry that would extract
+    outside of the extraction destination (a zip-slip attack)."""
+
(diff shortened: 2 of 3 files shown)


sfc-gh-olorek:
Closing without merging.  CPython's zipfile._extract_member already strips .. and absolute paths on all supported versions (3.10+), so the cited ../../../../home/victim/.bashrc payload doesn't actually escape the destination. Verified on Py 3.10/3.11/3.12/3.13. The PR's "rejects relative paths" tests pass with or without the patch for the same reason.

PR #2944, opened 2026-05-05
Add strict/immutable options for Snowpark functions and procedures (#2226)

### LINKED ISSUE #2226: Snowpark Project definition: support marking UDFs as "STRICT" and "IMMUTABLE"
### Description

It would be useful if we could specify these options in the Snowpark YAML project definition. These options are already supported in the Snowpark Python API.

Hypothetical example:

```yaml
entities:

  hello_function:
    type: function
    identifier:
      name: hello_function
    handler: functions.hello_function
    signature:
      - name: name
        type: string
    returns: string
    strict: true
    immutable: true
    meta:
      use_mixins:
        - snowpark_shared
```

### Context

`RETURNS NULL ON NULL INPUT` / `STRICT` makes UDF handler development easier by allowing the author to ignore null inputs. `IMMUTABLE` can be important for optimization in certain situations.

sfc-gh-olorek:
### Pre-review checklist

### Changes description

Closes #2226.

Adds two new optional boolean fields to Snowpark `function` and `procedure` entities in `snowflake.yml`:

* `strict` — emits `RETURNS NULL ON NULL INPUT` in the generated `CREATE FUNCTION`/`CREATE PROCEDURE` statement, so the handler is skipped when any input argument is NULL.
* `immutable` — emits `IMMUTABLE`, letting Snowflake reuse results for identical inputs. Only safe when the handler has no side effects.

Both default to `false`, so existing projects deploy unchanged.

Example:

```yaml
entities:
  my_udf:
    type: function
    handler: app.my_udf
    returns: string
    signature: "(name string)"
    stage: dev
    artifacts: [src]
    strict: true
    immutable: true
```

#### Scope
- Added the two fields to `SnowparkEntityModel` (v2) and `_CallableBase` (v1).
- Forwarded them through the v1→v2 conversion layer so users on either schema version can opt in.
- Updated both SQL-emitting code paths:
  - `SnowparkObjectManager.create_or_replace` in `_plugins/snowpark/common.py` (used by `snow snowpark deploy`).
  - `SnowparkEntity.get_deploy_sql` in `_plugins/snowpark/snowpark_entity.py` (used by the workspace/native-app-children flow).
- Extended `_check_if_replace_is_required` to compare the project-declared values against `null handling` / `volatility` in `DESCRIBE FUNCTION/PROCEDURE` output, so a change in either field triggers a replace instead of silently no-op'ing.
- Added parametrized test cases for the new replace-detection branches and refreshed the affected v1 schema snapshot.
- Mentioned the feature under **New additions** in `RELEASE-NOTES.md`.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -19,6 +19,7 @@
 ## Deprecations
 
 ## New additions
+* Added `strict` and `immutable` options for Snowpark `function` and `procedure` entities in `snowflake.yml`. When set, `snow snowpark deploy` emits `RETURNS NULL ON NULL INPUT` and/or `IMMUTABLE` in the generated `CREATE FUNCTION`/`CREATE PROCEDURE` statement and treats a change in either value as a reason to replace the object.
 
 ## Fixes and improvements
 * Fixed `SELECT *` output being corrupted when joined tables share column names. Duplicate column names are now disambiguated by appending a numeric suffix (e.g. `NAME`, `NAME_2`).
--- a/src/snowflake/cli/_plugins/snowpark/common.py
+++ b/src/snowflake/cli/_plugins/snowpark/common.py
@@ -123,6 +123,12 @@ class SnowparkObjectManager(SqlExecutionMixin):
             )
             query.append(f"RESOURCE_CONSTRAINT=({constraints})")
 
+        if entity.strict:
+            query.append("returns null on null input")
+
+        if entity.immutable:
+            query.append("immutable")
+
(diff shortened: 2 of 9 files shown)


sfc-gh-olorek:
Closing this stale draft. Internal development moved to the private repository in July 2026 and PR head branches are not mirrored outbound, so this draft cannot land here — anything still wanted has to be re-opened there. Nothing is tracked against this beyond the draft itself; closing rather than leaving it open indefinitely. The branch is preserved.

PR #2946, opened 2026-05-05
fix: [SNOW-3417002] escape $$ in SPCS service spec to prevent SQL injection

sfc-gh-olorek:
### Pre-review checklist

### Changes description

Resolves [SNOW-3417002](https://snowflakecomputing.atlassian.net/browse/SNOW-3417002) — SQL injection via dollar-quote breakout in SPCS service spec.

#### The bug

`ServiceManager._read_yaml()` (`src/snowflake/cli/_plugins/spcs/services/manager.py:413`) reads a YAML spec with `yaml.safe_load` and returns `json.dumps(data)`. That string is then interpolated into `$$...$$` dollar-quoted SQL literals by three call sites:

- `create()` — `FROM SPECIFICATION $$ {spec} $$` (manager.py:75-80)
- `_execute_job_service()` — `FROM SPECIFICATION $$ {spec_json} $$` (manager.py:248-253)
- `upgrade_spec()` — `from specification $$ {spec} $$` (manager.py:691-693)

`json.dumps` does not escape `$` (it has no meaning in JSON). A spec string value containing `$$` — e.g.

```yaml
spec:
  containers:
    - env:
        PAYLOAD: "$$ ); GRANT ROLE ACCOUNTADMIN TO USER attacker; --"
```

survives serialization verbatim. Snowflake's SQL parser treats the inner `$$` as the closing delimiter of the dollar-quoted literal, so everything after it is parsed as SQL. `execute_query` routes through `_execute_string` → `execute_stream` (`sql_execution.py:85`), which splits on `;` and executes each statement independently, so the injected `GRANT` runs with the caller's session privileges.

Vulnerable commands: `snow spcs service create`, `snow spcs service execute-job`, `snow spcs service upgrade`. `snow spcs service deploy` is not affected because it passes `SPECIFICATION_FILE = '<stage path>'` rather than inlining spec contents.

#### The fix

Replace `$$` with `$ $` in the serialized spec before embedding. The substitution only changes the JSON-level bytes being inlined in SQL — `$$` has no meaning in JSON or in the spec schema, so the parsed spec values Snowflake receives are unchanged for any legitimate input. Extracted to `ServiceManager._serialize_spec` so `build_image()` (which also hands a JSON spec to `_execute_job_service`) gets the same treatment.

#### Tests

- `test_read_yaml_escapes_dollar_quote_breakout` — asserts a spec value containing `$$` is neutralized in `_read_yaml`'s output.
- `test_service_spec_dollar_quote_not_broken_out` — parametrized over `create` / `execute_job` / `upgrade_spec`, asserts the final SQL contains exactly two `$$` tokens (the opening and closing delimiters) and that the payload is retained inside the literal rather than executed.
- Existing `test_read_yaml`, `test_create_service`, `test_create_service_if_not_exists`, `test_upgrade_spec`, `test_execute_job_service`, `test_build_image` continue to pass unchanged — the existing fixture values contain no `$`, so the escape is a no-op for them.

Full `tests/spcs/` suite passes locally (177 tests).

#### References

- CWE-89 (SQL Injection), OWASP A03:2021 (Injection)
- Sinks: `manager.py:79`, `manager.py:252`, `manager.py:693`
- Source: `manager.py:413` (`_read_yaml`)

[SNOW-3417002]: https://snowflakecomputing.atlassian.net/browse/SNOW-3417002?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -43,6 +43,9 @@
 
 ## Fixes and improvements
 * Encrypted private key files no longer require `PRIVATE_KEY_PASSPHRASE` to be set in the environment. The passphrase can now be read from `private_key_file_pwd` (the name used by `snowflake-connector-python`) or `private_key_passphrase` in `connections.toml` / `config.toml`. The `PRIVATE_KEY_PASSPHRASE` environment variable continues to take precedence when set. This also fixes a regression in 3.17.0 where commands using key-pair authentication with `private_key_passphrase` in `connections.toml` failed with `argument 'password': Cannot convert "<class 'str'>" instance to a buffer`.
+* Fixed `SELECT *` output being corrupted when joined tables share column names. Duplicate column names are now disambiguated by appending a numeric suffix (e.g. `NAME`, `NAME_2`).
+* Fixed `snow connection generate-jwt` and `snow connection generate-workload-identity-token` failing with `Connection None is not configured` when used with `--temporary-connection`.
+* Fixed SQL injection in `snow spcs service create`, `execute-job`, and `upgrade` where a `$$` sequence in a YAML spec file could break out of the dollar-quoted SQL literal and execute arbitrary SQL with the caller's session privileges. `$$` sequences in spec content are now neutralized before the spec is embedded in SQL.
 
 
 # v3.17.0
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -32,6 +32,7 @@
 * Fixed SQL injection via `FQN.sql_identifier`.
 * Fixed Snowsight URL generation (used by `snow streamlit deploy`, `snow streamlit get-url`, `snow app run`, `snow notebook`, and similar commands) for accounts whose host is 4-part (e.g. `<account>.us-east-1.snowflakecomputing.com`) or 5-part with a cloud suffix (e.g. `<account>.<region>.aws.snowflakecomputing.com`). These hosts now resolve to the correct regioned Snowsight URL instead of raising `"host (...) was missing or not in the expected format"`.
 * Fixed boolean connection parameters (`client_store_temporary_credential`, `oauth_disable_pkce`, `oauth_enable_refresh_tokens`, `oauth_enable_single_use_refresh_tokens`) being passed to the connector as raw strings when supplied via `SNOWFLAKE_*` or `SNOWFLAKE_CONNECTIONS_<name>_*` environment variables. Values like `false` / `0` are now correctly interpreted as `False` rather than truthy strings.
+* Fixed SQL injection in `snow spcs service create`, `execute-job`, and `upgrade` where a `$$` sequence in a YAML spec file could break out of the dollar-quoted SQL literal and execute arbitrary SQL with the caller's session privileges. `$$` sequences in spec content are now neutralized before the spec is embedded in SQL.
 
 
 # v3.17.1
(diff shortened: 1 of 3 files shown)


src/snowflake/cli/_plugins/spcs/services/manager.py
```diff
@@ -414,7 +414,14 @@ def _read_yaml(self, path: Path) -> str:
...
             data = yaml.safe_load(fh)
-        return json.dumps(data)
+        return self._serialize_spec(data)
+
+    @staticmethod
+    def _serialize_spec(data) -> str:
+        # The returned JSON is embedded inside $$...$$ dollar-quoted SQL
+        # literals; json.dumps does not escape $, so a spec value containing
+        # $$ would close the literal early and allow SQL injection.
+        return json.dumps(data).replace("$$", "$ $")
```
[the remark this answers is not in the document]
sfc-gh-olorek: Good catch — you're right that `.replace("$$", "$ $")` is vulnerable on odd-length runs (`$$$` → `$ $$`, still breaks out).

Fixed in 5c7d60e3 with a lookahead regex rather than a blanket `.replace("$", "$ ")`:

```python
return re.sub(r"\$(?=\$)", "$ ", json.dumps(data))
```

This breaks up every adjacent `$$` pair, handling runs of any length, while preserving legitimate single-`$` values like `$HOME` or `p$w0rd` (which escaping every `$` would mangle). Tests now parametrize over `$$`, `$$$`, `$$$$`, `$$$$$`, plus a single-`$` preserve-unchanged test.

RELEASE-NOTES.md
```diff
@@ -41,6 +41,9 @@
...
+* Fixed `SELECT *` output being corrupted when joined tables share column names. Duplicate column names are now disambiguated by appending a numeric suffix (e.g. `NAME`, `NAME_2`).
+* Fixed `snow connection generate-jwt` and `snow connection generate-workload-identity-token` failing with `Connection None is not configured` when used with `--temporary-connection`.
+* Fixed SQL injection in `snow spcs service create`, `execute-job`, and `upgrade` where a `$$` sequence in a YAML spec file could break out of the dollar-quoted SQL literal and execute arbitrary SQL with the caller's session privileges. `$$` sequences in spec content are now neutralized before the spec is embedded in SQL.
```
sfc-gh-jwilkowski: misplaced release notes. First 2 bullet points were released with 3.17.0 afair

sfc-gh-olorek: Good catch — fixed in e6f01540. The two bullets you flagged (SELECT \* column dedup, generate-jwt with --temporary-connection) are already under v3.17.0 in main, so I dropped them from v3.17.1. The SPCS SQL injection bullet moves up under "Unreleased version" since that's where this PR's fix actually lands.

src/snowflake/cli/_plugins/spcs/services/manager.py
```diff
@@ -414,7 +415,17 @@ def _read_yaml(self, path: Path) -> str:
...
+
+    @staticmethod
+    def _serialize_spec(data) -> str:
+        # The returned JSON is embedded inside $$...$$ dollar-quoted SQL
+        # literals; json.dumps does not escape $, so any run of two or more
+        # $ in spec content would close the literal early and allow SQL
+        # injection. Break up every adjacent $$ pair — this handles runs of
+        # any length ($$, $$$, $$$$, ...) without mutating legitimate
+        # single-$ values.
+        return re.sub(r"\$(?=\$)", "$ ", json.dumps(data))
```
sfc-gh-jwilkowski: Can you give test this approach instead?
```
return re.sub(r"\$(?=\$)", r"\\u0024", json.dumps(data))
```

sfc-gh-jwilkowski (changes requested):

src/snowflake/cli/_plugins/spcs/services/manager.py
sfc-gh-olorek: Done in ff1b510e — switched to `re.sub(r"\$(?=\$)", r"\\u0024", json.dumps(data))`.

Tested both forms end-to-end: same security guarantee (no `$$` survives in the SQL bytes), and the `\u0024` form is strictly better because it preserves spec values — e.g. `"$$100"` now reaches Snowflake unchanged instead of being silently corrupted to `"$ $100"`. Added round-trip assertions to lock that in. Full `tests/spcs/` suite (197) passes.

sfc-gh-jwilkowski (approved):

PR #2948, opened 2026-05-05
feat: [SNOW-2067317] drop removed snowpark entities on `deploy --prune`

### LINKED ISSUE #2218: snow snowpark deploy --prune should drop functions and procedures
### Description

`snow snowpark deploy --prune` should drop functions and procedures that were removed from the snowflake.yml.

### Context

These functions and procedures will not work anymore, since `--prune` removed the required code. Currently the user has to clean these up manually.

sfc-gh-olorek:
### Pre-review checklist

### Changes description

Closes #2218.

`snow snowpark deploy --prune` previously only cleared stage contents — procedures and functions that had been removed from `snowflake.yml` lingered in Snowflake until dropped manually. This PR extends `--prune` so the server-side catalog matches the declared entities too.

**Behavior**

After artifacts are uploaded and before entities are (re)created, the deploy flow now:

1. Groups declared Snowpark entities by `(type, database, schema)` using the identifier already resolved in the project definition.
2. For each group, runs `SHOW USER <FUNCTIONS|PROCEDURES> IN SCHEMA <db>.<schema>` and parses the returned `arguments` column.
3. Drops every object in that schema of the same type that does not match any declared entity by `(name, arg_types)`. Overloads are preserved — matching is done per signature, not per name.

**Safety**

- Only schemas that contain at least one declared entity of the same type are scanned — schemas the project does not touch are never queried.
- Builtin functions (`is_builtin = Y`) are skipped.
- Type aliases are handled via the existing `same_type()` helper so `STRING` vs `VARCHAR`, `INT` vs `NUMBER(38,0)`, etc. do not cause spurious drops.
- Nested parens like `NUMBER(38,0)` are parsed correctly (manual depth tracking, not a naïve regex).
- Original-case database/schema from the declared identifier is preserved in the `DROP` statement.
- `DROP ... IF EXISTS` is used so a concurrent drop does not fail the deploy.
- No behavior change when `--prune` is not passed.

**Tests**

- Unit tests in `tests/snowpark/test_common.py`:
  - `test_parse_arg_types_from_show` — parametrized coverage of arg-type parsing including `NUMBER(38,0)` and no-argument cases.
  - `test_drop_removed_snowpark_entities_drops_unlisted_objects` — stale entities get dropped, builtins are skipped, signatures are included in the drop FQN.
  - `test_drop_removed_snowpark_entities_keeps_matching_overloads` — only the non-declared overload is dropped.
  - `test_drop_removed_snowpark_entities_does_not_touch_other_schemas` — `SHOW` scope is limited to declared schemas.
- Integration test in `tests/snowpark/test_procedure.py`:
  - `test_deploy_procedure_prune_drops_removed_entities` — end-to-end `snow snowpark deploy --prune` flow with mocked `ObjectManager.show/describe/drop`, verifies case preservation and signature passing.

### DIFF
--- a/src/snowflake/cli/_plugins/snowpark/commands.py
+++ b/src/snowflake/cli/_plugins/snowpark/commands.py
@@ -16,7 +16,7 @@ from __future__ import annotations
 
 import logging
 from collections import defaultdict
-from typing import Dict, Optional, Set, Tuple
+from typing import Dict, Iterable, List, Optional, Set, Tuple
 
 import typer
 from click import ClickException, UsageError
@@ -42,6 +42,7 @@ from snowflake.cli._plugins.snowpark.common import (
     SnowparkObjectManager,
     StageToArtifactMapping,
     map_path_mapping_to_artifact,
+    same_type,
     zip_and_copy_artifacts_to_deploy,
 )
 from snowflake.cli._plugins.snowpark.package.anaconda_packages import (
@@ -137,7 +138,8 @@ def deploy(
     ),
     force_replace: bool = ForceReplaceOption(),
     prune: bool = PruneOption(
-        help="Remove contents of the stage before uploading artifacts."
+        help="Remove contents of the stage before uploading artifacts, and drop procedures "
+        "and functions that exist in the target schemas but are not declared in the project."
     ),
     **options,
 ) -> CommandResult:
@@ -181,6 +183,10 @@ def deploy(
         )
 
         create_stages_and_upload_artifacts(stages_to_artifact_map, prune=prune)
+
+    if prune:
+        with cli_console.phase("Pruning removed procedures and functions"):
+            drop_removed_snowpark_entities(om, snowpark_entities)
 
     # Create snowpark entities
     with cli_console.phase("Creating Snowpark entities"):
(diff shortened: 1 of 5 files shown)


src/snowflake/cli/_plugins/snowpark/commands.py
```diff
@@ -275,6 +281,177 @@ def create_stages_and_upload_artifacts(
...
+    #        list of declared (upper-name, [type,...]) tuples).
+    # Database/schema cases are preserved so DROP statements use the same
+    # spelling the user declared. The upper-cased keys only group duplicates.
+    declared: Dict[
+        Tuple[str, Tuple[Optional[str], Optional[str]]],
+        Tuple[Optional[str], Optional[str], List[Tuple[str, List[str]]]],
+    ] = {}
+    for entity in snowpark_entities.values():
+        fqn = entity.fqn
+        if not fqn.database or not fqn.schema:
```
sfc-gh-kolszewski, line 306: Is this `if` block actually reachable in a way that produces a different result? My reading is that `entity.udf_sproc_identifier.identifier_with_arg_types` builds the identifier from the same declared `identifier` field that `entity.fqn` already reflects — so if `entity.fqn.database` is `None`, the fallback FQN parsed from `identifier_with_arg_types` would also have `database=None`. In that case `_list_schema_snowpark_objects` would return `[]` either way, and the branch is a no-op. Is there a setup where `udf_sproc_identifier` resolves a db/schema that `entity.fqn` does not? If so, a brief comment explaining that scenario would help a lot.

src/snowflake/cli/_plugins/snowpark/commands.py
```diff
@@ -275,6 +281,177 @@ def create_stages_and_upload_artifacts(
...
+        existing = _list_schema_snowpark_objects(om, object_type, database, schema)
+        for existing_name, existing_arg_types, existing_fqn in existing:
+            if _matches_any(existing_name, existing_arg_types, declared_entries):
+                continue
+            cli_console.step(f"Dropping {object_type} {existing_fqn.identifier}")
+            om.drop(object_type=object_type, fqn=existing_fqn, if_exists=True)
+
+
+def _entity_arg_types(entity) -> List[str]:
+    if not entity.signature or entity.signature == "null":
```
sfc-gh-kolszewski, line 332: The `== "null"` guard looks like it's defending against a pydantic/YAML serialisation quirk where the field comes through as the literal string `"null"` rather than `None`. Could you add a short inline comment explaining when `entity.signature` can be the string `"null"` rather than a falsy value? Without it, future readers will likely remove this as an obvious dead check.

src/snowflake/cli/_plugins/snowpark/commands.py
```diff
@@ -275,6 +281,177 @@ def create_stages_and_upload_artifacts(
...
+            schema=schema,
+            name=name,
+            signature=f"({', '.join(arg_types)})",
+        )
+        results.append((name.upper(), arg_types, fqn))
+    return results
+
+
+def _is_builtin(row: dict) -> bool:
+    for key in ("is_builtin", "is_builtin_"):
```
sfc-gh-kolszewski, line 391: Where does Snowflake actually return a column named `"is_builtin_"` (with a trailing underscore)? I searched the codebase and there's no prior reference to that key — including the test fixtures in this PR, which only use `"is_builtin"`. If this is defensive coverage for a known server variant (a specific region, connector version, or SHOW variant), a comment pointing to that would be helpful. If it's speculative, it might be cleaner to remove it and add it back if the variant is ever observed.

src/snowflake/cli/_plugins/snowpark/commands.py
```diff
@@ -275,6 +281,177 @@ def create_stages_and_upload_artifacts(
...
+    ) in declared.items():
+        existing = _list_schema_snowpark_objects(om, object_type, database, schema)
+        for existing_name, existing_arg_types, existing_fqn in existing:
+            if _matches_any(existing_name, existing_arg_types, declared_entries):
+                continue
+            cli_console.step(f"Dropping {object_type} {existing_fqn.identifier}")
+            om.drop(object_type=object_type, fqn=existing_fqn, if_exists=True)
+
+
+def _entity_arg_types(entity) -> List[str]:
```
sfc-gh-kolszewski, line 331: Minor: `entity` is untyped — could be annotated as `SnowparkEntityModel` (or whichever union covers both procedures and functions) to match the style of the rest of the file.

src/snowflake/cli/_plugins/snowpark/commands.py
```diff
@@ -275,6 +281,177 @@ def create_stages_and_upload_artifacts(
...
+    for (object_type, _upper_key), (
+        database,
+        schema,
+        declared_entries,
+    ) in declared.items():
+        existing = _list_schema_snowpark_objects(om, object_type, database, schema)
+        for existing_name, existing_arg_types, existing_fqn in existing:
+            if _matches_any(existing_name, existing_arg_types, declared_entries):
+                continue
+            cli_console.step(f"Dropping {object_type} {existing_fqn.identifier}")
```
sfc-gh-kolszewski, line 327: When there are no stale entities the prune phase starts but prints nothing. A user watching the output might wonder whether the step ran or was silently skipped. Something like `cli_console.step("No stale entities found.")` when the drop loop exits without any calls (or skipping the phase entirely in that case) would make the output cleaner.

tests/snowpark/test_procedure.py
```diff
@@ -788,6 +788,77 @@ def test_deploy_procedure_with_empty_default_value(
...
+    enable_snowpark_glob_support_feature_flag,
+):
+    mock_download.return_value = DownloadUnavailablePackagesResult()
+    mock_om_describe.side_effect = ProgrammingError(
+        errno=DOES_NOT_EXIST_OR_NOT_AUTHORIZED
+    )
+    # ObjectManager.show() is called for integrations, then for procedures.
+    # snowpark_procedures declares procedureName(name string) and test().
+    # Stale procedure stale_proc(int) should be dropped.
+    mock_om_show.side_effect = [
```
sfc-gh-kolszewski, line 821: The first `[]` in the `side_effect` list is the integrations SHOW call that fires before `drop_removed_snowpark_entities`. This works today, but if the order of SHOW calls changes or a new one is added between integrations and the prune step, the test will silently feed wrong mock data to the wrong call and may still pass. Dispatching on `object_type` would make the intent explicit and the test resilient:

```python
def _show(object_type, **kw):
    if object_type == "procedure":
        return [{...}, ...]
    return []
mock_om_show.side_effect = _show
```

sfc-gh-olorek:
Closing this stale draft. Internal development moved to the private repository in July 2026 and PR head branches are not mirrored outbound, so this draft cannot land here — anything still wanted has to be re-opened there. The work itself is still tracked in SNOW-2067317, which stays the source of truth. The branch is preserved.

Issue #2037, opened 2025-01-31
Add project definition file schemas


sfc-gh-vtimofeenko:
### Description

[JSON Schema Store project](https://www.schemastore.org/json/) provides the format for declaring configuration file schemas. These schemas can be plugged into [editors](https://www.schemastore.org/json/#editors) to provide completion and linting.

[Project definition files](https://docs.snowflake.com/en/developer-guide/snowflake-cli/project-definitions/about) currently:
1. Require having the online documentation open while working on them
2. May be syntactically correct and pass initial validation but Snowflake will refuse to create the object (example below)

<details>
<summary>Sample failing project definition file</summary>

```yaml
definition_version: '2'

mixins:
  snowpark_shared:
    artifacts:
      - dest: my_snowpark_project
        src: app/
    stage: dev_deployment

entities:

  hello_function:
    type: function
    identifier:
      name: hello_function
    handler: functions.hello_function
    signature:
      - name: name
        type: strring  # <--- typo here
    returns: string
    meta:
      use_mixins:
        - snowpark_shared

```

```shell
❯ snow snowpark build && snow snowpark deploy          
Resolving dependencies from requirements.txt
  No external dependencies.
Preparing artifacts for source code
  Creating: app.zip
Build done.
Performing initial validation
Checking remote state
Preparing required stages and artifacts
  Creating (if not exists) stage: dev_deployment
  Uploading app.zip to @DROPME.PUBLIC.dev_deployment/my_snowpark_project/
Creating Snowpark entities
  Creating function hello_function
╭─ Error ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ 002040 (42601): 01ba115d-0807-3c0f-0057-53030711148a: SQL compilation error:                                                                                │
│ Unsupported data type 'STRRING'.                                                                                                                            │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

</details>

Use cases:

1. Developers using supported editors could iterate on project definition files faster and with lower amount of errors
2. CI/CD pipelines could lint the project definition files early and stop the pipeline if the validation fails

PR #2950, opened 2026-05-05
feat: [SNOW-1904824] add `snow helpers generate-project-schema`

sfc-gh-olorek:
### Pre-review checklist
   * [x] I've confirmed that my changes are working by executing CLI's commands manually on Linux.

### Changes description

Closes [SNOW-1904824](https://snowflakecomputing.atlassian.net/browse/SNOW-1904824) / #2037.

Adds a new `snow helpers generate-project-schema` command that emits a [JSON Schema (Draft 2020-12)](https://json-schema.org/draft/2020-12/schema) for the `snowflake.yml` project definition file. The schema is generated directly from the pydantic `DefinitionV10` / `DefinitionV11` / `DefinitionV20` models, so it always matches what the CLI itself validates against.

**Why:** Today, `snowflake.yml` authors:
1. Need to keep the online documentation open while editing, and
2. Only find out about typos (e.g. `strring` instead of `string`) once Snowflake rejects the object server-side.

With a published JSON Schema, editors that support the [JSON Schema Store project](https://www.schemastore.org/json/) (VS Code via the YAML extension, IntelliJ, etc.) and CI pipelines can auto-complete valid keys and flag typos long before a deploy.

**Usage:**

```sh
# Emit the v2 schema to stdout (default)
snow helpers generate-project-schema

# Target a specific definition version
snow helpers generate-project-schema --version 1.1

# Write the schema to a file (for checking into a repo / referencing from editors)
snow helpers generate-project-schema --output-file schema.json
snow helpers generate-project-schema -o schema.json
```

The output includes `$schema`, `$id`, `title`, and `description` fields so it can be consumed directly by JSON Schema tooling without further post-processing.

**Scope:**
* `requires_connection=False` — runs fully offline.
* Supports the same three definition versions the CLI itself supports (`1`, `1.1`, `2`); an unsupported version raises a clear `ClickException`.
* Unit tests cover default-version behaviour, all supported versions, `--output-file`, `-o` short option, the error path, and JSON stability (same input → byte-identical output). An optional test validates a minimal v2 project against the generated schema when `jsonschema` is available.

[SNOW-1904824]: https://snowflakecomputing.atlassian.net/browse/SNOW-1904824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/snowflake/cli/_plugins/helpers/commands.py
+++ b/src/snowflake/cli/_plugins/helpers/commands.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 import json
 import logging
 import os
+from enum import Enum
 from pathlib import Path
 from typing import Any, List, Optional
 
@@ -24,6 +25,7 @@ import click
 import typer
 import yaml
 from snowflake.cli._plugins.helpers.snowsl_vars_reader import check_env_vars
+from snowflake.cli.api.cli_global_context import get_cli_context
 from snowflake.cli.api.commands.snow_typer import SnowTyperFactory
 from snowflake.cli.api.config import (
     ConnectionConfig,
@@ -38,15 +40,13 @@ from snowflake.cli.api.output.types import (
     CommandResult,
     MessageResult,
     MultipleResults,
+    ObjectResult,
 )
 from snowflake.cli.api.project.definition_conversion import (
     convert_project_definition_to_v2,
 )
 from snowflake.cli.api.project.definition_manager import DefinitionManager
 from snowflake.cli.api.project.schemas.project_definition import (
-    DefinitionV10,
-    DefinitionV11,
-    DefinitionV20,
     get_version_map,
 )
 from snowflake.cli.api.secure_path import SecurePath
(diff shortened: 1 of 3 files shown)


sfc-gh-mraba (dismissed):

src/snowflake/cli/_plugins/helpers/commands.py
```diff
@@ -373,3 +380,88 @@ def show_config_sources(
...
+    if output_file is not None:
+        payload = json.dumps(schema, indent=2, sort_keys=True)
+        SecurePath(output_file).write_text(payload + "\n")
+        return MessageResult(f"Project definition schema written to {output_file}.")
+
+    # Under a structured output format, return the schema as an object so the
+    # command emits the JSON Schema document itself rather than a stringified
+    # blob nested under a "message" key.
+    output_format = get_cli_context().output_format
+    if output_format is not None and output_format.is_json:
```
sfc-gh-kolszewski: `get_cli_context().output_format` is typed as `OutputFormat` (not `Optional[OutputFormat]`) and defaults to `OutputFormat.TABLE`, so the `is not None` guard is always `True`. We can simplify:

```python
if get_cli_context().output_format.is_json:
```

src/snowflake/cli/_plugins/helpers/commands.py
```diff
@@ -373,3 +380,88 @@ def show_config_sources(
...
+    schema mapping) or a CI pipeline to get completion and to catch typos and type
+    errors in snowflake.yml before a deploy. The schema is generated from the CLI's
+    own pydantic models, so it stays in sync with the structural rules the CLI
+    enforces; some cross-field/semantic checks are only applied at load/deploy time.
+    """
+    schema = _build_project_definition_schema(version.value)
+
+    if output_file is not None:
+        payload = json.dumps(schema, indent=2, sort_keys=True)
+        SecurePath(output_file).write_text(payload + "\n")
```
sfc-gh-kolszewski, line 461: When the parent directory of `output_file` does not exist, `SecurePath.write_text` calls `touch()` internally, which delegates to `Path.touch()` — that raises an unhandled `FileNotFoundError` rather than a clean `ClickException`. Typer's `writable=True` on the option only validates the file itself if it already exists, so it does not catch this case.

We can add an explicit check:

```python
if output_file is not None:
    if not output_file.parent.exists():
        raise ClickException(
            f"Directory '{output_file.parent}' does not exist."
        )
    payload = json.dumps(schema, indent=2, sort_keys=True)
    SecurePath(output_file).write_text(payload + "\n")
    return MessageResult(f"Project definition schema written to {output_file}.")
```

And a matching test in `test_generate_project_schema.py`:

```python
def test_output_file_missing_parent_raises(runner, tmp_path):
    out = tmp_path / "nonexistent_dir" / "schema.json"
    result = runner.invoke(
        ["helpers", COMMAND, "--definition-version", "2", "-o", str(out)]
    )
    assert result.exit_code != 0
    assert "nonexistent_dir" in result.output
```

src/snowflake/cli/_plugins/helpers/commands.py
sfc-gh-olorek: Done in 8337761d — simplified to `if get_cli_context().output_format.is_json:`. Confirmed the property is non-`Optional` (returns `OutputFormat`, defaults to `OutputFormat.TABLE`) and that the rest of the codebase already uses `.output_format.is_json` directly, so the `is not None` guard was dead.

src/snowflake/cli/_plugins/helpers/commands.py
sfc-gh-olorek: Done in 8337761d. Added the explicit parent-existence check (using `click.ClickException` to match the existing style in this file — see the `raise click.ClickException(...)` already at the top) plus a `test_output_file_missing_parent_raises` test.

One tweak vs. the suggested test: I assert on `"does not exist"` rather than `"nonexistent_dir"`, because the Rich error panel word-wraps the long temp path and splits the directory name across lines (`nonexistent_` / `dir`), so the literal substring is not present in the rendered output. Also added `assert not out.exists()` to confirm nothing is partially created.

sfc-gh-kolszewski (approved):

PR #2952, opened 2026-05-05
Snowflake Apps: support managed compute pools

sfc-gh-gbloom:
### Pre-review checklist

### Changes description

Adds support for the `ENABLE_APPLICATION_SERVICE_MANAGED_COMPUTE_POOL` backend parameter that let the server allocate the build and service compute pools instead of requiring them to be configured by the user.

* `snow app setup` skips both `build_compute_pool` and `service_compute_pool` when the parameter is enabled, omitting the blocks from the generated `snowflake.yml` and printing a notice that managed pools are in use.
* `snow app deploy` forwards an empty string as the 4th positional argument to `SYSTEM$SPCS_TEST_BUILD_APP_ARTIFACT_REPO()` and omits the `IN COMPUTE POOL` clause from `CREATE APPLICATION SERVICE`.
* If the user-supplied `snowflake.yml` still pins either pool, we check for `ENABLE_APPLICATION_SERVICE_MANAGED_COMPUTE_POOL_FALLBACK ` param as well during deploy. If the param is false, we ignore user's value and print a warning that a managed pool will be used instead.
* `_generate_snowflake_yml` now treats `build_compute_pool` and `service_compute_pool` as optional and omits the blocks when not set.

### DIFF
--- a/src/snowflake/cli/_plugins/apps/commands.py
+++ b/src/snowflake/cli/_plugins/apps/commands.py
@@ -122,6 +122,7 @@ def snowflake_app_setup(
     metrics = ctx.metrics
     with metrics.span("snowflake_app.setup.resolve_defaults"):
         params = manager.fetch_snow_apps_parameters()
+        managed_compute_pool_enabled = manager.is_managed_compute_pool_enabled()
 
     def _resolve(
         user_input=None,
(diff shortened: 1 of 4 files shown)


src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -459,6 +473,30 @@ def snowflake_app_deploy(
...
     build_eai = defaults["build_eai"]
 
+    # When the backend opts the account into managed compute pools, ignore
+    # any user-provided ``build_compute_pool`` / ``service_compute_pool``
+    # (warning the user) so the server allocates managed pools instead.
+    # ``build_compute_pool`` becomes an empty string so it is forwarded as
+    # the 4th argument to ``SYSTEM$SPCS_TEST_BUILD_APP_ARTIFACT_REPO``;
+    # ``service_compute_pool`` becomes ``None`` so the deploy step omits the
+    # ``IN COMPUTE POOL`` clause from ``CREATE APPLICATION SERVICE``.
+    managed_compute_pool_enabled = manager.is_managed_compute_pool_enabled()
```
sfc-gh-anwang: It makes sense for the snow app setup step to not generate those fields in the yaml file if managed compute pool is enabled, but I don't think we should ignore those fields in the yaml if user specified them. It is possible that we gonna enable `ENABLE_APPLICATION_SERVICE_MANAGED_COMPUTE_POOL_FALLBACK` by default which would allow user to specify user created pool in DDL.  
To keep the CLI cleaner, CLI should specify the `IN COMPUTE POOL` clause if the property is specified in the yaml field.

sfc-gh-gbloom: Added a check for `ENABLE_APPLICATION_SERVICE_MANAGED_COMPUTE_POOL_FALLBACK` as well.

We specify `IN COMPUTE POOL` if the user has specified a value in `snowflake.yml`.

We print a warning message if:
1. If the user has specified a value in `snowflake.yml` and
2. `ENABLE_APPLICATION_SERVICE_MANAGED_COMPUTE_POOL` is true and
3. `ENABLE_APPLICATION_SERVICE_MANAGED_COMPUTE_POOL_FALLBACK` is false

sfc-gh-anwang (approved):
LGTM

PR #2953, opened 2026-05-06
Snowflake Apps: emit app_flow in telemetry for `snow app *` commands

sfc-gh-gbloom:
### Pre-review checklist

### Changes description

This PR adds an `app_flow` field to the CLI telemetry payload (values: `native_app` or `snowflake_app`) so the two flows can be distinguished.

### DIFF
--- a/src/snowflake/cli/_app/telemetry.py
+++ b/src/snowflake/cli/_app/telemetry.py
@@ -86,6 +86,18 @@ class CLITelemetryField(Enum):
     # Project context
     PROJECT_DEFINITION_VERSION = "project_definition_version"
     MODE = "mode"
+    # Which "snow app" product flow ran for this command:
+    # "native_app" for Native App (application / application package entities),
+    # "snowflake_app" for Snowflake Apps Deploy (snowflake-app entities).
+    # Only emitted for "snow app *" commands; absent for everything else.
+    #
+    # Note: the field appears on TelemetryEvent.CMD_EXECUTION_RESULT and
+    # CMD_EXECUTION_ERROR but NOT on CMD_EXECUTION, because the routing
+    # decorators that resolve the flow run inside the command callable --
+    # i.e. AFTER log_command_usage fires in pre_execute. Consumers who need
+    # the flow on the executing_command event should join on
+    # COMMAND_EXECUTION_ID across the three events for an invocation.
+    APP_FLOW = "app_flow"
 
 
 class TelemetryEvent(Enum):
@@ -169,7 +181,7 @@ def _find_command_info() -> TelemetryDict:
     if format_value is None:
         format_value = OutputFormat.TABLE
 
-    return {
+    info: TelemetryDict = {
         CLITelemetryField.COMMAND: command_path,
         CLITelemetryField.COMMAND_GROUP: command_path[0],
         CLITelemetryField.COMMAND_FLAGS: command_flags,
@@ -177,6 +189,12 @@ def _find_command_info() -> TelemetryDict:
         CLITelemetryField.PROJECT_DEFINITION_VERSION: str(_get_definition_version()),
         CLITelemetryField.MODE: _get_cli_running_mode(),
     }
+
+    app_flow = _get_app_flow()
+    if app_flow is not None:
+        info[CLITelemetryField.APP_FLOW] = app_flow
+
+    return info
 
 
 def _get_cli_running_mode() -> str:
@@ -198,6 +216,23 @@ def _get_definition_version() -> str | None:
         # (especially for commands that don't normally load it)
         pass
     return None
+
+
+def _get_app_flow() -> str | None:
+    """Return the resolved "snow app" product flow for the current command.
+
+    See ``snowflake.cli._plugins.nativeapp.v2_conversions.compat.AppFlow``;
+    this returns ``"native_app"``, ``"snowflake_app"``, or ``None`` for any
+    command outside the ``snow app`` group (or when the routing helpers
+    haven't run yet).
+    """
+    try:
+        return get_cli_context().app_flow
+    except AttributeError:
+        # CLI context not yet populated (e.g. during early startup or in
+        # contrived test scenarios where the manager was reset). Treat as
+        # "no flow recorded" rather than failing the whole telemetry call.
+        return None
 
 
 def _is_env_truthy(name: str) -> bool:
--- a/src/snowflake/cli/_plugins/nativeapp/v2_conversions/compat.py
+++ b/src/snowflake/cli/_plugins/nativeapp/v2_conversions/compat.py
@@ -57,6 +57,16 @@ NATIVE_APP_ENTITY_TYPES = {"application", "application package"}
 SNOWFLAKE_APP_ENTITY_TYPES = {"snowflake-app"}
 
 
+def set_app_flow(flow: AppFlow) -> None:
+    """Record the resolved ``snow app`` product flow on the CLI context so
+    telemetry (see ``CLITelemetryField.APP_FLOW``) can distinguish Native App
+    invocations from Snowflake Apps Deploy invocations.
+
+    Safe to call multiple times; the last call wins.
+    """
+    get_cli_context_manager().app_flow = flow.value
+
+
 APP_AND_PACKAGE_OPTIONS = [
     inspect.Parameter(
         "package_entity_id",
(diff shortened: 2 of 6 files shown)


src/snowflake/cli/_app/telemetry.py
```diff
@@ -86,6 +86,11 @@ class CLITelemetryField(Enum):
     # Project context
     PROJECT_DEFINITION_VERSION = "project_definition_version"
     MODE = "mode"
+    # Which "snow app" product flow ran for this command:
+    # "native_app" for Native App (application / application package entities),
+    # "snowflake_app" for Snowflake Apps Deploy (snowflake-app entities).
+    # Only emitted for "snow app *" commands; absent for everything else.
+    APP_FLOW = "app_flow"
```
sfc-gh-olorek, line 100: Worth calling out in this comment: per the test docstring below, `app_flow` isn't present on the initial `executing_command` event — it only shows up on the terminal event, because the routing decorator hasn't run yet when `log_command_usage` fires. Consumers joining on `command_execution_id` across events will want to know that, and the field-level docstring is probably the first place they'll look, not the test.

src/snowflake/cli/_app/telemetry.py
```diff
@@ -200,6 +211,20 @@ def _get_definition_version() -> str | None:
...
+    """Return the resolved "snow app" product flow for the current command.
+
+    See ``snowflake.cli._plugins.nativeapp.v2_conversions.compat.AppFlow``;
+    this returns ``"native_app"``, ``"snowflake_app"``, or ``None`` for any
+    command outside the ``snow app`` group (or when the routing helpers
+    haven't run yet).
+    """
+    try:
+        return get_cli_context().app_flow
+    except Exception:
```
sfc-gh-olorek: nit: `except Exception` feels too broad for what we're guarding against here. The only realistic failure I can think of is the context attribute not being populated yet, which would be an `AttributeError`. Narrowing it makes the intent clearer and avoids silently swallowing an unrelated bug in `get_cli_context()` down the line.

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -84,6 +88,11 @@ def snowflake_app_setup(
     See the ``snow app setup`` command in
     :mod:`snowflake.cli._plugins.nativeapp.commands` for the CLI surface.
     """
+    # ``snow app setup`` is the only "snow app" subcommand that runs without
+    # going through the routing decorators in compat.py, so set the flow
+    # explicitly so telemetry can attribute it to Snowflake Apps Deploy.
+    set_app_flow(AppFlow.SNOWFLAKE_APP)
```
sfc-gh-olorek: The comment is honest about the gap, which I appreciate — but it also means the next person adding a `snow app` subcommand has to remember this manually or telemetry silently drops the field. Would it be worth setting a default at the group callback level (wherever `snow app` is wired up) so the baseline is `SNOWFLAKE_APP` and the routing decorators only override when they detect Native App? That way new commands are correct by default.

src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -566,7 +575,7 @@ def snowflake_app_deploy(
             )
             workspace_build_source_uri_str = str(workspace_build_source_uri)
             version_match = re.search(
-                r"/versions/([^/]+)/", workspace_build_source_uri_str
+                r"/versions/([^/]+)", workspace_build_source_uri_str
```
sfc-gh-olorek: This looks like a standalone bug fix (version URIs without a trailing slash) that slipped in with the telemetry work. Can we split it into its own PR, or at least mention it in the description? Anyone reviewing for telemetry changes will scroll right past this, and it won't make it into release notes.

src/snowflake/cli/_plugins/apps/generate.py
```diff
@@ -74,7 +74,9 @@ def _generate_snowflake_yml(
             f"{database}.{schema}.{DEFAULT_PERSONAL_WORKSPACE_NAME}\n"
         )
     else:
-        code_storage_block = f"\n            code_stage: {app_id.upper()}_CODE\n"
+        code_storage_block = (
+            f"\n            code_stage: {database}.{schema}.{app_id.upper()}_CODE\n"
```
sfc-gh-olorek: Separate concern from telemetry: this changes the generated `snowflake.yml` so `code_stage` is fully-qualified with `{database}.{schema}` rather than a bare name. Looks like a legit improvement (removes the implicit session dependency, and the test updates back it up), but it really belongs in its own PR. As-is, anyone scanning this PR for telemetry changes is going to miss that the generator's output just changed for all users.

sfc-gh-olorek (commented):
Left a few inline comments. The telemetry work itself looks good — my main concern is the number of unrelated behavior changes riding along in this PR, which will make it hard for anyone reading the git log or writing release notes later to figure out what actually happened here.

src/snowflake/cli/_plugins/apps/manager.py
```diff
@@ -642,7 +642,7 @@ def upload_to_workspace(
         workspace_fqn: FQN,
         target_subdirectory: Optional[str] = None,
         overwrite: bool = True,
-    ) -> Iterator[Dict[str, str]]:
+    ) -> List[Dict[str, str]]:
```
sfc-gh-olorek: Two things here, and neither of them are about telemetry:

1. This changes the semantics from streaming (generator) to eager (list). If any caller was wrapping this in a progress indicator, they'll now stare at a frozen UI until *every* file finishes uploading instead of getting per-file ticks. Worth sweeping the call sites to confirm nobody relied on the streaming behavior.
2. Like the regex and `code_stage` changes, this doesn't belong in a telemetry PR — please split it out or at least mention it in the description so release notes can capture it.

src/snowflake/cli/_plugins/nativeapp/v2_conversions/compat.py
```diff
@@ -209,6 +219,9 @@ def force_project_definition_v2(
     def decorator(func):
         @wraps(func)
         def wrapper(*args, **kwargs):
+            # Commands using this decorator are always Native App, regardless
+            # of whether the source definition is v1 or v2.
+            set_app_flow(AppFlow.NATIVE_APP)
```
sfc-gh-olorek: Setting `NATIVE_APP` unconditionally here hinges on an invariant — "this decorator is only ever applied to Native App commands" — that isn't enforced anywhere in code. If someone reuses `force_project_definition_v2` on a Snowflake App command down the road (even by accident), telemetry will silently misattribute and we won't notice until someone looks at the data. Worth either a test that locks in the set of callers, or something more discriminating at runtime (e.g. detecting the flow from the PDF the way `_detect_flow_from_project` does, rather than hardcoding).

src/snowflake/cli/_app/telemetry.py
sfc-gh-gbloom: Updated the docstring

src/snowflake/cli/_app/telemetry.py
sfc-gh-gbloom: Narrowed to `AttributeError`

src/snowflake/cli/_plugins/apps/commands.py
sfc-gh-gbloom: Added an `@app.callback()` that defaults `app_flow=snowflake_app` for every `snow app` invocation

src/snowflake/cli/_plugins/apps/commands.py
sfc-gh-gbloom: Agree, removed from this PR

src/snowflake/cli/_plugins/apps/generate.py
sfc-gh-gbloom: Removed from this PR

src/snowflake/cli/_plugins/apps/manager.py
sfc-gh-gbloom: Removed from this PR

src/snowflake/cli/_plugins/nativeapp/v2_conversions/compat.py
sfc-gh-gbloom: Moved `set_app_flow(NATIVE_APP)` to after the v1-to-v2 conversion and entity resolution succeeds. Adding a test with a closed set of callers.

sfc-gh-olorek (dismissed):
Thanks for addressing the comments — the three drive-by changes split out cleanly, the `force_project_definition_v2` stamping now happens after entity resolution, and the group-level callback means new `snow app` subcommands default correctly. LGTM.

sfc-gh-olorek (approved):

Issue #2954, opened 2026-05-06
SNOW-3480430: snow connection list fails with AttributeError: \'String\' object has no attribute \'items\'


sfc-gh-edendulk:
### SnowCLI version

3.15.0_1

### Python version

3.13

### Platform

macOS

### What happened

Running snow connection list crashes with AttributeError: 'String' object has no attribute 'items'. The CLI expects each connection entry in the config to be a dictionary/table, but encounters a plain string value instead. This causes ConnectionConfig.from_dict() to fail when iterating over what it assumes is a dict.

Environment:
• Snowflake CLI version: 3.15.0_1 (Homebrew)
• Python: 3.13
• OS: macOS (Apple Silicon)

### Console output

```Shell
edendulk@HP7JGV926Q sncf-validation-onboarding % snow connection list --debug
2026-05-06 06:54:22 DEBUG [snowflake.cli.api.commands.snow_typer] Executing command exception callback
2026-05-06 06:54:22 DEBUG [snowflake.cli.api.commands.snow_typer] Executing command exception callback
2026-05-06 06:54:22 DEBUG [snowflake.cli.api.commands.snow_typer] Executing command post execution callback
2026-05-06 06:54:22 DEBUG [snowflake.cli.api.commands.snow_typer] Executing command post execution callback
2026-05-06 06:54:22 DEBUG [snowflake.connector.network] Opentelemtry otel injection failed
Traceback (most recent call last):
  File "/opt/homebrew/Cellar/snowflake-cli/3.15.0_1/libexec/lib/python3.13/site-packages/snowflake/cli/api/commands/snow_typer.py", line 141, in command_callable_decorator
    raise exception
  File "/opt/homebrew/Cellar/snowflake-cli/3.15.0_1/libexec/lib/python3.13/site-packages/snowflake/cli/api/commands/snow_typer.py", line 135, in command_callable_decorator
    result = command_callable(*args, **kw)
  File "/opt/homebrew/Cellar/snowflake-cli/3.15.0_1/libexec/lib/python3.13/site-packages/snowflake/cli/api/commands/decorators.py", line 173, in wrapper
    return func(**options)
  File "/opt/homebrew/Cellar/snowflake-cli/3.15.0_1/libexec/lib/python3.13/site-packages/snowflake/cli/_plugins/connection/commands.py", line 101, in list_connections
    connections = get_all_connections()
[...]

PR #2955, opened 2026-05-06
fix(connection): skip non-table [connections] entries instead of crashing (SNOW-3480430)

sfc-gh-olorek:
### Pre-review checklist

### Changes description

Fixes SNOW-3480430 / #2954.

`snow connection list` crashed with `AttributeError: 'String' object has no attribute 'items'` when `config.toml` had a scalar value directly under `[connections]`, e.g.:

```toml
[connections]
stray = "not-a-table"

[connections.good]
account = "acc"
user = "u"
```

`ConnectionConfig.from_dict` iterated the scalar with `.items()` unconditionally, so the whole command aborted and the user could not list any of their (otherwise valid) connections.

**Fix**

1. `ConnectionConfig.from_dict` now raises a clear `ClickException` when the input is not a mapping, so direct callers — and anyone writing config programmatically — get an actionable error (\"Connection configuration must be a TOML table with key/value pairs…\") instead of a raw `AttributeError`.
2. Both providers filter non-table entries under `[connections]` with a warning, so `snow connection list` keeps working and the user can see their good connections while being told which entry is malformed:
   - `LegacyConfigProvider.get_all_connections()` now filters via a shared `_is_connection_table` helper.
   - `AlternativeConfigProvider.get_all_connections()` (the `--all` / env-connections path) filters via the same helper.
   - `AlternativeConfigProvider._get_file_based_connections()` already silently dropped non-dicts; it now uses the same helper so the skip is logged consistently.

**Tests**

- `tests/test_config.py`:
  - `test_connection_config_from_dict_rejects_non_mapping` — parametrized over `str`, `int`, `bool`, `list` to confirm the `ClickException` message mentions \"TOML table\" and the offending type.
  - `test_connection_config_from_dict_accepts_empty_mapping` — empty table is still valid.
  - `test_connection_config_from_dict_preserves_known_and_other_settings` — happy path still routes unknown keys into `_other_settings`.
- `tests/test_connection.py::test_connection_list_tolerates_non_table_entry` — parametrized over legacy vs `SNOWFLAKE_CLI_CONFIG_V2_ENABLED=1`; writes a `config.toml` with a stray scalar under `[connections]` plus one good connection, invokes `snow connection list --format json`, and asserts the command exits 0, the good connection is listed, and the stray one is not.

Full `tests/test_config.py`, `tests/test_connection.py`, `tests/test_config_provider_integration.py`, and `tests/config_ng/` suites pass locally (402 passed, 15 skipped).

### Release notes

Entry added under *Unreleased version → Fixes and improvements*.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -27,6 +27,7 @@
 * Fixed macOS arm64 installer incorrectly requiring Rosetta 2. The `Distribution.xml` package metadata now declares `hostArchitectures="arm64,x86_64"`, so the installer is recognized as native on Apple Silicon.
 * Fixed SQL string literal escaping in `identifier_to_show_like_pattern` and `to_string_literal` to use Snowflake's standard single-quote doubling (`''`) instead of backslash escaping, which is not interpreted under the default `STANDARD_ESCAPE_SEQUENCES=FALSE` session setting. This closes SQL injection vectors through `SHOW ... LIKE` queries driven by a project-controlled `snowflake.yml` or `manifest.yml`.
 * Upgraded `pip` to version 26.1.1.
+* Fixed `snow connection list` crashing with `AttributeError` when `config.toml` contains a scalar value directly under `[connections]`. Such entries are now skipped with a warning so valid connections are still listed.
 
 
 # v3.17.1
--- a/src/snowflake/cli/api/config.py
+++ b/src/snowflake/cli/api/config.py
@@ -34,6 +34,7 @@ import tomlkit
 from click import ClickException
 from snowflake.cli.api.exceptions import (
     ConfigFileTooWidePermissionsError,
+    InvalidConnectionConfigurationError,
     MissingConfigurationError,
     UnsupportedConfigSectionTypeError,
 )
@@ -139,6 +140,13 @@ class ConnectionConfig:
 
     @classmethod
     def from_dict(cls, config_dict: dict) -> ConnectionConfig:
+        if not isinstance(config_dict, dict):
+            raise InvalidConnectionConfigurationError(
+                "Expected a TOML table with key/value pairs "
+                f"(e.g. `[connections.my_conn]`), got {type(config_dict).__name__}. "
+                "Check your config.toml for a connection entry that was written as a "
+                "plain value instead of a table."
+            )
         known_settings = {}
         other_settings = {}
         for key, value in config_dict.items():
--- a/src/snowflake/cli/api/config_provider.py
+++ b/src/snowflake/cli/api/config_provider.py
@@ -32,6 +32,31 @@ ALTERNATIVE_CONFIG_ENV_VAR: Final[str] = "SNOWFLAKE_CLI_CONFIG_V2_ENABLED"
 log = logging.getLogger(__name__)
 
 
+def _is_connection_table(name: str, config: Any) -> bool:
+    """Return True if ``config`` is a mapping suitable for ``ConnectionConfig.from_dict``.
+
+    A connection entry in ``config.toml`` must be a TOML table (e.g.
+    ``[connections.my_conn]``). If a user writes a plain value under
+    ``[connections]`` — for example ``my_conn = "value"`` at the top of the
+    section — downstream code crashes with ``AttributeError: 'String' object
+    has no attribute 'items'``. Skip those entries with a warning so
+    ``snow connection list`` still surfaces the valid ones.
+
+    ``tomlkit.items.Table`` inherits from ``dict``, so ``isinstance(x, dict)``
+    accepts the types we actually get from parsed config files.
+    """
+    if isinstance(config, dict):
+        return True
+    from snowflake.cli.api.console import cli_console
+
+    cli_console.warning(
+        f"Skipping connection '{name}' in config: expected a table "
+        f"(e.g. [connections.{name}]), got {type(config).__name__}. "
+        "Fix or remove this entry to silence this warning."
+    )
+    return False
+
+
 class ConfigProvider(ABC):
     """
     Abstract base class for configuration providers.
@@ -143,6 +168,7 @@ class LegacyConfigProvider(ConfigProvider):
         return {
             name: ConnectionConfig.from_dict(config)
             for name, config in connections.items()
+            if _is_connection_table(name, config)
         }
 
 
@@ -490,6 +516,7 @@ class AlternativeConfigProvider(ConfigProvider):
         return {
             name: ConnectionConfig.from_dict(config)
             for name, config in connections_dict.items()
+            if _is_connection_table(name, config)
         }
 
     def _get_file_based_connections(self) -> dict:
@@ -551,7 +578,7 @@ class AlternativeConfigProvider(ConfigProvider):
         filtered_connections: Dict[str, Dict[str, Any]] = {}
         if isinstance(raw_connections, dict):
             for conn_name, conn_config in raw_connections.items():
-                if isinstance(conn_config, dict):
+                if _is_connection_table(conn_name, conn_config):
                     filtered_connections[conn_name] = conn_config
 
         if general_params:
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -21,6 +21,7 @@ import pytest
 from snowflake.cli.api.cli_global_context import fork_cli_context
 from snowflake.cli.api.config import (
     ConfigFileTooWidePermissionsError,
+    ConnectionConfig,
     config_init,
     get_config_section,
     get_connection_dict,
@@ -685,3 +686,33 @@ def test_corrupted_config_raises_human_friendly_error(
 )
 def test_get_env_variable_name(path, key, expected):
     assert get_env_variable_name(*path, key=key) == expected
+
+
+@pytest.mark.parametrize("bad_value", ["just-a-string", 42, True, ["a", "b"]])
+def test_connection_config_from_dict_rejects_non_mapping(bad_value):
+    from snowflake.cli.api.exceptions import InvalidConnectionConfigurationError
+
+    with pytest.raises(InvalidConnectionConfigurationError) as exc_info:
+        ConnectionConfig.from_dict(bad_value)
+    assert "TOML table" in exc_info.value.message
+    assert type(bad_value).__name__ in exc_info.value.message
+    assert "Invalid connection configuration." in exc_info.value.format_message()
+
+
+def test_connection_config_from_dict_accepts_empty_mapping():
+    config = ConnectionConfig.from_dict({})
+    assert config.account is None
+    assert config.user is None
+
+
+def test_connection_config_from_dict_preserves_known_and_other_settings():
+    config = ConnectionConfig.from_dict(
+        {"account": "acc", "user": "u", "custom_setting": "value"}
+    )
+    assert config.account == "acc"
+    assert config.user == "u"
+    assert config.to_dict_of_all_non_empty_values() == {
+        "account": "acc",
+        "user": "u",
+        "custom_setting": "value",
+    }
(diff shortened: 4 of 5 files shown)


src/snowflake/cli/api/config_provider.py
```diff
@@ -32,6 +32,30 @@
...
+
+    A connection entry in ``config.toml`` must be a TOML table (e.g.
+    ``[connections.my_conn]``). If a user writes a plain value under
+    ``[connections]`` — for example ``my_conn = "value"`` at the top of the
+    section — ``config.items()`` returns that entry with a non-table value and
+    downstream code crashes with ``AttributeError: 'String' object has no
+    attribute 'items'``. Skip those entries with a warning so
+    ``snow connection list`` still surfaces the valid ones.
+    """
+    if hasattr(config, "items"):
```
sfc-gh-jwilkowski: **Prefer `isinstance(config, dict)` over `hasattr(config, "items")`**

Every other guard in this file uses `isinstance(x, dict)`. `tomlkit.Table` is a `dict` subclass, so the stronger check accepts exactly the right types and is consistent with the existing convention throughout both this file and `config.py`.

src/snowflake/cli/api/config_provider.py
```diff
@@ -32,6 +32,30 @@
...
+        return True
+    log.warning(
```
sfc-gh-jwilkowski: **Warning is invisible to users under normal invocation**

`log.warning()` only reaches the terminal when `--verbose`/`--debug` is active — the console handler's default threshold is `ERROR`. A user whose connection is silently dropped will see nothing at the terminal.

Replace (or supplement) with `cli_console.warning()`:

```python
from snowflake.cli.api.console import cli_console
# …
cli_console.warning(
    f"Skipping connection '{name}' in config: expected a table "
    f"(e.g. [connections.{name}]), got {type(config).__name__}. "
    "Fix or remove this entry to silence this warning."
)
```

`cli_console.warning()` writes to the terminal unconditionally and runs `sanitize_for_terminal()` on the message automatically — which also addresses a secondary issue: if `name` contains a newline (possible from a crafted `config.toml`), Python's `FileHandler` does not escape it, enabling log injection. `cli_console` sanitizes before output.

src/snowflake/cli/api/config.py
```diff
@@ -136,6 +136,13 @@ class ConnectionConfig:
 
     @classmethod
     def from_dict(cls, config_dict: dict) -> ConnectionConfig:
+        if not hasattr(config_dict, "items"):
+            raise ClickException(
```
sfc-gh-jwilkowski: **Prefer `not isinstance(config_dict, dict)`**

Same convention issue as in `config_provider.py` — all existing guards in this file use `isinstance`. `tomlkit.Table` is a `dict` subclass so this is safe and more idiomatic.

src/snowflake/cli/api/config.py
```diff
@@ -136,6 +136,13 @@ class ConnectionConfig:
...
+                "Connection configuration must be a TOML table with key/value pairs "
```
sfc-gh-jwilkowski: **Use `InvalidConnectionConfigurationError` instead of `ClickException`**

`ClickException` is being migrated away from in favour of the domain hierarchy in `exceptions.py`. `InvalidConnectionConfigurationError` is already defined there, inherits `CliError → BaseCliError → ClickException`, and its `format_message()` prepends "Invalid connection configuration." which is semantically accurate here:

```python
from snowflake.cli.api.exceptions import InvalidConnectionConfigurationError
# …
raise InvalidConnectionConfigurationError(
    "Connection configuration must be a TOML table with key/value pairs "
    f"(e.g. `[connections.my_conn]`), got {type(config_dict).__name__}. "
    "Check your config.toml for a connection entry written as a plain value."
)
```

tests/test_connection.py
```diff
@@ -415,6 +415,47 @@ def test_mask_sensitive_parameters_masks_all_known_sensitive_keys():
     assert params["password"] == "hunter2"
 
 
+@pytest.mark.parametrize("config_v2", [False, True])
+def test_connection_list_tolerates_non_table_entry(
+    runner, caplog, monkeypatch, config_v2
```
sfc-gh-jwilkowski: **`caplog` is declared but never used**

The docstring says the bad entry is "skipped with a warning" but nothing asserts on it. A future change could silently remove the warning and this test would still pass. Either remove `caplog` from the signature, or add:

```python
import logging
with caplog.at_level(logging.WARNING, logger="snowflake.cli.api.config_provider"):
    result = runner.invoke_with_config_file(...)
assert "Skipping connection 'stray'" in caplog.text
```

tests/test_connection.py
```diff
@@ -415,6 +415,47 @@ def test_mask_sensitive_parameters_masks_all_known_sensitive_keys():
...
+
+        result = runner.invoke_with_config_file(
+            tmp_file.name, ["connection", "list", "--format", "json"]
+        )
+
+        assert result.exit_code == 0, result.output
+        payload = json.loads(result.output)
+        names = {entry["connection_name"] for entry in payload}
+        assert "good" in names
+        assert "stray" not in names
```
sfc-gh-jwilkowski, line 462: **`--all` path not covered**

The `_is_connection_table` guard added to `AlternativeConfigProvider.get_all_connections()` (the `include_env_connections=True` branch) is not exercised here. Consider adding a sub-case that invokes `["connection", "list", "--all", "--format", "json"]` against the same stray-scalar config and asserts exit code 0 with only the good connection present.

RELEASE-NOTES.md
```diff
@@ -24,6 +24,7 @@
 ## Fixes and improvements
 * Fixed `SELECT *` output being corrupted when joined tables share column names. Duplicate column names are now disambiguated by appending a numeric suffix (e.g. `NAME`, `NAME_2`).
 * Fixed `snow connection generate-jwt` and `snow connection generate-workload-identity-token` failing with `Connection None is not configured` when used with `--temporary-connection`.
+* Fixed `snow connection list` crashing with `AttributeError: 'String' object has no attribute 'items'` when `config.toml` contains a scalar value directly under `[connections]` (e.g. a stray `key = "value"` line). Such entries are now skipped with a warning so valid connections are still listed, and `ConnectionConfig.from_dict` raises a clear error if called directly with a non-table value.
```
sfc-gh-jwilkowski: Suggested shorter wording that drops the internal `ConnectionConfig.from_dict` detail:

```
* Fixed `snow connection list` crashing with `AttributeError` when `config.toml` contains a scalar value directly under `[connections]`. Such entries are now skipped with a warning so valid connections are still listed.
```

sfc-gh-jwilkowski (changes requested):
The core fix is correct — the two-layer defence (filter at the provider, guard in `from_dict`) is the right approach, and the `config_v2` parametrization covers both provider paths. Five items need attention before merge.

src/snowflake/cli/api/config_provider.py
sfc-gh-olorek: Done — switched to `isinstance(config, dict)` in `_is_connection_table`. `tomlkit.items.Table` subclasses `dict`, so this keeps matching the types we actually receive from parsed config files.

src/snowflake/cli/api/config_provider.py
sfc-gh-olorek: Done — swapped `log.warning` for `cli_console.warning`. One caveat worth noting: `cli_console` is auto-muted when `--format json`/`csv` is active (see `_should_force_mute_intermediate_output` in `cli_global_context.py`), so the warning is visible in the default table output but stays out of structured payloads. That seems correct here — downstream consumers of `connection list --format json` shouldn't have to strip log lines. The test now covers both: JSON for the structural assertion, default format for the warning assertion.

src/snowflake/cli/api/config.py
sfc-gh-olorek: Done — switched to `not isinstance(config_dict, dict)` in `ConnectionConfig.from_dict`.

src/snowflake/cli/api/config.py
sfc-gh-olorek: Done — now raising `InvalidConnectionConfigurationError`. Trimmed the prefix from the message so it doesn't duplicate the "Invalid connection configuration." that `format_message` prepends; final rendering reads "Invalid connection configuration. Expected a TOML table with key/value pairs (e.g. `[connections.my_conn]`), got <type>. ...".

tests/test_connection.py
sfc-gh-olorek: Good catch. Dropped `caplog` from the signature (it was leftover from an earlier draft) and added an explicit warning assertion by invoking `connection list` a second time without `--format json` and checking for `"Skipping connection 'stray'"` in `result.output`. JSON output still auto-mutes `cli_console`, so we can't use it to inspect the warning — the default (table) path is the right one to assert on.

tests/test_connection.py
sfc-gh-olorek: Done — added `@pytest.mark.parametrize("extra_args", [[], ["--all"]])` so each `config_v2` variant runs with and without `--all`. With `SNOWFLAKE_CLI_CONFIG_V2_ENABLED=1` + `--all`, this exercises `AlternativeConfigProvider.get_all_connections` (the `include_env_connections=True` branch); without `--all`, it exercises `_get_file_based_connections`. Both paths now go through `_is_connection_table` and the test asserts the good connection survives and the stray one is dropped.

RELEASE-NOTES.md
sfc-gh-olorek: Done — used the suggested shorter wording (dropped the `ConnectionConfig.from_dict` detail). The release notes entry now reads: "Fixed `snow connection list` crashing with `AttributeError` when `config.toml` contains a scalar value directly under `[connections]`. Such entries are now skipped with a warning so valid connections are still listed."

sfc-gh-olorek:
@sfc-gh-jwilkowski all five items addressed and the branch is rebased onto latest `main`. Force-pushed as `fbbed6ae`:

1. `isinstance(x, dict)` in both `config.py` (`ConnectionConfig.from_dict`) and `config_provider.py` (`_is_connection_table`).
2. `cli_console.warning` for the skip notice. Note: structured output (`--format json`/`csv`) auto-mutes `cli_console`, so the warning shows in the default table output and stays out of machine-readable payloads — seems like the right tradeoff, happy to revisit if you'd rather always route it through `stderr`.
3. `InvalidConnectionConfigurationError` instead of `ClickException`; message trimmed so `format_message`'s `"Invalid connection configuration."` prefix doesn't duplicate.
4. `caplog` dropped; test now invokes `connection list` a second time without `--format json` and asserts `"Skipping connection 'stray'"` in the output.
5. `--all` path covered by parametrizing `extra_args` over `[[], ["--all"]]` — exercises `AlternativeConfigProvider.get_all_connections(include_env_connections=True)` when combined with `config_v2=True`.
6. Release notes entry shortened to your suggested wording.

Full `tests/test_config.py`, `tests/test_connection.py`, `tests/test_config_provider_integration.py`, and `tests/config_ng/` pass locally (409 passed, 15 skipped).

tests/test_config.py
```diff
@@ -685,3 +686,33 @@ def test_corrupted_config_raises_human_friendly_error(
 )
 def test_get_env_variable_name(path, key, expected):
     assert get_env_variable_name(*path, key=key) == expected
+
+
+@pytest.mark.parametrize("bad_value", ["just-a-string", 42, True, ["a", "b"]])
+def test_connection_config_from_dict_rejects_non_mapping(bad_value):
+    from snowflake.cli.api.exceptions import InvalidConnectionConfigurationError
```
sfc-gh-jwilkowski: local import

sfc-gh-jwilkowski (changes requested):

tests/test_config.py
sfc-gh-olorek: Addressed

sfc-gh-jwilkowski (approved):

PR #2958, opened 2026-05-06
fix: treat missing-key cleanup as a no-op in OpenConnectionCache

sfc-gh-olorek:
### Pre-review checklist

### Changes description

`OpenConnectionCache._cleanup(key)` had a defensive `if key not in self.connections` check
(`src/snowflake/cli/api/connections.py:293`) that logged a debug line and then fell
through to `self.connections.pop(key).close()` — which raises `KeyError` on the exact
case the check was supposed to handle.

In practice the cleanup timer scheduled by `_touch()` can race with:

* `clear()` (which iterates a snapshot of `connections` and calls `_cleanup` for each key),
* a second `_cleanup(key)` for the same key, or
* an explicit eviction path that removes a connection before its timer fires.

Whichever caller wins removes the connection from `self.connections`, and the loser
fires later with a stale key and crashes on `pop(key)` rather than being a no-op.

The fix:

* Pop with a default of `None` and early-return when the connection is already gone,
  logging the same debug message.
* Move `_cancel_cleanup_future_if_exists(key)` before the early-return so a lingering
  timer for the missing key is always cancelled rather than leaked.

Added a unit test that invokes `_cleanup` on a key that was never inserted, and on a
key whose only remaining state is a pending cleanup future, verifying no exception is
raised and the future is cancelled.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -25,6 +25,7 @@
 ## Fixes and improvements
 * Fixed `SELECT *` output being corrupted when joined tables share column names. Duplicate column names are now disambiguated by appending a numeric suffix (e.g. `NAME`, `NAME_2`).
 * Fixed `snow connection generate-jwt` and `snow connection generate-workload-identity-token` failing with `Connection None is not configured` when used with `--temporary-connection`.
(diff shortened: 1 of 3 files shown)


sfc-gh-olorek:
Closing this stale draft. Internal development moved to the private repository in July 2026 and PR head branches are not mirrored outbound, so this draft cannot land here — anything still wanted has to be re-opened there. Nothing is tracked against this beyond the draft itself; closing rather than leaving it open indefinitely. The branch is preserved.

PR #2959, opened 2026-05-06
fix: guard get_account_identifier / spcs build-image against NULL org

sfc-gh-olorek:
### Pre-review checklist

### Changes description

Both `get_account_identifier()` (in `connection/util.py`) and `snow spcs service build-image` (in `spcs/services/manager.py`) call `cursor.fetchone()` on the result of `SELECT CURRENT_ORGANIZATION_NAME() …` / `SELECT CURRENT_ACCOUNT_NAME() …` and then immediately subscript or `.lower()` the returned row — without checking whether `fetchone()` returned `None` or whether the column values are `NULL` / empty.

In practice this crashes the CLI with either:
- `TypeError: 'NoneType' object is not subscriptable` (when `fetchone()` returns `None`), or
- `AttributeError: 'NoneType' object has no attribute 'lower'` (when the column value is `NULL`)

…instead of a user-visible, actionable error. The existing call site in `dcm/commands.py` already wraps `get_account_identifier()` in a broad `try/except`, which suggests this has been flaky in the field.

**Changes**

1. `connection/util.py::get_account_identifier` — guard against `fetchone() is None` and against `NULL`/empty `ORG`/`ACCT` values. Raise a `ClickException` with a clear message in both cases.
2. `spcs/services/manager.py::build_image` — replace the ad-hoc `SELECT CURRENT_ORGANIZATION_NAME()` + separate `get_account(…)` call (two SQL round-trips, two independent failure modes) with a single call to the hardened `get_account_identifier(…)` helper. One round-trip, consistent error handling.
3. Tests:
   - New parametrised test `test_get_account_identifier_missing_result_raises_click_exception` covers: `fetchone() is None`, `ORG=None`, `ACCT=None`, `ORG=""`, `ACCT=""`.
   - Existing `test_build_image` / `test_build_image_repository_validation` updated to mock the new two-column `{"ORG": …, "ACCT": …}` shape and drop the obsolete `get_account` patch.
4. `RELEASE-NOTES.md` — entry under Unreleased / Fixes and improvements.

**Risk**

Low. Pure error-path hardening + a one-round-trip refactor of an existing call site. No behaviour change on the happy path. Full unit-test suites for `tests/test_utils.py`, `tests/spcs/`, `tests/connection/`, `tests/config_ng/` pass locally.

### DIFF
--- a/src/snowflake/cli/_plugins/spcs/services/manager.py
+++ b/src/snowflake/cli/_plugins/spcs/services/manager.py
@@ -22,7 +22,7 @@ from pathlib import Path
 from typing import List, Optional
 
 import yaml
-from snowflake.cli._plugins.connection.util import get_account
+from snowflake.cli._plugins.connection.util import get_account_identifier
 from snowflake.cli._plugins.object.common import Tag
 from snowflake.cli._plugins.object.manager import ObjectManager
 from snowflake.cli._plugins.spcs.common import (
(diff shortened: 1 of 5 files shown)


src/snowflake/cli/_plugins/spcs/services/manager.py
```diff
@@ -332,15 +332,13 @@ def build_image(
...
-
-        # Get account name using existing utility function
-        account_name = get_account(self._conn)
+        # Resolve org/account via the shared helper — one round-trip, with a
+        # clear error if CURRENT_ORGANIZATION_NAME/CURRENT_ACCOUNT_NAME return
+        # NULL or no row (rather than an opaque ``TypeError: 'NoneType' is not
+        # subscriptable``).
+        account_identifier = get_account_identifier(self._conn)
+        org_name = account_identifier.organization_name.lower()
+        account_name = account_identifier.account_name.lower()
```
sfc-gh-prafalski, line 341: Is it safe to `lower()` those names? There is no explicit information in the docs, whether those values are case-sensitive. 

However, there is that small remark regarding the organization name:
https://docs.snowflake.com/en/user-guide/admin-account-identifier#organization-name
> [...] The name must start with a letter and can only contain letters (lowercase and uppercase) and numbers. The name can’t contain underscores or other delimiters. [...]

sfc-gh-olorek: get_account() already lowercased the account name and org_name was lowercased inline, so this just preserves existing behavior. Both values only feed the registry hostname ({org}-{account}.registry-local.snowflakecomputing.com/...), which is case-insensitive per DNS, so case-folding is safe.

sfc-gh-prafalski (approved):
LGTM

PR #2968, opened 2026-05-06
Snowflake Apps: use correct Snowsight URL for `snow app open --settings`

sfc-gh-gbloom:
### Pre-review checklist

### Changes description

`snow app open --settings` now opens `#/apps/app-service/<db>.<schema>.<app>/details` when the app is deployed as an application service.

The `#/apps/service/...` URL will still be used for legacy apps deployed as SPCS services.

### DIFF
--- a/src/snowflake/cli/_plugins/apps/commands.py
+++ b/src/snowflake/cli/_plugins/apps/commands.py
@@ -324,7 +324,12 @@ def snowflake_app_open(
             f".{identifier_for_url(schema)}"
             f".{identifier_for_url(fqn.name)}"
         )
-        url = make_snowsight_url(ctx.connection, f"#/apps/service/{app_id}/details")
+        service_fqn = FQN(database=db, schema=schema, name=fqn.name)
+        manager = SnowflakeAppManager()
+        segment = (
+            "app-service" if manager.is_application_service(service_fqn) else "service"
+        )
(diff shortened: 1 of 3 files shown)


src/snowflake/cli/_plugins/apps/commands.py
```diff
@@ -324,7 +324,12 @@ def snowflake_app_open(
             f".{identifier_for_url(schema)}"
             f".{identifier_for_url(fqn.name)}"
         )
-        url = make_snowsight_url(ctx.connection, f"#/apps/service/{app_id}/details")
+        service_fqn = FQN(database=db, schema=schema, name=fqn.name)
+        manager = SnowflakeAppManager()
+        segment = (
+            "app-service" if manager.is_application_service(service_fqn) else "service"
```
sfc-gh-jfloyd, line 330: Don't we know whether it's an application service or not based on the snowflake yaml?

sfc-gh-gbloom: Not reliably, it's an implementation detail that is not exposed as a field in `snowflake.yml`. We can use other fields as heuristics, but I figured it would be safer just to check the actual object. We will soon not need to support SPCS services, so this check will be removed in a few weeks.

sfc-gh-jfloyd (approved):

PR #2971, opened 2026-05-06
fix: [SNOW-3481851] skip main_file auto-insert when covered by an existing artifact

### LINKED ISSUE #2967: SNOW-3481851: `streamlit deploy` `TooManyFilesError` regression in 3.16.0
### SnowCLI version

3.16.0

### Python version

3.12.13

### Platform

macOS-26.4.1-arm64-arm-64bit

### What happened

`snow streamlit deploy` raises `TooManyFilesError` when the parent directory of `main_file` is also explicitly listed in `artifacts:`. The same `snowflake.yml` deploys successfully on 3.14.0, so this is a regression introduced between 3.14.0 → 3.16.0.

The trigger is `StreamlitEntity.bundle()` (`streamlit_entity.py`), which auto-inserts `main_file` as an artifact at index 0:

```python
if main_file and not any(artifact.src == main_file for artifact in artifacts):
    artifacts.insert(0, PathMapping(src=main_file))
```

When the explicit `artifacts:` list is the parent directory of `main_file`, the bundle map ends up with two mappings reaching the same destination — the auto-inserted file and the directory walk's discovery of the same file — and `_ArtifactPathMap.put()` raises.

PR #2780 (merged 2026-03-31) added stricter handling for "directory contains a file already mapped" in `_ArtifactPathMap.put()` and may already cover this scenario, but that fix isn't in any released version yet (latest is v3.16.0 from 2026-03-19).

### Console output

```Shell
╭─ Error ──────────────────────────────────────────────────────────────────────╮
│ Multiple file or directories were mapped to one output destination.          │
│                                                                              │
│ destination = apps/par_accuracy_monitoring_v0                                │


Trace under `--debug`:


File ".../snowflake/cli/_plugins/streamlit/streamlit_entity.py", line 108, in deploy
    bundle_map = self.bundle()
File ".../snowflake/cli/_plugins/streamlit/streamlit_entity.py", line 83, in bundle
    return build_bundle(...)
File ".../snowflake/cli/_plugins/nativeapp/artifacts.py", line 80, in bundle_artifacts
    bundle_map.add(artifact)
File ".../snowflake/cli/api/artifacts/bundle_map.py", line 177, in add
    self._add_mapping(mapping.src, mapping.dest)
File ".../snowflake/cli/api/artifacts/bundle_map.py", line 152, in _add_mapping
    self._add(...)
File ".../snowflake/cli/api/artifacts/bundle_map.py", line 110, in _add
    self._artifact_map.put(...)
File ".../snowflake/cli/api/artifacts/bundle_map.py", line 452, in put
    raise TooManyFilesError(dest)
snowflake.cli.api.artifacts.common.TooManyFilesError: Multiple file or directories were mapped to one output destination.

destination = apps/par_accuracy_monitoring_v0
```

### How to reproduce

1. Create a multi-file Streamlit app:
   ```
   apps/my_app/
   ├── app.py
   ├── helpers.py
   └── pages/
       └── home.py
   ```

2. Use this `snowflake.yml`:
   ```yaml
   definition_version: 2
   entities:
     my_app:
       type: streamlit
       identifier:
         name: my_app
         schema: my_schema
         database: my_database
       main_file: apps/my_app/app.py
       query_warehouse: my_warehouse
       compute_pool: my_pool
       runtime_name: SYSTEM$ST_CONTAINER_RUNTIME_PY3_11
       artifacts:
         - apps/my_app/
   ```

3. Run `snow streamlit deploy --replace`.

**Expected**: deploy succeeds (or skips the auto-inserted `main_file` since it overlaps with the directory artifact), as it does on 3.14.0.

**Actual**: `TooManyFilesError` with `destination = apps/my_app`.

**Workaround**: list children of the app dir individually instead of the parent dir, omitting `app.py` (the auto-inserted `main_file`):

```yaml
artifacts:
  - apps/my_app/helpers.py
  - apps/my_app/pages
```

arthurnw:
### Pre-review checklist

### Changes description

Closes #2967.

`StreamlitEntity.bundle()` auto-inserts `main_file` as an artifact unless an existing artifact's `src` is byte-equal to `main_file`. The byte-equality check is too strict: when `main_file = "apps/my_app/app.py"` and the user already lists `apps/my_app/` as an artifact, the auto-insert fires anyway. The bundle map then ends up with two mappings reaching the same destination (the auto-inserted file and the directory walk's discovery of the same file), and `_ArtifactPathMap.put()` raises `TooManyFilesError` on snow CLI ≥ 3.16.0.

This replaces the byte-equality predicate with a coverage check: `main_file` is considered already covered if some artifact's `src` equals `main_file`, or contains it as a descendant **and** the artifact's effective destination matches its source (so the directory walk lands `main_file` at the same canonical deploy path the auto-insert would). Artifacts with a different `dest` still need `main_file` auto-inserted, since the dir walk lands it elsewhere.

Implementation is three small helpers (`_main_file_covered_by_artifacts`, `_artifact_covers_path`, `_artifact_dest_root`) and a one-line predicate swap inside `bundle()`. Trailing-slash semantics in `_artifact_dest_root` mirror `bundle_map._specifies_directory()`; that's noted in the docstring.

#### Note on PR #2780

PR #2780 added directory/glob overlap dedup inside `_ArtifactPathMap.put()` itself. It's adjacent but doesn't cover the auto-insert path: the auto-inserted file and a parent-directory artifact are both files at the same destination, but the parent-directory artifact's `put()` call walks the dir and tries to insert `app.py` from a different source (`apps/my_app` vs `apps/my_app/app.py`), which still trips the file → file conflict branch. Removing the duplicate at insertion time is the cleaner fix and doesn't conflict with #2780.

#### Tests

`tests/streamlit/test_streamlit_entity.py` adds 8 new cases (one parametrized x2):

- `test_bundle_auto_inserts_main_file_when_no_artifacts` — preserves existing behavior.
- `test_bundle_skips_auto_insert_when_main_file_is_explicit_artifact` — preserves existing behavior.
- `test_bundle_skips_auto_insert_when_main_file_inside_directory_artifact` (parametrized: `apps/my_app/` and nested `src/`) — covers the bug.
- `test_bundle_auto_inserts_when_main_file_outside_artifact_directory` — non-overlap.
- `test_bundle_auto_inserts_when_artifact_has_different_dest` — `dest` differs from `src`, so `main_file` deploys to two locations.
- `test_bundle_skips_auto_insert_when_artifact_dest_equals_src` — self-referential `dest` (no trailing slash, equal to `src`).
- `test_bundle_auto_inserts_when_artifact_dest_is_deploy_root` — `dest='./'` lands children at root, different from auto-insert path.
- `test_bundle_handles_glob_overlap_with_main_file` — glob `src` (`pages/*.py`) intentionally falls through to `_ArtifactPathMap.put()` file-level dedup.

Full streamlit suite: 97 passed, 8 snapshots.

#### Manual verification on macOS

Reproduced the failure on stock 3.16.0 with the bug's `snowflake.yml` (deploy raised `TooManyFilesError`), then ran the same project against this branch via `uvx --from <local-checkout>` and confirmed deploy succeeds end-to-end.

#### Not handled here

- Glob artifacts like `pages/*.py` overlapping `main_file`: covered by #2780's downstream dedup at the file level (test included to lock in the fall-through behavior).
- Windows / case-insensitive FS quirks: `Path.relative_to` is path-component aware, not bytewise; should match the dir walk's existing behavior. No Windows-specific test added.


RELEASE-NOTES.md
```diff
@@ -23,9 +23,9 @@
 * Added `--force` flag to `snow spcs service drop` to allow dropping services that contain block storage volumes.
 
 ## Fixes and improvements
+* Fixed `snow streamlit deploy` skipping directory artifacts when `main_file` is already included by an artifact directory.
```
sfc-gh-jwilkowski: What do you think about such wording? Users affected by this bug are more likely searching for TooManyFilesError which is what they see in console. Also skipping might be slightly misleading as ultimately at file gets uploaded
```suggestion
* Fixed `TooManyFilesError` during `snow streamlit deploy` when `main_file` is a descendant of a directory listed in `artifacts`.
```

tests/streamlit/test_streamlit_entity.py
```diff
@@ -53,29 +80,192 @@ def test_bundle_auto_includes_main_file(self, project_directory):
...
+
+        bundle_map = entity.bundle()
+        output = self._bundle_output(tmp_path)
+
+        assert sorted(str(p) for p in bundle_map.all_sources()) == [
+            "pages/main.py",
+            "pages/page.py",
+        ]
+        assert (output / "pages" / "main.py").exists()
+        assert (output / "pages" / "page.py").exists()
```
sfc-gh-jwilkowski, line 287: No test covers main_file at the project root with a './' artifact - could you add one for that as well?

sfc-gh-jwilkowski (changes requested):
Thank you for your contribution! PR looks good for the most part, I've left 2 small requests.

RELEASE-NOTES.md
arthurnw: Thanks for the feedback - this makes sense.

tests/streamlit/test_streamlit_entity.py
arthurnw: Added, thanks for the feedback!

sfc-gh-jwilkowski:
Thanks @arthurnw for your response! Your PR looks good to me, but we have some issues with accepting forks. I've reopened this PR as https://github.com/snowflakedb/snowflake-cli/pull/3059/changes and cheery-picked your commits. Please give it a look and we'll close it

sfc-gh-jwilkowski:
Your fix has been merged @arthurnw thanks once again for your help

PR #2973, opened 2026-05-07
SNOW-3258719 Cleanup DCM v1 API support

sfc-gh-lkyc:
### Pre-review checklist

### Changes description
- Drop the legacy `refresh` response format (`refreshed_tables` / `dt_name` / string statistics):
  - delete `OldFormatExtractor`, `NewFormatExtractor`, the `RefreshDataExtractor`, and `_get_extractor_cls`
  -   inline the new-format extraction directly into `RefreshReporter.extract_data`; consume the validated `RefreshTableResult` pydantic models directly.
  - drop the unused `RefreshTableResult.data_timestamp` field — it was declared on the model but never read anywhere in production code.
- Drop the legacy `plan` response format and align `deploy` / `purge` / `plan` with the rest of the DCM commands:
  - delete `_process_plan_result` and have all three call `PlanReporter(...).process(result)`, the same pattern already used by `refresh` and `test`.
  -  Remove the `version < 2` rejection branch in `PlanReporter.extract_data`; the `version > 2` forward-compat log is retained.
  - Test fixtures updated:
    -   `TestDCMRefresh` payloads converted to the `dts_refresh_result` shape,
    -  legacy empty-plan fixtures (`[("[]",)]`) replaced with calls to a new `_plan_cursor(mock_cursor, row=...)` helper that defaults to `{"version": 2, "changeset": []}`,
    -   legacy `TestRefreshReporter` test class and its snapshots deleted, `TestNewRefreshReporter` renamed to `TestRefreshReporter`.
    - rename the mock cursor column from the legacy `"operations"` to `"result"`, matching the column name returned by the new backend API.


sfc-gh-olorek (commented):
Two inline comments on edge-case behavior and dead fallback code in the refresh path.

src/snowflake/cli/_plugins/dcm/commands.py
```diff
@@ -407,7 +355,8 @@ def deploy(
             skip_plan=skip_plan,
         )
 
-    return _process_plan_result(result, command_name="deploy", save_output=save_output)
+    reporter = PlanReporter(save_output=save_output, command_name="deploy")
```
sfc-gh-olorek, line 358: Minor behavioral change worth calling out: the old `_process_plan_result` returned `CollectionResult([])` and called `save_command_response(command_name, {})` for an empty cursor or an empty/None first cell. `Reporter.process` (in `reporters/base.py`) takes a different path:

- empty cursor → prints `"No data.\n"` and returns `None` (no `save_command_response` call)
- empty string in `row[0]` → `json.loads("")` raises `JSONDecodeError`, which is converted to `CliError("Could not process response.")`

If the v2 backend can still legitimately return an empty cell for “nothing to do,” this flips a soft empty result into a hard error. Worth confirming the backend contract never emits that now, and documenting the save-output behavior change (`out/` no longer written on empty response) in the release notes.

Applies equally to the `purge` and `plan` call sites below.

src/snowflake/cli/_plugins/dcm/reporters/refresh.py
```diff
@@ -283,20 +181,23 @@ def __init__(self, save_output: bool = False):
...
+            response = RefreshResponse.model_validate(result_json)
+        except ValidationError as e:
+            log.info("Failed to validate refresh response: %s", e)
+            raise CliError("Could not process response.")
+
+        if response.dts_refresh_result is None:
+            return []
+
+        return [
+            table.model_dump() for table in response.dts_refresh_result.refreshed_tables
```
sfc-gh-olorek: Now that input to `parse_data` always comes from `RefreshTableResult.model_dump()`, a lot of the defensive code in `RefreshRow.from_dict` is unreachable:

- `data.get(_TABLE_NAME_KEY, "UNKNOWN")` — `RefreshTableResult.table_name` already defaults to `"UNKNOWN"`
- `str(raw_table_name)` — Pydantic guarantees `str`
- `else: log.info("Unexpected statistics type: ...")` branch — `statistics` is typed `Optional[RefreshStatistics]`, so after `model_dump()` it is `None | dict`
- `_safe_int` + the `inserted`/`deleted` setters' fallbacks — `RefreshStatistics.inserted_rows`/`deleted_rows` are already `int`

Since this PR's theme is removing v1 compat shims, it feels like the right moment to either:

**(a)** skip the `model_dump()` round-trip and have `parse_data` consume `List[RefreshTableResult]` directly (requires loosening the `Reporter` base class generic, or just accepting the type mismatch like `PlanReporter` already does), or

**(b)** keep `from_dict` but drop the now-dead defensive branches.

Not a blocker.

sfc-gh-olorek (changes requested):
left some comments

src/snowflake/cli/_plugins/dcm/commands.py
sfc-gh-lkyc: The backend always returns non-empty row

sfc-gh-olorek (dismissed):
lgtm

src/snowflake/cli/_plugins/dcm/reporters/refresh.py
sfc-gh-lkyc: Good point — no, we don't. Done in b56c509e.

`RefreshReporter.extract_data` now returns `List[RefreshTableResult]` (the validated pydantic models) directly. `RefreshRow.from_dict` is replaced by `from_model(table: RefreshTableResult)` which consumes the model via attribute access — the `isinstance(data, dict)` defensive branch, the `Optional[RefreshRow]` return, and the unused `_TABLE_NAME_KEY` / `_STATISTICS_KEY` / `_INSERTED_KEY` / `_DELETED_KEY` constants are gone.

275 DCM tests still green.

sfc-gh-jwilkowski (approved):
thank you for cleaning that up!

PR #2976, opened 2026-05-07
fix: [SNOW-3411343] escape single quotes when building SHOW LIKE patterns

sfc-gh-olorek:
### Pre-review checklist

### Changes description

Fixes [SNOW-3411343](https://snowflakecomputing.atlassian.net/browse/SNOW-3411343).

`identifier_to_show_like_pattern` builds the SQL string literal used by every `SHOW ... LIKE '<pattern>'` query the CLI issues (object lookups, stage existence checks, version/`show-versions` queries, image-repository lookups — 5 call sites). It already escaped the LIKE wildcards `%` and `_`, but did not escape single quotes. A quoted identifier like `"x'; <payload>;--"` coming from a project file therefore produced the literal `'x'; <payload>;--'`, and because `execute_stream` tokenizes on `;`, the payload ran as a separate statement.

The fix doubles every embedded single quote before wrapping the pattern as a literal — Snowflake's canonical string-literal escape. That keeps the pattern a single well-formed literal regardless of what comes from `snowflake.yml`:

```python
pattern = escape_like_pattern(unquote_identifier(identifier))
pattern = pattern.replace("'", "''")
return f"'{pattern}'"
```

I considered the ticket's suggestion of routing through `to_string_literal`, but that function runs `codecs.encode(..., "unicode-escape")` on the input, which would re-encode the backslashes that `escape_like_pattern` already inserted for `%`/`_` and produce `\\\\_` in the final literal — breaking legitimate LIKE escapes. Doubling the single quote inline is the minimal change that preserves existing behaviour for every other caller.

New tests cover wildcard escaping, quoted vs. unquoted canonicalization, embedded single quotes, the double-quote unescaping path, and an explicit injection payload. All 903 tests across the 10 affected test modules pass locally.

[SNOW-3411343]: https://snowflakecomputing.atlassian.net/browse/SNOW-3411343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -26,6 +26,7 @@
 * Fixed `SELECT *` output being corrupted when joined tables share column names. Duplicate column names are now disambiguated by appending a numeric suffix (e.g. `NAME`, `NAME_2`).
 * Fixed `snow connection generate-jwt` and `snow connection generate-workload-identity-token` failing with `Connection None is not configured` when used with `--temporary-connection`.
 * The internal connection cache now remembers failed connect attempts and re-raises the original exception on subsequent accesses within the same process, instead of re-dialing Snowflake every time a command accesses the shared connection. This fixes, among other cases, the customer-visible duplicate `LOGIN_HISTORY` events (and `OVERFLOW_FAILURE_EVENTS_ELIDED`) previously emitted when a `snow` invocation was rejected by an authentication policy.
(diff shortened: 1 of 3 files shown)


sfc-gh-olorek:
Redundant: https://github.com/snowflakedb/snowflake-cli/pull/2930 fixed the issue

PR #2977, opened 2026-05-07
fix: [SNOW-2361367] notebook/streamlit deploy without CREATE STAGE privilege

### LINKED ISSUE #2627: SNOW-2361367: Notebook Deploy is unable to deploy without CREATE STAGE permission
### SnowCLI version

3.13.0.dev0

### Python version

Python 3.10.16

### Platform

Linux-6.14.0-29-generic-x86_64-with-glibc2.39

### What happened

**Issue Experienced**
When attempting to deploy a notebook , it will always run `CREATE STAGE IF NOT EXISTS stagename` even if the stage already exists. This causes notebook deploy to fail, even if the role has privileges to use the specified stage.

**Expected Behaviour**
I would expect the entity to explicitly check if the stage exists, first, before attempting `CREATE STAGE IF NOT EXISTS`, which will fail even if the stage exists, if the user does not have the privilege

In a way, running `CREATE STAGE IF NOT EXISTS` is convenient - if it does not fail, the deployer is guaranteed to be able to write the stage, as they will have ownership on it. But I believe it should be allowed to use an existing stage that the deployer has privilege to use already, and any missing write privileges should be raised up naturally

**Workaround**

I have to `GRANT CREATE STAGE ON SCHEMA...` to the deploy role. I would prefer to not do this.

**Notes**

As this is stemming from [src/snowflake/cli/api/entities/utils.py : sync_deploy_root_with_stage()](https://github.com/snowflakedb/snowflake-cli/blob/8d3b53d518e58013351fe08377b675ee80182d69/src/snowflake/cli/api/entities/utils.py#L129) method, this is probably affecting streamlit deploys as well.

In that method, there's a conditional:

```python
    if stage_path_parts.is_vstage:
        # vstages are created by FBE, so no need to do it manually
        pass
    elif not package_name:
        # default entity stage creation behaviour
        (code)
    else:
        # fall through behaviour to cater for native apps with sql_facade
        (code)
```

I wonder if it might be better restructuring that to

```python
    if stage_path_parts.is_vstage:
        # vstages are created by FBE, so no need to do it manually
        pass
    elif package_name:
        #  cater for native apps with sql_facade
        (code)
    else:
        # default entity stage creation behaviour
        (code)
```

and let the default entity behaviour incorporate a check to see if the stage already exists (perhaps similar to how it is done in sql_facade)

### Console output

```Shell

```

### How to reproduce

_No response_

sfc-gh-olorek:
### Pre-review checklist

### Changes description

Entity artifact deploys that are not backed by a native-app package — `snow notebook deploy`, `snow streamlit deploy`, `snow dcm deploy`, and anything else that goes through `sync_deploy_root_with_stage` with `package_name=None` — always issued `CREATE STAGE IF NOT EXISTS <stage>` before uploading files. That statement requires `CREATE STAGE` on the parent schema even when it's a no-op, so a role that had only `USAGE`/`READ`/`WRITE` on an already-provisioned stage could not deploy without also being granted `CREATE STAGE`.

This change mirrors the existing native-app path:

1. Adds `StageManager.stage_exists(fqn)` that runs `show stages like …` and returns `False` on `DOES_NOT_EXIST_OR_CANNOT_BE_PERFORMED` or `INSUFFICIENT_PRIVILEGES` (so callers that genuinely can't see the stage still fall through to `create`).
2. Teaches `sync_deploy_root_with_stage`'s generic branch (`not package_name`) to call `stage_exists` first and skip `create` when the stage is already visible. Temporary stages keep the old behavior — always create — because a prior session's temp stage isn't visible anyway.

### Tests
* New unit tests in `tests/stage/test_stage.py` cover the new `stage_exists` helper (SHOW LIKE patterning with/without schema/database, quoted identifiers, underscore escaping, and swallowing of `DOES_NOT_EXIST_OR_CANNOT_BE_PERFORMED` / `INSUFFICIENT_PRIVILEGES`).
* New unit tests in `tests/nativeapp/test_manager.py` cover the generic branch of `sync_deploy_root_with_stage`: `create` is skipped when the stage exists, called when it doesn't, and always called (no pre-check) for temporary stages.
* Notebook snapshot tests updated to reflect the new SHOW LIKE pre-check (the conftest mock cursor returns a row, so `create` is correctly skipped — this is exactly the bug-fix path).
* Streamlit test fixtures patched to return `stage_exists=False` so existing `mock_create_stage.assert_called_once()` assertions continue to hold.

### References
* Fixes #2627
* Jira: SNOW-2361367

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -26,6 +26,7 @@
 * Fixed `SELECT *` output being corrupted when joined tables share column names. Duplicate column names are now disambiguated by appending a numeric suffix (e.g. `NAME`, `NAME_2`).
 * Fixed `snow connection generate-jwt` and `snow connection generate-workload-identity-token` failing with `Connection None is not configured` when used with `--temporary-connection`.
 * The internal connection cache now remembers failed connect attempts and re-raises the original exception on subsequent accesses within the same process, instead of re-dialing Snowflake every time a command accesses the shared connection. This fixes, among other cases, the customer-visible duplicate `LOGIN_HISTORY` events (and `OVERFLOW_FAILURE_EVENTS_ELIDED`) previously emitted when a `snow` invocation was rejected by an authentication policy.
+* Entity deploys (`snow notebook deploy`, `snow streamlit deploy`, `snow dcm deploy`, and other commands that sync artifacts to a stage) now check whether the target stage already exists before issuing `CREATE STAGE IF NOT EXISTS`. This means a role that holds only `USAGE`/`READ`/`WRITE` on a pre-existing stage can deploy without also being granted `CREATE STAGE` on the schema. In particular, `snow notebook deploy` no longer fails for roles that were previously forced to hold `CREATE STAGE` purely to satisfy the no-op `IF NOT EXISTS` path.
 
 
 # v3.17.0
--- a/src/snowflake/cli/_plugins/stage/manager.py
+++ b/src/snowflake/cli/_plugins/stage/manager.py
@@ -40,9 +40,18 @@ from snowflake.cli.api.commands.common import (
 )
 from snowflake.cli.api.console import cli_console
 from snowflake.cli.api.constants import PYTHON_3_12
+from snowflake.cli.api.errno import (
+    DOES_NOT_EXIST_OR_CANNOT_BE_PERFORMED,
+    INSUFFICIENT_PRIVILEGES,
+)
 from snowflake.cli.api.exceptions import CliError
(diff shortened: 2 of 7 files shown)


sfc-gh-olorek:
Closing this stale draft. Internal development moved to the private repository in July 2026 and PR head branches are not mirrored outbound, so this draft cannot land here — anything still wanted has to be re-opened there. The work itself is still tracked in SNOW-2361367, which stays the source of truth. The branch is preserved.

PR #2999, opened 2026-05-08
fix: [SNOW-1528909] decode SQL files by BOM before assuming UTF-8

### LINKED ISSUE #1303: SNOW-1528909: Snowflake CLI cannot handle UTF-16LE encoded text files
### SnowCLI version

2.6.0rc0

### Python version

Python 3.11.9

### Platform

macOS-14.5-arm64-arm-64bit

### What happened

Powershell redirects (e.g. command > file) by default encode output using UTF-16LE. Unfortunately, Snowflake CLI in a lot of paths is assuming utf-8 encoding, which makes common workflows fail there. Here's an example PR that simply changes the input for a snow sql -f command to use that encoding, showing the failure: #1299 

### Console output

```Shell
src/snowflake/cli/api/commands/snow_typer.py:96: in command_callable_decorator
    result = command_callable(*args, **kw)
src/snowflake/cli/api/commands/decorators.py:158: in wrapper
    return func(**options)
src/snowflake/cli/api/commands/decorators.py:158: in wrapper
    return func(**options)
src/snowflake/cli/plugins/sql/commands.py:82: in execute_sql
    single_statement, cursors = SqlManager().execute(query, files, std_in, data=data)
src/snowflake/cli/plugins/sql/manager.py:60: in execute
    query_from_file = SecurePath(file).read_text(
src/snowflake/cli/api/secure_path.py:157: in read_text
    return self._path.read_text(*args, **kwargs)
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/pathlib.py:1059: in read_text
    return f.read()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <encodings.utf_8.IncrementalDecoder object at 0x7f7815a95f90>
input = b'\xff\xfe/\x00*\x00\n\x00 \x00C\x00o\x00p\x00y\x00r\x00i\x00g\x00h\x00t\x00 \x00(\x00c\x00)\x00 \x002\x000\x002\x004\...00e\x00c\x00t\x00 \x00r\x00o\x00u\x00n\x00d\x00(\x00l\x00n\x00(\x001\x000\x000\x00)\x00,\x00 \x004\x00)\x00;\x00\n\x00'
final = True

>   ???
E   UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte

<frozen codecs>:322: UnicodeDecodeError
```
```


### How to reproduce

1. Encode a file using UTF-16LE
2. Use it as `snowflake.yml`, as a post-deploy hook, or as an input to `snow sql -f`
3. Observe a utf-8 codec error

sfc-gh-olorek:
### Pre-review checklist

### Changes description

Fixes https://github.com/snowflakedb/snowflake-cli/issues/1303 (SNOW-1528909).

`snow sql -f <file>` and the `!source <file>` include directive called `path.read_text()` / `path.open("r")` without an explicit encoding, so the decoder defaulted to whatever the platform locale happened to be. Two real-world inputs then broke on Windows:

- PowerShell's `>` redirect writes UTF-16 LE with a BOM. Reading it as UTF-8 raises `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0` before the first statement ever reaches Snowflake — the exact stack trace in the issue.
- A UTF-8 file with a BOM (common from Notepad, Visual Studio, many CI setups) decodes successfully but leaves a stray U+FEFF as the first character of the first statement, which Snowflake rejects as a syntax error.

This PR introduces `_read_sql_file(path)` in `src/snowflake/cli/_plugins/sql/statement_reader.py`:

- Opens the file in binary mode.
- Checks for a UTF-32 / UTF-16 / UTF-8 BOM in descending length order (UTF-32 BOMs start with the same two bytes as UTF-16 BOMs, so length order matters).
- Decodes with the matching codec: `utf-32`, `utf-16`, or `utf-8-sig`. Python's `utf-16` / `utf-32` codecs consume the BOM themselves, and `utf-8-sig` strips the UTF-8 BOM — so the returned text never contains a leading U+FEFF.
- Files without a BOM fall back to `utf-8`, matching how the CLI writes files elsewhere (`SecurePath.write_text`) and keeping the common case on the fast path.

Both entry points that previously read SQL files — `files_reader` for `snow sql -f` and `ParsedStatement.from_file` for `!source` — now go through this helper, so the behaviour is consistent between the two.

Note on approach: the issue thread discussed pulling in `charset_normalizer` for broader sniffing, but a follow-up comment (from a member of the SnowCLI team) preferred staying with BOM-only detection plus UTF-8 fallback. This change implements exactly that — no new dependencies, no heuristics that could guess the wrong encoding for a plain UTF-8 file that happens to contain ambiguous byte sequences.

### Testing

Added 13 regression tests to `tests/sql/test_statement_reader.py`:

- `_read_sql_file` correctly decodes UTF-8-sig, UTF-16 (with BOM from the codec), UTF-16-LE/BE (manually prepended BOM — what PowerShell and Notepad actually write), UTF-32, and UTF-32-LE/BE.
- `files_reader` (the `snow sql -f` path) and the `!source` directive (`ParsedStatement.from_file`) each round-trip BOM-prefixed files.
- Regression guards that plain UTF-8 files still decode on the fast path and that a UTF-8 BOM never leaks into the first statement.

Also manually verified the reproducer from the issue — a file written by `"/* copyright */\nselect round(ln(100), 4);".encode("utf-16")` (what PowerShell `>` produces) — now decodes and parses cleanly instead of crashing with `UnicodeDecodeError`.

Full `tests/sql/` suite: **225/225 passing locally**.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -26,6 +26,7 @@
 * Fixed `SELECT *` output being corrupted when joined tables share column names. Duplicate column names are now disambiguated by appending a numeric suffix (e.g. `NAME`, `NAME_2`).
 * Fixed `snow connection generate-jwt` and `snow connection generate-workload-identity-token` failing with `Connection None is not configured` when used with `--temporary-connection`.
 * The internal connection cache now remembers failed connect attempts and re-raises the original exception on subsequent accesses within the same process, instead of re-dialing Snowflake every time a command accesses the shared connection. This fixes, among other cases, the customer-visible duplicate `LOGIN_HISTORY` events (and `OVERFLOW_FAILURE_EVENTS_ELIDED`) previously emitted when a `snow` invocation was rejected by an authentication policy.
+* `snow sql -f <file>` and the `!source <file>` include directive now honor a UTF-8 / UTF-16 / UTF-32 byte-order mark at the start of a SQL file and decode it using the matching encoding. This fixes `snow sql` on Windows when the input was produced by a PowerShell `>` redirect (which writes UTF-16 LE with a BOM) or by an editor that prepends a UTF-8 BOM; previously these files crashed with `UnicodeDecodeError` or leaked U+FEFF into the first statement.
 
 
 # v3.17.0
--- a/src/snowflake/cli/_plugins/sql/statement_reader.py
+++ b/src/snowflake/cli/_plugins/sql/statement_reader.py
@@ -1,3 +1,4 @@
+import codecs
 import enum
 import io
 import re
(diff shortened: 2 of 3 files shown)


sfc-gh-olorek:
Closing this draft: SNOW-1528909 is already Done, so this change is superseded. Internal development also moved to the private repository in July 2026 — PR head branches are not mirrored, so a draft here cannot land. The branch is preserved if any of this is worth revisiting.

PR #3023, opened 2026-05-12
fix: correct app_flow telemetry for all snow app code paths

sfc-gh-mborins:
The _app_group_callback set app_flow='snowflake_app' before the command body runs, causing the executing_command telemetry event to report an incorrect app_flow for native-app projects. The result_executing_command event was correct because routing decorators (with_app_flow_routing, native_app_only) set the real value inside the command callable.

This PR:

1. Removes the premature default from _app_group_callback (aligning with the documented contract in telemetry.py lines 94-99: app_flow should NOT appear on executing_command because the routing decorators haven't resolved it yet).

2. Adds explicit `set_app_flow(AppFlow.SNOWFLAKE_APP)` to `snow app setup`, which is Snowflake-Apps-only but has no routing decorator.

3. Adds `TestAppFlowTelemetryAttribution`, a comprehensive test class that verifies all major routing paths stamp the correct app_flow on result/error events while keeping it absent from executing_command events.

### Pre-review checklist

### Changes description
- Remove premature app_flow default from _app_group_callback
- Add set_app_flow(AppFlow.SNOWFLAKE_APP) to snow app setup command
- Add TestAppFlowTelemetryAttribution test class covering native_app and snowflake_app paths

### DIFF
--- a/tests/app/test_telemetry.py
+++ b/tests/app/test_telemetry.py
@@ -407,16 +407,118 @@ def test_app_flow_native_app_in_telemetry_data(mock_conn, project_directory, run
     with project_directory("napp_post_deploy_missing_file"):
         runner.invoke(["app", "run"], catch_exceptions=False)
 
-    error_command_event = (
-        mock_conn.return_value._telemetry.try_add_log_to_batch.call_args_list[  # noqa: SLF001
-            1
-        ]
-        .args[0]
-        .to_dict()
-    )
+    events = _get_telemetry_events(mock_conn)
+    _assert_executing_event_has_no_app_flow(events)
+    _assert_result_or_error_event_has_app_flow(events, "native_app")
 
-    assert error_command_event["message"]["type"] == "error_executing_command"
-    assert error_command_event["message"]["app_flow"] == "native_app"
+
+# ── Comprehensive app_flow telemetry attribution tests ─────────────────
+
+
+def _get_telemetry_events(mock_conn):
+    """Extract all telemetry event dicts from the mock connector."""
+    telemetry = mock_conn.return_value._telemetry  # noqa: SLF001
+    return [
+        call.args[0].to_dict() for call in telemetry.try_add_log_to_batch.call_args_list
+    ]
+
+
+def _assert_executing_event_has_no_app_flow(events):
+    """The executing_command event fires before routing decorators run,
+    so app_flow must be absent (matching the contract in telemetry.py)."""
+    executing = [e for e in events if e["message"]["type"] == "executing_command"]
+    assert executing, "Expected at least one executing_command event"
+    for event in executing:
+        assert (
+            "app_flow" not in event["message"]
+        ), f"executing_command must not contain app_flow, got: {event['message']}"
+
+
+def _assert_result_or_error_event_has_app_flow(events, expected_flow):
+    """The result or error event (whichever fires) must contain the
+    expected app_flow value."""
+    post_events = [
+        e
+        for e in events
+        if e["message"]["type"]
+        in ("result_executing_command", "error_executing_command")
+    ]
+    assert post_events, "Expected a result or error telemetry event"
+    for event in post_events:
+        assert event["message"].get("app_flow") == expected_flow, (
+            f"Expected app_flow={expected_flow!r} on {event['message']['type']}, "
+            f"got: {event['message'].get('app_flow')!r}"
+        )
+
+
+class TestAppFlowTelemetryAttribution:
+    """Verify that all snow app code paths stamp the correct app_flow on
+    result/error telemetry events, and that the executing_command event
+    never contains app_flow (since routing resolves after pre_execute)."""
+
+    @mock.patch("snowflake.connector.connect")
+    def test_native_app_bundle(self, mock_conn, project_directory, runner):
+        """with_app_flow_routing detects native_app from v1 project."""
+        with project_directory("napp_post_deploy_missing_file"):
+            runner.invoke(["app", "bundle"], catch_exceptions=False)
+
+        events = _get_telemetry_events(mock_conn)
+        _assert_executing_event_has_no_app_flow(events)
+        _assert_result_or_error_event_has_app_flow(events, "native_app")
+
+    @mock.patch("snowflake.connector.connect")
+    def test_snowflake_app_bundle(self, mock_conn, project_directory, runner):
+        """with_app_flow_routing detects snowflake_app from entity type."""
+        with project_directory("snowflake_app_v2"):
+            runner.invoke(["app", "bundle"], catch_exceptions=False)
+
+        events = _get_telemetry_events(mock_conn)
+        _assert_executing_event_has_no_app_flow(events)
+        _assert_result_or_error_event_has_app_flow(events, "snowflake_app")
+
+    @mock.patch("snowflake.connector.connect")
+    def test_snowflake_app_deploy(self, mock_conn, project_directory, runner):
+        """with_app_flow_routing detects snowflake_app for deploy.
+        The command will fail trying to execute SQL (the mock connector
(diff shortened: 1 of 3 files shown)


sfc-gh-gbloom (commented):
Good catch! I agree that setting the default `app_flow` in `_app_group_callback` is not correct, but I suspect that simply removing it is not enough and might lead to snowflake-app only paths (like `snow app setup`) not being attributed correctly.

Let's add a unit test that verifies all native-apps/snowflake-apps code paths are stamped correctly.

sfc-gh-mborins:
Thanks for the review! You were right that simply removing the default wasn't sufficient.

I've pushed a follow-up commit that:

1. **Fixes `snow app setup`**: Added explicit `set_app_flow(AppFlow.SNOWFLAKE_APP)` in the `app_setup` function body, since this command has no routing decorator and is Snowflake-Apps-only.

2. **Adds `TestAppFlowTelemetryAttribution`**: A comprehensive test class in `tests/app/test_telemetry.py` that verifies all major code paths:
   - `native_app_only` decorator (via `app run`)
   - `with_app_flow_routing` on Native App projects (via `app bundle`, `app deploy`)
   - `with_app_flow_routing` on Snowflake Apps projects (via `app bundle`, `app deploy`)
   - Direct `set_app_flow` in `app setup`

   Each test asserts both sides of the contract:
   - `executing_command` event does NOT contain `app_flow`
   - `result/error_executing_command` event DOES contain the correct value

I verified the test catches the regression: removing `set_app_flow` from `app_setup` fails with `AssertionError: Expected app_flow='snowflake_app' on result_executing_command, got: None`.

*This response was authored with Cortex Code* 🤖

tests/app/test_telemetry.py
```diff
@@ -419,6 +419,158 @@ def test_app_flow_native_app_in_telemetry_data(mock_conn, project_directory, run
...
+    artifacts:
+      - src: "*"
+        dest: ./
+    manifest: ./manifest.yml
+  native_app:
+    type: application
+    identifier: my_app
+    from:
+      target: pkg
+"""
```
sfc-gh-olorek: `_NATIVE_APP_YML` is defined but never referenced — the native-app tests below use the `project_directory('napp_post_deploy_missing_file')` fixture instead. Safe to delete.

tests/app/test_telemetry.py
```diff
@@ -419,6 +419,158 @@ def test_app_flow_native_app_in_telemetry_data(mock_conn, project_directory, run
...
+        events = _get_telemetry_events(mock_conn)
+        _assert_executing_event_has_no_app_flow(events)
+        _assert_result_or_error_event_has_app_flow(events, "native_app")
+
+    @mock.patch("snowflake.connector.connect")
+    def test_snowflake_app_bundle(self, mock_conn, tmp_path, runner):
+        """with_app_flow_routing detects snowflake_app from entity type."""
+        (tmp_path / "snowflake.yml").write_text(_SNOWFLAKE_APP_YML)
+
+        with mock.patch("os.getcwd", return_value=str(tmp_path)):
```
sfc-gh-olorek: `mock.patch('os.getcwd', ...)` is fragile — it only covers call sites that explicitly call `os.getcwd()` (e.g. `definition_manager.py:52`), and misses `Path.cwd()`, subprocesses, or absolute-path logic that could resolve against the real cwd. The rest of this file uses the `project_directory` fixture (or the underlying `temporary_directory` fixture from conftest.py), which actually `chdir`s into the directory. Consider adding a `snowflake_app_v2` fixture under `tests/test_data/projects/` so these three tests (`test_snowflake_app_bundle`, `test_snowflake_app_deploy`, `test_snowflake_app_setup`) can use the same pattern as their native-app siblings.

tests/app/test_telemetry.py
```diff
@@ -419,6 +419,158 @@ def test_app_flow_native_app_in_telemetry_data(mock_conn, project_directory, run
...
+        """with_app_flow_routing detects snowflake_app for deploy.
+        The command will fail trying to execute SQL, but the error event
+        still carries app_flow because routing resolves before the SQL call."""
+        (tmp_path / "snowflake.yml").write_text(_SNOWFLAKE_APP_YML)
+
+        with mock.patch("os.getcwd", return_value=str(tmp_path)):
+            try:
+                runner.invoke(["app", "deploy"])
+            except (ValueError, Exception):
+                pass  # Expected: SQL execution fails with mocked connector
```
sfc-gh-olorek: The try/except here is unnecessary and a bit misleading:

- `runner.invoke()` captures exceptions and returns a `Result` by default — it won't propagate unless you pass `catch_exceptions=False`.
- `(ValueError, Exception)` is redundant because `Exception` already covers `ValueError`.

Since the other tests in this class just call `runner.invoke([...], catch_exceptions=False)` and let the assertion layer validate the event shape, this one can match:

```python
runner.invoke(["app", "deploy"], catch_exceptions=False)
```

If the SQL step really does raise something `runner.invoke` won't catch, keep the try/except but narrow it to that specific exception type and add a comment explaining which call is expected to raise.

tests/app/test_telemetry.py
```diff
@@ -419,6 +419,158 @@ def test_app_flow_native_app_in_telemetry_data(mock_conn, project_directory, run
...
+
+    @mock.patch("snowflake.connector.connect")
+    def test_native_app_run(self, mock_conn, project_directory, runner):
+        """native_app_only decorator stamps native_app."""
+        with project_directory("napp_post_deploy_missing_file"):
+            runner.invoke(["app", "run"], catch_exceptions=False)
+
+        events = _get_telemetry_events(mock_conn)
+        _assert_executing_event_has_no_app_flow(events)
+        _assert_result_or_error_event_has_app_flow(events, "native_app")
```
sfc-gh-olorek: This duplicates `test_app_flow_native_app_in_telemetry_data` (line 394 above): same command (`snow app run`), same fixture (`napp_post_deploy_missing_file`), same expected flow (`native_app`). The one new thing this test asserts is that `executing_command` has no `app_flow` — which is worth keeping.

Suggest either folding that assertion into the existing test and deleting this one, or deleting the older test now that this class has broader coverage. Two tests for the same path will drift apart over time.

tests/app/test_telemetry.py
sfc-gh-mborins: Done, removed. Thanks for catching that.

*This response was authored with Cortex Code* 🤖

tests/app/test_telemetry.py
sfc-gh-mborins: Good call. Replaced all `mock.patch('os.getcwd', ...)` with `project_directory('snowflake_app_v2')` using a new fixture at `tests/test_data/projects/snowflake_app_v2/snowflake.yml`. Now matches the pattern used by the native-app tests.

*This response was authored with Cortex Code* 🤖

tests/app/test_telemetry.py
sfc-gh-mborins: Agreed on the redundancy. Narrowed to `except ValueError` with a comment explaining why it's needed: the `SnowCLIRunner` forces `catch_exceptions=False` (line 243 in conftest.py), so the ValueError from the mocked SQL execution propagates up. Telemetry is still logged before the re-raise (via the `BaseException` handler in `snow_typer.py:138`).

*This response was authored with Cortex Code* 🤖

tests/app/test_telemetry.py
sfc-gh-mborins: Agreed. Removed `test_native_app_run` from the class and folded the `_assert_executing_event_has_no_app_flow` assertion into the existing `test_app_flow_native_app_in_telemetry_data` test. One test per path, no drift.

*This response was authored with Cortex Code* 🤖

sfc-gh-olorek (approved):
lgtm

PR #3025, opened 2026-05-13
chore: auto-resolve RELEASE-NOTES.md conflicts with merge=union + dup check

sfc-gh-olorek:
### Pre-review checklist

### Changes description

Contributors keep hitting merge conflicts on `RELEASE-NOTES.md` — almost always because two PRs each appended a bullet under the same section. This PR does two things that together auto-resolve that class of conflict without changing the contributor workflow.

**1. `RELEASE-NOTES.md merge=union` in `.gitattributes`**

Git's built-in union merge driver concatenates both sides' additions instead of producing a conflict. Works for local `git merge` / `git rebase` / `git pull` and for GitHub's "Merge pull request" button. Takes effect on any merge that runs after the attribute is checked out on the merging branch — so in-flight PRs get the benefit on their next `git merge origin/main` and beyond.

**2. `check-release-notes-duplicates` pre-commit hook**

Union merge has one silent failure mode: when a release cut moves bullets from "Unreleased version" into a new `# vX.Y.Z` section, an open PR that still has those bullets under Unreleased will end up with the same bullet in two sections after merging main. Union merge will not flag it.

The new stdlib-only script (`scripts/check_release_notes_duplicates.py`) parses `RELEASE-NOTES.md`, extracts bullets per section, and fails if any normalized bullet text appears in both "Unreleased version" and a released section, or is duplicated within Unreleased. It runs as a local pre-commit hook scoped to `RELEASE-NOTES.md`, so it only fires when that file is in the changeset.

The existing `lint.yaml` workflow already runs `pre-commit run --all-files`, so no CI config change is needed — the new hook runs in PR CI automatically.

**Why not towncrier / custom merge driver**

- Towncrier (fragment files like pytest/pip use) is the idiomatic long-term answer but changes every contributor's PR flow; worth a separate discussion.
- A custom merge driver would handle the release-cut case but only runs where installed client-side, and GitHub's support for custom drivers on the web merge is spotty.
- `merge=union` + a small CI trip-wire is ~60 lines total, keeps the "just edit RELEASE-NOTES.md" workflow, and closes the main conflict problem ~99%.

**Release notes**

Skipping a release-notes bullet intentionally: this change is dev-infra only with no user-visible behavior change.

### Files changed

- `.gitattributes` — one line enabling union merge for RELEASE-NOTES.md
- `scripts/check_release_notes_duplicates.py` — new stdlib-only detection script
- `.pre-commit-config.yaml` — registers the new hook in the existing `- repo: local` block
- `tests/test_check_release_notes_duplicates.py` — 7 unit tests
- `CONTRIBUTING.md` — new "Release notes" section explaining the workflow and the one edge case

### Testing

- `python scripts/check_release_notes_duplicates.py` against the current file exits 0.
- `hatch run test tests/test_check_release_notes_duplicates.py` — 7/7 pass.
- `pre-commit run check-release-notes-duplicates --files RELEASE-NOTES.md` passes.
- Manually verified the three error cases (cross-section duplicate, within-Unreleased duplicate, whitespace-normalized duplicate) all produce exit 1 with the offending bullet named.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -89,7 +89,7 @@ repos:
       - id: check-release-notes-duplicates
         name: "Check RELEASE-NOTES.md for duplicated bullets"
         language: system
-        entry: python scripts/check_release_notes_duplicates.py
+        entry: python3 scripts/check_release_notes_duplicates.py
         pass_filenames: false
         files: ^RELEASE-NOTES\.md$
   - repo: https://github.com/Lucas-C/pre-commit-hooks
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -86,6 +86,12 @@ repos:
           ^import requirements.*
         pass_filenames: true
         exclude: ^src/snowflake/cli/_plugins/snowpark/models.py$
+      - id: check-release-notes-duplicates
+        name: "Check RELEASE-NOTES.md for duplicated bullets"
+        language: system
+        entry: python scripts/check_release_notes_duplicates.py
+        pass_filenames: false
+        files: ^RELEASE-NOTES\.md$
   - repo: https://github.com/Lucas-C/pre-commit-hooks
     rev: v1.5.5
     hooks:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -57,10 +57,6 @@ Please keep in mind that you need these python versions available in your `$PATH
 
 Every user-visible change should get one line in `RELEASE-NOTES.md` under `# Unreleased version`, in the appropriate subsection (`## Backward incompatibility`, `## Deprecations`, `## New additions`, or `## Fixes and improvements`).
 
-Conflicts between PRs that each append a bullet are resolved automatically by the `merge=union` attribute in `.gitattributes` — no manual resolution needed.
-
-The one case the auto-merge cannot handle is a **release cut overlap**: if your branch added bullets under "Unreleased version" and `main` has since moved those bullets into a new `# vX.Y.Z` section, merging `main` can leave the same bullet in both sections. The `check-release-notes-duplicates` pre-commit hook flags this; fix it by removing the duplicated bullet from "Unreleased version" (it's already documented under the released version).
-
 ## Unit tests
 
 Unit tests are executed in random order. If tests fail after your change, you can re-execute them in the same order using `pytest --randomly-seed=<number>`, where number is a seed printed at the beginning of the test execution output.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,6 +53,14 @@ hatch env show
 
 Please keep in mind that you need these python versions available in your `$PATH`. You can install them using `hatch` or other tool like [pyenv](https://github.com/pyenv/pyenv)
 
+## Release notes
+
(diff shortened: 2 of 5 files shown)


sfc-gh-jwilkowski (changes requested):
The union-merge + dup-check approach is the right call — no new dependencies, no contributor workflow change, and the CI path works without any config changes.

One issue to fix before merging:

`entry: python scripts/check_release_notes_duplicates.py` uses bare `python`, but on Linux and many CI environments only `python3` is in `PATH` (`python` may be absent entirely). Both existing `language: system` hooks in this file use `hatch` as the entry — this is the only one using a bare interpreter name. The hook would silently fail with `FileNotFoundError` in environments where `python` is not found, defeating the duplicate check precisely where it matters most. Change to `python3`.

.pre-commit-config.yaml
```diff
@@ -86,6 +86,12 @@ repos:
           ^import requirements.*
         pass_filenames: true
         exclude: ^src/snowflake/cli/_plugins/snowpark/models.py$
+      - id: check-release-notes-duplicates
+        name: "Check RELEASE-NOTES.md for duplicated bullets"
+        language: system
```
sfc-gh-jwilkowski, line 91: [WORTH FIXING] `entry: python scripts/check_release_notes_duplicates.py` uses bare `python`. On Linux and many CI environments only `python3` is in `PATH` — `python` may not exist at all, causing the hook to silently fail with `FileNotFoundError` instead of blocking the commit. Both existing `language: system` hooks in this file (`lock-dependencies`, `dependencies-sync`) use `hatch` as their entry; this is the only one relying on a bare interpreter name.

Change to:
```yaml
entry: python3 scripts/check_release_notes_duplicates.py
```

sfc-gh-jwilkowski: or follow similar path as `lock-dependencies` hook does

sfc-gh-olorek: Good catch — switched to `python3`. I went with that over the `hatch run` path since this script is intentionally stdlib-only, so routing through a hatch env would add resolution overhead for no gain. Pushed as c872e01f.

CONTRIBUTING.md
```diff
@@ -53,6 +53,14 @@ hatch env show
...
 Please keep in mind that you need these python versions available in your `$PATH`. You can install them using `hatch` or other tool like [pyenv](https://github.com/pyenv/pyenv)
 
+## Release notes
+
+Every user-visible change should get one line in `RELEASE-NOTES.md` under `# Unreleased version`, in the appropriate subsection (`## Backward incompatibility`, `## Deprecations`, `## New additions`, or `## Fixes and improvements`).
+
+Conflicts between PRs that each append a bullet are resolved automatically by the `merge=union` attribute in `.gitattributes` — no manual resolution needed.
+
+The one case the auto-merge cannot handle is a **release cut overlap**: if your branch added bullets under "Unreleased version" and `main` has since moved those bullets into a new `# vX.Y.Z` section, merging `main` can leave the same bullet in both sections. The `check-release-notes-duplicates` pre-commit hook flags this; fix it by removing the duplicated bullet from "Unreleased version" (it's already documented under the released version).
+
```
sfc-gh-jwilkowski: Are those implementation details really worth describing here? I'd keep just the first sentence as this is our expectation and rest is imo not needed

sfc-gh-olorek: Trimmed to just the first sentence — pushed as b3286194.

sfc-gh-jwilkowski (approved):

PR #3031, opened 2026-05-18
Jw/snow 3511654 revamp contributor guidelines

sfc-gh-jwilkowski:
### Pre-review checklist

### Changes description
https://snowflakecomputing.atlassian.net/browse/SNOW-3511654

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,13 +1,12 @@
 ### Pre-review checklist
-   * [ ] I've confirmed that instructions included in README.md are still correct after my changes in the codebase.
-   * [ ] I've added or updated automated unit tests to verify correctness of my new code.
-   * [ ] I've added or updated integration tests to verify correctness of my new code.
-   * [ ] I've confirmed that my changes are working by executing CLI's commands manually on MacOS.
-   * [ ] I've confirmed that my changes are working by executing CLI's commands manually on Windows.
-   * [ ] I've confirmed that my changes are up-to-date with the target branch.
-   * [ ] I've described my changes in the release notes.
-   * [ ] I've described my changes in the section below.
-   * [ ] I've described my changes in the documentation.
+- [ ] I've confirmed that instructions included in README.md are still correct after my changes in the codebase.
+- [ ] I've added or updated unit tests to verify correctness of my new code.
+- [ ] I've added or updated integration tests to verify correctness of my new code.
+- [ ] I've confirmed my changes are working by executing CLI commands manually on macOS.
+- [ ] I've confirmed my changes are working by executing CLI commands manually on Windows.
+- [ ] I've confirmed my changes are up-to-date with the target branch.
+- [ ] I've described my changes in `RELEASE-NOTES.md` (see [when and how](docs/contributing/lifecycle.md#release-notes)).
+- [ ] I've updated documentation if behavior changed.
 
 ### Changes description
 ...
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,7 +5,7 @@
    * [ ] Manually tested on macOS
    * [ ] Manually tested on Windows
    * [ ] I've confirmed my changes are up-to-date with the target branch.
-   * [ ] I've described my changes in `RELEASE-NOTES.md` (see [when and how](docs/contributing/lifecycle.md#release-notes)).
+   * [ ] I've described my changes in `RELEASE-NOTES.md` (see [when and how](docs/contributing/lifecycle.md#release-notes-format)).
    * [ ] I've updated documentation if behavior changed.
 
 ### Changes description
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,8 +2,8 @@
 - [ ] I've confirmed that instructions included in README.md are still correct after my changes in the codebase.
 - [ ] I've added or updated unit tests to verify correctness of my new code.
 - [ ] I've added or updated integration tests to verify correctness of my new code.
-- [ ] I've confirmed my changes are working by executing CLI commands manually on macOS.
-- [ ] I've confirmed my changes are working by executing CLI commands manually on Windows.
+- [ ] Manually tested on macOS
+- [ ] Manually tested on Windows
 - [ ] I've confirmed my changes are up-to-date with the target branch.
 - [ ] I've described my changes in `RELEASE-NOTES.md` (see [when and how](docs/contributing/lifecycle.md#release-notes)).
 - [ ] I've updated documentation if behavior changed.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,12 +1,12 @@
 ### Pre-review checklist
-- [ ] I've confirmed that instructions included in README.md are still correct after my changes in the codebase.
-- [ ] I've added or updated unit tests to verify correctness of my new code.
-- [ ] I've added or updated integration tests to verify correctness of my new code.
-- [ ] Manually tested on macOS
-- [ ] Manually tested on Windows
-- [ ] I've confirmed my changes are up-to-date with the target branch.
-- [ ] I've described my changes in `RELEASE-NOTES.md` (see [when and how](docs/contributing/lifecycle.md#release-notes)).
-- [ ] I've updated documentation if behavior changed.
+   * [ ] I've confirmed that instructions included in README.md are still correct after my changes in the codebase.
+   * [ ] I've added or updated unit tests to verify correctness of my new code.
+   * [ ] I've added or updated integration tests to verify correctness of my new code.
+   * [ ] Manually tested on macOS
+   * [ ] Manually tested on Windows
+   * [ ] I've confirmed my changes are up-to-date with the target branch.
+   * [ ] I've described my changes in `RELEASE-NOTES.md` (see [when and how](docs/contributing/lifecycle.md#release-notes)).
+   * [ ] I've updated documentation if behavior changed.
 
 ### Changes description
 ...
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,7 +5,7 @@
    * [ ] Manually tested on macOS
    * [ ] Manually tested on Windows
    * [ ] I've confirmed my changes are up-to-date with the target branch.
-   * [ ] I've described my changes in `RELEASE-NOTES.md` (see [when and how](docs/contributing/lifecycle.md#release-notes-format)).
+   * [ ] I've described my changes in `RELEASE-NOTES.md` (see [when and how](https://github.com/snowflakedb/snowflake-cli/blob/main/docs/contributing/lifecycle.md#release-notes-format)).
    * [ ] I've updated documentation if behavior changed.
 
 ### Changes description
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,153 +16,29 @@
 
 # Contributing to Snowflake CLI
 
-There are two ways to contribute code to the repository: directly or by use of forks. For best practices for the second approach, refer to the section on forks below. Right now, there is limited access to contributing to the repository directly, and hence using forks is the recommended approach.
+## Quick start
 
-## Setup a development environment
-If you are interested in contributing, you will need to instantiate dev virtual environments and the pre-commit logic to help with formatting and linting of commits.
-
-We use [hatch](https://hatch.pypa.io/latest/) to manage and created development environments.
-Default environment will use the python version in your shell.
 ```bash
 pip install -U hatch==1.15.1 virtualenv==20.39.1
-hatch run pre-commit
-```
-This will spawn new shell with environment and all required packages installed.
-This will also install snowflake cli package in editable mode.
-
-To enter environment use following command.
-```bash
 hatch shell
-```
-This will spawn new shell with virtual environment enables. To leave just press ^D.
-
-
-Currently, the required Python version for development is Python 3.10+. For local development we recommend to use
-a wrapper for virtual environments like [pyenv](https://github.com/pyenv/pyenv).
-
-If you wish to setup environment with specific version ie. 3.10 you can use following command:
-
-```bash
-hatch env create local.py3.10```
-
-You can see all locally supported environments with
-
-```bash
-hatch env show
-```
-
-Please keep in mind that you need these python versions available in your `$PATH`. You can install them using `hatch` or other tool like [pyenv](https://github.com/pyenv/pyenv)
-
-## Release notes
-
-Every user-visible change should get one line in `RELEASE-NOTES.md` under `# Unreleased version`, in the appropriate subsection (`## Backward incompatibility`, `## Deprecations`, `## New additions`, or `## Fixes and improvements`).
-
-## Unit tests
-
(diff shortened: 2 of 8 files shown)


.github/pull_request_template.md
```diff
@@ -1,13 +1,12 @@
...
+   * [ ] I've added or updated unit tests to verify correctness of my new code.
    * [ ] I've added or updated integration tests to verify correctness of my new code.
-   * [ ] I've confirmed that my changes are working by executing CLI's commands manually on MacOS.
-   * [ ] I've confirmed that my changes are working by executing CLI's commands manually on Windows.
-   * [ ] I've confirmed that my changes are up-to-date with the target branch.
-   * [ ] I've described my changes in the release notes.
-   * [ ] I've described my changes in the section below.
-   * [ ] I've described my changes in the documentation.
+   * [ ] Manually tested on macOS
+   * [ ] Manually tested on Windows
```
sfc-gh-jwilkowski, line 6: reworded that, so that it's more informational rather than indicating that we expect both to happen

CONTRIBUTING.md
```diff
@@ -16,244 +16,128 @@
...
-hatch run pre-commit
-```
-This will spawn new shell with environment and all required packages installed.
-This will also install snowflake cli package in editable mode.
-
-To enter environment use following command.
-```bash
 hatch shell
+hatch run test
 ```
```
sfc-gh-olorek, line 38: The longer `## Development environment` section below keeps `pre-commit install`, but Quick start drops it. Anyone following Quick start commits without hooks and then trips on the first push when CI flags formatting. Probably worth pulling `pre-commit install` up here too — Quick start is supposed to be the path that Just Works.

docs/contributing/conventions.md
```diff
@@ -0,0 +1,188 @@
...
+`IDENTIFIER('...')` so Snowflake resolves it correctly:
+
+```python
+# WRONG
+cursor.execute(f"CREATE TABLE {database}.{schema}.{table_name} ...")
+
+# CORRECT
+from snowflake.cli.api.identifiers import FQN
+fqn = FQN.from_string(f"{database}.{schema}.{table_name}")
+cursor.execute(f"CREATE TABLE {fqn.sql_identifier} ...")
```
sfc-gh-olorek, line 41: I think the "CORRECT" example here still has the same issue as the WRONG one — `table_name` (and the others) get f-string-concatenated *before* `FQN.from_string` sees them. If `table_name` is `"foo'); DROP TABLE x;--"`, the wrap with `IDENTIFIER('...')` happens around an already-broken string. The protection is real for the `IDENTIFIER` resolution side of things but doesn't cover structural injection in the components.

Two options: either use the explicit constructor `FQN(database=database, schema=schema, name=table_name)` (the class supports this — `src/snowflake/cli/api/identifiers.py:46`), or call out here that the individual components must already be validated identifiers. I'd lean toward showing the field-by-field constructor — it's the form that's actually safe with untrusted input.

docs/contributing/conventions.md
```diff
@@ -0,0 +1,188 @@
...
+match `'myXdb'`:
+
+```python
+from snowflake.cli.api.project.util import identifier_to_show_like_pattern
+cursor.execute(f"SHOW DATABASES LIKE {identifier_to_show_like_pattern(name)}")
+```
+
+### Bind parameters
+
+For scalar values in DML statements (`SELECT`, `INSERT`, `UPDATE`), prefer bind
```
sfc-gh-olorek: Tiny pedantic thing, but `SELECT` is DQL, not DML. Since this page is the team's terminology authority I'd write "scalar values in queries and DML (`SELECT`, `INSERT`, `UPDATE`, `DELETE`)" or just "in queries and write statements".

docs/contributing/lifecycle.md
```diff
@@ -0,0 +1,262 @@
...
+- One-line release note required when entering PuPr (under `## New additions`):
+  ```
+  * <name> is now available in preview.
+  ```
+- Breaking changes are no longer acceptable from this point
+- Full backward compatibility rules apply going forward
+
+### GA (Generally Available)
+
+- Feature flag removed from `feature_flags.py`
```
sfc-gh-olorek: Nit — line 57 (and the rest of the doc) uses the full `src/snowflake/cli/api/feature_flags.py`; this bullet drops it to just `feature_flags.py`. Probably worth being consistent so people can grep their way to the file.

docs/contributing/lifecycle.md
```diff
@@ -0,0 +1,262 @@
...
+    ...
+```
+
+**Gate a feature so it cannot be invoked at all** (use sparingly — only when
+running without the flag would be unsafe or meaningless):
+```python
+@app.command(is_enabled=FeatureFlag.ENABLE_MY_FEATURE.is_enabled)
+def my_command():
+    ...
+```
```
sfc-gh-olorek, line 124: **Missing content — flagging here because this page is the natural home.**

The whole lifecycle story is about how features *graduate* (PrPr → PuPr → GA), but the docs never say how a feature is born. A first-time contributor reading all five files still doesn't know:

- Where command files live (`src/snowflake/cli/_plugins/<name>/commands.py` is the convention, right?)
- How `SnowTyperFactory` registration works — what makes a new group actually appear under `snow`?
- The `api/` vs `_plugins/` split — what goes where, and what does `SqlExecutionMixin` give you?
- How `CommandResult` shapes interact with `--format json|table|csv`

Most real "how do I add a new command?" questions land in one of those four buckets. Could either fit a short "Adding a new command" section into this file (since it would dovetail with the feature-flag scoping examples already here), or split it out into a small architecture page. Not a blocker for this PR but I'd flag it as the biggest gap in the new docs.

docs/contributing/testing.md
```diff
@@ -0,0 +1,155 @@
...
+- The happy path
+- Error paths — assert `result.exit_code != 0` and check the error message
+- `result.exit_code` assertion on every invoke — do not only check output
+
+## Fixtures
+
+Unit-only fixtures belong in `tests/conftest.py`. Fixtures shared between unit
+and integration tests belong in `tests_common/conftest.py`. Do not import
+fixtures from `tests_integration/` into unit tests — the two directories have
+different assumptions about what's available.
```
sfc-gh-olorek, line 102: **Missing content — fixture catalog.**

This section says where unit fixtures live but doesn't enumerate the ones already there. `tests/conftest.py` defines `runner`, `mock_connect`, `mock_ctx`, `mock_cursor` and a bunch of others that show up everywhere — without a list, contributors reach for `@mock.patch` decorators and re-mock things that already have a fixture (I've personally been called out for this). Even a short "the fixtures you'll usually want" table here would save a lot of review back-and-forth.

docs/contributing/testing.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Pytest markers
+
+| Marker | Meaning |
+|--------|---------|
+| `integration` | Requires live Snowflake connection |
+| `e2e` | Runs against a freshly installed CLI in a clean venv |
+| `performance` | Performance benchmarks |
+| `spcs` | Snowpark Container Services (requires SPCS account) |
+| `loaded_modules` | Checks which modules are imported at startup |
```
sfc-gh-olorek: A few things off in this table vs `pyproject.toml`:

- `spcs` is listed here but isn't actually registered in `[tool.pytest.ini_options].markers` — it only appears inside `addopts`. So pytest will warn on `@pytest.mark.spcs` usage. Either we add it to the markers list (probably the right fix) or drop it from this table.
- Missing from the table: `patch_app_version`, `no_qa`, `qa_only` — `qa_only` and `no_qa` in particular are ones contributors actually reach for.

The "default `hatch run test` excludes …" line below matches `addopts` exactly, so that part's good.

docs/contributing/conventions.md
```diff
@@ -0,0 +1,188 @@
...
+```python
+from snowflake.cli.api.console import cli_console as cc
+
+cc.step("Doing something...")          # regular progress line
+cc.warning("Something looks off")      # warning line
+
+with cc.phase("Building...", "Done."): # grouped output block
+    cc.step("Step A")
+    cc.step("Step B")
+```
```
sfc-gh-olorek, line 201: **Missing content — linting & formatting.**

This page covers imports, logging, errors, output, SQL safety — but says nothing about black, ruff, line length, what `.ruff.toml` enforces, or which of these pre-commit will fail you on. That's the *first* thing a contributor's PR is going to hit, and it's the section they're most likely to grep for. "Code conventions" without a tooling block feels half-finished — even a 5-line section pointing at `.pre-commit-config.yaml` and saying "black + ruff with these rules; line length is N; pre-commit will fix the auto-fixable ones" would do it.

sfc-gh-olorek (commented):
Big +1 on the restructure — splitting the lifecycle / conventions / testing / process / debugging stories out of CONTRIBUTING.md is overdue, and the lifecycle page in particular captures stuff that was tribal knowledge before. A bunch of the inline comments are nits, but a few are real (FQN example, markers table, missing `pre-commit install`).

The one structural thing I'd push on is what's *not* in the diff — these docs cover how a feature graduates (PrPr → PuPr → GA) but never how a feature is born. The most common new-contributor task is "add a new command" and after reading all five files they still don't know where commands live, how `SnowTyperFactory` gets registered, or how plugin entry-points wire things up. I left a comment on lifecycle.md with more detail. The other two missing pieces I'd really like to see: a fixture catalog in testing.md (people keep re-mocking `mock_connect` / `mock_ctx` / `mock_cursor` because they don't know they exist), and a linting/formatting section in conventions.md (no mention of black, ruff, line length, or what pre-commit will fail you on).

None of the missing-content stuff has to block this PR — happy for it to land as-is and follow up — but flagging in case you want to fold any of it in.

.github/pull_request_template.md
```diff
@@ -1,13 +1,12 @@
...
-   * [ ] I've described my changes in the release notes.
-   * [ ] I've described my changes in the section below.
-   * [ ] I've described my changes in the documentation.
+- [ ] I've confirmed that instructions included in README.md are still correct after my changes in the codebase.
+- [ ] I've added or updated unit tests to verify correctness of my new code.
+- [ ] I've added or updated integration tests to verify correctness of my new code.
+- [ ] Manually tested on macOS
+- [ ] Manually tested on Windows
+- [ ] I've confirmed my changes are up-to-date with the target branch.
+- [ ] I've described my changes in `RELEASE-NOTES.md` (see [when and how](docs/contributing/lifecycle.md#release-notes)).
```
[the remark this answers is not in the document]
sfc-gh-jwilkowski: done

CONTRIBUTING.md
sfc-gh-jwilkowski: done

docs/contributing/conventions.md
sfc-gh-jwilkowski: done

docs/contributing/lifecycle.md
sfc-gh-jwilkowski: done

docs/contributing/conventions.md
sfc-gh-jwilkowski: Good catch. Updated the example to use the direct constructor `FQN(database=database, schema=schema, name=table_name)` and added a note clarifying that `FQN.from_string()` remains appropriate when the full dotted name arrives as a single already-validated value (e.g. a parsed CLI argument).

docs/contributing/lifecycle.md
sfc-gh-jwilkowski: Agreed it's the biggest gap. Keeping it out of this PR — it deserves its own focused doc (command registration, `api/` vs `_plugins/` split, `CommandResult` shapes) rather than being appended here. Will track as a follow-up.

docs/contributing/testing.md
sfc-gh-jwilkowski: Added a fixture table covering the ones contributors reach for most: `runner`, `mock_cursor`, `mock_ctx`, `mock_connect`, `mock_statement_success`, `temporary_directory`, `project_directory`, `config_manager`.

docs/contributing/testing.md
sfc-gh-jwilkowski: Fixed `spcs` (not in the `markers` list in `pyproject.toml`). For the additional markers: took a stricter cut and documented only the 4 with meaningful real-world usage (`integration`, `e2e`, `no_qa`, `qa_only`). Markers with a single occurrence and zero-occurrence ones aren't worth advertising — the `addopts` excludes line below the table still enumerates them all since it mirrors `pyproject.toml` directly.

docs/contributing/conventions.md
sfc-gh-jwilkowski: Added a section. Kept it deliberately light — points at `.pre-commit-config.yaml` as the source of truth rather than listing specific rules or line lengths. Those details will go stale, and pre-commit surfaces violations automatically on commit anyway.

sfc-gh-jwilkowski:
Fixture catalog is done (see testing.md). For the command architecture guide — agreed it's the most impactful missing piece and I'll follow up with a dedicated doc. The linting section is also added to conventions.md, kept minimal by design (see inline reply).

sfc-gh-olorek (dismissed):
lgtm

sfc-gh-jbilek:
I checked the update, here are a few things:

1. The file links to the new docs are listed at the bottom of `CONTRIBUTING.md`. They should be at the top and serve as a navigation hub for all the content (faster scan and navigation).

2. `README.md` should also link to `CONTRIBUTING.md`. There's already a section for it at the bottom.

3. Unit tests, snapshots, and integration tests are now in both `CONTRIBUTING.md` and `testing.md`. If we remove the duplication, `CONTRIBUTING.md` could be really simple: just the linked files, a quick start for getting the repo running, and how to open a PR.

4. Separately, on PuPr: right now it stays behind a feature flag, which makes it the same user-facing surface as PrPr. If PuPr is not visible, what's the point of it? A single "opt-in for previews" toggle would protect users who want stability while letting PuPr reach a wider audience.

5. As a next step we should outline the process for design/spec and approval.

sfc-gh-jwilkowski:
> I checked the update, here are a few things:
> 
> 1. The file links to the new docs are listed at the bottom of `CONTRIBUTING.md`. They should be at the top and serve as a navigation hub for all the content (faster scan and navigation).
> 2. `README.md` should also link to `CONTRIBUTING.md`. There's already a section for it at the bottom.
> 3. Unit tests, snapshots, and integration tests are now in both `CONTRIBUTING.md` and `testing.md`. If we remove the duplication, `CONTRIBUTING.md` could be really simple: just the linked files, a quick start for getting the repo running, and how to open a PR.

Thanks, I've improved navigation in [dc94335](https://github.com/snowflakedb/snowflake-cli/pull/3031/commits/dc9433513837529a4424802999d1c9a8c363d920)

> 4. Separately, on PuPr: right now it stays behind a feature flag, which makes it the same user-facing surface as PrPr. If PuPr is not visible, what's the point of it? A single "opt-in for previews" toggle would protect users who want stability while letting PuPr reach a wider audience.

That's a good point and we should think of it as followup action item. As of now a feature being in PuPr mostly means a guarantee that we can't break it.

> 5. As a next step we should outline the process for design/spec and approval.

Yes, that's something I plan to tackle next.

sfc-gh-jbilek (approved):
✅

sfc-gh-olorek (approved):
lgtm

PR #3033, opened 2026-05-19
chore: catch release-notes bullets duplicated across released sections

sfc-gh-olorek:
### Pre-review checklist

### Changes description

The existing `check-release-notes-duplicates` pre-commit hook (added in #3025) catches bullets duplicated between "Unreleased version" and a released section. It does not catch the rebase-drift case: when an author's patch hunks are keyed on adjacent bullets as context, `git rebase` will silently apply the hunk under whichever header those context lines now live under. If main has moved bullets into a new `# vX.Y.Z` section since the branch was authored, the rebased patch can land in that section AND drag in context bullets that already shipped in an older release. Hit this on #2932 — two bullets ended up duplicated across `# v3.17.1` and `# v3.17.0` and the existing hook reported the file clean.

This adds a third rule to `find_duplicates`: a bullet appearing in both the most-recent `# vX.Y.Z` section and any older released section is flagged. The check is intentionally scoped to most-recent-vs-older (not all pairs of released sections) so legacy within-release patterns in old notes — e.g. v2.2.0 repeating `* snow snowpark package create:` as a header bullet under both Deprecations and New additions — do not false-positive.

The hook's error message is updated to name both failure modes so authors know whether to remove the bullet from Unreleased (post-release-cut case) or move a new bullet there from a released section (rebase-drift case).

No release-notes entry — this is internal tooling, matching #3025.

### DIFF
--- a/scripts/check_release_notes_duplicates.py
+++ b/scripts/check_release_notes_duplicates.py
@@ -110,7 +110,10 @@ def find_duplicates(path: Path) -> list[str]:
         # in an older released section is the rebase-context-drift case:
         # the author's patch dragged context bullets that already shipped
         # in a prior release into the current release-staging section.
-        if most_recent_released:
+        # Skip when the bullet is also in Unreleased — the check above
+        # already flagged it, and firing here would emit a contradictory
+        # second message ("move to Unreleased" vs. "remove from Unreleased").
+        if most_recent_released and not in_unreleased:
             in_recent = [p for p in places if p[0] == most_recent_released]
             in_older_released = [
                 p
--- a/scripts/check_release_notes_duplicates.py
+++ b/scripts/check_release_notes_duplicates.py
@@ -16,9 +16,16 @@
 
 Companion to the `RELEASE-NOTES.md merge=union` git attribute. Union merge
 auto-resolves the "two PRs each append a bullet" conflict but silently keeps
-both copies of a bullet that the release cut has already moved from
-"Unreleased version" into a `# vX.Y.Z` section. This script flags that case
-so the author removes the already-released bullet from Unreleased.
+both copies of a bullet in scenarios that produce an invalid file:
+
+* The release cut moved a bullet from "Unreleased version" into a `# vX.Y.Z`
+  section, but the author's branch still has the bullet under Unreleased.
+* A rebase keyed on adjacent bullets as patch context dropped the author's
+  new bullet under the most-recent released section instead of Unreleased,
+  carrying along context bullets that are also present in an older released
+  section.
+
+The script flags both cases so the author can remove the misplaced bullet.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


sfc-gh-jwilkowski (dismissed):
**Request changes:** one blocking issue.

**W1 — double-fire on three-way duplicate**
When a bullet appears in Unreleased + most-recent released + an older released section, both the existing `if in_unreleased and in_released:` check and the new `if in_recent and in_older_released:` check fire independently, producing two error messages with contradictory remediation instructions for the same bullet. Fix: add `and not in_unreleased` to the new guard so the rebase-drift check only fires when the Unreleased check hasn't already flagged the bullet.

Otherwise the logic is correct, the legacy false-positive avoidance is well-tested, and the updated error message is actionable. Clean change overall.

scripts/check_release_notes_duplicates.py
```diff
@@ -88,6 +106,25 @@ def find_duplicates(path: Path) -> list[str]:
...
+        # the author's patch dragged context bullets that already shipped
+        # in a prior release into the current release-staging section.
+        if most_recent_released:
+            in_recent = [p for p in places if p[0] == most_recent_released]
+            in_older_released = [
+                p
+                for p in places
+                if _RELEASED_SECTION_RE.match(p[0]) and p[0] != most_recent_released
+            ]
+            if in_recent and in_older_released:
```
sfc-gh-jwilkowski, line 123: [WORTH FIXING] When a bullet exists in Unreleased, the most-recent released section, and an older released section simultaneously, both this block and the `if in_unreleased and in_released:` check above (line 101) fire independently. The author receives two messages — one saying "remove from Unreleased" and another saying "move new bullet to Unreleased" — which is contradictory for the same bullet.

Guard this block so it only fires when the first check hasn't already fired:
```python
if in_recent and in_older_released and not in_unreleased:
```

sfc-gh-olorek: Good catch — fixed in afe11a87. Added a regression test for the triple-section case (Unreleased + most-recent released + older released) that asserts only the Unreleased-vs-released error fires.

tests/test_check_release_notes_duplicates.py
```diff
@@ -151,3 +151,91 @@ def test_main_returns_one_on_dirty_file(notes_file, capsys):
     assert rc == 1
     err = capsys.readouterr().err
     assert "Duplicate bullet" in err
+
```
sfc-gh-prafalski, line 154: The newly added rule flag bullet points that appear in both, the most-recent version and any older version. Currently, we allow duplicates between older, already released versions. Let's document that behavior with one more test:

```
def test_duplicate_across_two_older_releases_is_not_flagged(notes_file):
      path = notes_file(
          """
          # v3.17.2
          ## Fixes and improvements
          * Fixed the newest thing.

          # v3.17.1
          ## Fixes and improvements
          * Fixed an old shared thing.

          # v3.17.0
          ## Fixes and improvements
          * Fixed an old shared thing.
          """
      )
      assert check.find_duplicates(path) == []
```

sfc-gh-olorek: Added in 8353a424.

sfc-gh-prafalski (approved):
LGTM.

PR #3053, opened 2026-05-25
Snowflake Apps: preserve case of personal database name

sfc-gh-gbloom:
### Pre-review checklist

### Changes description

`SnowflakeAppManager.get_personal_database` returned `('USER$' || CURRENT_USER()).upper()`, which silently breaks any Snowflake user whose login was created as a quoted, mixed/lower-case identifier (e.g. `"first.last@domain.com"`). For those users:

- The real personal database is `USER$first.last@domain.com`.
- `app_fqn` later wraps the value in a **case-sensitive** quoted identifier (`"USER$FIRST.LAST@DOMAIN.COM"`).
- As a result, every deploy-time statement that defaults to the personal DB — `CREATE WORKSPACE`, `USE DATABASE`, the artifact-repo `IDENTIFIER(...)` calls, etc. — silently targets a non-existent database and fails (or worse, creates a brand-new wrongly-cased database).

`CURRENT_USER()` already returns the value in whatever case Snowflake stores: uppercase-folded for unquoted users, original case for quoted users. The defensive `.upper()` is unnecessary for the common case and actively wrong for quoted-identifier users.

### DIFF
--- a/src/snowflake/cli/_plugins/apps/manager.py
+++ b/src/snowflake/cli/_plugins/apps/manager.py
@@ -496,6 +496,15 @@ class SnowflakeAppManager(SqlExecutionMixin):
         Runs ``SELECT 'USER$' || CURRENT_USER() AS personal_database`` and
         returns the result.  Returns ``None`` when the query fails or the
         current user is not set (e.g. in unauthenticated contexts).
+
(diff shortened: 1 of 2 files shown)


sfc-gh-npasala (approved):
LGTM.

PR #3060, opened 2026-05-28
Jw/snow 3570963 contributing add docs about adding new commands

sfc-gh-jwilkowski:
### Pre-review checklist

### Changes description
https://snowflakecomputing.atlassian.net/browse/SNOW-3570963

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -16,5 +16,13 @@
 
 # Agent Instructions
 
-Before writing any code, read `CONTRIBUTING.md` and every guide it links to.
+Before writing any code, read each of the following guides in full:
+
+- [CONTRIBUTING.md](CONTRIBUTING.md)
+- [docs/contributing/adding-commands.md](docs/contributing/adding-commands.md)
+- [docs/contributing/conventions.md](docs/contributing/conventions.md)
+- [docs/contributing/lifecycle.md](docs/contributing/lifecycle.md)
+- [docs/contributing/testing.md](docs/contributing/testing.md)
+- [docs/contributing/process.md](docs/contributing/process.md)
+
 All requirements in those guides are mandatory, unless explicitly marked otherwise.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,19 @@
+<!--
+ Copyright (c) 2024 Snowflake Inc.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ -->
+
+# Agent Instructions
+
+When contributing to this repository, follow the guidelines in `CONTRIBUTING.md`.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -16,4 +16,5 @@
 
 # Agent Instructions
 
-When contributing to this repository, follow the guidelines in `CONTRIBUTING.md`.
+Before writing any code, read `CONTRIBUTING.md` and every guide it links to.
+All requirements in those guides are mandatory, unless explicitly marked otherwise.
--- a/docs/contributing/adding-commands.md
+++ b/docs/contributing/adding-commands.md
@@ -189,6 +189,8 @@ All types are in `src/snowflake/cli/api/output/types.py`:
 | `CollectionResult` | Iterable of dicts with no cursor (locally constructed results) |
 | `ObjectResult` | Single dict with no cursor |
 
+Other specialized types (`MultipleResults`, `StreamResult`, `EmptyResult`) exist for less common cases — see `src/snowflake/cli/api/output/types.py`.
+
 ### Lifecycle and visibility
 
 New commands start in PrPr behind a feature flag. See [lifecycle.md](lifecycle.md)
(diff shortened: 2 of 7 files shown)


docs/contributing/adding-commands.md
```diff
@@ -0,0 +1,233 @@
...
+```python
+@app.command()
+def create(name: str):
+    """Creates a new table."""
+    ...
+```
+
+### Return types
+
+All types are in `src/snowflake/cli/api/output/types.py`:
```
sfc-gh-olorek, line 182: Small thing — the way this reads, it sounds like these five are the whole set, but `output/types.py` also has `MultipleResults`, `StreamResult`, and `EmptyResult`. Could you add a one-liner like "other specialized types (streaming, multi-statement, empty) exist — see the file" so the next person doesn't think they have to pick from exactly these five?

docs/contributing/adding-commands.md
```diff
@@ -0,0 +1,233 @@
...
+
+Commands that modify or delete existing resources should use `ForceOption` and
+`InteractiveOption` from `snowflake.cli.api.commands.flags`.
+
+Behavior matrix:
+
+| `--force` | `--interactive` | Result |
+|-----------|-----------------|--------|
+| unset | False (non-TTY default) | Abort — safe default for scripts |
+| unset | True (TTY default) | Prompt the user |
```
sfc-gh-olorek: nit: "non-TTY default" / "TTY default" is a bit jargony on a first read — someone new to the project may not immediately parse what "non-TTY" means in this context. Maybe something like `unset (no terminal — e.g. CI, piped input)` and `unset (interactive terminal)`? Same content, just spelled out for the case when you're reading this without that mental shortcut.

docs/contributing/conventions.md
```diff
@@ -87,6 +87,49 @@ cursor.execute("SELECT * FROM my_table WHERE id = %s", (user_id,))
...
+## Handling sensitive values
+
+When a command option accepts a secret (password, token, key, etc.), declare it
+with `click_type=SecretTypeParser()` from `snowflake.cli.api.commands.flags`.
+This wraps the raw string in a `SecretType` at parse time so the value is never
+accidentally logged or printed (`__str__` returns `***`):
+
+```python
+from snowflake.cli.api.commands.flags import SecretTypeParser
+from snowflake.cli.api.secret import SecretType
```
sfc-gh-olorek: Tiny clarification suggestion — when I first read the snippet I assumed the `SecretType` import was needed at runtime (i.e. that I had to wrap something myself), but `SecretTypeParser()` already does the wrapping at parse time. The `SecretType` import is just for the type annotation. Worth a half-line note saying so, e.g. "`SecretTypeParser` wraps the raw input at parse time — `SecretType` is here only for the annotation".

docs/contributing/lifecycle.md
```diff
@@ -154,7 +154,8 @@ def my_command(
 
 ### PuPr → GA
 
-1. Remove the flag entry from `src/snowflake/cli/api/feature_flags.py`
+1. Remove the flag entry from `src/snowflake/cli/api/feature_flags.py` and all
+   remaining references to it in the codebase.
```
sfc-gh-olorek, line 158: Good catch on calling this out explicitly — flag removal is a real footgun. While we're here, worth mentioning that flag-hidden commands still show up in snapshot files, so going PuPr→GA also requires a snapshot regen even though nothing user-visible changed? It's the same gotcha as the snapshot section further down, just from a different angle.

AGENTS.md
```diff
@@ -0,0 +1,19 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ -->
+
+# Agent Instructions
+
+When contributing to this repository, follow the guidelines in `CONTRIBUTING.md`.
```
sfc-gh-olorek: Minor — as it stands this is one line of guidance under 14 lines of license header. Not a blocker, but I wonder if it's worth either folding the pointer into `CONTRIBUTING.md` directly, or expanding `AGENTS.md` with a couple of agent-specific gotchas (snapshot regen on hidden-command changes, design sign-off requirement, the `skip-release-notes` label rule). Right now anything that sends an agent here just bounces them to `CONTRIBUTING.md` anyway.

sfc-gh-olorek (commented):
Docs LGTM overall — checked the technical claims (flat/nested layouts, plugin_spec wiring, `SecretTypeParser`, return types, `ForceOption`/`InteractiveOption`) against `main` and they all hold up. A few minor inline nits below; none blocking.

docs/contributing/adding-commands.md
sfc-gh-jwilkowski: fixed

docs/contributing/adding-commands.md
sfc-gh-jwilkowski: fixed

docs/contributing/conventions.md
sfc-gh-jwilkowski: fixed

docs/contributing/lifecycle.md
sfc-gh-jwilkowski: fixed

AGENTS.md
sfc-gh-jwilkowski: well that was an interesting thing. Apparently when this included only a reference to CONTRIBUTING.md agents pretty much ignored it. After a couple iterations and telling explicitly that some files have to be read before any implementation I'm getting much better results that follow our guidelines

sfc-gh-olorek (approved):
lgtm

sfc-gh-prafalski (approved):
LGTM

sfc-gh-jbilek (approved):

PR #3032, opened 2026-05-18
Add @snowflakedb/client-side-interfaces to CODEOWNERS

sfc-gh-olorek:
### Pre-review checklist
   * [x] I've confirmed that instructions included in README.md are still correct after my changes in the codebase.
   * [x] I've added or updated automated unit tests to verify correctness of my new code.
   * [x] I've added or updated integration tests to verify correctness of my new code.
   * [x] I've confirmed that my changes are working by executing CLI's commands manually on MacOS.
   * [x] I've confirmed that my changes are working by executing CLI's commands manually on Windows.
   * [x] I've confirmed that my changes are up-to-date with the target branch.
   * [x] I've described my changes in the release notes.
   * [x] I've described my changes in the section below.
   * [x] I've described my changes in the documentation.

   _CODEOWNERS-only change; no code, tests, or docs are affected._

### Changes description

Adds the new `@snowflakedb/client-side-interfaces` team alongside the existing `@snowflakedb/snowcli` global owner. The new team is the master team for Client-Side Services (Snowflake CLI, Terraform Provider, CI/CD integrations) under Developer Platform. Additive change — `@snowflakedb/snowcli` retains ownership.

PR #3064, opened 2026-05-28
ci: [NOSNOW] unblock doc-only PRs by replacing paths-ignore with a should-run gate

sfc-gh-olorek:
### Pre-review checklist
        * N/A — CI-only change with no user-visible effect.

### Changes description

The `Testing`, `Integration testing` and `E2E testing` workflows use `paths-ignore` to skip PRs that only touch docs / `LICENSE` / `.github/CODEOWNERS`. The matrix-expanded jobs they produce (`tests (ubuntu-latest, 3.10)`, `integration-trusted (...) / tests-trusted`, `e2e-trusted (...) / tests-trusted`) are listed as **required** status checks on `main`. When a workflow is skipped via `paths-ignore`, GitHub does not post the matching required status check, so branch protection sees it as missing and `mergeStateStatus` stays `BLOCKED` forever.

Example: #3032 only modifies `.github/CODEOWNERS` and is unmergeable for this reason.

#### Approach

Drop `paths-ignore` from the trigger and replace it with an explicit `should-run` gate:

- New `changes.yaml` reusable workflow runs `dorny/paths-filter@v3.0.2` (pinned by SHA) and exposes a `code` output that is `true` whenever the PR touches anything that is **not** `**/*.md`, `docs/**`, `LICENSE`, or `.github/CODEOWNERS`.
- `test.yaml` always starts the `tests` matrix job, but each heavy step is gated by `if: needs.changes.outputs.code == 'true'`. The job still completes (and reports success), so the required check posts.
- `test_integration.yaml` and `test_e2e.yaml` always start their reusable callers and pass `should-run: ${{ needs.changes.outputs.code == 'true' || github.event_name != 'pull_request' }}` to `test_trusted.yaml`. On non-PR events (`push`, `schedule`) the heavy work always runs, matching today's behavior.
- `test_trusted.yaml` gains a `should-run` input (default `"true"`) that gates Python setup / dependency install / test invocation.

#### What this means in practice

- Doc-only PRs (like #3032): required checks complete in seconds with no real work, PR becomes mergeable.
- Code PRs: behavior is unchanged — `paths-filter` returns `code=true`, all steps run.
- `push` / `schedule` runs: always run the full job, same as before.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -24,6 +24,7 @@ jobs:
     uses: ./.github/workflows/changes.yaml
   tests:
     needs: [define-matrix, changes]
+    if: needs.changes.outputs.code == 'true'
     strategy:
       fail-fast: true
       matrix:
@@ -34,21 +35,17 @@ jobs:
       - uses: actions/checkout@v4
         with:
           persist-credentials: false
-      - if: needs.changes.outputs.code == 'true'
-        name: Set up Python ${{ matrix.python-version }}
+      - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
-      - if: needs.changes.outputs.code == 'true'
-        name: Install hatch
+      - name: Install hatch
         run: |
           pip install -U click==8.2.1 hatch==1.15.1 virtualenv==20.39.1
           hatch env create default
-      - if: needs.changes.outputs.code == 'true'
-        name: Test with hatch
+      - name: Test with hatch
         run: hatch run test-cov
-      - if: needs.changes.outputs.code == 'true'
-        uses: codecov/codecov-action@18283e04ce6e62d37312384ff67231eb8fd56d24
+      - uses: codecov/codecov-action@18283e04ce6e62d37312384ff67231eb8fd56d24
 
   # Temporarily disabled; re-enable once config v2 (NG) is ready.
   # tests-config-ng:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -4,11 +4,6 @@ on:
   pull_request:
     branches:
       - "**"
-    paths-ignore:
-      - "*.md"
-      - "docs/**"
-      - "LICENSE"
-      - ".github/CODEOWNERS"
   push:
     tags:
       - "v*" # Push events to matching v*, i.e. v1.0, v20.15.10
@@ -25,8 +20,10 @@ permissions:
 jobs:
   define-matrix:
     uses: ./.github/workflows/matrix.yaml
+  changes:
+    uses: ./.github/workflows/changes.yaml
   tests:
-    needs: define-matrix
+    needs: [define-matrix, changes]
     strategy:
       fail-fast: true
       matrix:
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yaml
```diff
@@ -37,17 +34,21 @@ jobs:
       - uses: actions/checkout@v4
         with:
           persist-credentials: false
-      - name: Set up Python ${{ matrix.python-version }}
+      - if: needs.changes.outputs.code == 'true'
+        name: Set up Python ${{ matrix.python-version }}
```
sfc-gh-jwilkowski: These four per-step `if:` conditions can be replaced by a single job-level `if: needs.changes.outputs.code == 'true'` on the `tests` job. GitHub has treated skipped jobs (via `if:`) as satisfying required checks since its 2022 behaviour change — the job doesn't need to run-and-do-nothing, a skip is sufficient to post the green check.

This would also eliminate the `should-run` input and step conditions added to `test_trusted.yaml`.

.github/workflows/test_trusted.yaml
```diff
@@ -15,6 +15,12 @@ on:
       hatch-run:
         required: true
         type: string
+      should-run:
+        # Posts the required check as success but skips the heavy steps when
```
sfc-gh-jwilkowski: Not needed if the callers use a job-level `if:` instead — that keeps this reusable workflow unmodified. See the suggestion on `test.yaml`.

.github/workflows/test_integration.yaml
```diff
@@ -49,6 +46,7 @@ jobs:
       python-version: ${{ matrix.python-version }}
       python-env: integration
       hatch-run: integration:test
+      should-run: ${{ needs.changes.outputs.code == 'true' || github.event_name != 'pull_request' }}
     secrets: inherit
```
sfc-gh-jwilkowski, line 52: Can be dropped in favour of extending the existing job-level `if:` on `integration-trusted` with `&& (needs.changes.outputs.code == 'true' || github.event_name != 'pull_request')`.

.github/workflows/test_e2e.yaml
```diff
@@ -48,6 +45,7 @@ jobs:
       python-version: ${{ matrix.python-version }}
       python-env: e2e
       hatch-run: e2e:test
+      should-run: ${{ needs.changes.outputs.code == 'true' || github.event_name != 'pull_request' }}
     secrets: inherit
```
sfc-gh-jwilkowski, line 51: Same as `test_integration.yaml` — fold the gate into the job-level `if:` on `e2e-trusted`.

sfc-gh-jwilkowski (commented):
Fix is correct — `paths-ignore` skipping the whole workflow is the root cause, and `changes.yaml` as a reusable workflow is the right abstraction. A few places can be simplified; see inline comments.

.github/workflows/test.yaml
sfc-gh-olorek: Done in 5d651e50 — replaced with a single job-level `if: needs.changes.outputs.code == 'true'` on `tests`.

.github/workflows/test_trusted.yaml
sfc-gh-olorek: Done in 5d651e50 — `should-run` input and step-level conditions removed; `test_trusted.yaml` is back to its pre-PR shape.

.github/workflows/test_integration.yaml
sfc-gh-olorek: Done in 5d651e50 — folded `(needs.changes.outputs.code == 'true' || github.event_name != 'pull_request')` into the existing job-level `if:` on `integration-trusted`.

.github/workflows/test_e2e.yaml
sfc-gh-olorek: Done in 5d651e50 — same treatment as `test_integration.yaml`, gate folded into the `e2e-trusted` job-level `if:`.

sfc-gh-jwilkowski (approved):

PR #3065, opened 2026-05-28
feat: [SNOW-3579570] replace dbt-version regex with server-driven validation

sfc-gh-afaulds:
### Pre-review checklist

### Changes description

The `--dbt-version` option had a hard-coded client-side regex (`^\d+\.\d+\.\d+(-[a-zA-Z0-9]+)?$`) that rejected dots in the prerelease suffix. This breaks for the dbt Fusion GA binary, which reports its version as `2.0.0-preview.175`, and is fragile to any future server-side naming change. Discussed with @sfc-gh-jwilkowski (original author of the regex) — preferred approach is server-driven validation so the CLI cannot drift out of sync with the supported-versions list.

**`snow dbt deploy`** — `DBTManager.deploy()` now calls `SELECT SYSTEM$SUPPORTED_DBT_VERSIONS()` first thing whenever `--dbt-version` is set. If the user-supplied value isn't in the returned list, it raises before stage creation / file upload, so the user fails fast without paying for a doomed upload. The error message lists the actual supported versions returned by the server.

**`snow dbt execute`** — no client-side validation at all. The server already rejects unknown versions immediately with `Invalid value specified for property DBT_VERSION, supported versions are: ...`, and execute has no expensive pre-step worth protecting.

**`commands.py`** — removed `SemanticVersionType`, `SEMANTIC_VERSION_PATTERN`, and the `re` / `click` imports. Both `--dbt-version` options now flow through unvalidated client-side; the manager (deploy) or the server (execute) is the only validator.

### Testing

- Unit tests: 84 dbt tests pass. New `TestValidateDbtVersion` class plus two new `TestDeploy` cases (`test_deploy_aborts_before_upload_when_version_unsupported`, `test_deploy_does_not_validate_when_version_not_specified`) cover the new validation flow end-to-end at the manager layer. The three command-level tests that asserted the old regex error message were updated to reflect that those values now flow through to the manager.
- Manual end-to-end testing on a writeable QA Snowflake account:
  - `--dbt-version=99.99.99` → fails before upload, lists `1.9.4, 1.10.15, 2.0.0-preview`.
  - `--dbt-version=2.0.0-preview.175` (the original motivator) → fails before upload, same message.


src/snowflake/cli/_plugins/dbt/commands.py
```diff
@@ -200,7 +175,6 @@ def before_callback(
     dbt_version: Optional[str] = typer.Option(
         None,
         "--dbt-version",
-        click_type=SemanticVersionType(),
```
sfc-gh-jwilkowski, line 209: If you remove this underlying validation of version string you're becoming prone to sql injections. Use `to_string_literal()` to sanitize this input for execute commands

src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -106,6 +110,26 @@ def get_dbt_object_attributes(name: FQN) -> Optional[DBTObjectEditableAttributes
             dbt_version=row_dict.get("dbt_version"),
         )
 
+    def _get_supported_dbt_versions(self) -> List[str]:
+        row = self.execute_query(SUPPORTED_DBT_VERSIONS_QUERY).fetchone()
+        if row is None or row[0] is None:
+            raise CliError("Could not fetch supported dbt versions from server.")
+        try:
+            entries = json.loads(row[0])
+        except json.JSONDecodeError as exc:
```
sfc-gh-jwilkowski:
```suggestion
        except (json.JSONDecodeError, TypeError) as exc:
```

src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -106,6 +110,26 @@ def get_dbt_object_attributes(name: FQN) -> Optional[DBTObjectEditableAttributes
...
+            raise CliError(
+                "Could not parse supported dbt versions from server."
+            ) from exc
+        return [e["dbt_version"] for e in entries]
```
sfc-gh-jwilkowski: please wrap it in try/except (KeyError, TypeError)

tests/dbt/test_dbt_commands.py
```diff
@@ -355,15 +355,16 @@ def test_deploy_with_invalid_dbt_version_fails(
                 "deploy",
                 "TEST_PIPELINE",
                 f"--source={dbt_project_path}",
-                "--dbt-version=1.9",
+                "--dbt-version=1.9.4",
             ]
         )
 
-        assert result.exit_code == 2, result.output
```
sfc-gh-jwilkowski, line 362: You silently changes exit code when the version is wrong. I think a test should be added that confirms that we return the same exit code on invalid version. Most likely the error to raise is `CliArgumentError`

src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -114,6 +138,9 @@ def deploy(
         force: bool,
         attrs: DBTDeployAttributes,
     ) -> SnowflakeCursor:
+        if attrs.dbt_version:
```
sfc-gh-jwilkowski: You can move this validation to L158, so that the "cheap" validations of profiles are done first and only afterwards anything that calls Snowflake

RELEASE-NOTES.md
```diff
@@ -24,6 +24,7 @@
 ## Fixes and improvements
 * Fixed `TooManyFilesError` during `snow streamlit deploy` when `main_file` is a descendant of a directory listed in `artifacts`.
 * Upgraded `snowflake-connector-python` from 4.5.0 to 4.6.0.
+* `snow dbt deploy --dbt-version` now validates the requested version against the server's supported list (`SYSTEM$SUPPORTED_DBT_VERSIONS()`) before uploading project files, instead of a hard-coded client-side regex. Unsupported versions fail fast with a message listing the actual supported versions, and the CLI no longer rejects valid versions that don't match its old regex (e.g. `2.0.0-preview.175`). `snow dbt execute --dbt-version` no longer performs client-side validation either; the server returns the same supported-versions message if the value is unknown.
```
sfc-gh-jwilkowski: nit: this message is relatively long and detailed. Try to summarise it in 1-2 sentences so that readers quickly understand what is the user impact (`snow dbt deploy --dbt-version` now validates the requested version against the server's supported list) and not necessarily implementation details (`SYSTEM$SUPPORTED_DBT_VERSIONS()`)

sfc-gh-jwilkowski (changes requested):
Thank you for your contribution. It looks good, but I've left some comments to look at. Also please take a look at our improved CONTRIBUTING.md, which should explain some of our conventions (RELEASE-NOTES.md expectations) and important details (sanitizing user input)

src/snowflake/cli/_plugins/dbt/commands.py
sfc-gh-afaulds: Sql injection ... using the same user's connection/credentials who is running the command. They could just run whatever they inject via `snow sql -q` anyway, couldn't they? And it would run on the same connection - user/roles etc.

I'll address it anyway.

sfc-gh-afaulds: Now using _to_string_literal

tests/dbt/test_dbt_commands.py
sfc-gh-afaulds: Moved down in the file to test_deploy_with_invalid_dbt_version_returns_exit_code_2

src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -106,6 +111,31 @@ def get_dbt_object_attributes(name: FQN) -> Optional[DBTObjectEditableAttributes
             dbt_version=row_dict.get("dbt_version"),
         )
 
+    def _get_supported_dbt_versions(self) -> List[str]:
+        row = self.execute_query(SUPPORTED_DBT_VERSIONS_QUERY).fetchone()
```
sfc-gh-mraba: The `ProgrammingError` is never caught

```suggestion
    from snowflake.connector.errors import ProgrammingError

    def _get_supported_dbt_versions(self) -> List[str]:
        try:
            row = self.execute_query(SUPPORTED_DBT_VERSIONS_QUERY).fetchone()
        except ProgrammingError as exc:
            raise CliError(
                "Could not fetch supported dbt versions from server. "
                "Ensure your Snowflake account supports SYSTEM$SUPPORTED_DBT_VERSIONS()."
            ) from exc
```

src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -389,7 +422,7 @@ def execute(
         dbt_command_escaped = dbt_command.replace("'", "\\'")
         query = f"EXECUTE DBT PROJECT {name}"
         if dbt_version:
-            query += f" dbt_version='{dbt_version}'"
+            query += f" dbt_version={to_string_literal(dbt_version)}"
```
sfc-gh-mraba, line 480: The same approach could be applied to L:389 above 
```
    dbt_command_escaped = dbt_command.replace("'", "\\'")
    query += f" args='{dbt_command_escaped}'"
```

This isn't introduced by the PR but is inconsistent now that `dbt_version` uses the safer form.

WDYT?

tests/dbt/test_dbt_commands.py
```diff
@@ -390,14 +392,41 @@ def test_deploy_with_prerelease_version_passes_to_manager(
...
+                    "deploy",
+                    "TEST_PIPELINE",
+                    f"--source={dbt_project_path}",
+                    "--dbt-version=99.99.99",
+                    "--enhanced-exit-codes",
+                ]
+            )
+
+        assert result.exit_code == 2, result.output
+        assert "Invalid value '99.99.99'" in result.output
```
sfc-gh-mraba, line 432: `test_deploy_with_invalid_dbt_version_returns_exit_code_2` doesn't exercise the real code path

```python
    # test_dbt_commands.py (PR)
    def raise_invalid_version(*args, **kwargs):
        raise CliArgumentError(...)

    with mock.patch("...DBTManager.deploy", side_effect=raise_invalid_version):
        ...
```

 This mocks `deploy` entirely, so it only tests that the CLI handler converts `CliArgumentError` to exit code 2 — not that `_validate_dbt_version` is wired up correctly inside deploy. The manager-level test `test_deploy_aborts_before_upload_when_version_unsupported` in `test_manager.py` covers the real path, so coverage is fine, but the command test is less meaningful than it looks.

src/snowflake/cli/_plugins/dbt/manager.py
sfc-gh-afaulds: Good catch — applied in 670109c0. Also added a unit test (test_get_supported_dbt_versions_raises_on_programming_error) covering the catch path.

src/snowflake/cli/_plugins/dbt/manager.py
sfc-gh-afaulds: Agreed — applied in 670109c0. Updated the 7 affected TestDBTExecute parametrized cases to expect '' doubled-quote escaping.

tests/dbt/test_dbt_commands.py
sfc-gh-afaulds: Fair point, reduced the mocking a bit but test_deploy_aborts_before_upload_when_version_unsupported is still needed for the full coverage.

src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -106,6 +111,37 @@ def get_dbt_object_attributes(name: FQN) -> Optional[DBTObjectEditableAttributes
...
+        try:
+            return [e["dbt_version"] for e in entries]
+        except (KeyError, TypeError) as exc:
+            raise CliError(
+                "Could not parse supported dbt versions from server."
+            ) from exc
+
+    def _validate_dbt_version(self, dbt_version: str) -> None:
+        supported = self._get_supported_dbt_versions()
+        if dbt_version not in supported:
```
sfc-gh-jwilkowski, line 187: In theory we could receive an empty list of supported versions, which would look weird with such handling. Might be a good idea to handle that specifically

src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -185,7 +224,7 @@ def _deploy_alter(
         # lock it to a patch as well. If target version is provided, it's better to just
         # apply it.
         if attrs.dbt_version:
-            set_properties.append(f"DBT_VERSION='{attrs.dbt_version}'")
+            set_properties.append(f"DBT_VERSION={to_string_literal(attrs.dbt_version)}")
```
sfc-gh-jwilkowski, line 275: what do you think about running this sanitization when DBTDeployAttributes is instantiated? That would be less repetitive and error prone

sfc-gh-jwilkowski (commented):
2 minor comment, otherwise looks good!

src/snowflake/cli/_plugins/dbt/manager.py
sfc-gh-afaulds: We also check attrs.dbt_version in _validate_dbt_version against the server-returned version list, which is not to_string_literal'd. I think it makes sense to keep this attr as the plain version, and convert to string literal when creating SQL commands, but let me know if you have a better idea. to_string_literal'ing the returned server versions feels kind of weird but pretty subjective.

src/snowflake/cli/_plugins/dbt/manager.py
sfc-gh-afaulds: Good point - addressed

src/snowflake/cli/_plugins/dbt/manager.py
sfc-gh-jwilkowski: ok, that makes sense, let's keep it as is 👌

sfc-gh-jwilkowski (dismissed):

sfc-gh-jwilkowski (dismissed):

sfc-gh-jwilkowski (approved):

PR #3073, opened 2026-06-01
ci: [NOSNOW] gate docs-only PRs at step level so matrix required checks post

sfc-gh-olorek:
### Pre-review checklist
   * [x] If my changes add or modify user-facing interface (commands, flags, output formats), I consulted maintainers and got sign-off beforehand — N/A, CI-only.
   * [x] I've confirmed that instructions included in README.md are still correct after my changes in the codebase — N/A, CI-only.
   * [x] I've added or updated unit tests to verify correctness of my new code — N/A, GitHub Actions YAML, validated end-to-end on this PR's own checks.
   * [x] I've added or updated integration tests to verify correctness of my new code — N/A, GitHub Actions YAML.
   * [x] I've described my changes in `RELEASE-NOTES.md` — `skip-release-notes` (CI plumbing only, no user-facing change).
   * [x] I've updated documentation if behavior changed — refreshed the header comment in `changes.yaml`.

### Changes description

Follow-up to #3064. That PR moved the docs-only short-circuit from `paths-ignore` to a job-level `if: needs.changes.outputs.code == 'true'` on the matrix-bearing jobs. That works for the un-expanded job name (e.g. `tests`) but GitHub evaluates job-level `if:` **before** matrix expansion, so the per-cell required check names branch protection wants — `tests (ubuntu-latest, 3.10)`, `tests (windows-latest, 3.10)`, `e2e-trusted (...) / tests-trusted`, `integration-trusted (...) / tests-trusted` — are never posted. Docs-only PRs (e.g. CODEOWNERS-only #3032) end up stuck at *"Expected — Waiting for status to be reported"*, the same deadlock #3064 was supposed to fix.

This PR pushes the gate down to the **step** level so the matrix expands and every per-cell check posts as a (no-op) success when `changes.code == 'false'`:

- `test.yaml`: drop the job-level `if:` on `tests`; gate every step (except `actions/checkout`) on `needs.changes.outputs.code == 'true'`.
- `test_trusted.yaml`: reintroduce the `should-run` input (string, default `"true"`) and gate the heavy steps on it.
- `test_e2e.yaml` / `test_integration.yaml`: keep only the trust gate at job level (so fork events still bypass the secret-using path) and pass the docs-only condition into the reusable workflow as `should-run`.
- `changes.yaml`: rewrite the header comment to spell out why the gate has to live at the step level — to avoid the "moved back to job-level" regression that produced this bug.

The `actions/checkout` step still always runs so each per-cell job has at least one step and reports a green check. Other workflows (Build, Code quality, CodeQL, Release notes check) are unaffected.

### Verification

Will be visible directly on this PR's checks: `tests (ubuntu-latest, 3.10)`, `tests (windows-latest, 3.10)`, and the `*-trusted (...) / tests-trusted` rows should all post (running the heavy path here, since this PR touches workflow files which are not in the ignore set). To verify the docs-only path, the next CODEOWNERS-only PR after this lands should see all of them as green skipped checks rather than `Expected — Waiting`.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -22,9 +22,13 @@ jobs:
     uses: ./.github/workflows/matrix.yaml
   changes:
     uses: ./.github/workflows/changes.yaml
+  # Required check `tests (<os>, <py>)` must report on every PR, including
+  # docs/CODEOWNERS-only ones. A job-level `if:` on a matrix job prevents the
+  # matrix from expanding, so the per-cell required checks are never posted
+  # and the PR sits BLOCKED. Keep the gate at the step level instead — the
+  # matrix still expands and each cell posts a (no-op) success.
   tests:
     needs: [define-matrix, changes]
-    if: needs.changes.outputs.code == 'true'
     strategy:
       fail-fast: true
       matrix:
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yaml
```diff
@@ -35,17 +39,21 @@ jobs:
...
         run: |
           pip install -U click==8.2.1 hatch==1.15.1 virtualenv==20.39.1
           hatch env create default
-      - name: Test with hatch
+      - if: needs.changes.outputs.code == 'true'
+        name: Test with hatch
         run: hatch run test-cov
-      - uses: codecov/codecov-action@18283e04ce6e62d37312384ff67231eb8fd56d24
+      - if: needs.changes.outputs.code == 'true'
+        uses: codecov/codecov-action@18283e04ce6e62d37312384ff67231eb8fd56d24
```
sfc-gh-prafalski, line 56: for future reference: 
> On docs-only PRs the codecov/codecov-action step is now skipped (if: needs.changes.outputs.code == 'true'). If Codecov is configured to post a required commit status (codecov/project, codecov/patch), it will never report on a docs-only PR → same Waiting deadlock, on a check this PR doesn't touch. Confirm Codecov isn't in the required set (or is set to non-blocking / informational).

sfc-gh-prafalski (approved):
LGTM

PR #3096, opened 2026-06-08
feat: [SNOW-3563127] add --env and --env-vars to snow dbt execute

sfc-gh-yuding:
### Pre-review checklist

### Changes description
`--env` selects a named environment from `env.yml` at execution time (`NO_ENV` skips the file entirely). `--env-vars` accepts a YAML/JSON object of overrides applied on top of the selected environment, mirroring `dbt run --vars` syntax. Values must be strings; numbers, booleans, null, nested objects, and arrays are rejected. Inline `DBT_ENV_SECRET_*` keys emit a warning pointing users at the safer `secrets:` block in `env.yml`.

The CLI flag is `--env` rather than `--environment` because `--environment` is a legacy alias of the global `ConnectionOption` and would be intercepted as a connection lookup. The emitted SQL is unchanged: `ENVIRONMENT='...'`.

## Help text & feature-flag gating

| # | Command | Behavior | Expected? |
|---|---|---|---|
| H1 | `snow dbt execute build --help` (flag off) | `--env` and `--env-vars` NOT in help output | Yes - hidden by default |
| H2 | `SNOWFLAKE_CLI_FEATURES_ENABLE_DBT_PROJECT_ENV_VARS=true snow dbt execute build --help` | Both flags appear in help | Yes - feature flag works |
| F1 | `snow dbt execute --env=dev <name> run` (flag NOT set) | Runs successfully, applies env.yml dev block | Yes - `hidden=True` only affects help rendering |

## End-to-end (real `EXECUTE DBT PROJECT` against qa6)

Fixture project reads `DBT_FOO` and `DBT_BAR` via dbt `env_var()` with `"unset"` default. `env.yml` defines `dev` (`DBT_FOO=dev_foo`, `DBT_BAR=dev_bar`) and `prod` (`DBT_FOO=prod_foo`, `DBT_BAR=prod_bar`). `default_environment: dev`.

| # | Command | Resulting `(foo, bar)` | Expected? |
|---|---|---|---|
| 13 | `snow dbt execute --env=prod <name> run` | `("prod_foo", "prod_bar")` | Yes - prod block applied |
| 14 | `snow dbt execute --env=prod --env-vars '{"DBT_FOO": "override_foo"}' <name> run` | `("override_foo", "prod_bar")` | Yes - `--env-vars` overrides `DBT_FOO`; `DBT_BAR` still from env.yml |
| 15 | `snow dbt execute --env=NO_ENV <name> run` | `("unset", "unset")` | Yes - env.yml skipped, `env_var()` default fires |
| 16 | `snow dbt execute --env-vars '{"DBT_FOO": "only_foo"}' <name> run` (no `--env`) | `("only_foo", "dev_bar")` | Yes - `default_environment: dev` provided `DBT_BAR`; override applied `DBT_FOO` |
| 17 | `snow dbt execute --env-vars "{DBT_FOO: 'yaml_foo', DBT_BAR: 'yaml_bar'}" <name> run` | `("yaml_foo", "yaml_bar")` | Yes - YAML syntax with quoted scalars accepted |
| NEW | `snow dbt execute --env-vars '{}' <name> run` | `("dev_foo", "dev_bar")` | Yes - empty dict treated as silent no-op; falls back to `default_environment: dev` |

## Client-side validation (no server interaction)

| # | Command | Behavior | Expected? |
|---|---|---|---|
| V1 | `--env-vars '{not: valid: yaml: at: all'` | Error: `--env-vars must be valid YAML/JSON: ...` | Yes |
| V2 | `--env-vars '"just_a_string"'` | Error: `--env-vars must be a YAML/JSON object, got str` | Yes |
| V3 | `--env-vars '{"DBT_X": null}'` | Error: `value for 'DBT_X' must not be null` | Yes |
| V4 | `--env-vars '{"DBT_X": 1}'` | Error: `must be a string, got int; quote scalars in YAML/JSON` | Yes |
| V5 | `--env-vars '{"FOO": "1"}'` | Error: `key 'FOO' must start with 'DBT_'` | Yes |
| V6 | `--env-vars '{"DBT-FOO": "1"}'` | Error: `must contain only ASCII letters, digits, and underscores` | Yes |
| V7 | `--env-vars '{"DBT_FOO": "1", "DBT_FOO": "2"}'` | Error: `--env-vars contains a duplicate key: duplicate key 'DBT_FOO'` | Yes |
| V8 | `--env-vars '{"": "v"}'` | Error: `--env-vars key must not be empty` | Yes |
| V9 | `--env-vars '{"DBT_X": {"nested": "1"}}'` | Error: `value for 'DBT_X' must be a string, got dict` | Yes |
| NEW V10 | `--env-vars '{"DBT_X": "value\nwith\nnewlines"}'` (JSON `\n` parses to actual newlines) | Error: `value for 'DBT_X' must not contain control characters (newlines, tabs, etc.)` | Yes - control-char rejection in `_parse_env_vars` |
| NEW V11 | `--env=$'dev\nprod'` (bash `$'...'` interprets `\n` as actual newline) | Error: `--env must not contain control characters (newlines, tabs, etc.)` | Yes - control-char rejection in `_env_callback` (command layer) |

## Other

| # | Command | Behavior | Expected? |
|---|---|---|---|
| Sec | `--env-vars '{"DBT_ENV_SECRET_TOKEN": "raw_xyz"}'` | Warning printed: `--env-vars contains key(s) with the DBT_ENV_SECRET_ prefix (DBT_ENV_SECRET_TOKEN); these values will appear in the SQL text and query history. To avoid that, use the secrets: block in env.yml referencing a Snowflake SECRET object.` Then proceeds. | Yes - NEW reworded text per review comment #9 |
| Async | `--run-async --env-vars '{"DBT_FOO": "async_foo"}'` | `Command submitted. ... query_id ...` | Yes - async path works with new flags |

### DIFF
--- a/src/snowflake/cli/_plugins/dbt/commands.py
+++ b/src/snowflake/cli/_plugins/dbt/commands.py
@@ -30,6 +30,7 @@ from snowflake.cli._plugins.dbt.constants import (
 from snowflake.cli._plugins.dbt.manager import (
     DBTDeployAttributes,
     DBTManager,
+    _reject_control_chars,
 )
 from snowflake.cli._plugins.object.command_aliases import add_object_command_aliases
 from snowflake.cli._plugins.object.commands import scope_option
@@ -40,6 +41,7 @@ from snowflake.cli.api.commands.snow_typer import SnowTyperFactory
 from snowflake.cli.api.console.console import cli_console
 from snowflake.cli.api.constants import ObjectType
 from snowflake.cli.api.exceptions import CliError
+from snowflake.cli.api.feature_flags import FeatureFlag
 from snowflake.cli.api.identifiers import FQN
 from snowflake.cli.api.output.types import (
     CommandResult,
@@ -86,6 +88,10 @@ add_object_command_aliases(
 
 
 SEMANTIC_VERSION_PATTERN = re.compile(r"^\d+\.\d+\.\d+(-[a-zA-Z0-9]+)?$")
+
+
+def _env_callback(value: Optional[str]) -> Optional[str]:
+    return _reject_control_chars(value, "--env")
 
 
 class SemanticVersionType(click.ParamType):
@@ -208,6 +214,8 @@ def before_callback(
         None,
         "--env",
         show_default=False,
+        callback=_env_callback,
+        hidden=not FeatureFlag.ENABLE_DBT_PROJECT_ENV_VARS.is_enabled(),
         help="Selects the target environment from env.yml at execution time. "
         "Use 'NO_ENV' to skip env.yml entirely.",
     ),
@@ -215,14 +223,15 @@ def before_callback(
         None,
         "--env-vars",
         show_default=False,
+        hidden=not FeatureFlag.ENABLE_DBT_PROJECT_ENV_VARS.is_enabled(),
         help="Environment variable overrides as a YAML/JSON object, e.g. "
         '\'{"DBT_FOO": "1", "DBT_BAR": "2"}\'. '
         "Values must be strings; numbers, booleans, null, nested objects, "
         "and arrays are rejected (quote scalars, e.g. 'DBT_FOO: \"1\"'). "
         "Keys must start with 'DBT_' and contain only letters, digits, and "
-        "underscores. Values with the DBT_ENV_SECRET_ prefix are accepted "
-        "but appear in the SQL text and query history; for real secrets "
-        "prefer the secrets: block in env.yml.",
+        "underscores. Variables with the DBT_ENV_SECRET_ prefix are accepted "
+        "but appear in the SQL text and query history; to avoid that, use "
+        "the secrets: block in env.yml.",
     ),
     **options,
 ):
--- a/src/snowflake/cli/_plugins/dbt/commands.py
+++ b/src/snowflake/cli/_plugins/dbt/commands.py
@@ -204,6 +204,26 @@ def before_callback(
         show_default=False,
         help="dbt Core version to use for execution (ephemeral, does not change project configuration). Full list of supported versions can be found at https://docs.snowflake.com/en/user-guide/data-engineering/dbt-projects-on-snowflake-dbt-core-versions",
     ),
+    environment: Optional[str] = typer.Option(
+        None,
+        "--env",
+        show_default=False,
+        help="Selects the target environment from env.yml at execution time. "
+        "Use 'NO_ENV' to skip env.yml entirely.",
+    ),
+    env_vars: Optional[str] = typer.Option(
+        None,
+        "--env-vars",
+        show_default=False,
+        help="Environment variable overrides as a YAML/JSON object, e.g. "
+        '\'{"DBT_FOO": "1", "DBT_BAR": "2"}\'. '
+        "Values must be strings; numbers, booleans, null, nested objects, "
+        "and arrays are rejected (quote scalars, e.g. 'DBT_FOO: \"1\"'). "
+        "Keys must start with 'DBT_' and contain only letters, digits, and "
+        "underscores. Values with the DBT_ENV_SECRET_ prefix are accepted "
+        "but appear in the SQL text and query history; for real secrets "
+        "prefer the secrets: block in env.yml.",
+    ),
     **options,
 ):
     """Handles global options passed before the command and takes pipeline name to be accessed through child context later."""
@@ -228,7 +248,17 @@ for cmd in DBT_COMMANDS:
         name = FQN.from_string(ctx.parent.params["name"])
         run_async = ctx.parent.params["run_async"]
         dbt_version = ctx.parent.params.get("dbt_version")
-        execute_args = (dbt_command, name, run_async, dbt_version, *dbt_cli_args)
+        environment = ctx.parent.params.get("environment")
+        env_vars = ctx.parent.params.get("env_vars")
+        execute_args = (
+            dbt_command,
+            name,
+            run_async,
+            dbt_version,
+            environment,
+            env_vars,
+            *dbt_cli_args,
+        )
         dbt_manager = DBTManager()
 
         if run_async is True:
--- a/src/snowflake/cli/_plugins/dbt/manager.py
+++ b/src/snowflake/cli/_plugins/dbt/manager.py
@@ -29,6 +29,7 @@ from snowflake.cli.api.console import cli_console
 from snowflake.cli.api.constants import DEFAULT_SIZE_LIMIT_MB, ObjectType
 from snowflake.cli.api.exceptions import CliError
 from snowflake.cli.api.identifiers import FQN
+from snowflake.cli.api.project.util import to_string_literal
 from snowflake.cli.api.secure_path import SecurePath
 from snowflake.cli.api.sql_execution import SqlExecutionMixin
 from snowflake.connector.cursor import SnowflakeCursor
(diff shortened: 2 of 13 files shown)


tests_integration/test_dbt.py
```diff
@@ -707,3 +707,92 @@ def test_execute_with_dbt_version(
...
+    Idempotent and parallel-safe — every xdist worker writes the same value,
+    and the parameter is intentionally not unset on teardown (see
+    tests_integration/apps/test_snowflake_apps.py for the same pattern).
+    """
+    rows = snowflake_session.execute_string("SELECT CURRENT_USER()")
+    user = rows[-1].fetchone()[0]
+    snowflake_session.execute_string(f"ALTER USER {user} SET ENABLE_DBT_ENV_VARS=TRUE")
+    yield
+
+
```
sfc-gh-yuding: TODO: check with snowcli team, how to enable GS parameter which gates the feature I am testing against.

tests_integration/test_dbt.py
```diff
@@ -707,3 +707,79 @@ def test_execute_with_dbt_version(
         assert result.exit_code == 0, result.output
         assert "Running with dbt=1.10.15" in result.output
         assert "Done. PASS=2 WARN=0 ERROR=0 SKIP=0 NO-OP=0 TOTAL=2" in result.output
+
+
+@pytest.mark.integration
```
sfc-gh-jwilkowski, line 712: please also add `@pytest.mark.qa_only` so that it won't run on every PR

src/snowflake/cli/_plugins/dbt/commands.py
```diff
@@ -204,6 +204,26 @@ def before_callback(
...
+    env_vars: Optional[str] = typer.Option(
+        None,
+        "--env-vars",
+        show_default=False,
+        help="Environment variable overrides as a YAML/JSON object, e.g. "
+        '\'{"DBT_FOO": "1", "DBT_BAR": "2"}\'. '
+        "Values must be strings; numbers, booleans, null, nested objects, "
+        "and arrays are rejected (quote scalars, e.g. 'DBT_FOO: \"1\"'). "
+        "Keys must start with 'DBT_' and contain only letters, digits, and "
+        "underscores. Values with the DBT_ENV_SECRET_ prefix are accepted "
```
sfc-gh-jwilkowski:
```suggestion
        "underscores. Variables with the DBT_ENV_SECRET_ prefix are accepted "
```

sfc-gh-jwilkowski: or Keys

src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -390,9 +427,73 @@ def execute(
...
+        if secret_keys:
+            cli_console.warning(
+                f"--env-vars contains key(s) with the {DBT_ENV_SECRET_PREFIX} prefix "
+                f"({', '.join(secret_keys)}); these values will appear in the SQL "
+                f"text and query history. For real secrets, declare them in the "
+                f"secrets: block of env.yml referencing a Snowflake SECRET object."
+            )
+
+        def _escape(s: str) -> str:
+            return s.replace("'", "\\'")
```
sfc-gh-jwilkowski: please use `to_string_literal()` for escaping

src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -390,9 +427,73 @@ def execute(
         query = f"EXECUTE DBT PROJECT {name}"
         if dbt_version:
             query += f" dbt_version='{dbt_version}'"
+        if environment:
+            environment_escaped = environment.replace("'", "\\'")
+            query += f" ENVIRONMENT='{environment_escaped}'"
```
sfc-gh-jwilkowski: please use `to_string_literal()` for escaping. Also it might be worth adding some extra validations for what this string might actually be (ie remove newlines, control characters etc), ideally in command layer

src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -390,9 +427,73 @@ def execute(
...
+        if env_vars_clause:
+            query += env_vars_clause
         query += f" args='{dbt_command_escaped}'"
         return self.execute_query(query, _exec_async=run_async)
 
+    @staticmethod
+    def _format_env_vars_clause(env_vars: Optional[str]) -> str:
+        if not env_vars:
+            return ""
+        pairs = DBTManager._parse_env_vars(env_vars)
```
sfc-gh-jwilkowski, line 453: is that an acceptable value? `--env-vars '{}'`

tests/dbt/test_dbt_commands.py
```diff
@@ -723,3 +723,281 @@ def test_dbt_execute_with_prerelease_version(
...
+                "dbt",
+                "execute",
+                "--env-vars",
+                raw_value,
+                "pipeline_name",
+                "run",
+            ]
+        )
+
+        assert result.exit_code != 0
```
sfc-gh-jwilkowski: I'd suggest asserting on exact exit code. This is part of our contract and can be used in downstream scripts, so we'd prefer to lock it

src/snowflake/cli/_plugins/dbt/commands.py
```diff
@@ -204,6 +204,26 @@ def before_callback(
...
+        "but appear in the SQL text and query history; for real secrets "
```
sfc-gh-jwilkowski: This feels a bit unclear - what are real secrets? What are not real secrets? What do you think about something like: 
```
Values with the DBT_ENV_SECRET_ prefix are accepted but appear in the SQL text and query history; to avoid that, use `secrets: block` in env.yml.
```

src/snowflake/cli/_plugins/dbt/commands.py
```diff
@@ -204,6 +204,26 @@ def before_callback(
...
         help="dbt Core version to use for execution (ephemeral, does not change project configuration). Full list of supported versions can be found at https://docs.snowflake.com/en/user-guide/data-engineering/dbt-projects-on-snowflake-dbt-core-versions",
     ),
+    environment: Optional[str] = typer.Option(
+        None,
+        "--env",
+        show_default=False,
+        help="Selects the target environment from env.yml at execution time. "
+        "Use 'NO_ENV' to skip env.yml entirely.",
+    ),
+    env_vars: Optional[str] = typer.Option(
```
sfc-gh-jwilkowski, line 222: As far as I understand those 2 should be hidden behind feature flags on CLI side as well before this functionality is live? See https://github.com/snowflakedb/snowflake-cli/blob/main/docs/contributing/lifecycle.md#feature-flags

sfc-gh-jwilkowski (changes requested):
Leaving a couple of comments to look at.

src/snowflake/cli/_plugins/dbt/manager.py
sfc-gh-yuding: Yes, `--env-vars '{}'` is acceptable, same as in SQL query, we accept empty `ENV_VARS`

src/snowflake/cli/_plugins/dbt/commands.py
```diff
@@ -86,6 +87,20 @@
 
 
 SEMANTIC_VERSION_PATTERN = re.compile(r"^\d+\.\d+\.\d+(-[a-zA-Z0-9]+)?$")
+_CONTROL_CHAR_RE = re.compile(r"[\x00-\x1f\x7f]")
```
sfc-gh-jwilkowski: I see the same pattern in manager.py - would be great to have only 1.

src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -390,9 +429,77 @@ def execute(
         query = f"EXECUTE DBT PROJECT {name}"
         if dbt_version:
             query += f" dbt_version='{dbt_version}'"
+        if environment:
+            query += f" ENVIRONMENT={to_string_literal(environment)}"
+        env_vars_clause = self._format_env_vars_clause(env_vars)
+        if env_vars_clause:
+            query += env_vars_clause
```
sfc-gh-jwilkowski, line 445: I just want to make sure - If I happen to have env.yml in my project, it won't be used once this PR gets merged? `--env` and `--env-vars` will default to None, so nothing will be loaded behind the scenes?

sfc-gh-jwilkowski (dismissed):
PR looks good to me! I've left one suggestion and a question, so please make that change in #3097 after resolving conflicts from this PR. Thanks!

src/snowflake/cli/_plugins/dbt/manager.py
sfc-gh-yuding: After the PR merged, when --env and --env-vars are unset, the CLI emits no ENVIRONMENT= and no ENV_VARS=() clauses on EXECUTE DBT PROJECT. So the CLI itself is loading nothing.

But on server-side, when the GS param gating this feature is enabled on the account, even if EXECUTE DBT PROJECT with no ENVIRONMENT= and no ENV_VARS=() clauses, it will read `env.yml` from the staged project and use env_config.default_environment in `env.yml` for the dbt subprocess.

sfc-gh-jwilkowski (approved):

PR #3097, opened 2026-06-08
feat: [SNOW-3444080] add env.yml support to snow dbt deploy

sfc-gh-yuding:
### Pre-review checklist

### Changes description
Adds three flags to `snow dbt deploy` that bring env.yml into the deploy flow alongside profiles.yml:

- `--env-file-dir <path>`: inject env.yml from a separate directory into the staged project root, mirroring `--profiles-dir`.
- `--default-env <name>`: set DEFAULT_ENVIRONMENT on the DBT PROJECT object (CREATE / CREATE OR REPLACE / ALTER SET).
- `--unset-default-env`: clear DEFAULT_ENVIRONMENT on alter, mutually exclusive with --default-environment.

env.yml is staged like profiles.yml: copied from the source directory by default, optionally overridden via `--env-file-dir`, and round-tripped through pyyaml so YAML comments don't leak onto the stage. The CLI does no client-side schema validation — the server enforces env-var prefixes, secret-key rules, and reserved environment names.

Server support for these properties is gated by ENABLE_DBT_ENV_VARS (GS PR #404443).
                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                            
 Tested live on **qa6** (`ENABLE_DBT_ENV_VARS = true`)

| # | Command | Behavior | Expected? |
|---|---|---|---|
| 1 | `snow dbt deploy SMOKE --source ./project --env-file-dir ./envs --default-env dev --unset-default-env` | exit 2; mutex error `Parameters '--unset-default-env' and '--default-env' are incompatible and cannot be used simultaneously.`; no SQL sent | ✅ |
| 2 | `snow dbt deploy SMOKE --source ./project --env-file-dir ./empty_envs` | `CliError: env.yml does not exist in directory ./empty_envs.`; no stage created | ✅ |
| 3 | `snow dbt deploy SMOKE --source ./project --env-file-dir ./dup_envs --default-env dev` (env.yml has `DBT_FOO` twice) | `CliError: env.yml has duplicate key 'DBT_FOO'`; **fails before "Creating temporary stage"** (no network roundtrip) | ✅ |
| 4 | `snow dbt deploy SMOKE --source ./project --env-file-dir ./bad_envs --default-env dev` (malformed YAML) | `CliError: env.yml is not valid YAML: mapping values are not allowed here …`; fails before stage creation | ✅ |
| 5 | `snow dbt deploy SMOKE --source ./project --env-file-dir ./envs --default-env staging` (no `staging` block in env.yml) | server rejects: `Invalid env.yml: environment 'staging' not found.` | ✅ |
| 6 | `snow dbt deploy SMOKE --source ./src_dup --default-env dev` (duplicate-key env.yml in the **source dir**, no `--env-file-dir`) | `CliError: env.yml has duplicate key 'DBT_FOO'`; fails before stage creation — implicit env.yml is sanitized/validated just like an explicit one | ✅ |
| 7 | `snow dbt deploy SMOKE --source ./src_bad --default-env dev` (malformed env.yml in the **source dir**, no `--env-file-dir`) | `CliError: env.yml is not valid YAML: …`; fails before stage creation | ✅ |
| 8 | `snow dbt deploy SMOKE --source ./project --env-file-dir ./lc_envs --default-env dev` (lowercase var name `dbt_foo`) | server rejects: `Invalid env.yml: 'dbt_foo' in env_config.environments['dev'].env must start with 'DBT_'.` (variable-name rules enforced server-side, by design) | ✅ |
| 9 | `snow dbt deploy SMOKE --source ./project --env-file-dir ./envs --default-env dev` (first deploy) | `CREATE DBT PROJECT … DEFAULT_ENVIRONMENT='dev'`; `dbt list` → `default_environment: "dev"` | ✅ |
| 10 | `snow dbt deploy SMOKE --source ./project --env-file-dir ./envs --default-env prod` (redeploy) | `ALTER DBT PROJECT … SET DEFAULT_ENVIRONMENT='prod'` + `ADD VERSION`; `default_environment: "prod"` | ✅ |
| 11 | `snow dbt deploy SMOKE --source ./project --env-file-dir ./envs --unset-default-env` (redeploy) | `ALTER DBT PROJECT … UNSET DEFAULT_ENVIRONMENT` + `ADD VERSION`; `default_environment: null` | ✅ |
| 12 | `snow dbt deploy SMOKE --source ./project --env-file-dir ./envs --unset-default-env` (redeploy when already null) | UNSET still emitted; server treats as no-op; `default_environment: null` | ✅ |
| 13 | `snow dbt deploy SMOKE --source ./project --env-file-dir ./envs --default-env dev --force` | `CREATE OR REPLACE DBT PROJECT … DEFAULT_ENVIRONMENT='dev'`; `default_environment: "dev"` | ✅ |
| 14 | `snow dbt deploy SMOKE --source ./project --env-file-dir ./envs --default-env dev --default-target prod` | `CREATE DBT PROJECT … DEFAULT_TARGET='prod' DEFAULT_ENVIRONMENT='dev'`; both fields set | ✅ |
| 15 | `snow dbt deploy SMOKE --source ./project_with_env --env-file-dir ./override_envs --default-env prod` (source env.yml has only `dev`; override has only `prod`) | succeeds; `default_environment: "prod"` — `--env-file-dir` overrides the source's env.yml on the stage | ✅ |
| 16 | `snow dbt deploy SMOKE --source ./project_with_env --default-env dev` (clean env.yml inside source dir, no `--env-file-dir`) | succeeds; `default_environment: "dev"` — env.yml auto-staged from the source dir | ✅ |
| 17 | `snow dbt deploy SMOKE --source ./project --env-file-dir ./envs --default-env ""` (empty string) | succeeds; `default_environment: null` — empty value treated as not-provided (consistent with `--default-target ""`, `--dbt-version ""`) | ✅ |
| 18 | `snow dbt deploy SMOKE --source ./project --env-file-dir ./envs --default-env $'dev\nX'` (control char) | `CliError: --default-env must not contain control characters (newlines, tabs, etc.)`; no SQL sent | ✅ |
| 19 | `snow dbt deploy SMOKE --source ./project --env-file-dir ./nope_xyz --default-env dev` (missing dir) | `CliError: env.yml does not exist in directory ./nope_xyz.` | ✅ |
| 20 | `snow dbt deploy SMOKE --source ./project` (no env.yml anywhere, no env flags) | succeeds; `default_environment: null` — env.yml is optional, deploy proceeds without one | ✅ |
| 21 | `snow dbt deploy --help` with feature flag OFF vs ON | flag OFF → `--default-env`, `--unset-default-env`, `--env-file-dir` hidden (0 matches); flag ON → all three visible (3 matches) | ✅ |


sfc-gh-jwilkowski (changes requested):
Please apply similar changes as listed in https://github.com/snowflakedb/snowflake-cli/pull/3096#pullrequestreview-4458703064

src/snowflake/cli/_plugins/dbt/commands.py
```diff
@@ -139,6 +164,20 @@ def deploy_dbt(
     unset_default_target: Optional[bool] = UnsetDefaultTargetOption(
         help="Unset the default target for the dbt project. Mutually exclusive with --default-target.",
     ),
+    default_environment: Optional[str] = DefaultEnvironmentOption(
+        help=(
+            f"Default environment for the dbt project (DEFAULT_ENVIRONMENT). "
```
sfc-gh-jwilkowski: I don't understand why that info needs to be duplicated?
```suggestion
            f"Default environment for the dbt project. "
```

src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -421,6 +452,17 @@ def _prepare_profiles_file(profiles_path: Path, tmp_path: Path):
...
+    @staticmethod
+    def _prepare_env_file(env_path: Path, tmp_path: Path):
+        source_env_file = SecurePath(env_path / ENV_FILENAME)
+        target_env_file = SecurePath(tmp_path / ENV_FILENAME)
+        if target_env_file.exists():
+            target_env_file.unlink()
+        with source_env_file.open(
+            read_file_limit_mb=DEFAULT_SIZE_LIMIT_MB
+        ) as sfd, target_env_file.open(mode="w") as tfd:
+            yaml.safe_dump(yaml.safe_load(sfd), tfd)
```
sfc-gh-jwilkowski: Shouldn't you use `_NoDuplicatesSafeLoader`?

tests_integration/test_dbt.py
```diff
@@ -373,6 +384,99 @@ def test_deploy_with_default_target(
...
+        env_yml = {
+            "env_config": {
+                "default_environment": "dev",
+                "environments": [
+                    {"name": "dev", "env": {"DBT_FOO": "bar"}},
+                    {"name": "prod", "env": {"DBT_FOO": "baz"}},
+                ],
+            }
+        }
+        (root_dir / ENV_FILENAME).write_text(yaml.dump(env_yml))
```
sfc-gh-jwilkowski, line 413: is this needed when we test only setting the attribute on dbt project level here?

sfc-gh-jwilkowski (commented):
Mostly looks good to me, but I'm leaving some questions. Please rebase your changes as going through changes from this PR and 3097 makes it more difficult to understand :)

src/snowflake/cli/_plugins/dbt/manager.py
sfc-gh-yuding: Yes, we should use _NoDuplicatesSafeLoader here. Added a test case for duplicate key in env.yml as well.

tests_integration/test_dbt.py
sfc-gh-yuding: Yes, we need this env.yaml setup, given server-side validation enforces that DEFAULT_ENVIRONMENT='dev' requires an env.yml containing an environment named dev on the stage. Without it, CREATE fails with `Invalid env.yml: environment 'dev' not found`.

sfc-gh-jwilkowski: I think this tests shows a potentially dangerous behaviour - as a user I'm not adding any flags/args that would tell the app to upload env.yaml that I might for some reason have in my project directory. `_prepare_env_file` for example takes care of duplicates, but it runs only when I provide `env_file_path` flag. In case of this test I think that if my env.yml contained duplicate keys, the deploy would proceed and then something would fail on server side.

2 questions then:
1) shouldn't `_prepare_env_file()` always run if we can find any env.yaml (even if user did not explicitly provide it?)
2) Is it ok/safe to deploy your env.yml implicitly if it happens to be in your dbt source directory (my intuition says it's ok as you pointed at this directory or ran snow dbt deploy from it, knowing it'll be uploading files)

src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -227,6 +240,8 @@ def deploy(
                 tmp_path = Path(tmp)
                 stage_manager.copy_to_tmp_dir(path.path, tmp_path)
                 self._prepare_profiles_file(profiles_path.path, tmp_path)
+                if env_file_path is not None:
+                    self._prepare_env_file(env_file_path.path, tmp_path)
```
sfc-gh-jwilkowski: notice that validations of dbt version or profiles are handled before copying files or creating temp stage. We might raise errors faster without network roundtrips if we moved some validations before this context

tests_integration/test_dbt.py
sfc-gh-yuding:
```
shouldn't _prepare_env_file() always run if we can find any env.yaml (even if user did not explicitly provide it?)
```
Yes!! That makes sense, thanks for pointing it out, otherwise the behavior will be inconsistent.

```
Is it ok/safe to deploy your env.yml implicitly if it happens to be in your dbt source directory (my intuition says it's ok as you pointed at this directory or ran snow dbt deploy from it, knowing it'll be uploading files)
``` 
Yes, I think it is ok. That is the exact same behavior profile.yml, it will be implicitly deployed too as long as it is in the source folder and no --profiles-dir specified.

src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -511,7 +567,7 @@ def _parse_env_vars(raw: str) -> Dict[str, str]:
         try:
             parsed = yaml.load(raw, Loader=_NoDuplicatesSafeLoader)
         except yaml.constructor.ConstructorError as e:
-            raise CliError(f"--env-vars contains a duplicate key: {e.problem}")
+            raise CliError(f"--env-vars has {e.problem}")
```
sfc-gh-jwilkowski: What are the examples of such issues? I'm wondering if the error strings are readable for regular user? Otherwise we could see `Error: --env-vars has foobarbaz`. I liked the previous phrasing more that told you what might be the source of issue and provided extra detail:
```
raise CliError(f"Failed to parse value from --env-vars: {e.problem}")
```

And to be quite honest the error below is also a bit misleading -

sfc-gh-yuding: I removed the `contains a duplicate key` because `{e.problem}` will already include some message like `duplicate key 'DBT_FOO'`. Sounds good, I will change to this error message.

sfc-gh-jwilkowski (approved):

PR #3100, opened 2026-06-09
feat: add --server-session-keep-alive flag for connections

sfc-gh-rgaunker:
## Summary

- Add new `--server-session-keep-alive` CLI flag that prevents Snowflake from closing idle sessions
- Works with both named connections and temporary connections (`-x`)
- Can also be set via `SNOWFLAKE_SERVER_SESSION_KEEP_ALIVE` env var
- Passes `server_session_keep_alive=True` to the Snowflake connector

## Test plan

- [x] Unit tests pass (7572 passed in docker)
- [x] Verified flag appears in `--help` output
- [x] Tested with temporary connection: `snow sql -q "SELECT 1" -x --server-session-keep-alive ...`
- [x] Verified connector logs `Parameter server_session_keep_alive was set to True - skipping session logout`

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -46,6 +46,7 @@
 ## New additions
 * Added a `--protocol` option (plus matching `SNOWFLAKE_PROTOCOL` env var and `protocol` config key) to `snow connection add` and the global connection overrides. This allows selecting `http` or `https` as the connection protocol without editing `config.toml`, which is primarily useful for local development against `http` deployments.
 * Added `snow helpers generate-project-schema` command to emit a JSON Schema for the `snowflake.yml` project definition file. The output follows [JSON Schema Draft 2020-12](https://json-schema.org/draft/2020-12/schema) and can be plugged into supported editors (e.g. VS Code via the YAML extension) or CI pipelines to get completion and to catch structural mistakes (unknown keys, wrong types, missing required fields) before a deploy. The schema is generated from the CLI's own models, so some cross-field and semantic checks are still only applied at load/deploy time. Use `--definition-version` to select the project definition version (`1`, `1.1`, or `2`; defaults to `2`) and `--output-file`/`-o` to write the schema to a file.
+* Added `--server-session-keep-alive` global connection flag (plus matching `SNOWFLAKE_SERVER_SESSION_KEEP_ALIVE` env var and `server_session_keep_alive` config key) that prevents Snowflake from closing idle sessions. Useful for long-running operations or connections held open between multiple operations.
 
 ## Fixes and improvements
 * Fixed `TooManyFilesError` during `snow streamlit deploy` when `main_file` is a descendant of a directory listed in `artifacts`.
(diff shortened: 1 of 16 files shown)


src/snowflake/cli/_app/snow_connector.py
```diff
@@ -90,6 +90,7 @@
     "oauth_enable_single_use_refresh_tokens",
     "client_store_temporary_credential",
     "secondary_roles",
+    "server_session_keep_alive",
```
sfc-gh-jwilkowski, line 93: Those definitions are also repeated in src/snowflake/cli/api/config_ng/sources.py. Please update _ENV_CONFIG_KEYS and _BOOLEAN_ENV_CONFIG_KEYS there as well

sfc-gh-jwilkowski (changes requested):
Thank you for this PR I've left some inline comments:
1) would be great to reflect those changes to config_ng module
2) `snow connection add` (src/snowflake/cli/_plugins/connection/commands.py) guides users through connection setup. Adding `server session keep alive` there would be consistent IMO
3) would be great to add a unit test that at least checks that this connection parameter is properly sent to connector. 
4) In `_avoid_closing_the_connection_if_it_was_shared()` we implicitly set `server_session_keep_alive` to True. If for any reason users would invoke cli with session/master token and server_session_keep_alive=false, then we'd overwrite it without any warning. I think that at least a log message/warning should be added in such case

RELEASE-NOTES.md
```diff
@@ -43,6 +43,7 @@
...
 * Added `snow helpers generate-project-schema` command to emit a JSON Schema for the `snowflake.yml` project definition file. The output follows [JSON Schema Draft 2020-12](https://json-schema.org/draft/2020-12/schema) and can be plugged into supported editors (e.g. VS Code via the YAML extension) or CI pipelines to get completion and to catch structural mistakes (unknown keys, wrong types, missing required fields) before a deploy. The schema is generated from the CLI's own models, so some cross-field and semantic checks are still only applied at load/deploy time. Use `--definition-version` to select the project definition version (`1`, `1.1`, or `2`; defaults to `2`) and `--output-file`/`-o` to write the schema to a file.
+* Added `--server-session-keep-alive` global connection flag (plus matching `SNOWFLAKE_SERVER_SESSION_KEEP_ALIVE` env var and `server_session_keep_alive` config key) that prevents Snowflake from closing idle sessions. Useful for long-running operations or connections held open between multiple operations.
```
sfc-gh-jwilkowski: You're adding this release note to an already released version, please move it down to `unreleased version` section

sfc-gh-jwilkowski (commented):
1 small change requested

RELEASE-NOTES.md
sfc-gh-rgaunker: addressed. Thanks

sfc-gh-jwilkowski (approved):
LGTM, thank you!

PR #3102, opened 2026-06-10
Snowflake Apps: skip system-default parameters in fetch_snow_apps_parameters

sfc-gh-mborins:
## Summary

- `SHOW PARAMETERS` returns a non-empty `value` with an empty `level` when a parameter has never been explicitly set (system default). The CLI was treating these as admin-configured values, causing `build_compute_pool` and `service_compute_pool` to appear in `snow app setup --dry-run` output even after `ALTER ACCOUNT UNSET`.
- Fix: skip rows where `level` is empty in `fetch_snow_apps_parameters`. Only `ACCOUNT` or `USER` level values are treated as explicitly configured.
- Updated existing tests to include `level` in mock cursor rows; added `test_ignores_system_default_level_parameters` to cover the regression directly.

## Test plan

- [x] `uv run pytest tests/apps/test_commands.py::TestFetchSnowAppsParameters` passes (7/7)
- [x] Run `snow app setup --dry-run` on an account where compute pool params have been unset: neither field should appear in output

### DIFF
--- a/src/snowflake/cli/_plugins/apps/manager.py
+++ b/src/snowflake/cli/_plugins/apps/manager.py
@@ -1093,8 +1093,15 @@ class SnowflakeAppManager(SqlExecutionMixin):
             for row in cursor:
                 param_name = (row.get("key") or row.get("KEY") or "").upper()
                 param_value = row.get("value") or row.get("VALUE") or ""
+                # Skip parameters at the system-default level. Snowflake
+                # returns an empty string for ``level`` when a parameter has
+                # never been explicitly set at the account or user level;
+                # a non-empty ``value`` in that case is merely the built-in
+                # default (e.g. ``SYSTEM_COMPUTE_POOL_CPU``) and should not
+                # be treated as an admin-configured value.
+                param_level = row.get("level") or row.get("LEVEL") or ""
                 mapped_key = _SNOW_APPS_PARAM_MAP.get(param_name)
-                if mapped_key and param_value:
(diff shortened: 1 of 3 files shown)


src/snowflake/cli/_plugins/apps/manager.py
```diff
@@ -1093,8 +1093,15 @@ def fetch_snow_apps_parameters(self) -> Dict[str, str]:
...
+                # Skip parameters at the system-default level. Snowflake
+                # returns an empty string for ``level`` when a parameter has
+                # never been explicitly set at the account or user level;
+                # a non-empty ``value`` in that case is merely the built-in
+                # default (e.g. ``SYSTEM_COMPUTE_POOL_CPU``) and should not
+                # be treated as an admin-configured value.
+                param_level = row.get("level") or row.get("LEVEL") or ""
                 mapped_key = _SNOW_APPS_PARAM_MAP.get(param_name)
-                if mapped_key and param_value:
+                if mapped_key and param_value and param_level:
```
sfc-gh-kbienkowski, line 1104: We should remove the SYSTEM_COMPUTE_POOL_CPU as the default on the GS side. That default is no longer valid after we moved to the managed compute pool, is it?

I'm not sure how reliable it is to depend on the level here. If it is reliable, it's reasonable to ship this patch while removing the hard-coded defaults from GS in parallel.

sfc-gh-gbloom (approved):

PR #3103, opened 2026-06-10
Add --no-prompt-exit-repl option to snow sql REPL

sfc-gh-jwilkowski:
### Pre-review checklist

### Changes description
Community contribution from @remisalmon (cherry-picked from #2758, preserving original authorship).

Closes #2758

### DIFF
--- a/src/snowflake/cli/_plugins/sql/commands.py
+++ b/src/snowflake/cli/_plugins/sql/commands.py
@@ -29,6 +29,7 @@ from snowflake.cli.api.commands.flags import (
 )
 from snowflake.cli.api.commands.overrideable_parameter import OverrideableOption
 from snowflake.cli.api.commands.snow_typer import SnowTyperFactory
+from snowflake.cli.api.config import get_config_value
 from snowflake.cli.api.exceptions import CliArgumentError
 from snowflake.cli.api.output.types import (
     CommandResult,
@@ -135,10 +136,12 @@ def execute_sql(
         ),
         is_flag=True,
     ),
-    no_prompt_exit: bool = typer.Option(
-        False,
+    no_prompt_exit: Optional[bool] = typer.Option(
+        None,
         "--no-prompt-exit",
         help="Do not prompt before exiting.",
+        envvar="SNOWFLAKE_NO_PROMPT_EXIT",
+        show_default=False,
     ),
     **options,
 ) -> CommandResult:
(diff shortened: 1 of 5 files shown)


src/snowflake/cli/_plugins/sql/repl.py
```diff
@@ -196,17 +198,6 @@ def repl_prompt(self, msg: str = " > ") -> str:
...
-    def yn_prompt(self, msg: str) -> str:
-        """Yes/No prompt."""
-        return self.session.prompt(
-            msg,
-            lexer=None,
-            completer=None,
-            multiline=False,
-            wrap_lines=False,
-            key_bindings=self._yes_no_keybindings,
-        )
```
sfc-gh-olorek, line 208: Unrelated to the `--no-prompt-exit-repl` feature: this deletes `yn_prompt` and inlines its body into `ask_yn`. It's harmless — `ask_yn` was the only caller — but it's scope creep that enlarges the diff of an otherwise small feature PR. Suggest either dropping it here or noting it explicitly as a deliberate drive-by cleanup so reviewers know it's intentional.

src/snowflake/cli/_plugins/sql/commands.py
```diff
@@ -135,6 +136,12 @@ def execute_sql(
         ),
         is_flag=True,
     ),
+    no_prompt_exit_repl: Optional[bool] = typer.Option(
+        None,
+        "--no-prompt-exit-repl",
+        help="Do not prompt before exiting the REPL.",
+        show_default=False,
+    ),
```
sfc-gh-olorek, line 144: Heads up that this makes the override one-way: it's an `Optional[bool]` with a single flag name, so once `cli.no_prompt_exit_repl = true` (config or `SNOWFLAKE_CLI_NO_PROMPT_EXIT_REPL`) is set, there's no CLI flag to re-enable the confirmation prompt for a single session. That's a defensible tradeoff — the `None` default is what drives the flag > env > config precedence, and a two-sided `--prompt-exit-repl/--no-prompt-exit-repl` pair would collide with it — just flagging it so it's a conscious decision rather than an accident.

src/snowflake/cli/_plugins/sql/repl.py
sfc-gh-jwilkowski: it's just a small refactor, I'd keep it as author intended.

src/snowflake/cli/_plugins/sql/commands.py
sfc-gh-jwilkowski: The one-way behavior is intentional. The feature exists to escape an annoying prompt; someone who has opted out globally is extremely unlikely to want to re-enable it for a single session. Adding --prompt-exit-repl would complicate the API for a use case that imo doesn't exist in practice.

sfc-gh-olorek (approved):
lgtm

PR #3105, opened 2026-06-11
SNOW-3388436 Link or copy DCM files to upload in the temporary dir

sfc-gh-lkyc:
### Pre-review checklist

### Changes description
Before, all DCM Project files were temporarily linked/copied to the `output` folder in the project folder. 
The problems with it:
- It’s confusing for users that the DCM command create `out` folder to store command output, and the `output` folder
- The `output` folder is removed after the operation, but it may happen that the folder remains - for example, in a scenario of a file upload error or a temporary stage creation error
- If users have the `output` folder with some files in their project folder (for whatever reason), these files get removed after DCM command execution - we remove customers' files 

This PR solves the problem by creating a new temporary folder at the system-determined temporary location using the `TemporaryDirectory` class.


src/snowflake/cli/_plugins/dcm/manager.py
```diff
@@ -293,13 +288,11 @@ def sync_local_files(
 
                 DCMProjectManager._report_files_to_be_deployed(plan)
 
-                cli_console.step(
-                    f"Uploading files from local {project_paths.bundle_root} directory to temporary stage."
-                )
+                cli_console.step("Uploading files to temporary stage.")
                 for result in stage_manager.put_recursive(
-                    local_path=project_paths.bundle_root,
+                    local_path=source_path.path,
```
sfc-gh-jwilkowski:
```suggestion
                    local_path=tmp_path,
```
`put_recursive` got somewhat confusing. When temp_directory is provided, there's an assumption that it contains all the files already (which could be confusing!). I'd suggest to pass tmp_path as local_path as well, so that we indicate that source_path no longer participates in this flow (files were copied from source_path to tmp path in DCMProjectManager._bundle_definition_files, right?)

tests/dcm/test_manager.py
```diff
@@ -699,8 +707,8 @@ def test_sync_local_files_collects_manifest_and_sources(
             project_identifier=TEST_PROJECT, source_directory=str(source_dir)
         )
 
-        mock_bundle_artifacts.assert_called_once()
-        artifacts = mock_bundle_artifacts.call_args.args[1]
+        mock_bundle.assert_called_once()
+        artifacts = mock_bundle.call_args.kwargs["artifacts"]
```
sfc-gh-jwilkowski, line 711: It might be worth adding an assertion to at least one test here that /output/bundle/ is indeed not used, especially in case of faulty uploads

sfc-gh-jwilkowski (changes requested):
Left some comments

tests/dcm/test_manager.py
sfc-gh-lkyc: Do you think it's worth doing it? Long-term, it won't have much value. It would be as valuable as checking that we don't create a folder with any other name that we can imagine :)

src/snowflake/cli/_plugins/dcm/manager.py
sfc-gh-lkyc: Agreed. Done. 

> (files were copied from source_path to tmp path in DCMProjectManager._bundle_definition_files, right?)

Yes

tests/dcm/test_manager.py
sfc-gh-jwilkowski: fair point, that would age badly :)

sfc-gh-jwilkowski (dismissed):
LGTM

sfc-gh-jbilek (approved):

PR #3108, opened 2026-06-12
feat: [SNOW-3444081] add --use-shell-env-vars to snow dbt execute

sfc-gh-yuding:
### Pre-review checklist

### Changes description
Adds a new --use-shell-env-vars option to `snow dbt execute` that forwards exported shell environment variables (DBT_* prefix, excluding DBT_ENV_SECRET_*) into the dbt execution context as ENV_VARS=(). Designed for CI/CD pipelines where env vars are already exported and enumerating each one with --env-vars KEY=VALUE is cumbersome.

Precedence (highest to lowest, all merged into one ENV_VARS=() clause):
  1. --env-vars (explicit per-key overrides)
  2. --use-shell-env-vars (DBT_* shell vars, excluding DBT_ENV_SECRET_*)
 
Behavior:
  * Shell-derived dict is sorted client-side for deterministic SQL.
  * DBT_ENV_SECRET_* keys are dropped from the shell with an info log; forward secrets via the env.yml `secrets:` block instead.
  * Empty-state info message lists per-shell export syntax (bash/zsh, fish, PowerShell, cmd.exe) plus a sudo -E hint.
  * Forwarded-vars warning fires once per invocation, calling out that values appear in query history and must not contain credentials, tokens, or other confidential data.

The flag is hidden behind FeatureFlag.ENABLE_DBT_PROJECT_ENV_VARS.

End-to-end against a live Snowflake account (`qa6` / DEXQA6) that supports `DBT PROJECT`. Each run started from a fresh shell with all pre-existing `DBT_*` env vars stripped. `SNOWFLAKE_CLI_FEATURES_ENABLE_DBT_PROJECT_ENV_VARS=true` was exported throughout. The same 7-case sequence was executed from **bash, zsh, sh (POSIX/dash), and fish** — 4 shells × 7 cases = 28 runs.

## Results: identical across all 4 shells

| # | Setup | Command (after `snow dbt execute -c qa6 --database $DB --schema PUBLIC`) | Server-observed row | Expected? |
|---|---|---|---|---|
| R1 | `DBT_FOO=shell_foo, DBT_BAR=shell_bar` exported | `--use-shell-env-vars --env=prod $PROJ run` | `FOO=shell_foo, BAR=shell_bar` — shell wins over env.yml's prod block (`ENV_VARS=()` overrides env.yml `env:` server-side) | ✅ |
| R2 | `DBT_FOO=shell_foo, DBT_BAR=shell_bar` exported | `--use-shell-env-vars --env-vars '{"DBT_FOO":"explicit_foo"}' --env=prod $PROJ run` | `FOO=explicit_foo, BAR=shell_bar` — explicit `--env-vars` beats shell on collision; shell still provides BAR | ✅ |
| R3 | `DBT_FOO=shell_foo` exported (no DBT_BAR) | `--use-shell-env-vars --env=NO_ENV $PROJ run` | `FOO=shell_foo, BAR=unset` — NO_ENV skips env.yml; shell still forwards; missing keys fall back to `env_var()` default | ✅ |
| R4 | `DBT_FOO=shell_foo, DBT_ENV_SECRET_TOKEN=should_not_appear` exported | `--use-shell-env-vars --env=NO_ENV $PROJ run` | `FOO=shell_foo, BAR=unset` — secret-prefixed key dropped client-side; never sent to server | ✅ |
| R5 | empty shell (no `DBT_*`) | `--use-shell-env-vars --env=prod $PROJ run` | `FOO=prod_foo, BAR=prod_bar` — empty-state info message emitted; env.yml prod block applies; pipeline succeeds | ✅ |
| R6 | `DBT_FOO=fromshell, DBT_BAR=fromshell` exported | `--use-shell-env-vars --env-vars '{"DBT_BAR":"explicit","DBT_NEW":"new"}' --env=NO_ENV $PROJ run` | `FOO=fromshell, BAR=explicit` — collision overwrites in place; new key appended (DBT_NEW visible in query history but not selected by model) | ✅ |
| R7 | `DBT_FOO=shell_foo, DBT_ENV_SECRET_TOKEN=NEVER_LEAK_<shell>` exported | `--use-shell-env-vars --env=NO_ENV $PROJ run` | `FOO=shell_foo, BAR=unset` — secret dropped; *AND* query history confirms no `EXECUTE DBT PROJECT` statement ever contained `NEVER_LEAK_*` | ✅ |

### DIFF
--- a/src/snowflake/cli/_plugins/dbt/manager.py
+++ b/src/snowflake/cli/_plugins/dbt/manager.py
@@ -56,22 +56,37 @@ def _reject_control_chars(value: Optional[str], flag_name: str) -> Optional[str]
     return value
 
 
-def _collect_shell_env_vars() -> tuple[Dict[str, str], int]:
+def _collect_shell_env_vars() -> tuple[Dict[str, str], int, int]:
     """Collect DBT_* environment variables from os.environ for --use-shell-env-vars.
 
     Returns (forwarded vars sorted by key, count of dropped DBT_ENV_SECRET_*
-    keys). Sorting makes the resulting SQL text deterministic across shells.
+    keys, count of skipped malformed keys). Only fully-uppercase, DBT_-prefixed
+    keys with valid characters and control-char-free values are forwarded —
+    matching exactly what the server accepts in ENV_VARS=(), so a malformed
+    shell var never triggers a server-side rejection of the whole run.
+    Secret-prefixed keys are detected case-insensitively and dropped (never
+    sent), so a secret cannot leak into query history. Sorting makes the
+    resulting SQL text deterministic across shells.
     """
     forwarded: Dict[str, str] = {}
     dropped_secret_count = 0
+    skipped_count = 0
     for key, value in os.environ.items():
-        if not key.startswith(_ENV_VAR_KEY_PREFIX):
+        upper = key.upper()
+        if not upper.startswith(_ENV_VAR_KEY_PREFIX):
             continue
-        if key.startswith(DBT_ENV_SECRET_PREFIX):
+        if upper.startswith(DBT_ENV_SECRET_PREFIX):
             dropped_secret_count += 1
             continue
-        forwarded[key] = value
-    return dict(sorted(forwarded.items())), dropped_secret_count
+        if (
+            key == upper
+            and _ENV_VAR_KEY_RE.match(key)
+            and not _CONTROL_CHAR_RE.search(value)
+        ):
+            forwarded[key] = value
+        else:
+            skipped_count += 1
+    return dict(sorted(forwarded.items())), dropped_secret_count, skipped_count
 
 
 class _NoDuplicatesSafeLoader(yaml.SafeLoader):
@@ -545,8 +560,8 @@ class DBTManager(SqlExecutionMixin):
         dbt_version: Optional[str] = None,
         environment: Optional[str] = None,
         env_vars: Optional[str] = None,
-        use_shell_env_vars: bool = False,
         *dbt_cli_args,
+        use_shell_env_vars: bool = False,
     ) -> SnowflakeCursor:
         if dbt_cli_args:
             processed_args = self._process_dbt_args(dbt_cli_args)
@@ -559,7 +574,7 @@ class DBTManager(SqlExecutionMixin):
 
         merged: Dict[str, str] = {}
         if use_shell_env_vars:
-            shell_vars, dropped_secret_count = _collect_shell_env_vars()
+            shell_vars, dropped_secret_count, skipped_count = _collect_shell_env_vars()
             if dropped_secret_count:
                 cli_console.message(
                     f"--use-shell-env-vars: dropped {dropped_secret_count} "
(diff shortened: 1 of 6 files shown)


src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -54,6 +55,24 @@ def _reject_control_chars(value: Optional[str], flag_name: str) -> Optional[str]
...
+    keys). Sorting makes the resulting SQL text deterministic across shells.
+    """
+    forwarded: Dict[str, str] = {}
+    dropped_secret_count = 0
+    for key, value in os.environ.items():
+        if not key.startswith(_ENV_VAR_KEY_PREFIX):
+            continue
+        if key.startswith(DBT_ENV_SECRET_PREFIX):
+            dropped_secret_count += 1
+            continue
```
sfc-gh-jwilkowski: I'd go more defensive here - `DBT_Env_Secret_TOKEN` would not be dropped. It looks like a mistake to set that in the first place, but even so I'd try to protect the user from simple mistakes

src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -54,6 +55,24 @@ def _reject_control_chars(value: Optional[str], flag_name: str) -> Optional[str]
...
+        forwarded[key] = value
```
sfc-gh-jwilkowski: worth sanitizing the values as well

src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -480,17 +500,48 @@ def execute(
...
+            if shell_vars:
+                cli_console.warning(
+                    f"--use-shell-env-vars: forwarded {len(shell_vars)} shell "
+                    "environment variable(s) into query text. Never put "
+                    "credentials, tokens, passwords, or other confidential "
+                    "data in shell environment variables starting with the "
+                    "DBT_* prefix."
+                )
+            else:
+                cli_console.message(
```
sfc-gh-jwilkowski, line 602: nit: if all the env vars were secrets, this message would be not very useful - somebody knows how to set them, but for some reason everything was set as secrets. I'm also wondering if we shouldn't just advise on consulting documentation of your shell. This list could grow and I don't this we want to maintain it

sfc-gh-jwilkowski (changes requested):

src/snowflake/cli/_plugins/dbt/manager.py
sfc-gh-yuding: Thanks for bringing this up! Server-side, all variable keys inside `ENV_VARS` must be uppercase, so an entry like `DBT_Env_Secret_TOKEN` would ultimately be rejected anyway.

I realized we should add this uppercase validation to the CLI so it fails fast. Since this is a minor change, I went ahead and included it in this PR. But please feel free to let me know if you'd prefer I split it out!

src/snowflake/cli/_plugins/dbt/manager.py
sfc-gh-yuding: Got it! Yes, advising consulting the doc would make more sense. Updated. Also only showing this message when we got exactly nothing starting with DBT_* from the shell.

src/snowflake/cli/_plugins/dbt/manager.py
sfc-gh-yuding: Yes,  `os.environ` values are always non-null str so I just added the additional control char check here.

src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -536,17 +571,50 @@ def execute(
...
+        merged: Dict[str, str] = {}
+        if use_shell_env_vars:
+            shell_vars, dropped_secret_count, skipped_count = _collect_shell_env_vars()
+            if dropped_secret_count:
+                cli_console.message(
+                    f"--use-shell-env-vars: dropped {dropped_secret_count} "
+                    f"{DBT_ENV_SECRET_PREFIX}* environment variable(s) from "
+                    "shell. To forward secrets, use the secrets: block in "
+                    "env.yml referencing a Snowflake SECRET object, or pass "
+                    "them explicitly via --env-vars."
```
sfc-gh-jwilkowski:
```suggestion
                    "env.yml referencing a Snowflake SECRET object, or if those are not sensitive pass "
                    "them explicitly via --env-vars."
```

sfc-gh-jwilkowski (approved):
LGTM, thanks!

PR #3110, opened 2026-06-15
Snowflake Apps: escape glob metacharacters in deploy PUT sources

sfc-gh-gbloom:
### Pre-review checklist

### Changes description

**Bug.** `snow app deploy` fails with `ProgrammingError 253006` while uploading app code when a file lives under a directory whose name contains glob metacharacters — most commonly a Next.js dynamic-route directory such as `[id]` or `[...slug]`.

**Root cause.** The Python connector expands every `PUT` source through `glob.glob` before uploading. The workspace upload built the `file://` source with `_local_path_to_file_uri`, which did **not** escape glob metacharacters, so a literal path like `.../[id]/page.tsx` was interpreted as a pattern: it matched nothing (`253006: File doesn't exist`) or, when a same-named sibling existed, resolved to a directory (`253006: Not a file but a directory`).

**Fix.**
- Escape the local path with `glob.escape` in `_local_path_to_file_uri`, so the connector matches the literal file. This is the actual bug fix and covers every per-file `PUT` (workspace and stage).
- Rewrite the stage upload to send each file with its own `PUT` (preserving the directory structure), mirroring the workspace path, instead of `StageManager.put_recursive` with `temp_directory=<bundle root>`. The previous recursive helper issued `PUT <dir>/*` (a directory glob) and destructively `rmtree`-d the local bundle as it walked the tree; the per-file form avoids the directory-matching glob entirely and no longer mutates the bundle.


src/snowflake/cli/_plugins/apps/manager.py
```diff
@@ -994,6 +1005,51 @@ def upload_to_workspace(
...
+        overwrite: bool = True,
+    ) -> Iterator[Dict[str, str]]:
+        """Recursively upload *local_root*'s contents into a stage.
+
+        Each file under *local_root* is uploaded with its own ``PUT``
+        statement, preserving the relative directory structure under
+        ``@<stage>``.  Files are uploaded one-at-a-time (rather than via
+        ``PUT <dir>/*``) because the glob form also matches subdirectories,
+        and the Snowflake PUT endpoint rejects directories with ``253006:
+        Not a file but a directory``.  This mirrors :meth:`upload_to_workspace`
```
sfc-gh-jfloyd, line 1021: Is this going to make the upload performance too slow?

sfc-gh-gbloom: I don't think so, we don't see a significant upload time difference between stages and workspaces (already file by file today).
We do lose the connection and require an additional round trip for each file, but in the ~30 file uploads we see today the difference is not significant. Also the file upload part is relatively small of the entire deploy (median 8 sec out of 2 minutes).

We are planning to improve upload times, but didn't implement yet as it's not currently an issue.

sfc-gh-jfloyd: How are we planning to improve upload times?

sfc-gh-jfloyd (approved):

PR #3113, opened 2026-06-16
fix: make REST error handling work on connector v4 and Universal Driver v5

sfc-gh-turbaszek:
## Why

The integration suite runs against both `snowflake-connector-python` v4.x and the **Universal Driver v5** (the `ud` hatch env). The object-create REST path imports HTTP error handling from the legacy connector's vendored `requests`:

```python
from snowflake.connector.vendored.requests.exceptions import HTTPError
```

UD v5 removed the `vendored` sub-package, so this raises `ModuleNotFoundError: No module named 'snowflake.connector.vendored'` — failing the 11 `test_object.py::test_create*` / `test_create_error*` integration tests under v5.

The two drivers also report HTTP failures differently:

| | connector v4.x | Universal Driver v5 |
|---|---|---|
| Exception raised | vendored `requests` `HTTPError` | `OperationalError` |
| 400 | `BadRequest` | `OperationalError` (`http_status == 400`) |
| Read status via | `err.response.status_code` | `err.http_status` |

## What

A driver-version-aware compat shim, with all three call sites routed through it:

- **`api/connector_errors.py`** (new) — detects the connector major version and exposes `HTTP_FAILURE_ERRORS` (the exception class to catch per driver) and `http_status_code(err)` (reads `http_status` on v5, `.response.status_code` on v4).
- **`api/rest_api.py`** — `get_endpoint_exists` / `_fetch_endpoint_exists` catch `HTTP_FAILURE_ERRORS` and read the status via the shim, preserving v4's `BadRequest`/400 semantics on both drivers.
- **`_plugins/object/manager.py`** — `_handle_create_error_codes` drops the vendored import and maps status codes via the shim (incl. a `case 400` for v5 parity).
- **`tests/api/test_rest_api.py`** — driver-agnostic `_make_http_failure` factory replacing direct vendored-`HTTPError` construction.

## Tests

`tests/api/test_rest_api.py` passes **13/13** against both connector v4.5.0 and a locally-built UD v5.

## Paired change (required)

The `http_status` attribute and the `snowflake.connector.network` backward-compat shim this PR depends on are added in a Universal Driver PR: snowflake-eng/universal-driver#274. The `ud` hatch env pulls UD from `@main`, so the 11 `test_object.py` integration tests go green once that lands.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/snowflake/cli/api/connector_errors.py
```diff
@@ -0,0 +1,74 @@
...
+"""
+
+from __future__ import annotations
+
+from typing import Optional, Tuple, Type
+
+import snowflake.connector
+
+
+def _connector_major() -> int:
```
sfc-gh-turbaszek: This should be something like `_is_v5_driver` or something

src/snowflake/cli/api/connector_errors.py
```diff
@@ -0,0 +1,74 @@
...
+    try:
+        return int(snowflake.connector.__version__.split(".", 1)[0])
```
sfc-gh-turbaszek: Instead of that import `version.VERSION` tuple
```suggestion
        return VERSION[0] >= 5
```

src/snowflake/cli/api/connector_errors.py
```diff
@@ -0,0 +1,74 @@
...
+    HTTP_FAILURE_ERRORS = (HTTPError,)
+
+
+def http_status_code(err: BaseException) -> Optional[int]:
+    """Return the HTTP status of a raw HTTP failure, or None if it is not one.
+
+    Works across connector v4 (vendored ``HTTPError`` with ``.response``) and
+    Universal Driver v5 (``OperationalError`` with ``.http_status``).
+    """
+    # Universal Driver v5: structured attribute set by SnowflakeRestful.
```
sfc-gh-turbaszek: Add if/else to the body based on driver version

src/snowflake/cli/api/connector_errors.py
```diff
@@ -0,0 +1,74 @@
...
+    # OperationalError is broad; only instances carrying an ``http_status`` are
+    # raw HTTP failures (see http_status_code). Others are re-raised by callers.
+    HTTP_FAILURE_ERRORS = (OperationalError,)
+else:
+    from snowflake.connector.vendored.requests.exceptions import HTTPError
+
+    HTTP_FAILURE_ERRORS = (HTTPError,)
+
+
+def http_status_code(err: BaseException) -> Optional[int]:
```
sfc-gh-turbaszek:
```suggestion
def get_http_status_code(err: BaseException) -> Optional[int]:
```

sfc-gh-turbaszek:
Seems to be working fine with UD https://github.com/snowflakedb/snowflake-cli/actions/runs/27681189473/job/81868795003

RELEASE-NOTES.md
```diff
@@ -23,6 +23,7 @@
 * Added `--no-prompt-exit-repl` option and configuration setting to skip the exit confirmation prompt in the SQL REPL.
 
 ## Fixes and improvements
+* Fixed REST API object operations (e.g. `snow object create`) crashing with `ModuleNotFoundError: No module named 'snowflake.connector.vendored'` when running against the Snowflake Universal Driver (connector-python v5). HTTP error handling now works on both connector v4.x and the Universal Driver v5.
```
sfc-gh-jwilkowski, line 26: nit: I don't think a release note for this change is really needed

sfc-gh-jwilkowski (approved):
LGTM, you can reconsider adding release note entry here

PR #3118, opened 2026-06-17
SNOW-2457402 Encoding handling fixes

sfc-gh-kolszewski:
### Pre-review checklist

### Changes description
Three fixes to the encoding configuration added in #2792:

1. Validate SNOWFLAKE_CLI_ENCODING_STDOUT env var before applying it

_apply_stdout_encoding_from_env was calling sys.stdout.reconfigure with unvalidated input. Invalid codec names now produce a clear warning on stderr instead of silently failing.

2. Restore platform-default encoding for connections.toml

_read_connections_toml and _update_connections_toml were hardcoded to utf-8 after #2792. They now use get_file_io_encoding(), which returns None when not configured — preserving the pre-#2792 platform-default behavior and avoiding potential breakage for users with non-UTF-8 encoded files.

3. Use env-var-only encoding for _read_config_file_toml

_read_config_file_toml now uses a new _get_file_io_encoding_env_only() helper that reads only from SNOWFLAKE_CLI_ENCODING_FILE_IO, not from the config file itself. This avoids a chicken-and-egg dependency and similarly restores the platform-default behavior when no env var is set.

No behavioral change for users who have not configured any encoding settings.

### DIFF
--- a/src/snowflake/cli/_app/__main__.py
+++ b/src/snowflake/cli/_app/__main__.py
@@ -14,6 +14,7 @@
 
 from __future__ import annotations
 
+import codecs
 import os
 import sys
 
@@ -29,6 +30,10 @@ def _apply_stdout_encoding_from_env() -> None:
     """
     enc = os.environ.get("SNOWFLAKE_CLI_ENCODING_STDOUT")
     if not enc:
+        return
+    try:
+        codecs.lookup(enc)
+    except LookupError:
         return
     try:
         sys.stdout.reconfigure(encoding=enc)  # type: ignore[attr-defined,union-attr]
--- a/src/snowflake/cli/_app/__main__.py
+++ b/src/snowflake/cli/_app/__main__.py
@@ -15,6 +15,7 @@
 from __future__ import annotations
 
 import codecs
+import io
 import os
 import sys
 
@@ -34,10 +35,13 @@ def _apply_stdout_encoding_from_env() -> None:
     try:
         codecs.lookup(enc)
     except LookupError:
+        # Silently skip — the invalid value will be caught again by
+        # get_stdout_encoding() → _validate_encoding() during config_init,
+        # which raises a proper ClickException with context.
         return
     try:
         sys.stdout.reconfigure(encoding=enc)  # type: ignore[attr-defined,union-attr]
-    except (AttributeError, Exception):
+    except (AttributeError, io.UnsupportedOperation):
         pass
 
 
--- a/src/snowflake/cli/api/config.py
+++ b/src/snowflake/cli/api/config.py
@@ -711,7 +711,9 @@ def _dump_config(config_and_connections: Dict):
         else:
             config_toml_dict.pop("connections", None)
 
-    with SecurePath(get_config_manager().file_path).open("w+") as fh:
+    with SecurePath(get_config_manager().file_path).open(
+        "w+", use_file_io_encoding=False
+    ) as fh:
         dump(config_toml_dict, fh)
 
 
@@ -783,32 +785,25 @@ def get_feature_flags_section() -> Dict[str, bool | Literal["UNKNOWN"]]:
     return {k: _bool_or_unknown(v) for k, v in flags.items()}
 
 
-def _get_file_io_encoding_env_only() -> Optional[str]:
-    """Read file I/O encoding from the env var only, without accessing the config file.
-
-    Used when opening the config file itself to avoid a chicken-and-egg problem
-    (can't read the encoding setting from the file we haven't opened yet).
-    Returns None when the env var is not set, which preserves the platform-default
-    behavior that existed before encoding support was added.
-    """
-    value = get_env_value(*ENCODING_SECTION_PATH, key="file_io")
-    return _validate_encoding(value, "cli.encoding.file_io")
-
-
 def _read_config_file_toml() -> dict:
-    return tomlkit.loads(
-        get_config_manager().file_path.read_text(
-            encoding=_get_file_io_encoding_env_only()
-        )
-    ).unwrap()
+    # config.toml is read by the connector (via read_config()) with platform
+    # default encoding, and _dump_config writes it with PLATFORM_DEFAULT_ENCODING
+    # for the same reason — keeping CLI read/write consistent with the connector.
+    return tomlkit.loads(get_config_manager().file_path.read_text()).unwrap()
 
 
 def _read_connections_toml() -> dict:
-    return tomlkit.loads(
-        get_connections_file().read_text(encoding=get_file_io_encoding())
-    ).unwrap()
+    # connections.toml is a shared file: the snowflake-connector reads it via
+    # its own read_config() using read_text() with no explicit encoding
+    # (platform default). Applying a custom file_io encoding here would make
+    # the CLI and connector read the same file with different codecs, corrupting
+    # non-ASCII content for one of them. Use platform default to stay consistent
+    # with the connector.
+    return tomlkit.loads(get_connections_file().read_text()).unwrap()
 
 
 def _update_connections_toml(connections: dict):
-    with open(get_connections_file(), "w", encoding=get_file_io_encoding()) as f:
+    # Same rationale as _read_connections_toml: platform default keeps CLI
+    # writes consistent with the connector's reads.
+    with open(get_connections_file(), "w") as f:
         f.write(tomlkit.dumps(connections))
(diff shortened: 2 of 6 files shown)


src/snowflake/cli/api/config.py
```diff
@@ -783,19 +783,32 @@ def _bool_or_unknown(value):
...
+    value = get_env_value(*ENCODING_SECTION_PATH, key="file_io")
+    return _validate_encoding(value, "cli.encoding.file_io")
+
+
 def _read_config_file_toml() -> dict:
-    # TOML files are always UTF-8 by spec; don't apply user's file_io encoding
     return tomlkit.loads(
-        get_config_manager().file_path.read_text(encoding="utf-8")
+        get_config_manager().file_path.read_text(
+            encoding=_get_file_io_encoding_env_only()
```
sfc-gh-olorek: `_read_config_file_toml` (read) now uses **env-only** encoding, but config.toml is *written* in `_dump_config` via `SecurePath(...).open("w+")`, whose `encoding` defaults to the **full** `get_file_io_encoding()` (config-file value **+** env var). So if someone sets `[cli.encoding] file_io` in config.toml but not the env var, config.toml gets written with that encoding and read back with the platform default — a potential mismatch for non-ASCII content.

The chicken-and-egg rationale doesn't quite hold at the only call site: `_read_config_file_toml` is called solely from `_dump_config`, which runs after `_read_config_file()` has already loaded config into the manager, so `get_file_io_encoding()` would resolve from memory without re-reading the file. Consider using `get_file_io_encoding()` here for read/write symmetry (matching `_read_connections_toml`), or documenting why config.toml read intentionally differs from its write.

src/snowflake/cli/api/config.py
```diff
@@ -783,19 +783,32 @@ def _bool_or_unknown(value):
...
+        )
     ).unwrap()
 
 
 def _read_connections_toml() -> dict:
-    # TOML files are always UTF-8 by spec; don't apply user's file_io encoding
-    return tomlkit.loads(get_connections_file().read_text(encoding="utf-8")).unwrap()
+    return tomlkit.loads(
+        get_connections_file().read_text(encoding=get_file_io_encoding())
```
sfc-gh-olorek: Nice — this makes connections.toml read/write consistent with `SecurePath`'s `file_io` handling. One thing worth confirming while the feature is still unreleased: connections.toml is also read by the snowflake-connector's own config manager. If the connector reads it as utf-8 (or a different default) while the CLI now writes with `get_file_io_encoding()`, a file round-tripped through both could mismatch for non-ASCII content. Do they resolve to the same encoding?

src/snowflake/cli/_app/__main__.py
```diff
@@ -30,6 +31,10 @@ def _apply_stdout_encoding_from_env() -> None:
     enc = os.environ.get("SNOWFLAKE_CLI_ENCODING_STDOUT")
     if not enc:
         return
+    try:
+        codecs.lookup(enc)
+    except LookupError:
+        return
```
sfc-gh-olorek, line 41: The PR description says invalid `SNOWFLAKE_CLI_ENCODING_STDOUT` values "produce a clear warning on stderr instead of silently failing," but this path silently `return`s (and `test_apply_stdout_encoding_from_env_invalid_skips` asserts `reconfigure` is not called). Note the `except (AttributeError, Exception)` below already swallowed the `LookupError` from `reconfigure`, so the runtime behavior of this function is effectively unchanged.

The clear error does surface later via `config_init` → `get_stdout_encoding()` → `_validate_encoding` (a `ClickException`), so users are protected — but either correct the description, or add a short comment here noting that invalid values are validated/reported later and skipped here to avoid failing before loggers/output are set up.

src/snowflake/cli/_app/__main__.py
```diff
@@ -30,6 +31,10 @@ def _apply_stdout_encoding_from_env() -> None:
...
     try:
         sys.stdout.reconfigure(encoding=enc)  # type: ignore[attr-defined,union-attr]
     except (AttributeError, Exception):
```
sfc-gh-olorek: `except (AttributeError, Exception)` is redundant — `Exception` already includes `AttributeError`. The sibling `apply_stdout_encoding()` in `config.py` catches narrowly (`except (AttributeError, io.UnsupportedOperation)`); worth matching that here instead of catching all of `Exception`.

tests/test_config.py
```diff
@@ -1044,6 +1044,125 @@ def test_invalid_stdout_encoding_from_config_file_raises_during_init(
...
+def test_get_file_io_encoding_env_only_rejects_invalid_encoding(monkeypatch):
+    """_get_file_io_encoding_env_only raises ClickException for an invalid codec."""
+    monkeypatch.setenv("SNOWFLAKE_CLI_ENCODING_FILE_IO", "not-a-real-encoding")
+    from snowflake.cli.api.config import _get_file_io_encoding_env_only
+
+    with pytest.raises(ClickException, match="Invalid encoding 'not-a-real-encoding'"):
+        _get_file_io_encoding_env_only()
+
+
+def test_read_config_file_toml_uses_env_only_encoding(monkeypatch, config_file):
```
sfc-gh-olorek: This test sets **both** the env var (utf-8) and the config-file value (latin-1) and asserts the env wins — good. But the case that actually demonstrates "env-only, ignores config" is missing: env **unset** + `[cli.encoding] file_io` set in config.toml → assert `read_text` is called with `None` (not the config value). That's also the exact scenario behind the read/write asymmetry noted on `_read_config_file_toml`. A behavioral round-trip test (write non-ASCII as cp1252, read it back, assert equality) would be stronger than asserting mock kwargs.

sfc-gh-olorek (commented):
Reviewed the encoding fixes. Direction looks right, and it's good that #2792 is still unreleased (per RELEASE-NOTES) so `skip-release-notes` is appropriate and there's no shipped behavior to break. This also brings these three functions in line with `SecurePath`, which already defaults to `get_file_io_encoding()`.

Main thing to decide on: the config.toml read/write encoding asymmetry (inline). The rest are minor — a description/behavior mismatch on the stdout change, a redundant exception catch, a connector-consistency question, and a small test gap. Details inline.

src/snowflake/cli/api/config.py
sfc-gh-kolszewski: Good catch on the call-site analysis. We dug into this further and found that the connector's `read_config()` reads both config.toml *and* connections.toml via the same `filep.read_text()` call (no explicit encoding, platform default). That made `get_file_io_encoding()` the wrong choice for both files: applying it to CLI reads/writes while the connector uses platform default would corrupt non-ASCII content for one of them.

Final approach: both config.toml and connections.toml use platform default unconditionally. For `_read_config_file_toml` that's plain `read_text()`. For the `_dump_config` write via `SecurePath.open()` we added a `use_file_io_encoding: bool = True` parameter to `SecurePath` — passing `False` skips the `get_file_io_encoding()` resolution and lets Python use its own default.

src/snowflake/cli/api/config.py
sfc-gh-kolszewski: Confirmed: the connector reads connections.toml in the same `read_config()` loop using `filep.read_text()` with no encoding argument — platform default. So `get_file_io_encoding()` was reverted here too. `_read_connections_toml` now uses plain `read_text()` and `_update_connections_toml` uses `open()` without an explicit encoding. Both match what the connector does.

src/snowflake/cli/_app/__main__.py
sfc-gh-kolszewski: Added a comment at the `LookupError` catch explaining the intent: invalid values are silently skipped here because they will surface as a proper `ClickException` later via `get_stdout_encoding()` → `_validate_encoding()` during `config_init`, once loggers and output are set up. The redundant `except` clause was fixed in the same commit (see below).

src/snowflake/cli/_app/__main__.py
sfc-gh-kolszewski: Fixed — narrowed to `except (AttributeError, io.UnsupportedOperation)` to match `apply_stdout_encoding()` in config.py.

tests/test_config.py
sfc-gh-kolszewski: The test was restructured. Since `_read_config_file_toml` now always uses platform default (no `get_file_io_encoding()` at all), the parametrized test with three encoding values was replaced by a single `test_read_config_file_toml_uses_platform_default_encoding` that asserts `read_text()` is called with no arguments. The `use_file_io_encoding=False` path in `SecurePath` is covered by a new parametrized case added to each of the three `test_*_encoding` tests in `test_secure_path.py`.

sfc-gh-olorek (approved):
lgtm

PR #3125, opened 2026-06-20
fix: [SNOW-3444081] skip role validation in profiles.yml for Jinja-templated values and improve output message

sfc-gh-yuding:
### Pre-review checklist

### Changes description
1. When profiles.yml uses a Jinja expression as the role value (e.g. {{ env_var('DBT_ROLE') }}), _validate_role tried to USE ROLE with the raw template string, which Snowflake rejects. Jinja-templated values are resolved by GS at CREATE/execute time — skip local validation for them and let the server surface any errors.
2. Improve `--use-shell-env-vars` output messages — replaces variable(s) with proper singular/plural, and adds "starting with DBT_*" to the forwarded-vars warning for clarity.

### DIFF
--- a/src/snowflake/cli/_plugins/dbt/manager.py
+++ b/src/snowflake/cli/_plugins/dbt/manager.py
@@ -45,6 +45,9 @@ DBT_ENV_SECRET_PREFIX = "DBT_ENV_SECRET_"
 _ENV_VAR_KEY_PREFIX = "DBT_"
 _ENV_VAR_KEY_RE = re.compile(r"^[A-Za-z0-9_]+$")
 _CONTROL_CHAR_RE = re.compile(r"[\x00-\x1f\x7f]")
+# Matches any Jinja expression ({{ ... }}). Used to skip local role validation for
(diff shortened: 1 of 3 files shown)


src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -576,27 +579,31 @@ def execute(
         if use_shell_env_vars:
             shell_vars, dropped_secret_count, skipped_count = _collect_shell_env_vars()
             if dropped_secret_count:
+                _var = "variable" if dropped_secret_count == 1 else "variables"
```
sfc-gh-jwilkowski, line 582: This logic is repeated 3 times, would be great to extract it to some helper functon

sfc-gh-jwilkowski (approved):

src/snowflake/cli/_plugins/dbt/manager.py
sfc-gh-yuding: Thank you! I will refactor it in next PR.

PR #3141, opened 2026-07-01
SNOW-3659937: dbt deploy — support dbt_projects_profiles.yml precedence

sfc-gh-nijain:
When ENABLE_FIX_3659937_DBT_PROJECTS_PROFILES_FILE is enabled, `snow dbt deploy` resolves the profiles file from --profiles-dir by precedence (`dbt_projects_profiles.yml` gets a higher precedence over  `profiles.yml`) and stages the winner under its own name, so the server can apply the precedence at EXECUTE DBT PROJECT time. The flag is off by default; with it off only profiles.yml is recognized and behavior is unchanged.

Precedence lives in one place (_candidate_profiles_filenames / _resolve_profiles_filename) shared by _validate_profiles and _prepare_profiles_file. Adds unit + deploy-level tests and a release note.

* This change will be released only after backend changes (https://github.com/snowflake-eng/snowflake/pull/466801) are released. Both these changes have their own independent parameters to turn the feature on/off. 

### Pre-review checklist

### Changes description

`snow dbt deploy` now supports a higher-precedence profiles file, `dbt_projects_profiles.yml`, in `--profiles-dir`. When present it is staged into the deployed project under its own name, taking precedence over `profiles.yml`; the server (SNOW-3659937) then materializes it as the effective `profiles.yml` at run time.

Gated behind the `ENABLE_FIX_3659937_DBT_PROJECTS_PROFILES_FILE` feature flag (default off), mirroring the server gate of the same name. With the flag off, only `profiles.yml` is recognized, so behavior is unchanged.

- Resolution precedence lives in one place (`_candidate_profiles_filenames` / `_resolve_profiles_filename`) shared by `_validate_profiles` and `_prepare_profiles_file`.
- The chosen file is staged under its own name (not renamed); any `profiles.yml` already copied from the source project is left in place.
- `--profiles-dir` help text mentions the precedence only when the flag is on.

> Do not merge until the backend gate `ENABLE_FIX_3659937_DBT_PROJECTS_PROFILES_FILE` is released and deployed. Until then the server ignores the file, so the CLI change would be a no-op for users.

PrPr (default-off flag), so no release note per lifecycle.md — `skip-release-notes` label applied.

### Testing done

Unit tests (`tests/dbt/test_manager.py`), full dbt suite green (210 passed).
The core is an 8-scenario matrix over flag on/off x each profiles file present/missing. "Before" is current `main`.

| # | Flag | dbt_projects_profiles.yml | profiles.yml | Before | After |
|---|------|---------------------------|--------------|--------|-------|
| 1 | off | present | present | stages profiles.yml | stages profiles.yml |
| 2 | off | present | missing | error: profiles.yml missing | error: profiles.yml missing |
| 3 | off | missing | present | stages profiles.yml | stages profiles.yml |
| 4 | off | missing | missing | error: profiles.yml missing | error: profiles.yml missing |
| 5 | on  | present | present | stages profiles.yml | stages dbt_projects_profiles.yml |
| 6 | on  | present | missing | error: profiles.yml missing | stages dbt_projects_profiles.yml |
| 7 | on  | missing | present | stages profiles.yml | stages profiles.yml |
| 8 | on  | missing | missing | error: profiles.yml missing | error names both files |

Flag-off rows (1-4) match `main` exactly, so this is backward compatible.

**Live verification with the real `snow dbt deploy`:**

Client side (deploy staging, from the connector upload log):
- flag off, only dbt_projects_profiles.yml -> fails locally "profiles.yml does not exist"
- flag on,  only dbt_projects_profiles.yml -> uploads dbt_projects_profiles.yml
- flag off, both present -> uploads profiles.yml
- flag on,  both present -> uploads dbt_projects_profiles.yml

Server side (reg SUT, project staged with both files targeting different schemas, then `EXECUTE DBT PROJECT ... run`):
- gate off at create -> model built in PROFILES_YML (used profiles.yml)
- gate on  at create -> model built in DBTPROJ_YML (used dbt_projects_profiles.yml)

Note: the effective profiles file is baked at `CREATE DBT PROJECT` time, so the gate must be set before create.


sfc-gh-nijain (commented):
Backend changes needed to be merged prior to merging this change: https://github.com/snowflake-eng/snowflake/pull/466801

src/snowflake/cli/_plugins/dbt/manager.py
```diff
@@ -523,11 +548,25 @@ def _validate_role(self, role_name: str) -> bool:
...
+        filename = DBTManager._resolve_profiles_filename(profiles_path)
+        if filename is None:
+            # _validate_profiles runs first and already errors when no profiles
+            # file exists; guard defensively in case this is called directly.
+            names = " or ".join(DBTManager._candidate_profiles_filenames())
+            raise CliError(
+                f"{names} does not exist in directory {profiles_path.absolute()}."
+            )
+        source_profiles_file = SecurePath(profiles_path / filename)
+        target_profiles_file = SecurePath(tmp_path / filename)
```
sfc-gh-yuding, line 569: Wanted to check on - when the flag is on and dbt_projects_profiles.yml wins, it looks like profiles.yml might still get uploaded to the stage un-redacted and un-validated. Just wanted to confirm whether   that's intended.                                                                                                                                                                                                            

If the idea is that only dbt_projects_profiles.yml matters when it's present, maybe dropping profiles.yml from the temp dir when it loses would work better? Could be I'm missing something though — let me know what you think!

tests/dbt/test_manager.py
```diff
@@ -1907,3 +1946,338 @@ def test_only_secrets_returns_empty_dict_with_count(self, clean_dbt_env):
         assert forwarded == {}
         assert dropped == 1
         assert skipped == 0
+
```
sfc-gh-yuding, line 1949: Would be nice to add an end-to-end integration test in test_dbt.py for the dbt_projects_profiles.yml precedence, since the behavior really only surfaces in the full deploy flow. Given the feature is still gated, we could mark it @pytest.mark.qa_only so CI skips it for now, and remove that marker once the feature fully launches.

sfc-gh-yuding (commented):
Mostly looks good to me! Just want to double check on the profiles.yml redact and validation part. Also the development for snowcli has been moved to the private repo - snowflake-eng/snowflake-cli last week, so you might need to open PR there.

tests/dbt/test_manager.py
sfc-gh-nijain: Good call, added

src/snowflake/cli/_plugins/dbt/manager.py
sfc-gh-nijain: Created a PR against `snowflake-eng/snowflake-cli` : https://github.com/snowflake-eng/snowflake-cli/pull/35

Updated the code to not upload the `profiles.yml` if `dbt_projects_profiles.yml` ends up winning.

